PR #1, opened 2024-12-16
Some more preliminary changes for the queijo repo

Vighnesh-V:
1. Cmake now produces compile_commands by default
2. test directory for basic lua files
3. Queijo.lib now links in Lua.VM
4. simple registration functions for the queijo runtime

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,5 @@
 __pycache__
 .cache
 .clangd
+compile_commands.json
+*~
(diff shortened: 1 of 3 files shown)


aatxe (changes requested):
Can you remove the filler libraries bits? I do wanna keep the cmake changes and some of the queijo.cpp changes. We should also add the `~` file gitignore rule too.

aatxe (approved):

PR #4, opened 2024-12-16
Refactor into separate libraries per domain.

aatxe:
It seems like it'd be better if we split things up into a single library per domain, both in terms of workflow right now and in terms of modularity longer-term. This is adding three initial libraries, `fs`, `net` and `task`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,22 +4,31 @@ project(Queijo LANGUAGES CXX C)
 add_subdirectory(luau)
 add_subdirectory(libuv)
 
-add_library(Queijo.Lib STATIC)
+add_library(Queijo.Fs STATIC)
+add_library(Queijo.Net STATIC)
+add_library(Queijo.Task STATIC)
 add_executable(Queijo.CLI)
 set_target_properties(Queijo.CLI PROPERTIES OUTPUT_NAME queijo)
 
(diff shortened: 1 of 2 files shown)


vegorov-rbx (approved):

CMakeLists.txt
```diff
@@ -4,22 +4,31 @@ project(Queijo LANGUAGES CXX C)
...
-
-target_compile_options(Queijo.Lib PRIVATE ${QUEIJO_OPTIONS})
+target_compile_features(Queijo.Fs PUBLIC cxx_std_17)
+target_compile_features(Queijo.Net PUBLIC cxx_std_17)
+target_compile_features(Queijo.Task PUBLIC cxx_std_17)
+target_include_directories(Queijo.Fs PUBLIC Fs/include ${LIBUV_INCLUDE_DIR})
+target_include_directories(Queijo.Net PUBLIC Net/include ${LIBUV_INCLUDE_DIR})
+target_include_directories(Queijo.Task PUBLIC Task/include ${LIBUV_INCLUDE_DIR})
+
+target_link_libraries(Queijo.Fs PRIVATE Luau.VM)
```
Vighnesh-V, line 24: I think Fs also needs uv

Vighnesh-V (approved):

PR #24, opened 2024-12-18
Example of async network requires with await library

vegorov-rbx:


examples/task.luau
```diff
@@ -0,0 +1,66 @@
+export type Task = {
+    success: boolean?,
+    result: any,
+    co: thread
+}
+
+local function task(f, ...): Task
```
aatxe: create

vegorov-rbx: yeah

aatxe (approved):

PR #32, opened 2024-12-19
Add a luau library with support for parsing.

aatxe:
This PR adds a `luau` library to queijo that allows the use of the Luau parser to get a table AST. As of the time of writing, this only supports expressions, but statement support is coming.


luau/src/luau.cpp
```diff
@@ -0,0 +1,888 @@
...
+        serialize(location.begin);
+        lua_setfield(L, -2, "begin");
+
+        serialize(location.end);
+        lua_setfield(L, -2, "end");
+    }
+
+    void serialize(Luau::AstName name)
+    {
+        lua_createtable(L, 0, 2);
```
vegorov-rbx:
```suggestion
        lua_createtable(L, 0, 1);
```

vegorov-rbx (approved):

PR #34, opened 2024-12-19
Queijo supports a `--check` option now, in order to run strict mode typechecking.

Vighnesh-V:
Followup work:
1. Requires - we need to write the appropriate require resolver to run the typecheck
2. Flesh out the queijo apis a bit more

### DIFF
--- a/cli/tc.cpp
+++ b/cli/tc.cpp
@@ -1,0 +1,240 @@
+#include "tc.h"
+#include "Luau/BuiltinDefinitions.h"
+#include "Luau/Error.h"
+#include "Luau/Transpiler.h"
+#include "Luau/TypeAttach.h"
+
+static const std::string kQueijoDefinitions = R"QUEIJO_TYPES(
+-- Net api
+declare net: {
+    get: (string) -> string,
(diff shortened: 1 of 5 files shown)


aatxe:
Can we not use the require resolver from Require.cpp? the same one we're using for the dynamic implementation?

Vighnesh-V:
> Can we not use the require resolver from Require.cpp? the same one we're using for the dynamic implementation?

We can. That was next up on my list :)

cli/tc.cpp
```diff
@@ -0,0 +1,240 @@
...
+#include "Luau/Error.h"
+#include "Luau/Transpiler.h"
+#include "Luau/TypeAttach.h"
+
+static const std::string kQueijoDefinitions = R"QUEIJO_TYPES(
+-- Net api
+declare net: {
+    get: (string) -> string,
+ -- is this right? I feel like we want a promise here ... 
+    getasync: (string) -> string,
```
vegorov-rbx: Case is wrong :)

vegorov-rbx (approved):

PR #42, opened 2024-12-20
Implement locals in the parser and build a simple linter example with it.

aatxe:
This adds support for locals to the parser, and includes a simple linter because I had to test this stuff somehow.

### DIFF
--- a/luau/src/luau.cpp
+++ b/luau/src/luau.cpp
@@ -108,6 +108,7 @@ struct AstSerialize : public Luau::AstVisitor
 
         if (lua_isnil(L, -1))
         {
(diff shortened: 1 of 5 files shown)


luau/src/luau.cpp
```diff
@@ -353,12 +395,13 @@ struct AstSerialize : public Luau::AstVisitor
 
         // TODO: attributes
 
-        // TODO: locals
-        lua_pushnil(L);
+        if (node->self)
+            serialize(node->self);
```
vegorov-rbx, line 399: Looks like a missing else?

vegorov-rbx (approved):

PR #62, opened 2025-02-04
Rename the project to `lute`.

aatxe:
[jackdotink](https://github.com/jackdotink) suggested the absolutely wonderful name of `lute`. It's much easier to pronounce, cuter, could lend itself well to good visual design language, and opens up a few different possible categories of user library names and so forth that are, finally, divorced from the moon: bardic things, medieval things, instruments, music things generally, etc. A wonderful name that I immediately passed the vibe check for me, the runtime name bikeshed is over.


jackhexed:
dibs on the library name "flute"

PR #67, opened 2025-02-07
Implement JSON serdes std library

bjcscat:
This JSON serializer/deserializer probably could do with some performance improvements but I believe it's adequate at the moment and avoids adding some dependency to parse JSON. I did some testing and I think it should be mostly compliant but more testing is definitely not a bad thing


aatxe (approved):
Thanks!

PR #70, opened 2025-02-13
Improve Performance of `read` and `readfiletostring` Functions

SnowyShiro:
This PR significantly improves the performance of the `read` and `readfiletostring` functions in the filesystem library by optimizing how data from `uv_fs_read` is stored.

The functions now use the same approach as `readasync`, storing the read result in a `std::vector<char>` by pushing each character of the `uv_fs_read` result. This replaces the previous method, which appended data to a Luau buffer. The original approach caused frequent buffer resizes, leading to poor performance, especially with large files.

Speed Benchmark:
| File Size | Original | Improved | Speedup |
| --------- | -------- | -------- | ------- |
| 1KB       | 0.0004s  | 0.0001s  | ~4x     |
| 1MB       | 0.3768s  | 0.0037s  | ~102x   |
| 100MB     | 39.4168s | 0.3982s  | ~99x    |

Benchmark Code:
```luau
local fs = require'@lute/fs';

local function benchmark(name, filename)
    local fd = fs.open(filename, 'r');

    local start = os.clock();
    local data = fs.read(fd);
    local took = os.clock() - start;

    fs.close(fd);

    assert(#data ~= 0, 'Error reading: ' .. filename);
    print(string.format('%s %.4f seconds', name, took));
end;

benchmark('1KB   :', '1kb.txt');
benchmark('1MB   :', '1mb.txt');
benchmark('100MB :', '100mb.txt');
```

### DIFF
--- a/fs/src/fs.cpp
+++ b/fs/src/fs.cpp
@@ -114,7 +114,7 @@ int close(lua_State* L)
     return 0;
 }
 
-static char readBuffer[5];
+static char readBuffer[1024];
 int read(lua_State* L)
 {
     memset(readBuffer, 0, sizeof(readBuffer));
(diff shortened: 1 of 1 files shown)


SnowyShiro:
Oh yea, are y'all open to people adding stuff to Lute? I wanted to add/refactor some stuff in the filesystem lib, like adding `fs.stat` and `fs.readDir`

aatxe:
Yeah, totally open to additions to lute. They'll obviously be reviewed for reasonableness, but directory support is pretty much a no-brainer. :)

aatxe (approved):

aatxe:
Thanks for the improvements, I actually meant to do this kind of thing myself, but you beat me to it! 😸

PR #71, opened 2025-02-18
Start making the CST usable in Lute

JohnnyMorganz:
Now that CST nodes exist in Luau, this PR extends the existing `@lute/luau` C++ module to return full-fidelity CST nodes to Luau, rather than just AST nodes. This is aimed to support future usages of the CST for tooling such as code modding, linting or formatting implemented fully in Luau.

We extend the serialization of some AST nodes with extra tokens, and create tests to check for validity of tokenization and the roundtrippability of the serialization state in Luau (i.e., `print(parse(source) == source` where `print` is fully defined in Luau). Future AST nodes will be added in follow up commits.

Alongside the CST data, a key concept that this PR introduces is the idea of trivia. Right now, the Luau CST does not preserve trivia itself, but retains enough information to compute spans of trivia, allowing later lookup in the original source code. To extract trivia, we keep the original input buffer around, and then take the span between the last token and the next token, and call this "trivia". Trivia is then individually tokenized and attached to "Token" nodes.

There are some key rules for trivia:
- [x] Trivia consists of whitespace, single line comments and multi line comments. Each kind is a separate trivia token
- [x] Whitespace trivia is split up into separate tokens based on the '\n' character. Other characters of whitespace are joined together up to and including the first newline character. i.e., `<space><space>\n\t\t\n\n` is split into `[<space><space>\n, \t\t\n, \n]` 
- [ ] Trivia on a token is split up into leading and trailing trivia. Leading / trailing split is based on the Roslyn Compiler trivia rules: trailing trivia consists of everything up to and including the first `\n` newline character, and all further trivia is leading trivia of the next token. The idea being that `local x = ...\n` can then be represented as a single line in trivia, to allow easier replacement (rather than the `\n` being part of the next statement).

The last rule is not yet implemented but will be introduced in a follow-up commit.

The API still remains experimental and is subject to change. In particular, I'm not fully sold on the naming scheme for nodes and fields (particularly the fields that match keyword names) and it may require some standardisation. Also, we are somewhat mixing Ast nodes with Cst nodes, and we may want to separate the two.


luau/src/luau.cpp
```diff
@@ -167,59 +343,10 @@ struct AstSerialize : public Luau::AstVisitor
...
-            break;
-        case Luau::AstExprBinary::Op::Or:
-            lua_pushstring(L, "or");
-            break;
-        case Luau::AstExprBinary::Op::Op__Count:
+        if (op == Luau::AstExprBinary::Op::Op__Count)
             luaL_error(L, "encountered illegal operator: Op__Count");
-        }
+
+        lua_pushstring(L, Luau::toString(op).data());
```
aatxe, line 349: lol, nice

aatxe (approved):

PR #84, opened 2025-02-21
Add various filesystem operations

### LINKED ISSUE #82: Add fs apis for reading a directory of files
Probably `fs.listdir` and a way to check if a path is a file / directory is useful to start with

bjcscat:
Adds `fs.listdir` which produces a table that has the name and type of folder entries, `fs.remove` which removes files and empty directories, and `fs.type` which returns the type of a file.

Closes #82

### DIFF
--- a/fs/src/fs.cpp
+++ b/fs/src/fs.cpp
@@ -4,11 +4,11 @@
 #include "lualib.h"
 #include "uv.h"
 
-#include "lute/ref.h"
 #include "lute/runtime.h"
 
 #include <cstdio>
 #include <cstring>
+#include <fcntl.h>
 #include <map>
 #include <memory>
 #include <optional>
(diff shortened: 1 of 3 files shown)


fs/src/fs.cpp
```diff
@@ -245,6 +288,135 @@ void cleanup(char* buffer, int size, const FileHandle& handle)
     uv_fs_close(uv_default_loop(), &closeReq, handle.fileDescriptor, nullptr);
 }
 
+int fs_remove(lua_State* L)
+{
+    std::filesystem::remove(luaL_checkstring(L, 1));
```
aatxe: Does libuv really not have a function to remove files?

aatxe (approved):

fs/src/fs.cpp
bjcscat: It doesn't for some reason

aatxe:
```c
int uv_fs_unlink(uv_loop_t* loop, uv_fs_t* req, const char* path, uv_fs_cb cb);
```
unlink, no?

PR #90, opened 2025-02-26
Fix argument parsing: Properly handle `--` to separate files and script arguments

### LINKED ISSUE #86: CLI tries to read program args as files to execute
```luau
-- program.luau
print(...)
```

```
$ lute program.luau -- 1 2 3
1       2       3
Error opening ./1
Error opening ./2
Error opening ./3
```

unfetchable:
Previously, lute treated arguments after -- as files instead of script arguments. This caused errors like:
```sh
Error opening ./1  
Error opening ./2  
Error opening ./3  
``` 
since ```1 2 3``` were misinterpreted as file paths.

This PR does the following:
- Ensures -- properly marks the transition between files and arguments.
- Converts argv[i] from char* to std::string before pushing it into files. (Misc)

Closes #86

### DIFF
--- a/cli/main.cpp
+++ b/cli/main.cpp
@@ -201,7 +201,28 @@ int main(int argc, char** argv)
     program_argc = argc - program_args;
     program_argv = &argv[program_args];
 
-    const std::vector<std::string> files = getSourceFiles(argc, argv);
+    std::vector<std::string> files;
+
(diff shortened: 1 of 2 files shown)


aatxe (approved):

aatxe:
Thanks!

PR #95, opened 2025-02-27
Implement CLI Argument Parser

### LINKED ISSUE #83: Create std library for argument parsing
Similar to `clap` in rust or `argparse` in python

unfetchable:
Derived from [frkcli](https://github.com/itsfrank/frkcli), cc @itsfrank

Closes #83

This provides a lightweight, flexible CLI argument parser for Luau. It allows handling positional arguments, options, and flags, making it easier to build scripts that are dependent on the CLI.


itsfrank:
Very cool that my library was helpful! 

I don't have time to look too deeply into the code, but my initial thought is that we probably want a different name than `argparse`

I am certain you are aware, but [argparse](https://docs.python.org/3/library/argparse.html) is the standard CLI parser for python and is massively popular. It has been re-implemented in a variety of languages, some of which re-use the name.

I think that we should only opt to use that name if we are explicitly re-implementing `argparse`'s interface as faithfully as reasonably possible in Luau.

After a quick glance it does not seem that this implementation is trying to match `argparse`'s interface, thus I would opt for a different name to avoid confusion.

aatxe:
Yeah, that's a valid point, Frank! Maybe something like `cliargs` or something. I'm not _really_ trying to name things in very "novel" ways for `batteries` since the point is kinda to just be boring utilities that people will almost certainly end up writing better versions of in the future.

batteries/argparse.luau
```diff
@@ -0,0 +1,137 @@
+local argparse = {}
+argparse.__index = argparse
+
+type ArgKind = "POSITIONAL" | "FLAG" | "OPTION"
```
aatxe:
```suggestion
type ArgKind = "positional" | "flag" | "option"
```

I don't really see any reason for us to all-caps the kinds. Feels unidiomatic to me.

batteries/argparse.luau
```diff
@@ -0,0 +1,137 @@
...
+	options: ArgOptions,
+}
+
+type ParseResult = {
+	values: { [string]: string },
+	flags: { [string]: boolean },
+	fwdArgs: { string },
+}
+
+function argparse.new()
```
aatxe: missing return annotation

batteries/argparse.luau
```diff
@@ -0,0 +1,137 @@
...
+	local self = setmetatable({}, argparse)
+	self.arguments = {}
+	self.positional = {}
+	self.parsed = { values = {}, flags = {}, fwdArgs = {} }
+	return self
+end
+
+function argparse:add(name, kind, options)
```
aatxe: type annotations, which probably also will require us defining this as
```suggestion
function argparse.add(self, name, kind, options)
```

batteries/argparse.luau
```diff
@@ -0,0 +1,137 @@
...
+		options = options or { aliases = {}, required = false },
+	}
+	table.insert(self.arguments, argument)
+
+	if kind == "POSITIONAL" then
+		table.insert(self.positional, argument)
+	end
+end
+
+function argparse:parse(args)
```
aatxe: top-level type annotations

batteries/argparse.luau
```diff
@@ -0,0 +1,137 @@
...
+		if argument.options.required and not self.parsed.values[argument.name] then
+			assert(self.parsed.values[argument.name], "Missing required argument: " .. argument.name)
+		end
+		if not self.parsed.values[argument.name] and argument.options.default then
+			self.parsed.values[argument.name] = argument.options.default
+		end
+	end
+end
+
+function argparse:get(name)
```
aatxe: top-level type annotations

batteries/argparse.luau
```diff
@@ -0,0 +1,137 @@
...
+	return self.parsed.values[name]
+end
+
+function argparse:has(name)
```
aatxe: top-level type annotations

batteries/argparse.luau
```diff
@@ -0,0 +1,137 @@
...
+	return self.parsed.flags[name] ~= nil
+end
+
+function argparse:getFwdArgs()
```
aatxe: top-level type annotations

batteries/argparse.luau
```diff
@@ -0,0 +1,137 @@
...
+	return self.parsed.fwdArgs
+end
+
+function argparse:help()
```
aatxe: top-level type annotations

batteries/argparse.luau
```diff
@@ -0,0 +1,137 @@
+local argparse = {}
+argparse.__index = argparse
```
aatxe: Check out `time.luau` in `std` for how to do the typing for this file generally, I've already done it with `duration` there. Should be something like:

```
type ModuleData = {
    -- whatever fields we're calling the data for the core type here
}

type ModuleInterface = typeof(argparse)
export type Module = setmetatable<ModuleData, ModuleInterface>
```

batteries/argparse.luau
```diff
@@ -0,0 +1,137 @@
...
+	local i = 1
+	local pos_index = 1
+	while i <= #args do
+		local arg = args[i]
+		if arg == "--" then
+			for j = i + 1, #args do
+				table.insert(self.parsed.fwdArgs, args[j])
+			end
+			break
+		elseif string.sub(arg, 1, 2) == "--" then
```
aatxe:
```suggestion
		if arg == "--" then
			for j = i + 1, #args do
				table.insert(self.parsed.fwdArgs, args[j])
			end
			break
		end
		
		if string.sub(arg, 1, 2) == "--" then
```

aatxe: Above is an example, but generally, I'd like to rewrite this to use early-break and/or early-return rather than just a big if-elseif chain with nested conditionals under it.

batteries/argparse.luau
```diff
@@ -0,0 +1,137 @@
...
+			local name = string.sub(arg, 3)
+			local found = false
+			for _, argument in self.arguments do
+				local aliases = argument.options.aliases or {}
+				if argument.name == name or table.find(aliases, name) then
+					found = true
+					if argument.kind == "OPTION" then
+						i += 1
+						assert(not (i > #args), "Missing value for argument: " .. argument.name)
```
aatxe:
```suggestion
						assert(i <= #args, "Missing value for argument: " .. argument.name)
```

batteries/argparse.luau
```diff
@@ -0,0 +1,137 @@
...
+			for j = 1, #flags do
+				local name = string.sub(flags, j, j)
+				local found = false
+				for _, argument in self.arguments do
+					local aliases = argument.options.aliases or {}
+					if argument.name == name or table.find(aliases, name) then
+						found = true
+						if argument.kind == "OPTION" then
+							i += 1
+							assert(not (i > #args), "Missing value for argument: " .. argument.name)
```
aatxe:
```suggestion
							assert(i <= #args, "Missing value for argument: " .. argument.name)
```

batteries/argparse.luau
```diff
@@ -0,0 +1,137 @@
...
+			else
+				table.insert(self.parsed.fwdArgs, arg)
+			end
+		end
+		i += 1
+	end
+
+	for _, argument in self.arguments do
+		assert(argument)
+		if argument.options.required and not self.parsed.values[argument.name] then
```
aatxe: I think this should be probably be `self.parsed.values[argument.name] == nil` instead.

batteries/argparse.luau
```diff
@@ -0,0 +1,137 @@
...
+			assert(self.parsed.values[argument.name], "Missing required argument: " .. argument.name)
+		end
+		if not self.parsed.values[argument.name] and argument.options.default then
```
aatxe: I think this should be probably be `self.parsed.values[argument.name] == nil` instead.

batteries/argparse.luau
```diff
@@ -0,0 +1,137 @@
...
+
+function argparse:getFwdArgs()
+	return self.parsed.fwdArgs
+end
+
+function argparse:help()
+	print("Usage:")
+	for _, argument in self.arguments do
+		local aliasStr = table.concat(argument.options.aliases or {}, ", ")
+		local reqStr = argument.options.required and "(required) " or ""
```
aatxe:
```suggestion
		local reqStr = if argument.options.required then "(required) " else ""
```

aatxe (changes requested):

batteries/cliargs.luau
```diff
@@ -0,0 +1,149 @@
...
+}
+
+type ParseData = {
+	arguments: { [number]: ArgData },
+	positional: { [number] : ArgData},
+	parsed: { [number]: ParseResult },
+}
+
+type ParseInterface = typeof(cliargs)
+export type ParserType = setmetatable<ParseData, ParseInterface>
```
aatxe, line 31:
```suggestion
export type Parser = setmetatable<ParseData, ParseInterface>
```

The one that is actually exposed should have an "ordinary" type name, don't need to overcomplicate it.

batteries/cliargs.luau
```diff
@@ -0,0 +1,149 @@
...
+
+function cliargs.add(self: ParserType, name: string, kind: ArgKind, options: ArgOptions)
+	local argument = {
+		name = name,
+		kind = kind,
+		options = options or { aliases = {}, required = false },
+	}
+	table.insert(self.arguments, argument)
+
+	if kind == "POSITIONAL" then
```
aatxe:
```suggestion
	if kind == "positional" then
```

Missed this one.

batteries/cliargs.luau
```diff
@@ -0,0 +1,149 @@
...
+		end
+		
+		if string.sub(arg, 1, 2) == "--" then
+			local name = string.sub(arg, 3)
+			local found = false
+			for _, argument in self.arguments do
+				local aliases = argument.options.aliases or {}
+				if argument.name == name or table.find(aliases, name) then
+					found = true
+					if argument.kind == "OPTION" then
```
aatxe:
```suggestion
					if argument.kind == "option" then
```

Missed this one.

batteries/cliargs.luau
```diff
@@ -0,0 +1,149 @@
...
+		elseif string.sub(arg, 1, 1) == "-" then
+			local flags = string.sub(arg, 2)
+			for j = 1, #flags do
+				local name = string.sub(flags, j, j)
+				local found = false
+				for _, argument in self.arguments do
+					local aliases = argument.options.aliases or {}
+					if argument.name == name or table.find(aliases, name) then
+						found = true
+						if argument.kind == "OPTION" then
```
aatxe:
```suggestion
						if argument.kind == "option" then
```

batteries/cliargs.luau
```diff
@@ -0,0 +1,149 @@
...
+function cliargs.parse(self: ParserType, args: {string}): ()
+	local i = 1
+	local pos_index = 1
+	while i <= #args do
+		local arg = args[i]
+		
+		if arg == "--" then
+			for j = i + 1, #args do
+				table.insert(self.parsed.fwdArgs, args[j])
+			end
```
aatxe:
```suggestion
			table.move(args, i + 1, #args, 1, self.parsed.fwdArgs)
```

examples/cliargs.luau
```diff
@@ -0,0 +1,15 @@
...
+parser:add("verbose", "flag", { help = "Enable verbose mode", aliases = { "v" } })
+parser:add("output", "option", { help = "Output file", aliases = { "o" }, default = "default.txt" })
+
+local args = { ... }
+
+parser:parse(args)
+
+print("Input File:", parser:get("input"))        -- "input.txt"
+print("Verbose Mode:", parser:has("verbose"))   -- true
+print("Output File:", parser:get("output"))     -- "result.txt"
```
aatxe:
```suggestion
print("Output File:", parser:get("output"))     -- "result.txt"

```

Every file should have a trailing newline.

aatxe (changes requested):

deviaze:
I'd just name it `cli` for simplicity.

aatxe (approved):
I'll do the file rename and type rename myself after it's merged in, I guess.

PR #99, opened 2025-03-01
Support and run analysis on CI

JohnnyMorganz:
This PR:
- Introduces require support to the `lute --check` command. This is copied from `luau-analyze`
- Creates a set of definition files in a definition folders for C-based apis under `@lute`. This alias is registered in the `.luaurc` file
- Adds a CI check to run analysis on std, batteries and examples

A couple points:
- Decided to only support new solver (`LuauSolverV2` is force-enabled) for dogfooding purposes.

Still need to fix some type errors


vm/include/lute/vm.h
```diff
@@ -16,7 +16,7 @@ namespace vm
 int lua_defer(lua_State* L);
 
 static const luaL_Reg lib[] = {
-    {"create", lua_spawn},
```
aatxe: This was actually intentional. I wanted consistency with other `create` constructors for builtins in Luau.

definitions/vm.luau
```diff
@@ -0,0 +1,7 @@
+local vm = {}
+
+function vm.spawn(file: string): any
```
aatxe:
```suggestion
function vm.create(file: string): any
```

vm/include/lute/vm.h
```diff
@@ -16,7 +16,7 @@ namespace vm
...
+    {"spawn", lua_spawn},
```
aatxe:
```suggestion
    {"create", lua_spawn},
```

aatxe (changes requested):
Looks good to me, besides the nit about vm.create

aatxe:
Looks like the CI setup isn't quite right though?

JohnnyMorganz:
> Looks like the CI setup isn't quite right though?

Was waiting on the other PRs. A rebase fixed it now, but we do get a lot of type errors. Will try and fix some of them.

vrn-sn:
Saw this PR was closed. Are we good to delete the branch?

JohnnyMorganz:
Yeah should be fine (would be nice to get analyze running in CI someday)

PR #103, opened 2025-03-10
Fix error handling for scandir call

bjcscat:
Fixes problems faced when a directory may not exist or any other issue as thats not handled


fs/src/fs.cpp
```diff
@@ -418,9 +419,12 @@ int listdir(lua_State* L)
 
                         lua_settable(L, -3);
                     }
-
+
```
aatxe:
```suggestion

```

aatxe (approved):

aatxe (approved):

PR #104, opened 2025-03-14
Release workflow

### LINKED ISSUE #93: Set up nightly or per-commit builds with GH Actions
Since Lute is still very in-production, it'll probably be a bit before we cut proper releases, but it would be nice to set up a nightly or per-commit flow for adventurous folks to try out. This can potentially be done with a luthier subcommand (`luthier publish lute` or something) that runs in a GH action, for instance, or it can just be done all in the action script itself. Either way, shouldn't be too awful to set up!

Nicell:
This adds a release workflow that can either be run manually to create a new release based on the version in `get_version.cmake` or will be run nightly and will produce pre-releases with the format of `0.1.0-nightly.20250314`. 

See the example run [here](https://github.com/Nicell/lute/actions/runs/13851701027) and the resulting release [here](https://github.com/Nicell/lute/releases/tag/0.1.0-nightly.20250314). (Two slight adjustments made after this run happened, no longer installing cmake on macos since it's already there and changed `-nightly-20250314` to `-nightly.20250314`.

The output is confirmed working with foreman and rokit.

Closes #93

Note: For nightly releases, with foreman you can just put `0.1.0-nightly`, but on rokit, you need to have the full tag written including date.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,130 @@
+name: Build and Release
+
+on:
+  schedule:
+    - cron: '0 0 * * *'
+  workflow_dispatch:
+
+jobs:
+  build:
+    strategy:
+      matrix:
+        include:
+          - os: ubuntu-latest
+            artifact_name: lute-linux-x86_64
+            options: --c-compiler=clang --cxx-compiler=clang++
+          - os: ubuntu-24.04-arm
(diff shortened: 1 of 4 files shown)


aatxe (approved):

aatxe:
This looks great, thanks so much!

PR #105, opened 2025-03-15
Add HTTP Server to `net`

Nicell:
Adds an http server to `net`. This server is backed by uWebSockets, a very fast, standards-compliant HTTP/WebSocket server that powers bun also. The API isn't fully thought through yet, but this is a start. Other things to think about include:

- WebSockets
- TLS
- Routes
- Port reuse (and errors when port binding fails)
- Request and response body streaming

The above things seem like they can be follow up PRs since we don't have a stable API. For now this is a basic API we can play around with and build off of.


aatxe (changes requested):
Can you delete the manually vendored uWebSockets and add it using `git subrepo`? You'll have to install it as a separate tool, but it makes the management of this stuff much less painful for us longer term.

aatxe:
We have learned that git-subrepo does not play nicely with submodules and uWebSockets, naturally, uses submodules and Make. I think the right way forward here is to build a little bit of dependency management into `luthier`. I'm going to look into that first.

CMakeLists.txt
```diff
@@ -54,32 +59,42 @@ set(WOLFSSL_LIBRARY wolfssl)
...
 
-set(LIBUV_INCLUDE_DIR ${PROJECT_SOURCE_DIR}/extern/libuv/include)
-
 include(Sources.cmake)
 
 target_compile_features(Lute.Runtime PUBLIC cxx_std_17)
 target_compile_features(Lute.Fs PUBLIC cxx_std_17)
 target_compile_features(Lute.Luau PUBLIC cxx_std_17)
-target_compile_features(Lute.Net PUBLIC cxx_std_17)
+target_compile_features(Lute.Net PUBLIC cxx_std_20)
```
aatxe: Wait, does uWebSockets require C++20?

.github/workflows/ci.yml
```diff
@@ -58,6 +58,8 @@ jobs:
         if: runner.os == 'Windows'
         run: |
           choco install ninja
+          vcpkg install zlib
```
aatxe: Conflicts with your other PR, so this can be cleaned up now, I think.

aatxe (commented):
Looks good so far, going to look more carefully through the serve implementation tomorrow before final approval though.

CMakeLists.txt
Nicell: yes, but I've moved this requirement to uWebSockets.cmake, which is better, hopefully 😅

aatxe: cursed, but yeah, limiting it to just their thing is better, i don't really want to maintain code that's in c++20

net/src/net.cpp
```diff
@@ -89,6 +90,322 @@ int getAsync(lua_State* L)
     return lua_yield(L, 0);
 }
 
+static std::unordered_map<int, std::shared_ptr<uWS::App>> serverInstances;
+static std::unordered_map<int, std::shared_ptr<struct ServerLoopState>> serverStates;
```
aatxe: Can you use `DenseHashMap` from Luau's common library instead? It should be directly available already, and it is just strictly more performant. You'll need a tombstone value, but it seems like you're already starting at `1`, so `0` will work great.

net/src/net.cpp
```diff
@@ -89,6 +90,322 @@ int getAsync(lua_State* L)
     return lua_yield(L, 0);
 }
 
+static std::unordered_map<int, std::shared_ptr<uWS::App>> serverInstances;
```
aatxe: Can we actually do `std::unique_ptr` here and plain borrows (`uWS:App*`) in `ServerLoopState`? This avoids any atomics, and I think the memory model here is sane.

net/src/net.cpp
```diff
@@ -89,6 +90,322 @@ int getAsync(lua_State* L)
...
+{
+    // Check if the response is a string or a table
+    if (lua_isstring(L, responseIndex))
+    {
+        std::string body = lua_tostring(L, responseIndex);
+        res->writeStatus("200 OK");
+        res->writeHeader("Content-Type", "text/html");
+        res->end(body);
+    } 
+    else if (lua_istable(L, responseIndex))
```
aatxe:
```suggestion
    // Check if the response is a string or a table
    if (lua_isstring(L, responseIndex))
    {
        std::string body = lua_tostring(L, responseIndex);
        res->writeStatus("200 OK");
        res->writeHeader("Content-Type", "text/html");
        res->end(body);
        return;
    }
 
    if (lua_istable(L, responseIndex))
```

generally prefer early returns.

i would also move the `500` server error up to an

`if (!lua_istable(L, responseIndex))` early exit as well, and then keep the main chunk of the implementation as the main flow of the function

aatxe (commented):
Some minor feedback on some of the net.cpp implementation, but otherwise I think this is good to go.

aatxe (approved):

PR #116, opened 2025-03-20
Create composite action for build

### LINKED ISSUE #114: Create a composite shared action for ci.yml and release.yml
The installation and build is the same with the slight difference in debug vs. release. We can share this in a composite action inside the repo for easier maintenance and preventing divergence between release and pr runs. Also maybe rename ci.yml to pr.yml?

Nicell:
Closes #114 

Also adds caching of the extern folder so CI should be a bit faster by not having to fetch every dependency each run.

Nightly builds are failing right now but should be fixed by this PR too


aatxe (approved):
Great stuff, thanks!

PR #129, opened 2025-03-29
Process

### LINKED ISSUE #115: API: Process Library
The runtime should expose a process API, it'll make a lot of important work better, and immediately unlock dogfooding opportunities in #89 now that we have a cli argument parsing library. Ostensibly, `libuv` already abstracts over this stuff neatly and deals with the sorts of concerns I was worried about.

Nicell:
Adds a basic `process.create` method to create a subprocess. Supports shell, cwd, and env options. See `examples/process.luau` for some usage.

```luau
type ProcessOptions = {
    shell: boolean | string?,
    cwd: string?,
    env: { [string]: string }?,
}

type ProcessResult = {
    ok: boolean,
    exitcode: number,
    stdout: string,
    stderr: string,
    signal: string?,
}

type process = {
    create: (args: string | {string}, options: ProcessOptions?) -> ProcessResult,
}
```

Closes #115


process/src/process.cpp
```diff
@@ -0,0 +1,398 @@
...
+    int64_t exit_code = -1;
+    int term_signal = 0;
+    bool completed = false;
+    ResumeToken resume_token;
+    std::shared_ptr<ProcessHandle> self;
+    std::atomic<int> pending_closes{0};
+
+    ~ProcessHandle() {}
+
+    void close_handles()
```
aatxe: Annoying nitpicky stuff, but all internal stuff should be camelCase, only the exposed API is snake_case. This is carrying over the same C++ style as Luau itself.

aatxe (changes requested):
Generally looks good, just need to fix the casing of the names of internal things to match Luau's code style requirements.

aatxe (approved):

process/src/process.cpp
Nicell: All good, I totally forgot about this, fixed it all up I think

aatxe:
Gonna have to bump CMake compatibility in this one lol, you lost the roulette

PR #130, opened 2025-03-29
Process: env

Nicell:
Builds upon #129 by adding `process.env`. It has a metatable that implements `__index`, `__newindex`, and `__iter` to act as a proxy for the process's environment.

### DIFF
--- a/process/src/process.cpp
+++ b/process/src/process.cpp
@@ -371,7 +371,7 @@ int create(lua_State* L)
     return lua_yield(L, 0);
 }
 
-static int env_index(lua_State* L)
+static int envIndex(lua_State* L)
 {
     const char* key = luaL_checkstring(L, 2);
     char value[1024];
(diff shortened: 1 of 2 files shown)


process/src/process.cpp
```diff
@@ -371,8 +371,137 @@ int create(lua_State* L)
...
+
+    return 0;
+}
+
+typedef struct
+{
+    uv_env_item_t* items;
+    int count;
+    int index;
+} EnvIter;
```
aatxe:
```suggestion
struct EnvIter
{
    uv_env_item_t* items;
    int count;
    int index;
}
```

We're writing C++ 😉

aatxe (changes requested):

process/src/process.cpp
Nicell: still learning what a C++ is, thanks!

aatxe: it's funny because you didn't do this in the other one, but I guess you were reading more C sample code this time.

aatxe (approved):

PR #135, opened 2025-04-08
Ansi battery improvements

gaymeowing:
* Replace use of the concat operator in `replaceClose` with table.concat, as table.concat is faster than repeatedly doing
`str ..= "meow"`
* Freeze the return table


batteries/ansi.luau
```diff
@@ -1,27 +1,46 @@
-local function replaceClose(str: string, close: string, replace: string, index: number?): string
-	local result = ""
+local RESULT_TBL = {}
+local PREV_LEN = 0
```
aatxe: global state for a function like this is very wrong.

aatxe (changes requested):

aatxe (changes requested):
I don't want us to ignore the code formatter just because "it's better." The whole point of the code formatter is that we don't have to have discussions about what is or isn't better.

aatxe:
I think this got superseded by the other PR for ansi, but feel free to reopen if some of this optimization is still needed.

PR #136, opened 2025-04-08
Improve luthier's git version extraction on macOS

vrn-sn:
Apple Git's `git --version` output on my machine, which broke luthier: `git version 2.39.3 (Apple Git-145)`.

I tested on a few different platforms, and the beginning of the output of `git --version` is always consistent: `git version X.Y.Z`.


aatxe (approved):
good catch, forgot about modified versions because i use git from homebrew

PR #138, opened 2025-04-15
Move Lute to Luau's new Luau.Require library

vrn-sn:
This PR moves Lute to the `Luau.Require` Luau library, allowing us to upgrade our Luau dependency version as well.

There are some hacks involved to support `@lute/*` requires and the `vm.create` function, but these will be replaced with simpler solutions once we update `Luau.Require`.

At a high level, luthier now generates a C++ file containing the Luau source code of the `@std` libraries. When the `@std` alias is seen by require-by-string, we switch to navigating a new virtualized filesystem that only contains the `@std` libraries.

For now, we take a similar approach for `@lute` libraries, but these will have a better solution soon where we can simply register each `@lute/*` path to a Luau table result.


tools/luthier.py
```diff
@@ -351,6 +411,8 @@ def main(argv):
         if not projectPathExists(args):
             check(configure(args))
 
+        generateStdLibFiles()
```
aatxe: I would do this inside of `configure` tbf, probably before we actually call out to cmake configure. This is more about "prepping for a build" than actually building. Doing so probably also means we'll want to add a dirty check to `build` for the `std` files to regenerate them as necessary (but we can also skip doing the full configure in that case and only do the regenerate).

tools/luthier.py
```diff
@@ -326,12 +385,12 @@ def main(argv):
     if subcommand == "fetch":
         return fetchDependencies(args)
     elif subcommand == "configure" or subcommand == "tune":
+        generateStdLibFiles()
         return configure(args)
     elif subcommand == "build" or subcommand == "craft":
         # auto configure if it's not already happened
         if not projectPathExists(args):
```
aatxe, line 394:
```suggestion
        if not projectPathExists(args):
            generateStdLibFiles()
```

aatxe: needs to happen in forced configure too

tools/luthier.py
```diff
@@ -161,6 +161,65 @@ def projectPathExists(args):
...
 
+def generateStdLibFiles():
+    restoredPath = os.getcwd()
+
+    os.chdir(os.path.join(getSourceRoot(), 'std'))
+    os.makedirs("src/generated", exist_ok=True)
+
+    with open("src/generated/modules.cpp", "w") as cpp:
+        cpp.writelines([
+            "// This file is auto-generated by luthier.py. Do not edit.\n",
```
aatxe, line 172:
```suggestion
            "// This file is auto-generated by luthier.py. Do not edit.\n",
            "// Instead, you should modify the luau source files in `std/libs`.\n",
```

aatxe (approved):

vrn-sn:
We don't have C++ unit testing set up yet, but once we do, I want to copy over the Luau require-by-string tests over to here. For now, I just used this PR's version of lute to manually run each of the test `.luau` files in the Luau repository to ensure require-by-string still works as intended for normal filesystem requires.

aatxe (approved):

PR #140, opened 2025-04-16
Add module name to error message when an invalid require happens

ffrostfall:
QoL improvement for dynamic requires when the stack trace isn't really there


aatxe:
Can you rebase this on `primary`?

PR #142, opened 2025-04-16
Remove vcpkg from CI

Nicell:


aatxe:
This change is totally fine (good even), but we need to also change the MSVC version to VS2015 minimum, which I think was happening coincidentally from that other MS action we were using.

aatxe (approved):

Issue #144, opened 2025-04-17
Runtime API for spawning threads in a way that supports full error messages and continuations


ffrostfall:
Something like `task.resume(thread)`?

vegorov-rbx:
Can this be specified in more detail, along the lines of:
* 'suppose we do this to accomplish that'
* 'we get this'
* 'here's a problem with what we got'
* (optional, in this case your linked PR) 'this is the solution I came up with'

As it is right now, it's hard to follow along both the problem and the current solution.

ffrostfall:
I'm going to close this issue and open up multiple, as I've found after doing the task.resume PR that there are multiple issues which need attention. Sorry for the confusion!

PR #146, opened 2025-04-18
Implement task.resume

ffrostfall:
Closes issue #144 

This PR:
- Changes the scheduler to close the process when all threads are dead
  - It does this by not returning in `Runtime::runToCompletion()`. These returns have been removed.
- Abstracts error handling logic into `Runtime::reportError(lua_State* L)`
- Implements `task.resume<T...>(thread, T...)`. This propagates errors to the output.

### DIFF
--- a/runtime/include/lute/runtime.h
+++ b/runtime/include/lute/runtime.h
@@ -32,13 +32,15 @@ struct Runtime
     bool hasContinuations();
 
     void schedule(std::function<void()> f);
-
+    
     // Resume thread with the specified error
     void scheduleLuauError(std::shared_ptr<Ref> ref, std::string error);
 
     // Resume thread with the results computed by the continuation
     void scheduleLuauResume(std::shared_ptr<Ref> ref, std::function<int(lua_State*)> cont);
-
+    
+    void reportError(lua_State* L);
+    
     // Run 'f' in a libuv work queue
     void runInWorkQueue(std::function<void()> f);
 
--- a/runtime/include/lute/runtime.h
+++ b/runtime/include/lute/runtime.h
@@ -32,7 +32,6 @@ struct Runtime
     bool hasContinuations();
 
     void schedule(std::function<void()> f);
-    
     // Resume thread with the specified error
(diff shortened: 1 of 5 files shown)


ffrostfall:
finished implementation (Actually this time), ready for review/merge @aatxe

runtime/include/lute/runtime.h
```diff
@@ -32,13 +32,15 @@ struct Runtime
     bool hasContinuations();
 
     void schedule(std::function<void()> f);
-
+
```
aatxe:
```suggestion

```

runtime/src/runtime.cpp
```diff
@@ -66,7 +66,6 @@ bool Runtime::runToCompletion()
         if (L == nullptr)
         {
             fprintf(stderr, "Cannot resume a non-thread reference");
-            return false;
```
aatxe: Feels very wrong that this would not return here?

ffrostfall: yeah you're right

vegorov-rbx:
Can you explain why this function is needed?

If you create use `coroutine.create` you should be able to continue using other coroutine library functions to interact with it.
I would have expected some `task` library problem to be solved with this method, not a `coroutine` one.

This almost looks like `coroutine.wrap`, but not quite and just drops an error to the output without the ability to handle it (and pretends to the caller that everything went ok?).

Issue that you linked doesn't really specify the problem.

ffrostfall:
> Can you explain why this function is needed?
> 
> If you create use `coroutine.create` you should be able to continue using other coroutine library functions to interact with it. I would have expected some `task` library problem to be solved with this method, not a `coroutine` one.
> 
> This almost looks like `coroutine.wrap`, but not quite and just drops an error to the output without the ability to handle it (and pretends to the caller that everything went ok?).
> 
> Issue that you linked doesn't really specify the problem.

After learning more about the specifics of this, I've decided to close this PR as it doesn't solve the issue, and make a series of issues accurately & fully explaining the problems with the scheduler and what I hoped to solve with `task.resume`

Issue #145, opened 2025-04-18
Remove `setfenv` and `getfenv` from the builtin environment.


aatxe:
Lute has no reason to retain `setfenv` and `getfenv`. They've been deprecated in Luau itself, and we don't need to support them for backwards compatibilty reasons in standalone Lute.

PR #148, opened 2025-04-18
Remove setfenv & getfenv

ffrostfall:
Closed old PR because I messed up and forgot to switch to main before making another branch

Closes issue #145


aatxe (approved):
Technically think I would prefer not registering them in the first place, but this works.

Issue #117, opened 2025-03-21
API: System


Nicell:
This library would surface information about the host system like platform for platform-specific logic, CPU core/logical thread count for better parallelism, and more.

PR #149, opened 2025-04-19
System API for getting os, architecture, processor count

ffrostfall:
TODO:
- ~~Get system info for POSIX compliant systems~~
- ~~Bikeshed on the field name for thread/logical processor/core count (lol)~~

Closes issue #117


ffrostfall:
ready for review/merge

system/src/system.cpp
```diff
@@ -0,0 +1,98 @@
+#include "lute/system.h"
+#include "lua.h"
+#include "uv.h"
+#include <iterator>
+#include <string>
+#include <vector>
+#include <winnt.h>
```
aatxe: Windows-only header.

aatxe:
```suggestion
```

system/src/system.cpp
```diff
@@ -0,0 +1,98 @@
...
+
+int luaopen_system()
+{
+    return 0;
+}
```
aatxe: Implement this?

system/src/system.cpp
```diff
@@ -0,0 +1,98 @@
...
+
+int luteopen_system(lua_State* L)
+{
+    lua_createtable(L, 0, std::size(system_lib::lib));
```
aatxe: This initial size isn’t accurate since we’re adding additional properties below.

aatxe (changes requested):

system/src/system.cpp
ffrostfall: Just caught that

system/src/system.cpp
ffrostfall: So that'd be std::size(systemlib::lib) + 3, right

aatxe: Yes, although having something that lists the properties in the header would probably be nice tbh. A uniform structure is nice here, and also makes it easier to look at the headers to see what they’re exposing.

system/src/system.cpp
```diff
@@ -0,0 +1,104 @@
+#include "lute/system.h"
+#include "lua.h"
+#include "uv.h"
+#include <iterator>
+#include <string>
+#include <vector>
+
+int luaopen_system(lua_State* L)
+{
+    luaL_register(L, "system", system_lib::lib);
```
aatxe: This doesn't register the properties. The correct way to implement this is by calling the one that pushes the table, and then setting it into the appropriate global.

system/src/system.cpp
```diff
@@ -0,0 +1,104 @@
...
+        lua_setfield(L, -2, "model");
+
+        lua_pushinteger(L, (int)cpuInfo.speed);
+        lua_setfield(L, -2, "speed");
+
+        // sys, user, idle, irq, nice
+        lua_createtable(L, 0, 5);
+
+        // cast to double cuz int is 32bit and lua numbers are 52bit
+        lua_pushnumber(L, (double)cpuInfo.cpu_times.sys);
```
aatxe:
```suggestion
        lua_pushnumber(L, static_cast<double>(cpuInfo.cpu_times.sys));
```

system/src/system.cpp
```diff
@@ -0,0 +1,105 @@
...
+
+        auto cpuInfo = cpus[i];
+
+        // model, speed, times
+        lua_createtable(L, 0, 2);
+
+        lua_pushstring(L, cpuInfo.model);
+        lua_setfield(L, -2, "model");
+
+        lua_pushinteger(L, (int)cpuInfo.speed);
```
aatxe, line 68:
```suggestion
        lua_pushinteger(L, static_cast<int>(cpuInfo.speed));
```

system/src/system.cpp
```diff
@@ -0,0 +1,105 @@
...
+
+        lua_pushnumber(L, static_cast<double>(cpuInfo.cpu_times.irq));
+        lua_setfield(L, -2, "irq");
+
+        lua_pushnumber(L, static_cast<double>(cpuInfo.cpu_times.nice));
+        lua_setfield(L, -2, "nice");
+
+        lua_pushnumber(L, static_cast<double>(cpuInfo.cpu_times.user));
+        lua_setfield(L, -2, "user");
+
```
aatxe: Wait, why are these `pushnumber` and the earlier one `pushinteger`? Surely, this should all be `pushinteger` if these are all 32-bit integers?

ffrostfall: Yeah. Not sure what I was thinking.

ffrostfall: The comment was wrong, the code was right. The cpu times are uint64, I cast to double to maintain as much precision as possible.

aatxe (approved):

PR #153, opened 2025-04-20
Add homedir & cwd process APIs

ffrostfall:


process/src/process.cpp
```diff
@@ -527,6 +527,21 @@ int luteopen_process(lua_State* L)
...
+    size_t homedir_size = 255;
+
+    uv_cwd(cwd_buff, &cwd_size);
+    uv_os_homedir(homedir_buff, &homedir_size);
+
+    lua_pushstring(L, cwd_buff);
+    lua_setfield(L, -2, "cwd");
+
+    lua_pushstring(L, homedir_buff);
+    lua_setfield(L, -2, "homedir");
```
aatxe: Both of these should be functions, not properties. They change dynamically.

aatxe (changes requested):

process/src/process.cpp
```diff
@@ -374,6 +374,30 @@ int create(lua_State* L)
     return lua_yield(L, 0);
 }
 
+int homedir(lua_State* L)
+{
+    char homedir_buff[255];
+    size_t homedir_size = 255;
```
aatxe: I almost wonder if we should consider having an optional argument for the max length? On one hand, it’s frustratingly low-level, and it would also prevent us from e.g. implementing some automatic resizing or something, but on the other hand, the behavior of it just dying with 256 byte paths is pretty bad (if they’re UTF-16, which I think they might be on Windows, that’s only 128 characters).

process/src/process.cpp
```diff
@@ -374,6 +374,30 @@ int create(lua_State* L)
...
+
+    uv_os_homedir(homedir_buff, &homedir_size);
```
aatxe: Error handling?

process/src/process.cpp
```diff
@@ -374,6 +374,30 @@ int create(lua_State* L)
...
+
+    return 1;
+}
+
+int cwd(lua_State* L)
+{
+    char cwd_buff[255];
+    size_t cwd_size = 255;
+
+    uv_cwd(cwd_buff, &cwd_size);
```
aatxe: Error handling?

aatxe (changes requested):

process/src/process.cpp
ffrostfall: After looking at the docs again, it seems that libuv will return the required length if it's not enough.

aatxe (approved):

PR #159, opened 2025-04-20
Add result to @batteries

ffrostfall:
Many operations involving the file system & networking can fail. One of the most common ways of handling this is pcall, but this is not type safe.

A common idiom is to have a Result type, somewhat similar to Rust's Result type. It's an extremely common and useful idiom. Any project written in Lute interacting with APIs which could error will end up encountering the problem of gracefully handling failure; the result battery significantly helps with that.

### DIFF
--- a/batteries/result.luau
+++ b/batteries/result.luau
@@ -1,6 +1,4 @@
-type T<Value> =
-    | { success: true, value: Value }
-    | { success: false, traceback: string, err: string }
+type T<Value> = | { success: true, value: Value } | { success: false, traceback: string, err: string }
 
 local result = {}
 
@@ -21,6 +19,18 @@ function result.fail<Value>(err: string): T<Value>
 		traceback = traceback,
 		err = err,
 	}
+end
+
+function result.pcall<R, T...>(f: (T...) -> R, ...: T...): T<R>
+	local success, value = pcall(f, ...)
+
+	if success then
+		return result.ok(value)
+	else
+		local err = value :: any
+
+		return result.fail(err)
+	end
 end
 
 return result
--- a/batteries/result.luau
+++ b/batteries/result.luau
@@ -1,0 +1,31 @@
+type T<Value> = {
+	success: true,
+	value: Value,
+} | {
+	success: false,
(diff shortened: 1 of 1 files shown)


batteries/result.luau
```diff
@@ -0,0 +1,31 @@
+type T<Value> = {
+	success: true,
+	value: Value,
+} | {
+	success: false,
+	traceback: string,
+	err: string,
+}
```
aatxe:
```suggestion
type T<Value> =
    | { success: true, value: Value }
    | { success: false, traceback: string, err: string }
```

batteries/result.luau
```diff
@@ -0,0 +1,31 @@
+type T<Value> = {
+	success: true,
+	value: Value,
+} | {
+	success: false,
+	traceback: string,
+	err: string,
```
aatxe: Do we want this over a form where `err`'s type can be specified?

aatxe (changes requested):
PR description?

batteries/result.luau
ffrostfall: Yes, kind of? The reason why is so that result failures can be appended. For example:
```luau
local function someOperation()
  return result(false, "failed to do generic operation")
end

local function meow()
  local call = someOperation()
  if not call then
    call.err = `failure in function "meow"; {call.err}`
  end
end

local function meow2()
  local call = someOperation()
  if not call then
    call.err = `failure in function "meow2"; {call.err}`
  end
end
```
err as string is *great* for propagating messages. Maybe we could attach something like `errMessage` and `err` to have both? I can see the value in, say, an error tagged enum.

batteries/result.luau
```diff
@@ -0,0 +1,26 @@
+type T<Value> =
+    | { success: true, value: Value }
+    | { success: false, traceback: string, err: string }
```
aatxe: stylua doesn't like this even though i think it's the nicest way to write this 😦

aatxe (approved):

aatxe (approved):

PR #160, opened 2025-04-20
Update ansi.luau 

ffrostfall:
- ansi.luau -> colorful.luau
- NO_COLOR support
- export Styler type
- add `combine(...)`, looks like `combine(colorful.red, colorful.bold)("This is red bold text!")`
- update to use string interpolation instead of concatenation in some places


batteries/colorful.luau
```diff
@@ -1,27 +1,69 @@
+local process = (require)("@lute/process") -- temporary ()
+
+export type Styler = (text: string) -> string
+
+-- https://no-color.org/
+local function colorDisabled(): boolean
+	-- is it really okay to index env every style call?
+	return process.env.NO_COLOR ~= nil
```
Nicell: I'd suggest looking at how picocolors handles NO_COLOR. it also returns a generator function where you can explicitly specify if you want colors or not (but default returns the inferred preference) https://github.com/alexeyraspopov/picocolors/blob/main/picocolors.js

examples/ansi.luau
```diff
@@ -5,4 +5,7 @@ print(ansi.bgRed("This is red background!"))
 print(ansi.bold("This is bold text!"))
 print(ansi.underline("This is underlined text!"))
 print(ansi.dim(ansi.italic("This is dim italic text!")))
+
 print(ansi.red(`red({ansi.blue(`blue({ansi.green("green")})`)})`))
+
+print(ansi.combine(ansi.red, ansi.bold)("This is bold red text!"))
```
Nicell: isn't this more verbose than `ansi.red(ansi.bold("This is bold red text!"))`

batteries/colorful.luau
```diff
@@ -1,27 +1,69 @@
...
+
+	local function created_formatter(str: string): string
+		if colorDisabled() then
+			return str
 		end
+
+		local index = searchForSubstringIn(str, close, #open + 1)
+		local encasedString = if index then replaceClose(str, close, replaceCloseWith, index) else str
+
+		return `{open}{encasedString}{close}`
```
Nicell: just a discussion point, why do we prefer string interpolation here? I don't personally find it easier to read and only ever use it when mixing with inline strings, but this doesn't have that.

ffrostfall: To be honest, I've always just kind of had this idea that it was better but never really thought about why. I don't think it matters. Personally I find it more readable but that's probably because of my own consistency

examples/ansi.luau
ffrostfall: It is, this is an artifact of using examples as testing.

batteries/colorful.luau
aatxe: I think I prefer just concat for things that are literally just concat.

batteries/colorful.luau
aatxe: Yeah, I think it'd be reasonable to return a function here, e.g.

```lua
function(colorsEnabled: boolean?)
    local colorsEnabled = colorsEnabled or (process.env.NO_COLOR ~= nil)

    return ...
end
```

examples/ansi.luau
```diff
@@ -5,4 +5,7 @@ print(ansi.bgRed("This is red background!"))
 print(ansi.bold("This is bold text!"))
 print(ansi.underline("This is underlined text!"))
 print(ansi.dim(ansi.italic("This is dim italic text!")))
+
 print(ansi.red(`red({ansi.blue(`blue({ansi.green("green")})`)})`))
```
aatxe: All of these are still referring to `ansi` but you renamed the module.

batteries/colorful.luau
```diff
@@ -61,52 +51,70 @@ local function combine(...: Styler): Styler
...
-	bgBlackBright = formatter("\x1b[100m", "\x1b[49m"),
-	bgRedBright = formatter("\x1b[101m", "\x1b[49m"),
-	bgGreenBright = formatter("\x1b[102m", "\x1b[49m"),
-	bgYellowBright = formatter("\x1b[103m", "\x1b[49m"),
-	bgBlueBright = formatter("\x1b[104m", "\x1b[49m"),
-	bgMagentaBright = formatter("\x1b[105m", "\x1b[49m"),
-	bgCyanBright = formatter("\x1b[106m", "\x1b[49m"),
-	bgWhiteBright = formatter("\x1b[107m", "\x1b[49m"),
-}
+local function colorful(colorEnabled: boolean)
```
aatxe: Shouldn't this be optional?
```suggestion
local function colorful(colorEnabled: boolean?)
```

aatxe (changes requested):

aatxe (approved):

PR #171, opened 2025-04-22
Create Luau version of Luthier

### LINKED ISSUE #89: Dogfooding: Rewrite Luthier in Lute.
There's still some additional work necessary for this to be possible (argparse #83, process API), but I want to put this down in an issue as an obvious thing we want to do.

Nicell:
Port Luthier to Luau. The only piece fully lacking is support for directories in `std`. This also highlights that we're lacking `os.walk` and `os.chdir` equivalents, so I had to work around that.

Closes #89


tools/luthier.luau
```diff
@@ -0,0 +1,373 @@
+local process = require("@lute/process")
+local system = require("@lute/system")
+local fs = require("@lute/fs")
+local cli = require("@batteries/cli")
```
aatxe, line 6:
```suggestion

local cli = require("@batteries/cli")
```

I am oddly picky about this.

tools/luthier.luau
```diff
@@ -0,0 +1,373 @@
...
+		check(
+			call(
+				{ "git", "submodule", "update", "--init", "--recursive", "--depth=1" },
+				projectRelative("extern", dependency.name)
+			)
+		)
+		return result
+	end
+
+	if gitVersionInfo.major >= 2 and gitVersionInfo.minor >= 49 then
```
aatxe: This is buggy logic, it actually doesn't work when version is 3 or higher. The correct logic is:
```suggestion
	if gitVersionInfo.major >= 3 or (gitVersionInfo.major == 2 and gitVersionInfo.minor >= 49) then
```

aatxe (changes requested):
Overall, looks pretty good, just a couple things.

tools/luthier.luau
Nicell: good catch, thanks

aatxe (approved):

tools/luthier.luau
```diff
@@ -0,0 +1,374 @@
...
+args:add("config", "option", { help = "configuration (default is debug)", default = "debug" })
+args:add("clean", "flag", { help = "perform a clean build" })
+args:add("which", "flag", { help = "print out the path to the compiled binary and exit", aliases = { "w" } })
+args:add("cxx-compiler", "option", { help = "C++ compiler to use" })
+args:add("c-compiler", "option", { help = "C compiler to use" })
+
+if isWindows then
+	args:add("vs2017", "flag", { help = "Build with vs2017 (default is 2019)" })
+	args:add("vs2022", "flag", { help = "Build with vs2022 (default is 2019)" })
+end
```
```suggestion
```

tools/luthier.luau
```diff
@@ -0,0 +1,374 @@
...
+
+local function getCompiler(): string
+	if isMac then
+		return "xcode"
+	elseif isWindows and args:has("vs2017") then
+		return "vs2017"
+	elseif isWindows and args:has("vs2022") then
+		return "vs2022"
+	else
+		return "vs2019"
```
```suggestion
	else
		return "vs2022"
```

aatxe (changes requested):
Based on the inclusion of BoringSSL, we have a hard requirement on VS 2022 or newer now.

.github/actions/setup-and-build/action.yml
```diff
@@ -56,20 +54,20 @@ runs:
     - name: Fetch Lute
       if: steps.cache-extern.outputs.cache-hit != 'true'
       shell: bash
-      run: python ./tools/luthier.py fetch lute
+      run: lute ./tools/luthier.luau -- fetch lute
```
Nicell:
```suggestion
      run: lute ./tools/luthier.luau fetch lute
```

.github/actions/setup-and-build/action.yml
```diff
@@ -56,20 +54,20 @@ runs:
...
 
     - name: Configure Lute
       shell: bash
-      run: python ./tools/luthier.py configure lute --config=${{ inputs.config }} ${{ inputs.options }}
+      run: lute ./tools/luthier.luau -- configure lute --config ${{ inputs.config }} ${{ inputs.options }}
```
Nicell:
```suggestion
      run: lute ./tools/luthier.luau configure lute --config ${{ inputs.config }} ${{ inputs.options }}
```

.github/actions/setup-and-build/action.yml
```diff
@@ -56,20 +54,20 @@ runs:
...
 
     - name: Build Lute
       shell: bash
-      run: python ./tools/luthier.py build lute --config=${{ inputs.config }} ${{ inputs.options }}
+      run: lute ./tools/luthier.luau -- build lute --config ${{ inputs.config }} ${{ inputs.options }}
```
Nicell:
```suggestion
      run: lute ./tools/luthier.luau build lute --config ${{ inputs.config }} ${{ inputs.options }}
```

.github/actions/setup-and-build/action.yml
```diff
@@ -56,20 +54,20 @@ runs:
...
 
     - name: Get Artifact Path
       id: artifact_path
       shell: bash
       run: |
-        exe_path=$(python ./tools/luthier.py build lute --config=${{ inputs.config }} --which)
+        exe_path=$(lute ./tools/luthier.luau -- build lute --config ${{ inputs.config }} --which)
```
Nicell:
```suggestion
        exe_path=$(lute ./tools/luthier.luau build lute --config ${{ inputs.config }} --which)
```

tools/luthier.luau
```diff
@@ -0,0 +1,437 @@
...
+	})
+
+	local libsPath = projectRelative("std", "libs")
+	local libs = fs.listdir(libsPath)
+
+	for _, lib in libs do
+		local aliasedPath = `@std/{lib.name}`
+		local libSrc = fs.readfiletostring(joinPath(libsPath, lib.name))
+		fs.write(cpp, `\n    \{"{aliasedPath}", R"({libSrc})"},\n`)
+	end
```
Nicell, line 226: Doesn't support directories in std atm. No equivalent to os.walk makes this annoying, and it's unnecessary at the moment.

tools/luthier.luau
```diff
@@ -0,0 +1,447 @@
...
+local targetMap: { [string]: { exeName: string } } = {
+	lute = {
+		exeName = "lute",
+	},
+}
+
+local args = cli.parser()
+
+args:add("help", "flag", { help = "show the help message" })
+args:add("executable", "positional", { help = "the lute executable", required = true }) -- should this be handled automatically?
```
aatxe: This is the target name, really, not executable. The reason it's not automatic is because we will eventually have e.g. some unit and integration test targets as well.

aatxe (approved):
Looks good, with a comment.

tools/luthier.luau
Nicell: This was poorly named, it's the script name. The first argument passed in is the script itself to align with other runtimes (updated when I changed the subcommand stuff in cli). I've renamed this to "luthier" since it holds the luthier script name. The comment at the end is more like, should cliargs battery automatically add the first positional arg for you since it will always be the script name? I could probably make that clearer.

aatxe: Ahh, yeah, that should be fixed.

aatxe (approved):

PR #172, opened 2025-04-22
Add optional raw string display mode and fix non-primitive table key handling in pp function

### LINKED ISSUE #107: pp prints backslash in newlines
pp("\n") returns a backslash and then a newline instead of just a newline

### LINKED ISSUE #176: Pretty print does not print out table fields where keys are table references
```luau
local value = {
    [{}] = true
}

print(require("@batteries/pp")(value)
```
prints an empty table `{}` even though there exists a table key with value "true"

unfetchable:
Closes #107 
Closes #176


unfetchable (commented):
With this modification, you could call:

- `pp(data)` for the current behavior with escaped characters
- `pp(data, true)` for raw string representation without escaping

JohnnyMorganz:
Maybe worth future proofing by accepting an `options` table with `rawStrings: boolean` as an entry? Added benefit of giving a name to the option as well, rather than it being a straight bool

unfetchable:
I probably should have made the non-primitive table key change a different PR, whoops.
Anyways, with these changes, the pretty printer will now correctly display tables with non-primitive keys.

batteries/pp.luau
```diff
@@ -21,20 +21,28 @@ local function isPrimitiveArray(array: { [unknown]: unknown })
 	return n == max
 end
 
-local function formatValue(value: unknown)
-	if typeof(value) ~= "string" then
+local function formatValue(value: unknown, options: { [string]: unknown }?)
```
aatxe: you should define an exported type `PrettyPrintOptions` and it should have the actual list of options currently supported, and that should be the type of this, not `{[string]: unknown}`.

aatxe (changes requested):

batteries/pp.luau
```diff
@@ -1,5 +1,9 @@
 local isPrimitiveType = { string = true, number = true, boolean = true }
 
+type existingOptions = {
+	["rawStrings"]: boolean?,
```
```suggestion
	rawStrings: boolean?,
```

batteries/pp.luau
```diff
@@ -1,5 +1,9 @@
 local isPrimitiveType = { string = true, number = true, boolean = true }
 
+type existingOptions = {
```
```suggestion
type ExistingOptions = {
```

aatxe (changes requested):

batteries/pp.luau
```diff
@@ -21,20 +25,28 @@ local function isPrimitiveArray(array: { [unknown]: unknown })
 	return n == max
 end
 
-local function formatValue(value: unknown)
-	if typeof(value) ~= "string" then
+local function formatValue(value: unknown, options: existingOptions?)
```
aatxe: Every single reference to the type _also_ needs to be updated.

aatxe (changes requested):

aatxe (approved):

PR #182, opened 2025-04-23
Switch from wolfssl to boringssl.

aatxe:
We're switching to boringssl since it has better licensing and is known to use formally verified cryptographic protocols which is very important to me.


CMakeLists.txt
```diff
@@ -55,14 +50,13 @@ set(BUILD_CURL_EXE OFF)
...
-set(CURL_USE_WOLFSSL ON)
-set(WOLFSSL_INCLUDE_DIR ${PROJECT_SOURCE_DIR}/extern/wolfssl/include)
-set(WOLFSSL_LIBRARY wolfssl)
+set(HAVE_BORINGSSL ON)
 
 add_subdirectory(extern/curl)
 
 # uWebSockets setup
-set(WITH_WOLFSSL ON)
+set(WITH_WOLFSSL OFF)
```
Nicell, line 58:
```suggestion
```
I think uWebsockets "prefers" boringssl, so just having that on should work

extern/uWebSockets.cmake
```diff
@@ -23,7 +23,7 @@ option(WITH_LIBDEFLATE "Build with libdeflate support" OFF)
 option(WITH_ZLIB "Build with zlib support" ON)
 option(WITH_PROXY "Build with PROXY Protocol v2 support" OFF)
 option(WITH_QUIC "Build with QUIC support" OFF)
-option(WITH_BORINGSSL "Build with BoringSSL support" OFF)
+option(WITH_BORINGSSL "Build with BoringSSL support" ON)
```
Nicell, line 26: is this necessary if you're turning it on in cmakelists?

aatxe: No, it's not, it's just changing the default, I got that from @ffrostfall

aatxe:
Closing this to start over from scratch.

PR #186, opened 2025-04-23
Create super basic docs site

Nicell:
Simple docs site made with vitepress. Should auto-deploy to github pages when pushed to `primary`.

image


aatxe (approved):
It is a place to start, I'll take it.

PR #199, opened 2025-04-24
Add stdio option to process.run

Nicell:
We need a way to inherit stdout/stderr with `process.run` to finish #171.

This adds support for `default` `inherit` and `none` options for `stdio`. Default buffers the output and returns stdout/stderr in the result table. Inherit uses the parent process stdout/stderr (needed for luthier). None just ignores stdout/stderr from the created process.

In the future it'd be great to:
- Add a `forward` stdio kind that both buffers the output and forwards it to the parent process stdout/stderr
- Support different kinds for stdout vs stderr and also support a string/buffer in for stdin


process/src/process.cpp
```diff
@@ -339,12 +354,31 @@ int run(lua_State* L)
...
+    stdio[0].flags = UV_IGNORE;
+    if (stdioKind == kStdioKindNone)
+    {
+        stdio[1].flags = UV_IGNORE;
+        stdio[2].flags = UV_IGNORE;
+    }
+    else if (stdioKind == kStdioKindInherit)
+    {
+        stdio[1].flags = UV_INHERIT_FD;
+        stdio[1].data.fd = 1;
```
aatxe: should we use `stdout` instead of hardcoding it here?

process/src/process.cpp
```diff
@@ -339,12 +354,31 @@ int run(lua_State* L)
...
+        stdio[2].flags = UV_INHERIT_FD;
+        stdio[2].data.fd = 2;
```
aatxe: ditto `stderr`?

aatxe (changes requested):

aatxe (approved):

PR #202, opened 2025-04-24
CST: Handle if statements (excluding elseif) and nils

JohnnyMorganz:
This PR adds support for serializing a full if statement as a CST node. For now, we exclude special handling of elseif blocks within if statements, this will be added in a future change.

Alongside this, we serialize `nil` expressions directly as a token. This allows us to successfully roundtrip the `writeFile.luau` example.

### DIFF
--- a/batteries/syntax/ast_types.luau
+++ b/batteries/syntax/ast_types.luau
@@ -43,6 +43,8 @@ export type AstLocal = {
 	name: Token<string>,
 }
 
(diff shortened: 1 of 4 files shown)


aatxe (approved):
Looks good, though at the end of all of this, we may want to make all the keys explicitly _not_ keywords.

JohnnyMorganz:
Yup agreed, the ergonomics of it is quite poor

PR #205, opened 2025-04-24
Generate SHA-256 hash for `std/libs` libraries to track stale generated files

vrn-sn:
The generated hash is stored in the .gitignored `std/src/generated/hash.txt` file. Files are processed in sorted order to keep the hash deterministic. In addition to providing each file's contents to the hash function, we also provide each file's path within the `std/libs` directory.

### DIFF
--- a/tools/luthier.py
+++ b/tools/luthier.py
@@ -257,8 +257,8 @@ def generateStdLibFilesIfNeeded():
             f"extern const std::pair<std::string_view, std::string_view> lutestdlib[{numItems}];\n",
         ])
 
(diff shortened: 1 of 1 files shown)


tools/luthier.py
```diff
@@ -7,6 +7,7 @@
 import sys
 
 from os import path
+from hashlib import sha256
```
aatxe: can you replace this with blake2b instead? it's a better hash _and_ it's more performant.

vrn-sn: Just made the change.

aatxe (approved):

PR #213, opened 2025-04-26
TLS on net.serve

### LINKED ISSUE #125: net: TLS on `net.serve`
An optional cert and key (and cert authorities?) should be able to be provided through `net.serve`'s configuration. If they're provided, the uWebSockets SSL App should be used instead of the default App.

Nicell:
Implements TLS on `net.serve` using `SSLApp`. `App` and `SSLApp` are unique templated classes. We can support both by storing them in a variant when required.

```luau
type TLS = {
  cert_file_name: string, -- fullchain certificate file path
  key_file_name: string, -- key file path
  passphrase: string?, -- optional passphrase for cert and key
  ca_file_name: string?, -- optional certificate authority file path for mTLS, defaults to no mTLS
}
```

Tested locally, seems to work as expected 😎 closes #125

### DIFF
--- a/net/src/net.cpp
+++ b/net/src/net.cpp
@@ -13,7 +13,6 @@
 #include <string>
 #include <utility>
 #include <vector>
-#include <variant>
 
 namespace net
 {
--- a/net/src/net.cpp
+++ b/net/src/net.cpp
@@ -12,6 +12,7 @@
 #include <string>
 #include <utility>
 #include <vector>
+#include <variant>
 
 namespace net
 {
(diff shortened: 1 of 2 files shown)


net/src/net.cpp
```diff
@@ -98,13 +99,13 @@ int getAsync(lua_State* L)
 }
 
 static const int kEmptyServerKey = 0;
-static Luau::DenseHashMap<int, std::unique_ptr<uWS::App>> serverInstances(kEmptyServerKey);
+static Luau::DenseHashMap<int, std::variant<std::unique_ptr<uWS::App>, std::unique_ptr<uWS::SSLApp>>> serverInstances(kEmptyServerKey);
```
aatxe: Use `Luau::Variant` instead of `std::variant`

aatxe: I'd also say you should define an alias for it as well.

aatxe (changes requested):

aatxe (approved):

Issue #166, opened 2025-04-21
Future-proof API for spawning threads


ffrostfall:
If any API ever ends up wanting to spawn a blocking C-invoked function (https://github.com/aatxe/lute/issues/164), then `coroutine.resume` will evaporate continuations. This isn't good behavior. In this example:
```luau
print("start")

local thread: thread

delay(1, function()
	coroutine.resume(thread)
end)

-- async, blocking C-sided invoke
runtimeApi.invoke(function()
	thread = coroutine.running()
	coroutine.yield()
end)

-- This will never print
print("finish")
```
If the `wait` function utilizes coroutine.resume, then it will not print finish. This isn't good!

Alongside this, coroutine.resume has some funky behavior for error reporting. It will not report errors on threads which have yielded, but the runtime will:
```luau
local function f()
	if math.random(1, 2) == 2 then
		task.defer()
	end

	error("meow")
end

-- Has a 50/50 chance of printing true (it ran, and yielded, then the runtime reported the error),
-- or printing false and hiding the error!
coroutine.resume(f)
```

While it is possible to handle reporting errors via coroutine.resume, it is cumbersome, and many times it's more desirable to just have the runtime report errors in threads for you.

ffrostfall:
This does not need to be in the `task` library, but rather it could be implemented in something like `@std/coroutine` or @std/task`.

Since Luau's API will soon support yielding and thread-level continuations, the only logic left is error reporting - that can be done on the Luau side.

PR #218, opened 2025-04-28
Implement task.resume

ffrostfall:
Implements issue #166 by implementing an error reporter method on the Runtime class, and implementing `task.resume(thread, ...)`.


task/src/task.cpp
```diff
@@ -11,6 +16,36 @@ int lua_defer(lua_State* L)
...
+        return 1;
+    };
+
+    lua_remove(L, 1);
+
+    int args = lua_gettop(L);
+    lua_xmove(L, thread, args);
+
+    int resumptionStatus = lua_resume(thread, L, args);
+    if ((resumptionStatus != LUA_OK) && (resumptionStatus != LUA_YIELD) && (resumptionStatus != LUA_BREAK))
```
aatxe:
```suggestion
    if (resumptionStatus != LUA_OK && resumptionStatus != LUA_YIELD && resumptionStatus != LUA_BREAK)
```

task/src/task.cpp
```diff
@@ -1,7 +1,12 @@
 #include "lute/task.h"
 
+#include "lua.h"
+#include "lute/ref.h"
 #include "lute/runtime.h"
 
+// taken from extern/luau/VM/lcorolib.cpp
+static const char* const statnames[] = {"running", "suspended", "normal", "dead", "dead"};
```
aatxe, line 8: Is this not exposed? If so, is there an API exposed already for getting the name for errors? cc @vegorov-rbx

aatxe (approved):
We can refactor the array of states later if need be.

Issue #75, opened 2025-02-20
API: Instants in `@std/time`


aatxe:
We should add an implementation of monotonic instants in `@std/time`. I think Luau's built-in os time already provides a monotonic clock, but if not, we'll need to provide a monotonic clock API in lute using [`clock_gettime`](https://linux.die.net/man/3/clock_gettime) to use for this.

aatxe:
`os.time` is not monotonic, so we will need a `@lute/time` library in the runtime to provide the primitive needed to make an `Instant`.

PR #220, opened 2025-04-28
Introduce @lute/time runtime library

ffrostfall:
Closes issue #75 by introducing a time library.

Includes a `time.now()` function, which uses libuv's cross-platform `clock_gettime` function for instants with nanosecond precision. Since libuv's return result is `sec` + `nsec`, we cast both into doubles and then divide down.

`time` isn't available as a namespace name due to conflicts in other include statements, so the namespace is `libtime`.


time/src/time.cpp
```diff
@@ -0,0 +1,47 @@
...
+{
+int lua_now(lua_State* L)
+{
+    uv_timespec64_t timespec;
+
+    int status = uv_clock_gettime(UV_CLOCK_MONOTONIC, &timespec);
+    assert(status == 0);
+
+    // both of these are ints, so we cast them to double.
+    lua_pushnumber(L, static_cast<double>(timespec.tv_sec) + static_cast<double>(timespec.tv_nsec / 1e9));
```
aatxe: We should not return an instant as a number to luau code. The goal here is to provide a userdata for the timespec instead, and the necessary functionality to support it interacting with durations appropriately.

aatxe (changes requested):

time/src/time.cpp
ffrostfall: It is now a userdata

time/include/lute/time.h
```diff
@@ -0,0 +1,58 @@
...
+#include "lualib.h"
+#include <string>
+
+// open the library as a standard global luau library
+int luaopen_time(lua_State* L);
+// open the library as a table on top of the stack
+int luteopen_time(lua_State* L);
+
+static const char kInstantType[] = "Instant";
+static const char kDurationType[] = "Duration";
```
aatxe:
```suggestion
static const char kInstantType[] = "instant";
static const char kDurationType[] = "duration";
```

time/src/time.cpp
```diff
@@ -0,0 +1,559 @@
...
+{
+int lua_nanoseconds(lua_State* L)
+{
+    // int32_t doesn't have enough precision for nanoseconds
+    int64_t nanoseconds = static_cast<int64_t>(luaL_checknumber(L, 1));
+    if (nanoseconds < 0)
+    {
+        luaL_error(L, "duration cannot be negative");
+        return 0;
+    }
```
aatxe:
```suggestion
        luaL_error(L, "duration cannot be negative");
```

`luaL_error` is noret.

time/src/time.cpp
```diff
@@ -0,0 +1,559 @@
...
+}
+
+int lua_microseconds(lua_State* L)
+{
+    double microseconds = luaL_checknumber(L, 1);
+    if (microseconds < 0)
+    {
+        luaL_error(L, "duration cannot be negative");
+        return 0;
+    }
```
aatxe: ditto.

time/src/time.cpp
```diff
@@ -0,0 +1,559 @@
...
+namespace libtime
+{
+int lua_now(lua_State* L)
+{
+    uv_timespec64_t now;
+
+    int status = uv_clock_gettime(UV_CLOCK_MONOTONIC, &now);
+    assert(status == 0);
+
+    uv_timespec64_t* timespec = static_cast<uv_timespec64_t*>(lua_newuserdatatagged(L, sizeof(uv_timespec64_t), 1));
```
aatxe: define the tag in a constant in this file, we will refactor all those constants into a runtime header later, e.g.
`kInstantTag`

time/src/time.cpp
```diff
@@ -0,0 +1,559 @@
...
+// returns the address of the timespec from the duration on the stack
+static uv_timespec64_t getTimespecFromDuration(lua_State* L, int idx)
+{
+    return *static_cast<uv_timespec64_t*>(luaL_checkudata(L, idx, kDurationType));
+}
+
+// creates a userdata, and returns a fresh timespec pointer to it
+static int createDurationFromTimespec(lua_State* L, uv_timespec64_t timespec)
+{
+    uv_timespec64_t* duration = static_cast<uv_timespec64_t*>(lua_newuserdatatagged(L, sizeof(uv_timespec64_t), 2));
```
aatxe: Likewise, constant for `kDurationTag`.

aatxe (commented):
Overall, looks pretty good, just needs a little more cleanup.

aatxe (approved):

PR #222, opened 2025-04-28
Enable IPO for Release builds

Nicell:
Adds extra optimization and creates smaller binaries. Understood if we want to wait on this until we have a test suite to make sure the final build still functions as expected.


Nicell:
This currently causes a linking error on ubuntu clang presumably because `ld` is being used. I'll look into this more later.

PR #227, opened 2025-04-29
Remove restriction on the type of require results.

aatxe:
We originally implemented a restriction on require results being a table or function so that scripts would intrinsically be forwards-compatible as they evolve. To my surprise, this is a controversial choice. As part of the as-of-yet-unwritten style guide of Lute, we'll never return anything that is not a table or function, but users are free to do how they see fit.


Dekkonot:
I'm normally not inclined to police tone, but I really do not appreciate the way you've communicated here as if you're _better_ than the users who gave feedback for a decision you made. It's disrespectful at best, and it's downright demeaning at worst.

PR #230, opened 2025-04-29
CST: Handle named function declarations

JohnnyMorganz:
Reapplication of https://github.com/luau-lang/lute/pull/225, looks like it was merged into the assignment branch and I accidentally force pushed over it

---

This PR extends the CST support to handle named function declarations (function name() or local function name()).

As part of this PR, we also correct the comma positioning that separates arguments in the function definition, by using the comma positions stored on the CST node.


JohnnyMorganz:
Going to merge this as the previous PR already had an approval

PR #233, opened 2025-04-29
Add a basic testing solution to Lute

ffrostfall:


cli/commands/test.luau
```diff
@@ -0,0 +1,56 @@
...
+		end
+	end
+
+	return paths
+end
+
+local function loadTestFile(path: string)
+	local contents = fs.readfiletostring(path)
+	local compiled = luau.compile(contents)
+	local fileFunction = luau.load(compiled)
```
vrn-sn, line 32: I'm not sure if I understood the issue you mentioned, but this might be a fix? Is `path` an absolute path?
```suggestion
	local fileFunction = luau.load(compiled, `@{path}`)
```

PR #236, opened 2025-04-30
Add robust error handling to HTTP server with port reuse support

### LINKED ISSUE #127: net: Properly handle ports on HTTP server
Right now, the HTTP server attempts to open on a port and does not error if it fails. This error should be added while still allowing for explicit port reuse for platforms that support it (Linux with SO_REUSEPORT).

unfetchable:
This PR introduces comprehensive error handling for the HTTP server implementation and adds support for port reuse on Linux systems.

Closes #127


net/src/net.cpp
```diff
@@ -597,4 +707,4 @@ int luteopen_net(lua_State* L)
     lua_setreadonly(L, -1, 1);
 
     return 1;
-}
+}
```
aatxe, line 710:
```suggestion
}

```

Do not strip the final line break.

net/include/lute/net.h
```diff
@@ -21,4 +21,4 @@ static const luaL_Reg lib[] = {
     {nullptr, nullptr},
 };
 
-} // namespace net
+} // namespace net
```
aatxe, line 24:
```suggestion
} // namespace net

```

Do not strip the final line break.

.vscode/settings.json
```diff
@@ -1,5 +1,63 @@
...
+        "stdexcept": "cpp",
+        "streambuf": "cpp",
+        "string": "cpp",
+        "string_view": "cpp",
+        "typeinfo": "cpp",
+        "unordered_map": "cpp",
+        "unordered_set": "cpp",
+        "vector": "cpp",
+        "algorithm": "cpp"
+    }
```
aatxe, line 62: I don't use VS Code, but what is this?

aatxe (changes requested):

.vscode/settings.json
Nicell: I think the cpp extension adds this automatically for some reason. Might be worth adding .vscode to gitignore.

net/src/net.cpp
```diff
@@ -57,6 +59,19 @@ int get(lua_State* L)
...
+        luaL_errorL(L, "network request failed: %s", error.c_str());
+
+    lua_pushlstring(L, data.data(), data.size());
+    return 1;
+}
+
+int getAsync(lua_State* L)
+{
+    std::string url = luaL_checkstring(L, 1);
+
```
Nicell, line 74: this should be reverted

net/src/net.cpp
```diff
@@ -237,7 +256,7 @@ static void processRequest(
     const std::string_view& body
 )
 {
-    lua_State* L = lua_newthread(state->runtime->GL);
+    lua_State* L = state->runtime->GL;
```
Nicell, line 259: ditto. seems like recent changes have been incorrectly merged

net/src/net.cpp
```diff
@@ -405,6 +481,13 @@ int lua_serve(lua_State* L)
         }
         lua_pop(L, 1);
 
+        lua_getfield(L, 1, "allow_port_reuse");
```
Nicell, line 484:
```suggestion
        lua_getfield(L, 1, "reuseport");
```
or `allowportreuse` to be luaucase

Nicell (changes requested):
I looked at this a bit more and it looks like uSockets just turns on reuse port by default with no simple configuration for it, which is kinda annoying. I don't think all the checks added here really help with the root problem I wanted to solve here, which is conditionally allowing port reuse on the actual socket being created. These checks added don't prevent another lute instance (or any program for that matter) from reusing the port randomly.

unfetchable:
I'll be taking some time to go over and rewrite this.
I'm not particularly satisfied with how this turned out, so I'll be taking a moment to see how I can properly go about these changes.
Thank you for your reviews

PR #243, opened 2025-05-02
Pretty Print Improvements

gaymeowing:
* Include table memory addresses when printing tables, useful for tables with circular references
* Add function return types
* Use `table.concat` for primitive arrays instead of repeatedly using the concat operator
* Use `table.insert` for populating the keys table instead of `[#t + 1] = v`
* Replace cases where `typeof` is used when `type` could be used instead
* Remove usage of the string metatable

### DIFF
--- a/batteries/pp.luau
+++ b/batteries/pp.luau
@@ -183,5 +183,5 @@ return function(data: unknown, options: ExistingOptions?): string
 	end
 
 	-- FIXME(luau): `{[unknown]: unknown} should be treated identically to `table`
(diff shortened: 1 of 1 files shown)


batteries/pp.luau
```diff
@@ -169,14 +174,14 @@ local function traverseTable(
 	return output
 end
 
-return function(data: unknown, options: ExistingOptions?)
-	options = if options then options else {} :: ExistingOptions
+return function(data: unknown, options: ExistingOptions?): string
+	options = options or {} :: any
```
```suggestion
	options = options or {} :: ExistingOptions
```

aatxe (approved):

PR #247, opened 2025-05-02
CST: Serialize and print EOF trivia properly

JohnnyMorganz:
The CST works by attaching trivia between tokens as trailing trivia of the previous token or leading trivia of the currently processed token. However, right now the CST does not serialize any trailing trivia of the file properly (e.g. whitespace such as newline at EOF, or files composed solely of comments). This is because we do not process a final token to attach leading trivia to.

This PR introduces a final "EOF" token that is composed solely of leading trivia at the end of the file. This is not a concept in the Luau AST itself, so we need to extend our parse result to carry the "eof" token separately, independent of the AstStatBlock.

An alternative approach could be to attach all trivia at the end of the file as trailing trivia of the last processed token. However, this ends up special casing this particular token and breaks the guarantee that the trailing trivia will only consist of trivia up to and including a single newline character. As such, downstream tooling would potentially need to handle final statement / expressions in files separately to all other statements / expressions if they cared about trivia processing (e.g., formatters). Also, this would not work for files consisting solely of trivia, as there is no token present.

### DIFF
--- a/batteries/syntax/parser.luau
+++ b/batteries/syntax/parser.luau
@@ -12,7 +12,18 @@ local function parseexpr(source: string): T.AstExpr
 	return luau.parseexpr(source)
 end
 
+export type ParseResult = {
+	root: T.AstStatBlock,
+	eof: T.Eof,
+}
+
+local function parsefile(source: string): ParseResult
+	local result = luau.parse(source)
+	return { root = result.root, eof = result.eof }
+end
+
 return {
 	parse = parse,
 	parseexpr = parseexpr,
+	parsefile = parsefile,
 }
--- a/batteries/syntax/ast_types.luau
+++ b/batteries/syntax/ast_types.luau
@@ -36,6 +36,8 @@ export type Token<Kind = string> = {
 	read trailingTrivia: { Trivia },
 }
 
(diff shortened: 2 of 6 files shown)


batteries/syntax/parser.luau
```diff
@@ -12,7 +12,18 @@ local function parseexpr(source: string): T.AstExpr
...
+
+local function parsefile(source: string): ParseResult
+	local result = luau.parse(source)
+	return { root = result.root, eof = result.eof }
+end
+
 return {
 	parse = parse,
 	parseexpr = parseexpr,
+	parsefile = parsefile,
```
JohnnyMorganz, line 28: Not 100% on this API name. It was unclear if this should've just replaced "parse", or if we still wanted the ability to parse a block independently of whether it was a whole file. Happy to change it if there's suggestions, or can do it in a followup when we standardise the whole API after it's complete.

aatxe: I think this is a fine name for now, i suppose

JohnnyMorganz: cool, will leave it alone for now, and revisit it afterwards

aatxe (approved):

PR #248, opened 2025-05-02
CST: Improve support for conditionals

JohnnyMorganz:
This PR:
- Fixes the handling of `elseif` clauses in conditional statements; and
- Introduces serialization for if-expressions

When serializing elseifs to Luau (both in the statement and expression case), we store it as an array of specific elseif nodes. Note that such an elseif node does not exist in the C++ representation of the AST: instead the C++ AST stores elseif chains in a "linked list" format: for `AstStatIf`, if `elsebody` is another `AstStatIf`, then it is actually an elseif.

However, the array representation is a simpler representation in Luau, allows us to differentiate between the `if` and `elseif` token correctly, and makes it easier to post-process in downstream consumers.

### DIFF
--- a/batteries/syntax/ast_types.luau
+++ b/batteries/syntax/ast_types.luau
@@ -168,6 +168,24 @@ export type AstExprTypeAssertion = {
 	annotation: AstType,
 }
 
+export type AstExprIfElseIfs = {
+	["elseif"]: Token<"elseif">,
+	condition: AstExpr,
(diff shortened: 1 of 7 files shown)


JohnnyMorganz:
This PR is rebased on https://github.com/luau-lang/lute/pull/247, marking as draft until the former is merged (can still be reviewed)

aatxe (approved):

PR #249, opened 2025-05-02
CST: Serialize union and intersection types

JohnnyMorganz:
This PR adds support for serializing union and intersection types. This relies on the Luau FFlags `LuauPreserveUnionIntersectionNodeForLeadingTokenSingleType` and `LuauParseOptionalAsNode2`.

Serializing intersection nodes is fairly straight forward: it can be treated as a punctuated list, where the separator is the `&` token. 

Serializing union nodes is a little bit more involved due to the special handling of the `string?` optional syntax. But in the end, we serialize it in the same way as a punctuated list.

In both cases, we also serialize an optional leading token.


JohnnyMorganz:
~~This PR is based on https://github.com/luau-lang/lute/pull/248, marking as draft until the former is merged (can still be reviewed)~~

aatxe (approved):

PR #253, opened 2025-05-03
Properly handle process.env GC

Nicell:
I didn't realize `__gc` wasn't available in Luau, so I don't think the destructor for the `EnvIter` was done properly at all. This should work now.


aatxe (approved):
I am unsure of how this slipped by me in review. Oops!

PR #254, opened 2025-05-04
Disable linux arm64 release builds for now

Nicell:
Foreman is broken on linux arm64 it looks like. I'm guessing it's because there isn't an arm64 build at all and running a x86 binary isn't going to work (although the error is confusing). 

I'll look into making a foreman arm64 build next week.


aatxe:
If that's the issue, shouldn't previous nightlies have also not worked...?

Nicell:
> If that's the issue, shouldn't previous nightlies have also not worked...?

We weren't using foreman for releases until #171 was merged

aatxe:
Ah, that would do it. I believed incorrectly that I changed it when I added the foreman config in the first place.

PR #255, opened 2025-05-05
CST: Handle table types

JohnnyMorganz:
This PR adds support for handling table types as part of the CST.

Type tables are serialized as a list of separate table items, split by a separator (either a comma or a semicolon). The separator is included as part of the item node.

We add a special case node `AstTypeArray` to handle the `{ number }` case. Although this node does not exist on the C++ side, we serialize it independently in Luau for easier ergonomics, given that an array type cannot be mixed with a general table type.

### DIFF
--- a/tests/astSerializerTests/type-tables-2.luau
+++ b/tests/astSerializerTests/type-tables-2.luau
@@ -1,0 +1,2 @@
+-- stylua: ignore
+type A = { [  "a type"]: string }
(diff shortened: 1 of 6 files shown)


tests/astSerializerTests/type-tables-2.luau
```diff
@@ -0,0 +1,2 @@
+-- stylua: ignore
+type A = { [  "a type"]: string }
```
aatxe, line 2:
```suggestion
type A = { [  "a type"]: string }

```

aatxe (approved):

PR #256, opened 2025-05-05
CST: Handle type functions

JohnnyMorganz:
This PR adds support for serializing type functions

```lua
type function foo()
end
```

### DIFF
--- a/tests/astSerializerTests/type-function-1.luau
+++ b/tests/astSerializerTests/type-function-1.luau
@@ -1,0 +1,5 @@
+type function Foo()
+end
+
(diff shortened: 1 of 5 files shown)


tests/astSerializerTests/type-function-1.luau
```diff
@@ -0,0 +1,6 @@
+-- stylua: ignore start
+type function Foo()
+end
+
+export type function Bar()
+end
```
aatxe, line 6:
```suggestion
end

```

aatxe (approved):

PR #257, opened 2025-05-05
Ansi color battery improvements

gaymeowing:
* Rename to richterm, play on words with rich text and isn't vague
* Reduce the size of the created closures via doing all of the formatting within a `format` function
* Use `table.concat` instead of repeadedly using the concat operator (`table.concat` is faster)
* Remove dependency on lutes process lib, and added a nocolor arg to every formatter that replicates this behavior. As its stated in the README that batteries aren't supposed to depend on lute.
* Rename bright formatters to be gramatically correct ie `redBright` is now `brightRed`

If the process lib is ever moved to the @std alias (it might already be under there idk), I'd expect there to be a pcall require situation to be added so that if the env variable for no color is set richterm respects it.


aatxe (approved):
I don't understand why the least interesting bit of code in this whole project is such a battleground, but sure.

PR #258, opened 2025-05-05
Re-enable Linux arm releases

Nicell:
Foreman 1.6.4 with Linux arm support was released today: https://github.com/Roblox/foreman/releases


aatxe (approved):
Thank you for your service o7

PR #266, opened 2025-05-09
Expose previously private stuff in lute

cheesycod:
For lutes scheduler to play nicely with mlua (im working on a mlua lute integration thing), these runtime internals need to be public so tokio can do its thing and then call into luteexts scheduler apis to progress with lute stuff


cheesycod:
Superseded by #272

PR #267, opened 2025-05-10
Add basic compile command

### LINKED ISSUE #137: Standalone executable creation
Add a `lute compile <file>` command to turn a Lute program into a single-file executable. Lute seems well-positioned for this, as the whole runtime is small compared to something like Node. This would make creating CLI tools more appealing.

A bonus follow-up would be cross-compilation to other operating systems and architectures.

Nicell:
This adds `lute compile <script.luau> [optional_exe_file_name]`. Does not support cross-compilation or multiple files. This is single file bytecode embedding using a very simple magic flag on the existing executable.

Closes #137


aatxe (approved):

aatxe:
I'd like to some day see this become a more principled linking setup, but I feel like it'll end up being this for a long time based on the available resourcing, and that's okay.

Issue #216, opened 2025-04-27
Hermetic Builds


aatxe:
The current build setup in cmake is _okay_, but we keep discovering bugs caused by tools using `find_packages` and it's kind of nightmarish. At some point, we should put everything in one build system. If it's going to be cmake still, that's probably okay (and a benefit is easier to embed lute in a lot of other projects that are also using cmake), but I would also accept replacing it with buck2 or nix. No other alternatives would be seriously considered at this time (at least not without discussion here).

aatxe:
I was thinking we could do without this, but I'm upping the priority on this and working on it myself.

aatxe:
In terms of urgent things that actually fit within the 0.1.0 milestone, as opposed to sort of broader nice-to-haves for build hermeticity, we need the following:
- [x] our own cmake file for curl that will use our boringssl and other static dependencies present in our build system
- [x] most likely zlib actually needs to be added as a statically linked dependency, and we'll need to write a cmake file for that as well
- [x] any other things that curl or uSockets need to actually be Useful™️ should be added as statically linked dependencies as well

aatxe:
In #271, we have done enough of this problem to call it done. We don't define all of the compilers in the build system, but all dependencies are defined in it, and that's the appropriate level of investment for lute IMO.

PR #269, opened 2025-05-11
Switching to CMake for dependency management and luau embedding

bjcscat:
This PR changes up the CMake file structure and adds CPM to handle dependencies and embedding luau source into the runtime for requires. Partially addresses #216


bjcscat:
Closed in favor of #271

PR #276, opened 2025-05-13
Explicit port reuse in net.serve

### LINKED ISSUE #127: net: Properly handle ports on HTTP server

Nicell:
Closes #127

### DIFF
--- a/net/src/net.cpp
+++ b/net/src/net.cpp
@@ -219,6 +219,7 @@ struct ServerLoopState
     std::shared_ptr<Ref> handlerRef;
     std::string hostname;
     int port;
+    bool reusePort = false;
 };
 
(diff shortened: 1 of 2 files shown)


net/src/net.cpp
```diff
@@ -526,6 +531,13 @@ int lua_serve(lua_State* L)
         }
         lua_pop(L, 1);
 
+        lua_getfield(L, 1, "reuseport");
```
Nicell, line 534: open to a different name for this field

aatxe: seems fine to me

aatxe (approved):

PR #277, opened 2025-05-13
Sync to upstream Luau 0.673

JohnnyMorganz:


vrn-sn (approved):
Approving for the require-by-string change—passes the tests in the main Luau repo.

runtime/src/require.cpp
```diff
@@ -204,8 +205,12 @@ static int load(lua_State* L, void* ctx, const char* path, const char* chunkname
     // new thread needs to have the globals sandboxed
     luaL_sandboxthread(ML);
 
+    std::optional<std::string> contents = readFile(loadname);
```
JohnnyMorganz: I might've got this wrong for `@std` actually

vrn-sn: Oof, we really need our own C++ unit tests—running against Luau's unit tests wouldn't have caught this 😞.

Does it work to just change it to this? (Can probably remove the `#include "Luau/FileUtils.h"` as well.)
```cpp
std::optional<std::string> contents = getFileContents(reqCtx->currentVFSType, std::string(loadname), "");
```

It's a little hacky, but I will be going through all this code again after Luau's 0.674 release with the new VFS abstraction.

JohnnyMorganz: The reason it worked fine during my local testing was because of the `std` alias present in `.luaurc`

PR #278, opened 2025-05-13
fix AstStatForIn's 'values' type in the syntax battery

TechHog8984:


JohnnyMorganz (approved):
Thanks!

Curious to already see users of the syntax library :)

PR #279, opened 2025-05-13
Make Lute a single-folder VS Code workspace

vrn-sn:
VS Code has varying levels of settings overrides. The most basic is user settings, then workspace settings, followed by project folder settings. Currently, we store project settings in `.vscode/settings.json`, which means that we can't add the `.vscode` directory to `.gitignore`.

This PR creates a [multi-root workspace](https://code.visualstudio.com/docs/editing/workspaces/multi-root-workspaces) containing only the root folder of this repository, which allows us to store these settings in `lute.code-workspace` instead. Now, `.vscode` is added to `.gitignore` and can be used for defining local-only launch and task configurations.

This also means that we can add workspace-level launch and task configurations (which would run luthier) that come packaged with this repository by default in `lute.code-workspace`.

As long as the repository is opened in VS Code as a workspace, this should behave the same as before.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@
 /cmake[.-]*/
 /coverage/
 /.vs/
+/.vscode/
 /fuzz/luau.pb.*
 /crash-*
 /default.prof*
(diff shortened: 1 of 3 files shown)


aatxe (approved):
I still don't use VS Code, but this looks broadly more principled to me, so I'm happy to do it this way.

PR #281, opened 2025-05-16
Add Lute.Test target for C++ unit tests

### LINKED ISSUE #197: Add doctest for unit testing C++.
We need to have a unit test binary using doctest for testing the runtime.

vrn-sn:
This PR adds the `doctest.h` header to our `tests` folder, along with a file containing a simple test. The `Lute.Test` target is now an option when running luthier.

Tests can be run like this:

1. `python3 tools/luthier.py configure Lute.Test`
2. `python3 tools/luthier.py build Lute.Test`
3. `python3 tools/luthier.py run Lute.Test`

Resolves #197.


ffrostfall:
purely inquisitive; why put doctest in tests over extern?

vrn-sn:
> purely inquisitive; why put doctest in tests over extern?

Ultimately up to @aatxe. My reasoning was that it's a single header file (~300KB) that's only used to enable testing, so I didn't think we needed to bundle it with the other `extern` libraries.

aatxe (commented):
Can you add the luthier targets for this? target name is `Lute.Test` (honestly, may as well add `Lute.CLI` as another name for the current target too). It should just be adding a new entry to the tables at the top.

vrn-sn:
@aatxe I added `Lute.CLI` (same as `lute`) and `Lute.Test`. Tests now run by passing `run Lute.Test` to luthier.

```
> python3 tools/luthier.py run Lute.Test
```

aatxe (approved):

PR #283, opened 2025-05-17
Implement system statistic functions into `@lute/system`

### LINKED ISSUE #194: `system.freememory()`, `system.totalmemory()`, `system.hostname()`, `system.uptime()` APIs.
These APIs are important for debugging & logging purposes for libraries/tools running on servers,

CompeyDev:
Closes #194.


system/src/system.cpp
```diff
@@ -66,6 +69,52 @@ int lua_threadcount(lua_State* L)
     return 1;
 }
 
+int lua_freememory(lua_State* L)
+{
+    lua_pushinteger(L, uv_get_free_memory()); // FIXME: this is a 64 bit int, deal with it?
```
ffrostfall: Use `lua_pushnumber` here, casting it to a double. Divide the uint64_t by the size of a megabyte, so it's in megabytes.

This will have enough precision.

system/src/system.cpp
```diff
@@ -66,6 +69,52 @@ int lua_threadcount(lua_State* L)
...
+}
+
+int lua_hostname(lua_State* L)
+{
+    size_t sz = UV_MAXHOSTNAMESIZE;
+    char* hostname = (char*) malloc(sz);
+
+    int res = uv_os_gethostname(hostname, &sz);
+    if (res != 0) {
+        lua_pushfstringL(L, "libuv error: %s", uv_strerror(res));
```
ffrostfall: You should properly handle this error, and not just use UV_MAXHOSTNAMESIZE.

You could look at the implementation of `process.cwd` for this - it does it properly.

system/src/system.cpp
```diff
@@ -66,6 +69,52 @@ int lua_threadcount(lua_State* L)
...
+        lua_error(L);
```
ffrostfall: Also, this can be `luaL_error` over `lua_error`.

system/src/system.cpp
```diff
@@ -66,6 +69,52 @@ int lua_threadcount(lua_State* L)
...
+{
+    lua_pushinteger(L, uv_get_total_memory()); // FIXME: this is a 64 bit int, deal with it?
+
+    return 1;
+}
+
+int lua_hostname(lua_State* L)
+{
+    size_t sz = UV_MAXHOSTNAMESIZE;
+    char* hostname = (char*) malloc(sz);
```
ffrostfall: Use std::string, don't make a C-style buffer. `process.cwd` is an example you could look to. You could also just copy the `cwd` implementation and place it here, they're extremely similar.

system/src/system.cpp
```diff
@@ -66,6 +69,52 @@ int lua_threadcount(lua_State* L)
...
+
+    return 1;
+}
+
+int lua_uptime(lua_State* L) {
+    double uptime = 0;
+    
+    int res = uv_uptime(&uptime);
+    if (res != 0) {
+        lua_pushfstringL(L, "libuv error: %s", uv_strerror(res));
```
ffrostfall: You could also use luaL_error here.

system/src/system.cpp
```diff
@@ -69,34 +70,41 @@ int lua_threadcount(lua_State* L)
     return 1;
 }
 
+#define BYTES_PER_MB 1048576 // 2^20 bytes
+
 int lua_freememory(lua_State* L)
 {
-    lua_pushinteger(L, uv_get_free_memory()); // FIXME: this is a 64 bit int, deal with it?
+    lua_pushnumber(L, (double) uv_get_free_memory() / BYTES_PER_MB);
```
ffrostfall: You should use a static_cast here

system/src/system.cpp
```diff
@@ -69,34 +70,41 @@ int lua_threadcount(lua_State* L)
...
 
     return 1;
 }
 
 int lua_totalmemory(lua_State* L)
 {
-    lua_pushinteger(L, uv_get_total_memory()); // FIXME: this is a 64 bit int, deal with it?
+    lua_pushnumber(L, (double) uv_get_total_memory() / BYTES_PER_MB);
```
ffrostfall: you should also use a static_cast here

system/src/system.cpp
```diff
@@ -69,34 +70,41 @@ int lua_threadcount(lua_State* L)
     return 1;
 }
 
+#define BYTES_PER_MB 1048576 // 2^20 bytes
```
ffrostfall: This shouldn't be a macro, but rather just a constant.

system/src/system.cpp
```diff
@@ -106,8 +114,7 @@ int lua_uptime(lua_State* L) {
     
     int res = uv_uptime(&uptime);
     if (res != 0) {
-        lua_pushfstringL(L, "libuv error: %s", uv_strerror(res));
-        lua_error(L);
+        luaL_error(L, "libuv error: %s", uv_strerror(res));
         return 0;
```
ffrostfall: You don't need to return - luaL_error will exit.

system/src/system.cpp
aatxe:
```suggestion
constexpr std::size_t BYTES_PER_MB = 1024 * 1024;
```

system/src/system.cpp
```diff
@@ -66,6 +70,59 @@ int lua_threadcount(lua_State* L)
...
+    int res = uv_uptime(&uptime);
+    if (res != 0)
+    {
+        luaL_error(L, "libuv error: %s", uv_strerror(res));
+    }
+
+    lua_pushnumber(L, uptime);
+
+    return 1;
+}
```
aatxe, line 125:
```suggestion
}

```

aatxe (approved):

PR #286, opened 2025-05-18
Implement `Runtime::runOnce()` and use it in the internal scheduler loop

### LINKED ISSUE #272: Expose a Runtime::runOneIteration API
Exposing a Runtime::runOneIteration API that runs one iteration of the scheduler and returns data + code (with code at top of stack) about the status would be really useful for me in terms of embedding lute.

ffrostfall:
This PR implements a C-sided API `Runtime::runOnce()` to run a singular iteration of the scheduler loop. This returns an `std::variant` between `StepErr` and `StepSuccess`, returning the `lua_State* L` which errored if the variant is a `StepErr`. This is then utilized in `Runtime::runToCompletion`, which simplifies some logic regarding exiting the runtime loop.

Fixes #272

### DIFF
--- a/runtime/include/lute/runtime.h
+++ b/runtime/include/lute/runtime.h
@@ -1,5 +1,6 @@
 #pragma once
 
+#include "Luau/Variant.h"
 #include "lua.h"
 #include "lute/ref.h"
 
@@ -10,7 +11,6 @@
 #include <mutex>
 #include <string>
 #include <thread>
-#include <variant>
 #include <vector>
 
 struct ThreadToContinue
@@ -35,7 +35,7 @@ struct StepEmpty
 {
 };
 
-typedef std::variant<StepSuccess, StepErr, StepEmpty> RuntimeStep;
+typedef Luau::Variant<StepSuccess, StepErr, StepEmpty> RuntimeStep;
 
 struct Runtime
 {
--- a/runtime/include/lute/runtime.h
+++ b/runtime/include/lute/runtime.h
@@ -28,9 +28,14 @@ struct StepErr
 
 struct StepSuccess
 {
+    lua_State* L;
 };
 
-typedef std::variant<StepSuccess, StepErr> RuntimeStep;
+struct StepEmpty
+{
+};
+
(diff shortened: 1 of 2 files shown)


ffrostfall:
I'm uncertain about the namespacing & organization related to `StepSuccess` and `StepErr`, and the usage of `holds_alternative`, but I think this is the right way to do it.

also side note, sorry for the horrific diff :/

cheesycod:
> I'm uncertain about the namespacing & organization related to `StepSuccess` and `StepErr`, and the usage of `holds_alternative`, but I think this is the right way to do it.
> 
> also side note, sorry for the horrific diff :/

It'd be nice if StepSuccess included lua_State of the thread and maybe even a reason enum/the status(?)

cheesycod:
I approve this PR personally

runtime/include/lute/runtime.h
```diff
@@ -19,19 +21,37 @@ struct ThreadToContinue
...
+struct StepSuccess
+{
+    lua_State* L;
+};
+
+struct StepEmpty
+{
+};
+
+typedef std::variant<StepSuccess, StepErr, StepEmpty> RuntimeStep;
```
aatxe:
```suggestion
using RuntimeStep = std::variant<StepSuccess, StepErr, StepEmpty>;
```

aatxe: Prefer `using` to `typedef`, categorically.

aatxe: Use `Luau::Variant`, not `std::variant`.

aatxe (changes requested):
Looks good, but use `Luau::Variant`, not `std::variant`.

runtime/src/runtime.cpp
```diff
@@ -35,73 +35,98 @@ Runtime::~Runtime()
...
+        else
+        {
+            continue;
+        }
+    };
 
     return true;
 }
 
+
```
aatxe:
```suggestion
```

aatxe (approved):

PR #287, opened 2025-05-18
Implement `task.wait()`

### LINKED ISSUE #76: task.delay/wait using durations
Self-explanatory really

ffrostfall:
Closes #76. Implements `task.wait` with three overloads: `task.wait(number)`, `task.wait(duration)`, and `task.wait()`.

This is done by utilizing the [libuv function `uv_timer_start`](https://docs.libuv.org/en/v1.x/timer.html#c.uv_timer_start). We also create a struct named `WaitData`, which contains three fields: the resumption token, the timer, and the start time.

## Future Work
Maybe we should implement functionality in the scheduler to check for threads yielding via task.wait to support better cancelling functionality?

We should also find out if utilizing uv_timer_start will result in high CPU usage during yielding. If so, that is a bigger motivator to switch to our own scheduler implementation.


aatxe:
>  if utilizing uv_timer_start will result in high CPU usage during yielding

I am fairly certain it shouldn't, but definitely would be interested in hearing otherwise if so.

aatxe (approved):

PR #292, opened 2025-05-21
build-system-improvements: fix BoringSSL MSVC paths

ffrostfall:
Fixes build paths on MSVC by checking to see if we're using MSVC and setting the paths accordingly. BoringSSL seems to compile to `*.lib` on Windows instead of `lib*.a`


ffrostfall:
oops. what wonky merge history.

Issue #60, opened 2025-01-25
lute Runtime Type Definitions


aatxe:
We should have a folder (probably `lute`) with definition files for the runtime's library surface. `tc.cpp` should be refactored to use these files as well, and we should likely include a configuration for luau-lsp to use them as well. This will considerably improvement the development experience of working with `lute`.

aatxe:
We do have some definitions already in `tc.cpp` to be clear, but we want to get them into files (so they can more readily be consumed by other tools). They should use the same generation scheme we're using to embed the std library as well.

ffrostfall:
Solved by #304

JohnnyMorganz:
Does #304 solve type definitions for `@std`?

aatxe:
I don't think it does, but making it do so should be pretty actionable since we can just... output the entire scripts for everything in `@std` into the appropriate folder. We have their full contents in the binary as-is.

aatxe:
This was fixed at some point, but whoever fixed it didn't actually come back and close it. We do include the `std` files now in the definitions that get set up.

PR #302, opened 2025-05-29
Add a few type definitions that I had to write today for the sake of demoing things.

aatxe:
We're working towards #60 by adding type definitions for `@lute/crypto`, @lute/vm`, and `@lute/net`, plus two missing functions from the `@lute/luau` definitions. These came out of making a short demo video of Lute for internal use.


definitions/net.luau
```diff
@@ -0,0 +1,34 @@
...
+	error("not implemented")
+end
+
+export type handler = (...any) -> ...any
+
+export type configuration = {
+	hostname: string?,
+	port: number?,
+	reuseport: boolean?,
+	tls: { cert_file_name: string, key_file_name: string, passphrase: string?, ca_file_name: string? },
```
Nicell, line 26: tls is optional

definitions/net.luau
```diff
@@ -0,0 +1,34 @@
...
+	headers: { [string]: string },
+	status: number,
+	ok: boolean,
+}
+
+function net.request(url: string, metadata: metadata?): request
+	error("not implemented")
+end
+
+export type handler = (...any) -> ...any
```
Nicell, line 20: I assume this is a placeholder, we can define the request argument in a follow up I suppose

aatxe: yeah, it was half-assed enough to make the demo not have squiggles if i'm being honest

PR #304, opened 2025-05-30
Runtime type definitions setup

ffrostfall:
This PR adds in `cli/commands/setup`, which puts type definitions in the user directory. This PR also adds some type definitions to the remaining Lute libraries.

Type definition files are put into `~/.lute/typedefs/0.1.0/`. We autogenerate the `definitions` file into a singular `generated-types.luau` file, containing a table `{ [fileName]: fileContent }`. Upon usage of the `setup` command, we write this table into the type definition directory.


tools/luthier.luau
```diff
@@ -417,10 +442,16 @@ local exitCode = 0
 
 if subcommand == "fetch" then
 	exitCode = fetchDependencies()
+elseif subcommand == "generate" then -- TODO cutesy name for generate
```
aatxe, line 728: "compose" ?

tools/luthier.luau
```diff
@@ -251,6 +254,28 @@ local function generateStdLibFilesIfNeeded()
 	fs.close(hash)
 end
 
+local function generateTypeDefinitionFiles()
```
aatxe, line 365: This needs to do a hash check on the files and only run if needed, doesn't need the "If Needed" in the name, and honestly you can remove it from the generateStdLibFiles one too if you want.

tools/luthier.luau
```diff
@@ -417,10 +442,16 @@ local exitCode = 0
 
 if subcommand == "fetch" then
 	exitCode = fetchDependencies()
+elseif subcommand == "generate" then -- TODO cutesy name for generate
+	generateTypeDefinitionFiles()
+	generateStdLibFilesIfNeeded()
```
aatxe: Might be worth pulling this into a single generateAnyFiles function that we call first in all these cases?

aatxe (commented):
Looks pretty good, I would say you can split the remaining definition file additions into its own PR though since they're kinda orthogonal changes.

.styluaignore
```diff
@@ -0,0 +1 @@
+**/generated-types.luau
```
aatxe: missing newline at eof

.gitignore
```diff
@@ -19,3 +19,4 @@ extern/libuv/libuv-static.pc
 extern/libuv/DartConfiguration.tcl
 extern/*/
 generated
+**/generated-types.luau
```
aatxe: missing newline at eof

tools/luthier.luau
ffrostfall: Would rather do this in another PR

checkraisefold:
All type definition luau changes can be removed from this PR - extracted to and expanded in #351

aatxe (approved):

PR #312, opened 2025-06-03
Fix uninitialized memory issue where `get_alias` is never initialized to nullptr, therefore causing a configuration error

ffrostfall:
Luau does not allow setting `get_config` and `get_alias` at the same time. Lute does not use `get_alias`, but we also never initialize it to `nullptr` when creating the require configuration. This causes runtime errors, at least on Windows.

### DIFF
--- a/runtime/src/require.cpp
+++ b/runtime/src/require.cpp
@@ -214,7 +214,6 @@ static int load(lua_State* L, void* ctx, const char* path, const char* chunkname
 
     RequireCtx* reqCtx = static_cast<RequireCtx*>(ctx);
     std::optional<std::string> contents = getFileContents(reqCtx->currentVFSType, std::string(loadname), "");
-    ;
     if (!contents)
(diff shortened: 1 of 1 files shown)


runtime/src/require.cpp
```diff
@@ -213,7 +213,8 @@ static int load(lua_State* L, void* ctx, const char* path, const char* chunkname
     luaL_sandboxthread(ML);
 
     RequireCtx* reqCtx = static_cast<RequireCtx*>(ctx);
-    std::optional<std::string> contents = getFileContents(reqCtx->currentVFSType, std::string(loadname), "");;
+    std::optional<std::string> contents = getFileContents(reqCtx->currentVFSType, std::string(loadname), "");
+    ;
```
vrn-sn:
```suggestion
```

vrn-sn (approved):
Just remove that redundant `;` and I can merge.

aatxe (approved):

PR #314, opened 2025-06-03
Add support for multi-file CLI command scripts embedded into the lute executable

vrn-sn:
## Summary

In this PR, we add support for running command-line Luau scripts that are stored in the `cli/commands` directory.

If we have `cli/commands/foo.luau` or `cli/commands/foo/init.luau`, running `lute foo` will automatically execute the contents of that script (with support for relative requires as well).

## Implementation details

We create a new `CliVfs` that can optionally be embedded in `RequireVfs`. The CLI enables this, but `vm.create` does not, as it does not need a custom VFS for command-line scripts. The actual implementation of the VFS is nearly identical to `StdLibVfs`.

The embedding of the CLI scripts is done using Luthier to generate source files for a new `Lute.CLI.Commands` target. This is similar to how the `@std` libraries were embedded into the `Lute.Std` target.
- https://github.com/luau-lang/lute/pull/138
- https://github.com/luau-lang/lute/pull/205

While working on this, I noticed that `luthier.luau` was incomplete. Due to the lack of an `os.walk` API like we have in Python, `luthier.luau` doesn't correctly handle nested directories that need to be embedded. To solve this, I have added a `traverseFileTree` function that effectively accomplishes the same thing.

Because of slight differences in the order files are processed by `os.walk` and `traverseFileTree`, `luthier.py` and `luthier.luau` spit out different hashes. I think this is fine. When only using one of the tools in isolation, the caching seems to work as expected.


CMakeLists.txt
```diff
@@ -20,6 +20,7 @@ add_library(Lute.VM STATIC)
 add_library(Lute.Process STATIC)
 add_library(Lute.System STATIC)
 add_library(Lute.Time STATIC)
+add_library(Lute.CLICommands STATIC)
```
aatxe:
```suggestion
add_library(Lute.CLI.Commands STATIC)
```

vrn-sn: Fixed

aatxe (approved):

cli/main.cpp
```diff
@@ -25,6 +27,7 @@
 
 #include "compile.h"
 #include "tc.h"
+#include <iostream>
```
vrn-sn, line 30: @aatxe 😭, leaked through

PR #327, opened 2025-06-05
Build releases on 22.04

Nicell:
24.04 is newer than we need and creates builds that don't work on older OS's unnecessarily from glibc that's too new. Let's build on 22.04. Maybe we can always stay behind one? For example 24.04 builds don't even work on latest debian!

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -56,10 +56,10 @@ jobs:
     strategy:
       matrix:
         include:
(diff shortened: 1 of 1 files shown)


aatxe:
sure, why not? we should also probably look eventually at packaging for apt, brew, and other major package managers, but that's a problem for the future

aatxe (approved):

PR #334, opened 2025-06-09
Pull Lute.Require out of Lute.Runtime

vrn-sn:
`Lute.Runtime` no longer contains all the require-by-string logic for Lute, as it is moved into its own `Lute.Require` library. This makes working with `Lute.Runtime` a bit simpler.

In a follow-up PR, I will be moving the CLI's state set-up logic into `Lute.Runtime`, allowing us to link `Lute.Test` against `Lute.Runtime` to properly test our Luau VM.


vrn-sn:
Using stacked PRs here. The base branch is intentionally not set to `primary`. Once #333 is merged, this PR should automatically update. (Important to leave this as a PR draft until then.)

aatxe (approved):

PR #335, opened 2025-06-11
Improve ModulePath API to facilitate building library VFS subtrees

vrn-sn:
Updates the `ModulePath` API to avoid needing to provide a position when defining the root directory to which a `ModulePath` can parent. Constructing a `ModulePath` is also simplified via the new `create` factory function that returns `std::nullopt` in case of an invalid starting `ModulePath`.

During this work, I realized that certain tests had been disabled unintentionally—most likely due to the build system PR. I've re-enabled the tests in `modulepath.test.cpp` now. I've also created `luteprojectroot.h` and `luteprojectroot.cpp`, as require-by-string tests (to be added soon) will also need these available.

### DIFF
--- a/lute/require/include/lute/modulepath.h
+++ b/lute/require/include/lute/modulepath.h
@@ -20,9 +20,15 @@ struct ResolvedRealPath
 class ModulePath
 {
 public:
-    ModulePath(
+    // (rootDirectory + '/' + filePath) is the full path to the initial file.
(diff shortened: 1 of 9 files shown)


vrn-sn:
Dependent on #334. Leaving this as a draft because I will need to rebase once that PR is merged.

aatxe (approved):

PR #336, opened 2025-06-11
Add unit tests for require-by-string implementation in lute

vrn-sn:
This PR has two main aspects:
1. We copy over the existing require-by-string test files from the Luau repository to Lute and add test cases in `require.test.cpp`. We also add some new tests for Lute-specific behavior `lute.luau` and `std.luau`.
2. We move the `setupState` logic that used to be under `Lute.CLI.lib` to be under `Lute.Runtime` instead.


vrn-sn:
Yet another stacked PR: dependent on #335. (As before, leave this in draft until dependencies are merged in, at which point I will need to rebase.)

vrn-sn:
@aatxe when you get a chance, this is ready as well.

aatxe (approved):

PR #337, opened 2025-06-11
Remove luthier.py since it's been superseded entirely by the Luau implementation.

aatxe:
We no longer need to maintain `luthier.py`, and it poses a frustrating maintenance risk right now to attempt to apply both together. It may be useful in the future to introduce an optional path in the CI to fetch everything from a tarball somewhere and then use cmake/ninja directly instead of running luthier, but we can deal with that when needed.


.github/actions/setup-and-build/action.yml
```diff
@@ -34,19 +29,8 @@ runs:
       id: run
       shell: bash
       run: |
-        if [ "${{ inputs.use-python-luthier }}" = "true" ]; then
-          echo "runtime=python" >> "$GITHUB_OUTPUT"
-          echo "ext=py" >> "$GITHUB_OUTPUT"
-        else
           echo "runtime=lute" >> "$GITHUB_OUTPUT"
           echo "ext=luau" >> "$GITHUB_OUTPUT"
```
Nicell: could just use lute/luau inline then instead of using variables

aatxe: good call, fixed that

PR #338, opened 2025-06-12
Run tests on Windows and MacOS

Nicell:


vrn-sn:
@aatxe seems like Lute.Test won't build on Windows. Maybe something related to the build system work that went in recently? I'm able to built it locally on my Windows PC.

aatxe:
I forget if it was zlib or zstd in the message right now, but yeah, we may need to make another `Find_X.cmake` override.

PR #339, opened 2025-06-12
Support executing `require` in dynamically generated, library-aware virtual filesystems

vrn-sn:
This PR introduces `Package::RequireVfs` as an alternative to `RequireVfs` (the current abstraction around the filesystem, `@std` libraries, `@lute` libraries, and [optionally] CLI commands and embedded/bundled VFS).

To accomplish this, we introduce `IRequireVfs` as a shared interface in `require.h` that both `RequireVfs` and `Package::RequireVfs` implement; `RequireCtx` now simply contains a `std::unique_ptr<IRequireVfs>`.

At a surface level, `Package::RequireVfs`'s logic is fairly similar to `RequireVfs`'s. Our `RequireVfs` implementation can be in any of the following VFS types at a given moment: `Disk`, `Std`, `Cli`, `Lute`. Similarly, `Package::RequireVfs` can be in any of the following: `UserlandVfs`, `Std`, `Lute`.

The default VFS type (for example, when code execution of a project begins) for `Package::RequireVfs` is `UserlandVfs`. In this case, `Package::RequireVfs` manages an internal `Package::UserlandVfs`, which itself can be in one of two VFS states: `Disk` or `Subtree`. Conceptually, the `Disk` case is used to execute user code, and `Subtree` is used for executing code in installed dependencies.

To enable executing code that relies on installed dependencies, `Package::UserlandVfs` must be constructed with two arguments:
1. `directDependencies`: the list of all libraries that user code directly depends on.
2. `libraries`: the list of all transitive dependencies of the user code, with their dependencies on each other also passed in explicitly.

What this enables: when executing user code (in `Disk` mode), we are able to dynamically inject aliases to the dependencies in `directDependencies` that, when followed, transition `Package::UserlandVfs` into `Subtree` mode. Then, when executing dependency code (in `Subtree` mode), we are able to use the transitive dependency information passed in via the `libraries` constructor argument to inject aliases to each dependency's direct dependencies.

Because this happens on a per-dependency basis, an alias `@dep` doesn't need to mean the same thing in `library1` and `library2`. Although it is still true that user code and any individual library cannot _directly_ depend on multiple versions of a dependency, multiple versions are completely fine in the set of transitive dependencies.


aatxe:
Conceptually, I'm not sure how I feel about naming this "Library" exactly, but the overarching choices here look good. I think we want to conceive of a "Luau executable" of sorts, with arbitrary numbers of modules in them, and modules are potentially under their discrete namespaces. I think it's fine if we call the separate module namespaces e.g. `@React` or `@Testing` or `@Foobar` as "libraries" but it seems like here we're calling the whole executable a "library?"

aatxe (approved):
Bikeshedding names aside, I think I'm happy with this. I closed the copilot stuff I don't care about, but left open the ones that are reasonable, but I'll leave it to your discretion if you want to do them or not.

lute/require/include/lute/libraryrequirevfs.h
```diff
@@ -0,0 +1,55 @@
...
+
+    bool isPrecompiled() const override
+    {
+        return false;
+    }
+
+private:
+    enum class VFSType
+    {
+        Library,
```
aatxe:
```suggestion
        Userland,
```

lute/require/include/lute/libraryrequirevfs.h
```diff
@@ -0,0 +1,55 @@
+#pragma once
+
+#include "lute/libraryvfs.h"
+#include "lute/require.h"
+#include "lute/stdlibvfs.h"
+
+namespace Library
```
aatxe:
```suggestion
namespace Package
```

PR #342, opened 2025-06-14
Add lineOffsets field to Luau parser output

christopher-buss:
This PR enhances the Luau parser to expose line offset information, enabling tooling to better access precise source location mapping for nodes.

**Changes Made**:

Exposes the computed 'lineOffsets' field to the Luau 'ParseResult'.
Modified exisitng ast test serializer tests to work with new apis.
Added new test to ensure that line offsets are exposed correctly.

```luau
local result = luau.parse("local x = 1\nlocal y = 2")
-- result.lineOffsets = { 0, 12 }  -- Line 0 starts at char 0, line 1 starts at char 12
```

### DIFF
--- a/lute/luau/src/luau.cpp
+++ b/lute/luau/src/luau.cpp
@@ -2580,10 +2580,12 @@ int luau_parse(lua_State* L)
     lua_setfield(L, -2, "lines");
 
     lua_createtable(L, serializer.lineOffsets.size(), 0);
+    int j = 0;
     for (size_t i = 0; i < serializer.lineOffsets.size(); i++)
     {
+        lua_pushinteger(L, ++j);
         lua_pushnumber(L, serializer.lineOffsets[i]);
-        lua_rawseti(L, -2, i + 1);
+        lua_settable(L, -3);
     }
     lua_setfield(L, -2, "lineOffsets");
 
--- a/lute/luau/src/luau.cpp
+++ b/lute/luau/src/luau.cpp
@@ -273,7 +273,7 @@ struct AstSerialize : public Luau::AstVisitor
         return result;
     }
 
-    // Splits a list of trivia into trailing trivia for the previos token, and leading trivia for the next token
+    // Splits a list of trivia into trailing trivia for the previous token, and leading trivia for the next token
     // The trailing trivia consists of all trivia up to and including the first '\n' character seen
     static std::pair<std::vector<Trivia>, std::vector<Trivia>> splitTrivia(std::vector<Trivia> trivia)
     {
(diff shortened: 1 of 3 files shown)


lute/luau/src/luau.cpp
```diff
@@ -2574,6 +2579,14 @@ int luau_parse(lua_State* L)
     lua_pushnumber(L, result.parseResult.lines);
     lua_setfield(L, -2, "lines");
 
+    lua_createtable(L, serializer.lineOffsets.size(), 0);
+    for (size_t i = 0; i < serializer.lineOffsets.size(); i++)
+    {
+        lua_pushnumber(L, serializer.lineOffsets[i]);
+        lua_rawseti(L, -2, i + 1);
```
ffrostfall: Are you sure you need to use rawseti here?

If this is for arrays, you should be able to just use `lua_settable`.
You can see how `system.cpus` does it here: https://github.com/luau-lang/lute/blob/primary/lute/system/src/system.cpp

lute/luau/src/luau.cpp
```diff
@@ -752,7 +756,7 @@ struct AstSerialize : public Luau::AstVisitor
         serializeToken(node->location.begin, node->local->name.value);
         lua_setfield(L, -2, "token"),
 
-        serialize(node->local);
+            serialize(node->local);
```
JohnnyMorganz, line 759: Not sure if intentional, can you revert this?

lute/luau/src/luau.cpp
```diff
@@ -2560,9 +2565,9 @@ int luau_parse(lua_State* L)
         luaL_error(L, "parsing failed:\n%s", fullError.c_str());
     }
 
-    lua_rawcheckstack(L, 3);
+    lua_rawcheckstack(L, 4);
 
-    lua_createtable(L, 0, 3);
+    lua_createtable(L, 0, 5);
```
JohnnyMorganz: Not sure these numbers are correct.

I think the checkstack should be 3 -> 6 (+1 for the lua_creattable, +2 for the key and value entries)

createtable should only be 3 -> 4, as there is only 1 new field added

lute/luau/src/luau.cpp
```diff
@@ -2574,6 +2579,16 @@ int luau_parse(lua_State* L)
     lua_pushnumber(L, result.parseResult.lines);
     lua_setfield(L, -2, "lines");
 
+    lua_createtable(L, serializer.lineOffsets.size(), 0);
+    int j = 0;
+    for (size_t i = 0; i < serializer.lineOffsets.size(); i++)
+    {
+        lua_pushinteger(L, ++j);
```
JohnnyMorganz: nit: could we just use `i + 1` here instead of a new variable `j`?

JohnnyMorganz (commented):
Looks good, thanks for the test fix as well. Seems like there are some unrelated formatting changes though?

christopher-buss:
I just used the clang format on the file - I think some aspects were previously not formatted which is why they've ended up being formatted here.

christopher-buss:
I've resolved the two issues.

I can resolve the formatting issues if needed - but right now the file isn't up to date with the clang formatter, so anyone else who makes changes to this file would have to revert rather than just being able to run the formatter.

JohnnyMorganz (approved):
Thanks!

PR #345, opened 2025-06-19
Use <= instead of < (visitInterpolatedString)

wmccrthy:
While working with the `visitor`, I noticed that interpolated strings were not properly traversed. It appears that is bc of this condition, which will not visit the last expression in the interpolated string node's expressions table.


BitwiseAndrea (approved):
My approval means nothing but i just want to be part of it all

JohnnyMorganz (approved):
Thank you!

PR #349, opened 2025-06-29
Add fs.stat, fix compilation with clang-cl

checkraisefold:
cmake apparently does not enable exceptions by default with clang-cl, so compilation fails without this flag; otherwise, compilation works perfectly with clang-cl on LLVM 19.1+ (fixed an issue with intrinsics which breaks libsodium/etc)

also adds fs.stat, being the main point of this PR. a caveat is durations are any typed; I plan on making a separate PR with way more complete type defs, including the ones from #304

### DIFF
--- a/lute/fs/src/fs.cpp
+++ b/lute/fs/src/fs.cpp
@@ -413,9 +413,19 @@ int fs_stat(lua_State* L)
     createDurationFromTimespec32(L, stat.st_mtim);
     lua_setfield(L, -2, "modified_at");
 
+    // permissions
+    lua_createtable(L, 0, 2);
+
+    // libuv writes this correctly cross-platform
+    bool canAnyWrite = stat.st_mode & 0222;
+    lua_pushboolean(L, !canAnyWrite);
+    lua_setfield(L, -2, "readonly");
+
+    lua_setfield(L, -2, "permissions");
+
     return 1;
 }
-  
+
 static void defaultCallback(uv_fs_t* req)
 {
     auto* request_state = static_cast<ResumeToken*>(req->data);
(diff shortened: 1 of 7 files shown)


checkraisefold:
this is a draft, but looking for feedback still wrt implementation and if anything else from `stat` should be added. 

I think file permissions could be valuable to add although I'm not sure what kind of interface to expose. just a read only flag in a table, which is compatible with all platforms; similar to Rust/Lune? do we also add the unix permissions to that table if the host os is POSIX?

aatxe:
I think read-only flags is the right starting place.

aatxe:
@checkraisefold is this ready for review now? I merged in primary.

checkraisefold:
> @checkraisefold is this ready for review now? I merged in primary.

yeah i dont have the read only thing in there yet but the rest is reviewable

checkraisefold:
tested and working on windows, thanks libuv. because its the same logic it should work the exact same on unix
there is no alternative to the octal 222 file mode unless we redefine the unix macros for it and use a bitflag since they arent defined on windows
explorer_xifWod1FxX
VSCodium_2RT1eXARiX

lute/fs/src/fs.cpp
```diff
@@ -334,6 +382,50 @@ int fs_rmdir(lua_State* L)
...
+{
+    const char* path = luaL_checkstring(L, 1);
+
+    uv_fs_t stat_req;
+    int err = uv_fs_stat(uv_default_loop(), &stat_req, path, nullptr);
+
+    if (err)
+        luaL_errorL(L, "%s", uv_strerror(err));
+
+    lua_newtable(L);
```
aatxe: We can allocate for this too, it's 6 record fields.

aatxe (approved):

lute/fs/src/fs.cpp
checkraisefold: done

PR #350, opened 2025-07-05
Move syntax library from batteries to std

JohnnyMorganz:
The syntax library is closely connected with `@lute/luau`. According to the readme:

> `batteries`: A collection of useful, standalone Luau libraries that do not depend on lute.

Let's move the syntax library into `std` instead, as it makes more sense there. This means the syntax library will also be vendored as part of the lute runtime itself


aatxe:
Blegh, Windows is failing in CI because MSVC is hitting length limits on the size of the string literals in the program. This is because the generated file is a constexpr pair of `std::string_view` and I think we need to use character arrays instead since that's what MSVC has "proper" support for arbitrarily-large generated literal values of.

JohnnyMorganz:
> I think we need to use character arrays instead since that's what MSVC has "proper" support for arbitrarily-large generated literal values of.

According to https://learn.microsoft.com/en-us/cpp/error-messages/compiler-errors-1/compiler-error-c2026 this would still occur for char arrays. But, adjacent concatenated strings do work, so I've switched luthier to generate that 🙃

aatxe (approved):

PR #357, opened 2025-08-06
fix(fs): Prevent string truncation on embedded null characters (#340)

### LINKED ISSUE #340: `fs.writeStringToFile` incorrectly truncates strings containing a null terminator
The `writeStringToFile` function prematurely truncates any input string at the position of a null terminator (\0).

**To reproducde:**
`require("@lute/fs").writestringtofile("test.txt", "Hi, this is \0 a message!")`

Image

Other filesystem functions probably have the same issue.

ActualMasterOogway:
Replaces `strlen` with `luaL_checklstring` to correctly get the length of Lua strings, which can contain null terminators. This ensures that the full string content is written to files.

Closes #340


Nicell (approved):
Thanks!

PR #358, opened 2025-08-07
Auto-generate reference docs on site

Nicell:
Docs site reference is out of date. We have a source of truth for types now, so let's make use of that.

This is a really basic script that generates decent markdown for the docs site. I know it's not great, but at least it's something to start.


aatxe (approved):

aatxe:
A proper documentation generator is definitely "on the list" at some point, but this is definitely an improvement over what we have right now.

PR #361, opened 2025-08-12
Add printStatement func to printer

wmccrthy:
There are print cases not covered by `printBlock` and `printExpr`, so I figure this is useful to have

If anything, we can replace `printBlock` with this function and export it as `print` instead, [since`visitStatement` encompasses blocks](https://github.com/luau-lang/lute/blob/primary/lute/std/libs/syntax/visitor.luau#L796), in addition to many statement types

### DIFF
--- a/lute/std/libs/syntax/printer.luau
+++ b/lute/std/libs/syntax/printer.luau
@@ -123,13 +123,6 @@ local function printVisitor()
 	return printer
 end
 
---- Returns a string representation of an AstStatBlock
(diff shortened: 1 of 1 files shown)


JohnnyMorganz (approved):
Thanks!

I think we should actually remove the `print` function itself - it is less descriptive than the other functions and it's actually a mistake to use it instead of `printfile` in most cases because printfile also handles the EOF character (and any trivia attached to it)

PR #369, opened 2025-09-02
Introduction of a new task based system for handling async behavior

bjcscat:
This PR seeks to provide a means to simplify the safe implementation of async behaviors in Lute.

To this end it creates a base class Task which serves as the basis for create async operations. Tasks are owned by the runtime and serve to create distinct instances of repeated behavior (async lute function calls). Tasks carry a start method as well as a resume method. Start methods are invoked immediately when added onto the runtime by a caller. The task may then call schedule to mark it as ready for resumption, which the runtime will perform. Tasks may also be cancelled or spawn off new instances of a task to complete their behavior.

At the time of writing the PR includes a fs library composed entirely of these Tasks


aatxe:
cc @vrn-sn I think this reasonably connects with #384?

Issue #362, opened 2025-08-15
Add a build system option for building without the standard library embedding


aatxe:
Supporting a build without `@std` embedded means we can more easily bootstrap without luthier, and that's good for long-term maintainability. Shouldn't be terribly hard to do either.

aatxe:
#386 is proposing a slight variation here where we embed an _empty_ standard library, and then use that to generate the proper standard library for a second build.

vrn-sn:
> embed an empty standard library, and then use that to generate the proper standard library for a second build

Bootstrapping at its finest 😄

aatxe:
I'm still thinking it'd be a worthwhile improvement to also provide a define that stubs out the code to do the embedding since it would simplify the bootstrapping script and the manual instructions for bootstrapping quite a bit.

PR #370, opened 2025-09-03
Use CMake to embed cli and std Luau files

bjcscat:
Hopefully this will help with bootstrapping concerns.

Type definition generation is left within luthier as that generates Luau files for consumption within Luau.

I believe this addresses #362 by sidestepping the luthier requirement


aatxe:
We actively do not want to maintain a big pile of string manipulation stuff in CMake, which is why the raised issue was to allow a build path without it, rather than to move it into CMake.

PR #371, opened 2025-09-04
RequireVfs: Remove all `default` blocks in favor of being explicit

vrn-sn:
When first implementing this, I didn't realize that the compiler checks for exhaustivity when switching over an enum. This is an improvement that helps us catch errors at compile-time by being more explicit.


aatxe (approved):
Yeah, exhaustivity checking is nice. I think MSVC sometimes has problems with it, but seems like it's working here.

PR #373, opened 2025-09-15
Allow required modules to return string and numeric values

vrn-sn:
Based on this PR in the Luau repository: https://github.com/luau-lang/luau/pull/1967.

I also spent a while trying to get newlines in error messages to play nicely, but we still have some issues that we'll need to address later. Main tension is that `lua_debugtrace` automatically puts a newline at the end of the trace, but a "regular" error on a thread is usually just a string with no newline.


vrn-sn:
Changes in `definitions/luau.luau` are just from running StyLua on that file. Not really sure why that file got out-of-date to begin with.

aatxe (approved):

PR #375, opened 2025-09-17
Boiler plate for getting module return type

wmccrthy:


aatxe:
#531 supersedes this.

PR #376, opened 2025-09-17
Switch fs.exists to use uv_fs_access

bjcscat:
This is more correct behavior, fs.exist currently uses stat and checks for an error


lute/fs/src/fs.cpp
```diff
@@ -13,6 +13,9 @@
 #ifdef _WIN32
 #include <direct.h>
 #endif
+#ifndef _WIN32
```
aatxe:
```suggestion
#else 
```

aatxe (approved):

PR #377, opened 2025-09-22
Add printType function

wmccrthy:
This PR adds a `printType` function to the `printer` API. I've been working with the `visitor`, `parser` and `printer` APIs a bunch, and the need for this has come up a few times.

This PR also consolidates some repeated logic used across the printing functions

### DIFF
--- a/lute/std/libs/syntax/printer.luau
+++ b/lute/std/libs/syntax/printer.luau
@@ -131,7 +131,7 @@ end
 
 --- Returns a string representation of an AstExpr
 function printExpr(expr: luau.AstExpr): string
-	return printNode(expr, visitor.visitExpression :: any)
+	return printNode(expr, visitor.visitExpression)
 end
 
 --- Returns a string representation of an AstStat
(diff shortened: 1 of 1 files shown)


lute/std/libs/syntax/printer.luau
```diff
@@ -123,18 +123,25 @@ local function printVisitor()
 	return printer
 end
 
---- Returns a string representation of an AstExpr
-function printExpr(block: luau.AstExpr): string
+function printNode(node: luau.AstExpr | luau.AstStat | luau.AstType, visit: any): string
```
aatxe: Can't we do a more precise type for `visit` here? Looks like it expects a node and a buffer, and returns nothing, so... `(node: ..., printer: buffer) -> ()`

aatxe (approved):
Looks good to me besides the type comment.

lute/std/libs/syntax/printer.luau
wmccrthy: Yea, we definitely can. How about:

`function printNode<T>(node: T, visit: (node: T, visitor: visitor.Visitor) -> ()): string`

Issue #378, opened 2025-09-28
net.request PUT/POST hangs when payload has '\0'


csobrinho:
Example:
```luau
local net = require("@lute/net")
local pp = require("@batteries/pp")

local function test(payload: string)
	local response = net.request("https://httpbin.org/put", {
		method = "PUT",
		headers = {
			["Content-Type"] = "application/octet-stream",
			["Content-Length"] = tostring(#payload),
		},
		body = payload,
	})
	print(response.status, pp(response.headers), response.ok, pp(response.body))
end

test("Hello World") -- works
test("Hello\0World") -- hangs
```

csobrinho:
PR #379 contains the fix.

aatxe:
Fixed by #379

PR #379, opened 2025-09-28
Manually get the full body including any \0 instead of relying on the string length

csobrinho:
Without this we would get a truncated body and curl would hang waiting for the missing bytes to send.

Test:

```luau

local function test(payload: string)
	local response = net.request("https://httpbin.org/post", {
		method = "POST",
		headers = {
			["Content-Type"] = "application/octet-stream",
		},
		body = payload,
	})
	print(response.status, pp(response.headers), response.ok, pp(response.body))
end

test("Hello World") -- works
test("Hello\0World") -- hangs
```

`Content-Length` is 11 for both.


csobrinho:
Related to #378.

aatxe (approved):

PR #383, opened 2025-10-01
infra: Bootstrap shell script for downloading lute dependencies on macOS/linux

### LINKED ISSUE #363: Add a bootstrap shell script solely to download external dependencies
The tune files all provide everything necessary to manually pull all the dependencies to do a build without luthier if needed, but doing so is annoying laborious. We should provide a shell script in tools that will do exactly this task, and nothing else from luthier, in order to provide a reasonable alternative to doing it manually. The code for luthier has the actual `git` invocations that are used to check out all the dependencies.

Vighnesh-V:
Resolves #363 .

This PR adds a bootstrap script for fetching the dependencies needed to built lute from scratch, which removes the requirement of needing a `lute` binary.

### DIFF
--- a/tools/bootstrap.sh
+++ b/tools/bootstrap.sh
@@ -1,6 +1,4 @@
-#!bin/bash
-
-fetch() {
+fetch_dep() {
   local dep_file="$1"
 
   # Ensure the file exists
@@ -28,7 +26,7 @@ fetch() {
     esac
   done < "$dep_file"
 
-  local target_dir="../extern/$name"
+  local target_dir="extern/$name"
 
   # Clone if not already present
   if [[ -d "$target_dir" ]]; then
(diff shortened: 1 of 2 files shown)


tools/bootstrap.sh
```diff
@@ -0,0 +1,58 @@
...
+  # Checkout specific revision if provided
+  if [[ -n "$revision" ]]; then
+    echo "Checking out revision $revision in $target_dir"
+    git -C "$target_dir" checkout "$revision" || return 1
+  fi
+}
+
+# Fetch all the dependencies
+echo "Fetching dependencies..."
+fetch "../extern/boringssl.tune"
```
vrn-sn: Maybe we just loop over the `.tune` files in this directory?

Vighnesh-V: made the script loop over the .tune files. For convenience moved this into the root directory + added instructions for building to the README.

vrn-sn: I think it might make more sense in `tools` next to luthier.

vrn-sn (approved):
I think we should just move the `bootstrap.sh` script into the tools directory but otherwise looks good to merge!

Vighnesh-V:
Just moved that script back into tools/!

PR #386, opened 2025-10-02
Update the bootstrap script to actually build two copies of lute

Vighnesh-V:
The bootstrap script didn't correctly generate the lute standard library headers that the `luthier` script did. This PR:
- generates a minimal header and implementation for the lute standard libraries during the bootstrap phase
- uses these headers to build a `bootstrapped-lute` binary
- uses that `bootstrapped-lute` to run `luthier.luau` which builds the actual `lute` executable with the `std` library correctly generated.
- accepts an `--install` argument to install this `lute` binary to `$HOME/.lute/bin`.

There are some additional improvements that have been made:
1. General README fixes
2. This script can now be run multiple times in sequence, and it correctly deletes artifacts generated by the bootstrapping process each time (e.g. generated std headers, dependencies).
3. Uses git commands to `clone` only the particular revision of the `extern` dependencies we care about, not the entire history.


aatxe (commented):
Overall looks good, my one extra ask would be: can you also add a GH Action config here for running the bootstrap script on Linux and macOS?

Vighnesh-V:
I have the bootstrap script working on MacOS and Ubuntu now, with a caveats:
1) MSVC doesn't like having an empty constexpr array, so the default stdlib that bootstrap generates is a pair of "" strings.
2) I have edited the build action to only run the bootstrap step on linux / mac.

aatxe (approved):

PR #389, opened 2025-10-03
Cross compilation WIP

Nicell:
Adds a `--target` flag to `lute compile`. It accepts one of our currently supported platforms. If you pass it, lute will download a release from github for the targeted platform and cache it for any future compile requests. That binary will then be used for the compilation process instead of the local binary.

TODO
- [ ] Need to embed the current version into the binary. I will put up a PR for that first (with `lute --version` support, too).
- [ ] Produce clearer output for when download is happening and for errors

### DIFF
--- a/lute/cli/include/lute/compile.h
+++ b/lute/cli/include/lute/compile.h
@@ -2,6 +2,9 @@
 
 #include "Luau/FileUtils.h"
 
+#include <filesystem>
+#include <optional>
+
 struct AppendedBytecodeResult
 {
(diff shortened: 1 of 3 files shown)


Vighnesh-V:
@Nicell Should I close this PR out - it's a bit out of date now!

Nicell:
@Vighnesh-V closed, thanks 😄 definitely would like to see this make it in at some point, do you think the general approach here could be used with the new way compile works? I haven't totally kept up.

Issue #382, opened 2025-10-01
Embedded code for `@std` and CLI commands should be embedded as bytecode


vrn-sn:
For simplicity, this code is embedded as source code and compiled upon being required. Instead we can pre-compile it to bytecode before embedding and modify the virtual filesystems responsible for requiring them to simply run the bytecode during the `load` phase of `require`.

This does have security implications: we could end up running bytecode not generated by the Luau compiler. We need a way to guard against this.

aatxe:
> This does have security implications: we could end up running bytecode not generated by the Luau compiler. We need a way to guard against this.

This is a pretty troublesome point if we want to retain that. The right mitigation here is code signing, but that adds a fairly significant degree of complexity for our release pipeline.

vrn-sn:
Agree. I think this could be a good long-term goal to have, but I'm not going to look into it seriously for a while. I spent some time today researching how we could do it, but it's not a strong enough need yet.

vrn-sn:
I talked with some security folks yesterday. The most important thing to do here is signing our release binaries to protect anything that's embedded in the binary in a "sane" way. That will cover the bytecode for the embedded `@std` and `@cli` scripts.

However, it doesn't cover stuff like the current append-bytes-to-binary embedding logic of `lute compile` since that data isn't included in the hash that's used for checking the signature. We could instead embed the scripts by adding them into a new PE/ELF section (we would strip away the old signature and optionally allow the caller of `lute compile` to resign it with their private key).

I honestly think that this approach could work pretty well. We could also then move most of the `lute compile` logic out of C++ and into a Luau `compile` CLI command, which would basically be using APIs like `luau.compile` for getting bytecode and `process.run` to call `objcopy --add-section` and `codesign --remove-signature`. That logic feels a lot easier to implement in Luau. I believe the only C++ logic we'd need would be something like "if the `.embedded` section contains data, reinterpret it as embedded bytecode and run it instead of running Lute normally."

cc: @aatxe, @Nicell what are your thoughts on this?

Nicell:
> That logic feels a lot easier to implement in Luau. I believe the only C++ logic we'd need would be something like "if the `.embedded` section contains data, reinterpret it as embedded bytecode and run it instead of running Lute normally."

[...]

PR #392, opened 2025-10-06
refactor: Support multiple luau files in a single payload for `lute compile`

Vighnesh-V:
This is a precursor PR needed to start making `lute compile` work with multiple files. This:

- Defines a Lute executable format that supports embedding multiple bytecode compiled Luau modules. The encoding looks roughly like: `[lute runtime] [all modules compressed bytecode] [compressed size] [uncompressed size] [num_files] [entry point path string] [entry point length] [lute byte]`
Where each module is just:
`[path len] [path data] [bytecode len] [bytecode data]`

We'll probably have to change the above in order to make requires work correctly.

- Adds some utilities for constructing / parsing these payloads out from a Lute Executable
- Adds some utilities for constructing the output Lute Executable

My general plan of attack for this is the following and should help with the issues discussed in https://github.com/luau-lang/lute/issues/382:
- [ ] get local requires working
- [ ] figure out how to embed std library / battery modules so that it works with requires
- [ ] extend LuteExecutable to work with binaries in multiple formats e,g, PE / ELF / Mach-O like [Sui](https://github.com/denoland/sui) does
- [ ] Figure out the codesigning piece of this

### DIFF
--- a/lute/cli/src/compile.cpp
+++ b/lute/cli/src/compile.cpp
@@ -1,121 +1,437 @@
 #include "lute/compile.h"
 
+#include "Luau/Compiler.h"
+#include "Luau/FileUtils.h"
 #include "lute/options.h"
 #include "uv.h"
-
+#include <zlib.h>
 #include <fstream>
+#include <filesystem>
+#include <cstring>
 
+namespace fs = std::filesystem;
 const char MAGIC_FLAG[] = "LUTEBYTE";
 const size_t MAGIC_FLAG_SIZE = sizeof(MAGIC_FLAG) - 1;
-const size_t BYTECODE_SIZE_FIELD_SIZE = sizeof(uint64_t);
 
-AppendedBytecodeResult checkForAppendedBytecode(const std::string& executablePath)
+void LuteEncodeResult::setError(PayloadEncodeStatus status, std::string errMessage)
 {
-    AppendedBytecodeResult result;
+    this->status.emplace(status);
+    this->errMessage.emplace(std::move(errMessage));
+}
+
+void LuteDecodeResult::setError(PayloadDecodeStatus status, std::string errMessage)
+{
+    this->status.emplace(status);
+    this->errMessage.emplace(std::move(errMessage));
+}
+
+LuteExePayload::LuteExePayload()
+{
+}
+
+bool LuteExePayload::add(const std::string& luauFilePath)
+{
+    // First file added becomes the entry point
+    if (filePaths.empty())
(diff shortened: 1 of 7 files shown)


lute/cli/src/compile.cpp
```diff
@@ -1,121 +1,323 @@
...
+        entryPointPath = luauFilePath;
     }
 
-    std::streampos fileSize = exeFile.tellg();
-    if (fileSize < static_cast<std::streampos>(MAGIC_FLAG_SIZE + BYTECODE_SIZE_FIELD_SIZE))
+    filePaths.push_back(luauFilePath);
+    return true;
+}
+
+std::string LuteExePayload::getInjectablePayload() const
```
vrn-sn: Wonder if we can refactor this at all to make it easy to add unit tests for the uncompressed encoding format? Maybe not returning errors directly to `stderr` and splitting up the uncompressed bundle creation from the compression for testing.

Also, maybe as a future work item, but we should have a way for users of `lute compile` to tweak the default `copts()` used for compilation.

Vighnesh-V: Agreed on the future work item - I'll add it as an issue, since maybe users want to override compile optimizations etc.

On the first 1 - since the compressed payload is just bytes, it doesn't care about the lute runtime, so we can just cat it out to a file and then run an expect test. E.g. given local.luau, bundle this up and compare it to the bundle on disk.

Re: testing the bundle format - I think the right thing to do here is make `LuteExePayload` support decoding and encoding given a byte array that represents what we concatenated on. That way we can just call `payload.decode(payload.encode())` and check that it round trips properly

Vighnesh-V: Ok, I've gone ahead and updated this with some tests - there's a round trippable test now. I suspect the format of the bundle will change a bit in the upcoming weeks, so I'll avoid writing an integration test that compares against a dump on disk until the format stabilizes.

tests/src/compile/main.luau
```diff
@@ -0,0 +1,5 @@
+-- Simple test file for compile bundling
```
aatxe, line 1: Surely we should have a multifile bundle in the tests, right?

Vighnesh-V: We have to do a little more work to make requires actually work for multi file bundles. I can make it bundle multiple files just for the purpose of the test though.

Vighnesh-V:
I can make this work, but I'm going to split my new PR up a bunch.

PR #393, opened 2025-10-06
stdlib: Introduce a small standard library utility for writing tests in lute

Vighnesh-V:
Based on [frktest](https://github.com/itsfrank/frktest)

This testing utility needs a bunch of assertions to be more useful, but this PR just scopes out what a simple, builtin testing api in Luau could look like.

For some tests that look like:
```
local check = test.assert
test.case("foo", function()
	   check.eq(3, 3)
	   check.eq(6, 1)
end)

test.case("bar", function()
	   
	   check.eq("a", "b")
end)

test.case("baz", function()
	   
	   check.neq("a", "b")
end)

test.suite("MySuite", function()
	test.case("subcase", function()
		check.eq(5 * 5, 3)
	end)
end)

test.run();
```

this will render as:
```
❯ ./build/xcode/debug/lute/cli/lute run examples/testing.luau

==================================================
TEST RESULTS
==================================================

Failed Tests (3):

❌ foo
   ./examples/testing.luau:5
        eq: 6 ~= 1

❌ bar
   ./examples/testing.luau:10
        eq: a ~= b

❌ MySuite.subcase
   ./examples/testing.luau:20
        eq: 25 ~= 3

--------------------------------------------------
Total:  4
Passed: 1 ✓
Failed: 3 ✗
==================================================
```
Currently reports:
- [x] Test suites and cases
- [x] Line numbers
- [x] Stack traces
- [x] The expression that failed.

### DIFF
--- a/lute/std/libs/test.luau
+++ b/lute/std/libs/test.luau
@@ -1,0 +1,166 @@
+--!strict
+
+type Test = () -> ()
+type TestCase = { name: string, case: Test }
+type TestCases = { TestCase }
+type TestSuite = { name: string, cases: TestCases }
+
+type FailedTest = {
+	test: string, -- name of the test case that failed
+	assertion: string, -- name of the assertion that failed (e.g., "assert.eq")
+	error: string, -- the error message from the assertion
+	filename: string, -- source file where the failure occurred
+	linenumber: number, -- line number of the failure
+}
+
+type TestRunResult = {
+	failures: { FailedTest },
+	total: number,
+	failed: number,
+	passed: number,
+}
+
+type TestReporter = (TestRunResult) -> ()
+
+local function printFailedTest(failed: FailedTest)
+	print(`❌ {failed.test}`)
+	print(`   {failed.filename}:{failed.linenumber}`)
+	print(`   	{failed.assertion}: {failed.error}`)
+	print()
+end
+
+local function simpleReporter(result: TestRunResult)
+	print("\n" .. string.rep("=", 50))
+	print("TEST RESULTS")
+	print(string.rep("=", 50))
+
+	if #result.failures > 0 then
+		print(`\nFailed Tests ({#result.failures}):\n`)
+		for _, failed in result.failures do
+			printFailedTest(failed)
+		end
+	end
+
+	print(string.rep("-", 50))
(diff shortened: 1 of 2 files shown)


vrn-sn (approved):

lute/std/libs/test.luau
```diff
@@ -0,0 +1,166 @@
+--!strict
+
+type Test = () -> ()
+type TestCase = { name: string, toRun: Test }
```
```suggestion
type TestCase = { name: string, case: Test }
```

luacase properties in API, so we should have a better name than toRun.

lute/std/libs/test.luau
```diff
@@ -0,0 +1,166 @@
...
+type TestCases = { TestCase }
+type TestSuite = { name: string, cases: TestCases }
+
+type FailedTest = {
+	testName: string, -- name of the test case that failed
+	assertName: string, -- name of the assertion that failed (e.g., "assert.eq")
+	errorMessage: string, -- the error message from the assertion
+	fileName: string, -- source file where the failure occurred
+	lineNumber: number, -- line number of the failure
```
```suggestion
	test: string, -- name of the test case that failed
	assertion: string, -- name of the assertion that failed (e.g., "assert.eq")
	error: string, -- the error message from the assertion
	filename: string, -- source file where the failure occurred
	line: number, -- line number of the failure
```

though i feel like line should actually be a location, similar to how the parser library gives locations as spans

lute/std/libs/test.luau
```diff
@@ -0,0 +1,166 @@
...
+}
+
+type TestRunResult = {
+	failedTests: { FailedTest },
+	totalTests: number,
+	testsFailed: number,
+	testsPassed: number,
```
```suggestion
	failures: { FailedTest },
	total: number,
	failed: number,
	passed: number,
```

lute/std/libs/test.luau
aatxe: I think this one actually isn't externally visible, so less important, but still.

lute/std/libs/test.luau
Vighnesh-V: Yeah, the Luau debug api doesn't provide column numbers unfortunately :(

Vighnesh-V: I've changed the names of these fields though!

aatxe (approved):

PR #396, opened 2025-10-08
infra: make some minor improvements to the bootstrap script and the build system

aatxe:
This PR is a bit of a grab bag of small fixes to both the build system and the bootstrap script. The important fix here is that we force `CURL_ZSTD` off so that we won't have linking failures if you _happen_ to have zstd present on your system for other reasons. Otherwise, we tweak the message a bit when printing the Lute version in CMake, and tweak the bootstrap script to echo the key commands it's running to make the whole output from it easier to interpret. We also change the default behavior for the bootstrap script to assume something vaguely POSIX-y without specialized needs, which makes the script work on additional Unix platforms like FreeBSD. Finally, we change `bootstrapped-lute` to `lute0` for the initial version of lute generated without the standard library embedded in it because the term `bootstrapped` here feels confusing as it reads like it's the _final_ version, rather than an intermediary. `lute0` is "the first build" and, god help us, if our process is ever multistaged for bootstrapping, it has an obvious extension to `lute1`, `lute2`, etc. This may be reasonable to do later for validation if only to make sure that `lute1` and `lute2` are the same executable, byte for byte.

### DIFF
--- a/tools/bootstrap.sh
+++ b/tools/bootstrap.sh
@@ -8,7 +8,7 @@ fetch_dep() {
 
   # Ensure the file exists
   if [[ ! -f "$dep_file" ]]; then
-    echo "(ERR) dependency not found: $dep_file"
+    echo "(EE) dependency not found: $dep_file"
     return 1
   fi
 
--- a/tools/bootstrap.sh
+++ b/tools/bootstrap.sh
@@ -1,11 +1,14 @@
 set -e
 
+# function to display commands before running them
+exe() { echo ": $@" ; "$@" ; }
+
 fetch_dep() {
   local dep_file="$1"
 
   # Ensure the file exists
   if [[ ! -f "$dep_file" ]]; then
-    echo "Dependency file not found: $dep_file"
+    echo "(ERR) dependency not found: $dep_file"
     return 1
   fi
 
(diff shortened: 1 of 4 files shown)


tools/bootstrap.sh
```diff
@@ -52,49 +55,46 @@ done
...
 elif [[ "$os_type" == MINGW* || "$os_type" == MSYS* || "$os_type" == CYGWIN* ]]; then
-    BUILD_PATH=build/vs2022/debug
-    EXE_PATH+=".exe"
-    OUT_BINARY+=.exe
+  BUILD_PATH=build/vs2022/debug
+  EXE_PATH+=".exe"
+  OUT_BINARY+=.exe
 else
-  echo "Unknown OS: $os_type, cannot bootstrap"
-  return 1
```
hgoldstein, line 80: Why are we killing this?

hgoldstein (approved):
High level feedback:
* This is getting complex enough that I wonder if you'd be better off making it a Python script? On the other hand it is _mostly_ just executing programs.
* I think you should just `set -x` at the top if you're going to start outputting commands you're running. That's what it's for.

aatxe:
`set -x` has terrible semantics, it prints every single line of the script. We only really want to print the _actual_ commands that run in this case.

Vighnesh-V (approved):
Only one small nit : can we update the README to describe prioritizing the foreman install over using bootstrap?

tools/bootstrap.sh
aatxe: discussed in person, but the short is "we should just have a sane default behavior here, this isn't a thing worth having a loud failure over"

PR #397, opened 2025-10-08
clicommands: typedef setup command should try to create the full path where the typedefs are stored

Vighnesh-V:
I ran into a small bug with this script where I had `~/.lute` on my device, but not the rest of the path. The setup script only checks if this directory exists and assumes that the rest of the path is there if `~/.lute` is there. This can cause errors with writing the type def files because the subpaths won't exist.

This change also refactors the code that makes the subdirs by letting you pass the full path and then making the subdirectories as you go.

### DIFF
--- a/lute/cli/commands/setup/init.luau
+++ b/lute/cli/commands/setup/init.luau
@@ -4,14 +4,19 @@ local process = require("@lute/process")
 
 local homedir = process.homedir()
 
--- Already exists
+function mkdirdashp(home: string, subpath: string)
+	local subdir = home
+	for _, part in string.split(subpath, "/") do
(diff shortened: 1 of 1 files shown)


vrn-sn (approved):
longer term, wonder if `fs` should just support `mkdir -p` natively.

Vighnesh-V:
We should probably provide a helper for that, bcause the raw mkdir syscall doesn't actually support the -p argument: https://man7.org/linux/man-pages/man2/mkdir.2.html (at least on linux).

vrn-sn:
I was talking about the user command: https://man7.org/linux/man-pages/man1/mkdir.1.html.

PR #401, opened 2025-10-08
Prevent merging failing PRs by making CI aggregator fail when `build`, `test`, or `check-format` jobs fail

vrn-sn:
### Problem

Our CI aggregator job was unconditionally running and passing. This wasn't too much of an issue because contributors weren't merging their PRs until all checks passed, but PRs set to auto-merge were being merged automatically.

Because of this, some PRs introduced bugs that got merged in: #386 and #396.

### Solution

The aggregator job's now depends on the success of the `build`, `test`, and `check-format` jobs. If any job fails, the aggregator fails, which will block PRs from being merged.

The bug in #396 was fixed in #400.

To fix the bug introduced in #386, this PR also reverts that PR's changes to our GitHub Actions setup (using the bootstrap script to create a `lute` binary). We already have Foreman installed on these runners and there's no need to use bootstrapping here—in fact, it makes each CI job quite a bit slower since two fresh builds are needed in order to run anything.

### DIFF
--- a/.github/actions/setup-and-build/action.yml
+++ b/.github/actions/setup-and-build/action.yml
@@ -60,33 +60,16 @@ runs:
         path: extern
         key: extern-${{ hashFiles('extern/*.tune') }}
 
-    # We can currently only bootstrap correctly in ci on mac/linux
-    - name: Bootstrap Lute (Darwin/Linux only)
-      if: runner.os != 'Windows'
-      shell: bash
-      run: |
-        ./tools/bootstrap.sh
-        correct_lute=$(./build/lute0 tools/luthier.luau build lute --config debug --which)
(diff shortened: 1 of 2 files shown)


aatxe (approved):

vrn-sn:
Got Gated Commits blocking my PR. Now I need to actually fix whatever broke in the last few days.

aatxe:
It's the bootstrap script:

```
mv: rename /Users/runner/work/lute/lute/build/xcode/debug/lute/cli/lute to ./build/lute: No such file or directory
```

vrn-sn:
Yeah, I just spent some time undoing the part of #386 that touched our workflow files. We're already using Foreman in our actions, so I don't really see why we need to bootstrap.

PR #402, opened 2025-10-08
Infra: stop running in-progress CI workflow when a new commit is pushed

vrn-sn:
We use [concurrency groups](https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/control-workflow-concurrency) to cancel in-progress CI runs when they're invalidated by a new commit.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,6 +6,10 @@ on:
   pull_request:
     branches: ["primary"]
 
+concurrency:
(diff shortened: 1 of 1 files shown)


vrn-sn:
Interrupt test (last commit just above this comment) works as intended. The initial workflow run for the first commit has explicitly failed instead of continuing to run in parallel.

hgoldstein (approved):

PR #404, opened 2025-10-09
[net.serve] Integrate app with main libuv event loop to prevent blocking the task scheduling system

### LINKED ISSUE #384: net.serve blocks the task scheduler without yielding the current coroutine
Net.serve appears to block the task scheduler until the server is closed. However - it does not yield the current coroutine. This means any subsequent defer to the task scheduler after calling `net.serve` will hang indefinitely. I don't think this is expected behavior.

```lua
net.serve({
    port = 8081,
    handler = function(request)
        return {
            status = 200
        }
    end
})
print(1) -- prints
task.wait()
print(2) -- does not print
```

It looks like we are scheduling the http server in the current thread (`run()` loops continuously and does not return).

```cpp
    state->loopFunction = [state]()
    {
        if (!state->running)
        {
            return;
        }
        Luau::visit(
            [](auto* appPtr)
            {
                if (appPtr)
                    appPtr->run();
            },
            state->app
        );
        state->runtime->schedule(state->loopFunction);
    };

    serverInstances[serverId] = std::move(app);
    serverStates[serverId] = state;

    runtime->schedule(state->loopFunction);
```

I think we should be either stepping the http server via the task scheduler, or if it needs to be a blocking loop, running it on a separate thread.

vrn-sn:
Resolves #384.

### Problem

The current `net.serve` logic is to schedule a continuation that runs every task scheduler step, which (1) explicitly calls `run()` on our `uWS::TemplatedApp<T>`, and (2) schedules itself to run the next frame. This would work if `run()` were actually `step()`, but uWS doesn't provide an easy way to step an app.

### Solution

After some investigation, because uWS is configured to use libuv under the hood, we can patch in the libuv event loop that our task system is managing by calling `uWS::Loop::get(uv_default_loop())` before creating the app object. We get many things for free here (e.g. reference counting for keeping the server alive, managed by libuv), and it's simpler to reason about. This also means we don't need to explicitly "run" the app; it simply gets registered to run on the same event loop.

To ensure this works, some minor changes were made to the task scheduler logic to ensure that we are consuming events from the libuv event loop if it is live and to ensure we aren't blocking on I/O.

### DIFF
--- a/lute/runtime/src/runtime.cpp
+++ b/lute/runtime/src/runtime.cpp
@@ -50,12 +50,16 @@ Runtime::~Runtime()
 
 bool Runtime::hasWork()
 {
-    return hasContinuations() || hasThreads() || activeTokens.load() != 0;
+    // TODO: activeTokens and uv_loop_alive have a decent amount of overlap.
+    // Unfortunately, we do currently have some places where we add/release
+    // tokens that don't correspond to libuv activity, so for now we keep both.
+    // uv_ref/unref could be used to patch tokens into the libuv loop itself.
+    return hasContinuations() || hasThreads() || activeTokens.load() != 0 || uv_loop_alive(uv_default_loop());
 }
 
 RuntimeStep Runtime::runOnce()
 {
-    uv_run(uv_default_loop(), UV_RUN_DEFAULT);
+    uv_run(uv_default_loop(), UV_RUN_NOWAIT);
 
     // Complete all C++ continuations
     std::vector<std::function<void()>> copy;
--- a/lute/net/src/net.cpp
+++ b/lute/net/src/net.cpp
@@ -4,12 +4,16 @@
 
 #include "curl/curl.h"
 #include "App.h"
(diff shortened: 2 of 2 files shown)


lute/runtime/src/runtime.cpp
```diff
@@ -50,12 +50,16 @@ Runtime::~Runtime()
 
 bool Runtime::hasWork()
 {
-    return hasContinuations() || hasThreads() || activeTokens.load() != 0;
+    // TODO: activeTokens and uv_loop_alive have a decent amount of overlap.
+    // Unfortunately, we do currently have some places where we add/release
+    // tokens that don't correspond to libuv activity, so for now we keep both.
+    // uv_ref/unref could be used to patch tokens into the libuv loop itself.
```
aatxe, line 56: this sounds like a good way forward, i think.

aatxe (approved):

Vighnesh-V (approved):

PR #406, opened 2025-10-09
bugfix: Lute's testing framework swaps tests passed and tests failed

Vighnesh-V:
We were printing the wrong result to stdout for tests passed vs failed.


skanosue (approved):
signal maxing 🤯

PR #407, opened 2025-10-09
[CLI] Use require-by-string semantics during file discovery

### LINKED ISSUE #306: Use require-by-string-like semantics when discovering files to run
- We should auto-append `.luau` when finding files to run
- We should run directories with an `init.luau` inside of them

vrn-sn:
Resolves #306.

Enables discovering files to run by using require-by-string semantics.

The bulk of the changes here enable creating a `ModulePath` object, which handles all the require-by-string semantics for us (with some minor changes made to `ModulePath`'s API to easily extract whether we're pointing at a file or directory).

Other than this, I removed `#include <filesystem>` to remain consistent with how the rest of Lute handles these operations. I also changed up some function signatures:
```cpp
// old (mutates filePath directly)
bool checkValidPath(std::filesystem::path& filePath)

// new (returns an optional with the new value)
std::optional<std::string> getValidPath(std::string filePath)
```

If we are unable to find a file to run using require-by-string semantics, we still fallback to checking `./.lute/file.luau` and `./.lute/file.lua`.

See test for expected behavior: `require_by_string_semantics_in_cli`.


lute/require/include/lute/modulepath.h
```diff
@@ -12,9 +12,16 @@ enum class NavigationStatus
...
+    enum class Type
+    {
+        File,
+        Directory
+    };
+
     NavigationStatus status;
     std::string realPath;
     std::optional<std::string> relativePath;
+    Type type;
```
Vighnesh-V: nit: PathType is a bit more descriptive.

Vighnesh-V (commented):
Left some small comments.

lute/require/include/lute/modulepath.h
vrn-sn: Done!

lute/cli/src/climain.cpp
```diff
@@ -211,38 +210,61 @@ static int assertionHandler(const char* expr, const char* file, int line, const
...
+        if (resolved.status != NavigationStatus::Success)
+            return std::nullopt;
 
-    // if the file has an explicit extension, dont do a fallback
-    if (filePath.has_extension())
+        if (resolved.type == ResolvedRealPath::PathType::File)
+            return resolved.realPath;
+
+        return std::nullopt;
+    };
```
aatxe: Does this need to be a lambda? Can't it just be a helper function? It doesn't seem like it captures anything at all.

aatxe (commented):
Overall looks good to me, but one bit that seems weird.

lute/cli/src/climain.cpp
vrn-sn: I'll change it, this was a remnant of an earlier version of the change.

aatxe (approved):

lute/cli/src/climain.cpp
vrn-sn: basically, the old version of the code was doing these in-place mutations, so I had written the lambda to do some similar side-effecty thing. then I realized I'm not insane and decided to rewrite the whole thing to passing by value

aatxe: this ended up much better for it, so thanks for doing that then lol

Issue #321, opened 2025-06-05
API: `luau.requirefrompath`


JohnnyMorganz:
More of a question about what's the best way to do this. Or if it's not possible, how to make it possible.

I have a tool that needs to dynamically require Luau code that is provided on the command line:

```
lute tool.luau <path/to/loaded/file.luau>
```

In tool.luau, it works like:

```luau
local arguments = { ... }
local path = arguments[1]

local func = require(path)
func() -- `path` returns a function that can get called
```

Right now, this doesn't work as expected when a file path is given. Due to the way `require` works, the path given needs to be relative to the `tool.luau` file, rather than the current working directory where `lute` was executed. Also, absolute paths are not supported, so I cannot do `process.cwd() .. "/" .. path`

i.e., in the following situation:
```
- project
    tool.luau
- current
    apply.luau
```

Where my cwd is `current`: `lute ../project/tool.luau ./apply.luau` fails, but `lute ../project/tool.luau ../current/apply` works

vrn-sn:
Yes, `require` doesn't support this on its own. In the `Luau.Require` library, we do provide `proxyrequire`, which allows passing in a chunkname as a second argument, overriding the location of the requirer.

Currently, the chunkname `=stdin` should give you the behavior you want (it corresponds to a module located at the psuedo-path `./stdin`).

We could expose `proxyrequire` directly so that you can do something like this:
```luau
local arguments = { ... }
local path = arguments[1]

local func = proxyrequire(path, "=stdin")
func() -- `path` returns a function that can get called
```

Another (probably better) solution would be to abstract away the `=stdin` piece and expose some sort of `luau.requirefromcwd` function in Lute's bulit-in Luau library that does this internally.

vrn-sn:
We could also have a `luau.requirefrompath` that simply prepends a `@` to the user-provided path and passes it to `proxyrequire`, similar to what we did for `luau.load` (#301).

@aatxe, what do you think?

aatxe:
I think adding something to the luau library for `requirefromcwd` would be good. Really, either of those options in the luau library seems reasonable enough to me to try.

JohnnyMorganz:
Would these cases work if the provided path is an absolute path?

aatxe:
The `requirefrompath` would, presumably, or could at least be made to.

vrn-sn:
[...]

PR #411, opened 2025-10-09
stdlib: Update the testing api with some more test lifecycle methods + testing assertions

Vighnesh-V:
This PR comes with some changes to the testing api to reduce the amount of global state we depend on - when making subcases, we now pass in the suite object on which the user can operate. Additionally, assertions are passed in with the test case itself, making them more accessible and reducing an additional line of code needed to start writing tests.

- [x] {before,after}x{all,each} lifecycle methods
- [x] Table assertions 

For example: 
image

### DIFF
--- a/lute/std/libs/assert.luau
+++ b/lute/std/libs/assert.luau
@@ -70,6 +70,6 @@ function assertions.tableeq(lhs: { [unknown]: unknown }, rhs: { [unknown]: unkno
 	end
 end
 
-export type Assert = typeof(assertions)
+export type Asserts = typeof(assertions)
 
 return table.freeze(assertions)
--- a/lute/std/libs/assert.luau
+++ b/lute/std/libs/assert.luau
@@ -1,0 +1,20 @@
+-- strict
+
+local assert = {}
+-- todo: we should add some more assertions for diffing tables!
+function assert.eq<T>(lhs: T, rhs: T)
+	if lhs ~= rhs then
+		error({ msg = `{lhs} ~= {rhs}` })
+	end
+end
+
+function assert.neq<T>(lhs: T, rhs: T)
+	if lhs == rhs then
+		error({ msg = `{lhs} == {rhs}` })
+	end
+end
+
+export type Asserts = typeof(assert)
+
+table.freeze(assert)
+return assert
--- a/lute/std/libs/assert.luau
+++ b/lute/std/libs/assert.luau
@@ -1,7 +1,5 @@
--- strict
-
 local assert = {}
--- todo: we should add some more assertions for diffing tables!
+
 function assert.eq<T>(lhs: T, rhs: T)
 	if lhs ~= rhs then
 		error({ msg = `{lhs} ~= {rhs}` })
@@ -14,7 +12,62 @@ function assert.neq<T>(lhs: T, rhs: T)
 	end
 end
 
+-- Helper to format values for error messages
+local function formatValue(val: any): string
+	if type(val) == "nil" then
+		return "nil"
+	elseif type(val) == "string" then
+		return `"{val}"`
+	elseif type(val) == "table" then
+		return "table"
+	else
+		return tostring(val)
+	end
+end
+
+-- Deep table comparison helper
+local function tablesEqual(t1: any, t2: any, path: string): (boolean, string?)
+	if type(t1) ~= "table" or type(t2) ~= "table" then
+		if t1 ~= t2 then
+			return false, `lhs{path} = {formatValue(t1)} but rhs{path} = {formatValue(t2)}`
+		end
+		return true, nil
+	end
+	-- Check all keys in t1
+	for k, v1 in t1 do
+		local v2 = t2[k]
+		local newPath = if path == "" then `[{k}]` else `{path}[{k}]`
+
+		-- Check if key exists in t2
+		if v2 == nil and t2[k] == nil then
+			return false, `lhs{newPath} = {formatValue(v1)} but rhs{newPath} does not exist`
+		end
+
+		local equal, msg = tablesEqual(v1, v2, newPath)
+		if not equal then
+			return false, msg
+		end
+	end
+
+	-- Check for keys in t2 that aren't in t1
+	for k, v in t2 do
+		if t1[k] == nil then
+			local newPath = if path == "" then `[{k}]` else `{path}[{k}]`
+			return false, `rhs{newPath} = {formatValue(v)} but lhs{newPath} does not exist`
+		end
+	end
+
+	return true, nil
+end
+
+-- Table equality assertion with deep comparison
+function assert.tableeq(lhs: { [any]: any }, rhs: { [any]: any })
+	local equal, msg = tablesEqual(lhs, rhs, "")
+	if not equal then
+		error({ msg = msg or "tables are not equal" })
+	end
+end
+
 export type Asserts = typeof(assert)
 
-table.freeze(assert)
-return assert
+return table.freeze(assert)
--- a/lute/std/libs/assert.luau
+++ b/lute/std/libs/assert.luau
@@ -70,6 +70,6 @@ function assertions.tableeq(lhs: { [unknown]: unknown }, rhs: { [unknown]: unkno
 	end
 end
 
-export type Assertionss = typeof(assertions)
+export type Assert = typeof(assertions)
 
 return table.freeze(assertions)
--- a/lute/std/libs/assert.luau
+++ b/lute/std/libs/assert.luau
@@ -1,12 +1,12 @@
-local assert = {}
+local assertions = {}
 
-function assert.eq<T>(lhs: T, rhs: T)
+function assertions.eq<T>(lhs: T, rhs: T)
 	if lhs ~= rhs then
 		error({ msg = `{lhs} ~= {rhs}` })
 	end
 end
 
-function assert.neq<T>(lhs: T, rhs: T)
+function assertions.neq<T>(lhs: T, rhs: T)
 	if lhs == rhs then
 		error({ msg = `{lhs} == {rhs}` })
 	end
@@ -25,21 +25,22 @@ local function formatValue(val: any): string
 	end
 end
 
--- Deep table comparison helper
-local function tablesEqual(t1: any, t2: any, path: string): (boolean, string?)
-	if type(t1) ~= "table" or type(t2) ~= "table" then
+-- Structural equality for two tables
+local function tablesEqual(t1: unknown, t2: unknown, path: string): (boolean, string?)
+	if typeof(t1) ~= "table" or typeof(t2) ~= "table" then
 		if t1 ~= t2 then
 			return false, `lhs{path} = {formatValue(t1)} but rhs{path} = {formatValue(t2)}`
 		end
 		return true, nil
 	end
+
 	-- Check all keys in t1
 	for k, v1 in t1 do
-		local v2 = t2[k]
+		local v2 = (t2 :: any)[k]
 		local newPath = if path == "" then `[{k}]` else `{path}[{k}]`
 
 		-- Check if key exists in t2
-		if v2 == nil and t2[k] == nil then
+		if v2 == nil then
 			return false, `lhs{newPath} = {formatValue(v1)} but rhs{newPath} does not exist`
 		end
 
@@ -51,7 +52,8 @@ local function tablesEqual(t1: any, t2: any, path: string): (boolean, string?)
 
 	-- Check for keys in t2 that aren't in t1
 	for k, v in t2 do
-		if t1[k] == nil then
+		local v1 = (t1 :: any)[k]
+		if v1 == nil then
 			local newPath = if path == "" then `[{k}]` else `{path}[{k}]`
 			return false, `rhs{newPath} = {formatValue(v)} but lhs{newPath} does not exist`
 		end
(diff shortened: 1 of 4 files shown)


lute/std/libs/assert.luau
```diff
@@ -0,0 +1,20 @@
...
+function assert.neq<T>(lhs: T, rhs: T)
+	if lhs == rhs then
+		error({ msg = `{lhs} == {rhs}` })
+	end
+end
+
+export type Asserts = typeof(assert)
+
+table.freeze(assert)
+return assert
```
aatxe:
```suggestion
return table.freeze(assert)
```

lute/std/libs/assert.luau
```diff
@@ -0,0 +1,20 @@
+-- strict
```
Vighnesh-V: Pulled this out because i know we'll want a lot of different assertions here, more than just eq and neq.

lute/std/libs/assert.luau
```diff
@@ -0,0 +1,20 @@
...
+
+local assert = {}
+-- todo: we should add some more assertions for diffing tables!
```
aatxe:
```suggestion
local assert = {}

-- todo: we should add some more assertions for diffing tables!
```

`-- strict` does nothing, but also luaurc is already set to have all these files be strict.

lute/std/libs/test.luau
```diff
@@ -20,16 +36,14 @@ type TestRunResult = {
...
-
 local function printFailedTest(failed: FailedTest)
 	print(`❌ {failed.test}`)
 	print(`   {failed.filename}:{failed.linenumber}`)
 	print(`   	{failed.assertion}: {failed.error}`)
 	print()
 end
 
-local function simpleReporter(result: TestRunResult)
+function simpleReporter(result: TestRunResult)
```
aatxe: Why is this function no longer a local?

lute/std/libs/test.luau
```diff
@@ -48,56 +62,38 @@ local function simpleReporter(result: TestRunResult)
...
+-- Testing Environment
 type TestEnvironment = {
-	anonymous: TestCases,
+	anonymous: { TestCase },
 	suites: { TestSuite },
 	reporter: TestReporter,
-	active_suite: TestSuite?,
 }
 
 local env: TestEnvironment = { anonymous = {}, suites = {}, reporter = simpleReporter }
```
aatxe, line 108:
```suggestion
```

Dead code now, right?

lute/std/libs/test.luau
```diff
@@ -1,10 +1,26 @@
...
 type TestCase = { name: string, case: Test }
-type TestCases = { TestCase }
-type TestSuite = { name: string, cases: TestCases }
 
+type TestSuite = { name: string, cases: { TestCase } }
+type Suite = {
+	case: (string, Test) -> (),
+}
+
+local function makesuite(name: string): (TestSuite, Suite)
```
```suggestion
local function makeSuite(name: string): (TestSuite, Suite)
```

this isn't part of the public API, right? it can be camelCase then

lute/std/libs/test.luau
```diff
@@ -1,10 +1,26 @@
...
+	local data: TestSuite = { name = name, cases = {} }
+	return data,
+		{
+			case = function(name: string, case: Test)
+				table.insert(data.cases, { name = name, case = case })
+			end,
+		}
+end
```
Vighnesh-V: I'm not a huge fan of how this is done, but the reason is that I want users to be able to consistently write tests like:

`test.case, suite.case, test.suite`.

If you do OOP here, you end up having to pass self and it becomes inconsistent - suite:case instead of suite.case.
If you write this full OCaml closure style, the internal library isn't able to know the structure of the test suite class, unless you write getters and setters for the methods as well, which also means that folks trying to register test suites will also get those methods in autocomplete, which they don't need.

If you do this, then the external users of `Suite` only know how to register test cases, e.g. `suite.case` and it's consistent with the rest of the api provided. The internal tracker of registered TestSuites knows the structure too.

lute/std/libs/test.luau
```diff
@@ -48,56 +62,38 @@ local function simpleReporter(result: TestRunResult)
 	print(string.rep("=", 50) .. "\n")
 end
 
+export type TestReporter = (TestRunResult) -> ()
+
+-- Testing Environment
 type TestEnvironment = {
-	anonymous: TestCases,
+	anonymous: { TestCase },
```
Vighnesh-V, line 103: I didn't think the pluralized array alias was that helpful, so I clipped it.

aatxe: this is the right call

lute/std/libs/test.luau
aatxe: what if we go another way? can we have all of them be colon calls?

lute/std/libs/test.luau
Vighnesh-V, line 108: Open to any suggestions on getting rid of this global state - I guess we could either do something like:

```
test.start(function(env)
end
)
```
and then all your tests get registered and run immediately in this context, but then your tests get a little more nested.

Alternatively we could do something like:
```
local test = require('@std/test')
local env = test.fresh()
env.suite(
    ....
)

env.case (...)
env.run()
```

lute/std/libs/test.luau
Vighnesh-V: I moved reporters into their own file and then back into `test`.  I've made this local again.

lute/std/libs/test.luau
aatxe: nevermind, i see it

aatxe: I think the answer is that in an ideal world, we'd have the testing library just define tests that are discoverable, and a lute subcommand `lute test` would do the work of discovering the tests and running them.

Vighnesh-V: Still working on figuring out how to do this correctly. The current issue I'm having is that registering tests and reporting them relies on this TestEnvironment.

```
test.case()
test.suite(....)
test.suite(....)
test.case()
```

With the new api we won't necessarily be able to get a report of all the tests without this `env` - we can only run each suite and test case independently. We would need some global state to do this. The alternative is what I suggested in my comment down below.

lute/std/libs/test.luau
Vighnesh-V: ok, I tried the colon calls, but I feel like that complicates the testing library a lot + it gives more autocomplete results than expected. Given that this does provide a really nice api for writing tests, with good autocomplete support (because the `Suite` that is visible to users has a very restrictive api) , I'm more comfortable with this change now.

lute/std/libs/test.luau
Vighnesh-V: I had a quick chat with Varun, and I made a small start on writing the `lute test` command. Unfortunately, we still need to extend require as discussed here: https://github.com/luau-lang/lute/issues/321 in order to support require'ing of test files as we can't use things like loadstring. 

I guess for now, we can leave the test.run() method in place, along with the env variable. Once we have this command, we can implement test discovery pretty easily, by swapping the env variable with a reference to the global "testenv" that we can define just for the `test` CLI command. This means the test definition API won't change more than what I've already implemented in this PR, and we're setup to migrate to the builtin `lute test` command.

lute/std/libs/test.luau
Vighnesh-V: I revisited the colon calls this morning - I think I like that you could write your tests imperatively or declaratively - either works.

lute/std/libs/test.luau
Vighnesh-V: I have a prototype of `lute test` working via registering a global for the `test` command. `@std/test` knows how to write to this global, and `lute test` finds `.test.luau` files and requires them to discover the test and can execute them.

Vighnesh-V:
Once this PR gets merged in, I'll extract the reporters written here into the test discovery branch I have and rewrite them to use full colour output.

lute/std/libs/test.luau
```diff
@@ -48,56 +97,38 @@ local function simpleReporter(result: TestRunResult)
...
-function test.setreporter(reporter: TestReporter)
-	test.reporter = reporter
+function test.suite(name: string, registerFn: (TestSuite) -> ())
+	local suite = TestSuite.new(name)
+	registerFn(suite)
+	table.insert(env.suites, suite)
 end
 
+-- run all the tests
 function test.run()
```
Vighnesh-V, line 131: All of this logic will eventually be moved to the test discovery and reporting `lute` command.

lute/std/libs/test.luau
```diff
@@ -1,10 +1,61 @@
 --!strict
 local ps = require("@lute/process")
-type Test = () -> ()
+local assert = require("./assert")
+-- Test Types
+type BeforeEach = () -> ()
```
skanosue: unused?

lute/std/libs/test.luau
```diff
@@ -122,22 +153,44 @@ function test.run()
...
 	-- Run tests from suites
 	for _, suite in env.suites do
+		-- Run beforeall hook once per suite
+		if suite._beforeall then
+			local beforeallSuccess, beforeallErr = pcall(suite._beforeall)
+			if not beforeallSuccess then
+				-- If beforeall fails, skip all tests in the suite
+				for _, tc in suite.cases do
+					total += 1
+					failed += 1
```
skanosue: perf wise, i think these could be lifted out of the loop?

lute/std/libs/test.luau
```diff
@@ -147,10 +200,63 @@ function test.run()
...
+
+			-- Run aftereach hook if it exists (runs even if test failed)
+			if suite._aftereach then
+				local afterSuccess, afterErr = pcall(suite._aftereach)
+				if not afterSuccess then
+					-- If test passed but aftereach failed, mark as failed
+					if success then
+						passed -= 1
+						failed += 1
+					end
```
skanosue: you could move lines 221-223 to after `aftereach` is handled and set `success = false` here instead of having this extra check

lute/std/libs/test.luau
Vighnesh-V: yup, getting rid of this

lute/std/libs/test.luau
Vighnesh-V: fixed - just added the length of the suite's cases to the totals

lute/std/libs/assert.luau
```diff
@@ -0,0 +1,73 @@
...
+	elseif type(val) == "string" then
+		return `"{val}"`
+	elseif type(val) == "table" then
+		return "table"
+	else
+		return tostring(val)
+	end
+end
+
+-- Deep table comparison helper
```
aatxe:
```suggestion
-- Structural equality for two tables
```

lute/std/libs/assert.luau
```diff
@@ -0,0 +1,73 @@
...
+local function tablesEqual(t1: any, t2: any, path: string): (boolean, string?)
```
aatxe:
```suggestion
local function tablesEqual(t1: unknown, t2: unknown, path: string): (boolean, string?)
```

You should make this work with `unknown` if it doesn't already typecheck, unless there's a good reason (existing limitation in Luau) it has to be `any`.

lute/std/libs/assert.luau
```diff
@@ -0,0 +1,73 @@
...
+			local newPath = if path == "" then `[{k}]` else `{path}[{k}]`
+			return false, `rhs{newPath} = {formatValue(v)} but lhs{newPath} does not exist`
+		end
+	end
+
+	return true, nil
+end
+
+-- Table equality assertion with deep comparison
+function assert.tableeq(lhs: { [any]: any }, rhs: { [any]: any })
```
aatxe:
```suggestion
function assert.tableeq(lhs: { [unknown]: unknown }, rhs: { [unknown]: unknown })
```

Ditto on not doing `any`s here if we can.

Vighnesh-V: Fixed!

lute/std/libs/assert.luau
Vighnesh-V: it doesn't have to be any - I've set it to unknown. Unrelated, but there is a problem when you iterate through one table's keys and index into the other table (even though the refinement says that t2 is a table). I've fixed this with an any cast.

tests/smoke.test.luau
skanosue, line 1: Do we still need this smoke test?

skanosue (approved):

tests/smoke.test.luau
Vighnesh-V: nah, we don't. since we have actual tests now :)

Vighnesh-V: or rather, since you will introduce actual tests :)

Vighnesh-V: I'll delete this in a follow up pr if that's okay?

lute/std/libs/test.luau
```diff
@@ -1,10 +1,60 @@
...
+	_beforeeach: (() -> ())?,
+	_beforeall: (() -> ())?,
+	_aftereach: (() -> ())?,
+	_afterall: (() -> ())?,
+	case: (self: TestSuite, string, Test) -> (),
+	beforeeach: (self: TestSuite, () -> ()) -> (),
+	beforeall: (self: TestSuite, () -> ()) -> (),
+	aftereach: (self: TestSuite, () -> ()) -> (),
+	afterall: (self: TestSuite, () -> ()) -> (),
+}
```
aatxe, line 20: This type is in the actual interface for the API meaning that all of these internal `_beforeeach`, etcs. show up in autocomplete when you're writing things. We should be using an interface type that only presents the interface of a test suite to users, and then internally casting to an implementation type.

lute/std/libs/test.luau
```diff
@@ -122,34 +152,107 @@ function test.run()
...
+				for _, tc in suite.cases do
+					local failedtest: FailedTest = {
+						test = `{suite.name}.{tc.name}`,
+						assertion = "beforeall",
+						error = `beforeall hook failed: {beforeallErr}`,
+						filename = "",
+						linenumber = 0,
+					}
+					table.insert(failures, failedtest)
+				end
```
aatxe: Can we rewrite these using xpcall? It'll be cleaner:
```suggestion
		    xpcall(suite._beforeall, function(error)
				failed += 1
				local failedtest: FailedTest = {
					test = `{suite.name}`,
					assertion = "beforeall",
					error = `beforeall hook failed: {error}`,
					filename = "",
					linenumber = 0,
				}
				table.insert(failures, failedtest)
			end)
```

aatxe (changes requested):

lute/std/libs/test.luau
Vighnesh-V: done! Rewritten using xpcall instead.

Vighnesh-V:
Sorry about force push, but I went back and re-edited history. All future PR's will be merge only :)

aatxe (approved):

PR #416, opened 2025-10-10
stdlib: add `system` and boolean environment checks

### LINKED ISSUE #167: @std extension: `system.win32`, `system.unix`, `system.macos`
To avoid patterns like `string.find(string.lower(system.os), "windows") ~= nil`, I propose we add simple boolean flags for a variety of different environments, so you can do something like `if system.win32 then` without needing to worry about any specifics with names.

annieetang:
### Problem

There was no std lib for `system` related functionalities. 

### Solution

We introduce a `std` library analogue for `system`, where we can now check the environment with boolean flags.
Closes #167

To test (with huge help from @vrn-sn), we wanted to write a C++ file that compares the "os" captured by the C++ executable and compares to the `@std/system` OS. To write that (and other similar output comparison test cases), we exposed the `CliRuntimeFixture` and a `runCode()` function that was used by `require.test.cpp` before but we added a `capture()` function based on the `Luau/ReplFixture` class.

Then, we changed some function signatures to ensure the program args (`argc` and `argv`) are not statically passed to avoid some tests failing.

### Example:

```lua

if system.linux then
    print("Running on Linux")
end
```

See `examples/system_environment_check.luau` for example use case


lute/std/libs/system.luau
```diff
@@ -0,0 +1,19 @@
...
+local osName = string.lower(system.os)
+
+local win32 = osName == "windows"
+local linux = osName == "linux"
+local macos = osName == "darwin" or osName == "macos"
+local unix = macos or linux or osName == "unix" or osName == "posix"
+
+return table.freeze({
+	os = system.os,
+
```
aatxe, line 18: There's a bunch of _other_ functionality in `@lute/system` that is not exported here, but should be.

tests/system.test.luau
```diff
@@ -0,0 +1,18 @@
...
+	local win32 = osName == "windows"
+	local linux = osName == "linux"
+	local macos = osName == "darwin" or osName == "macos"
+	local unix = macos or linux or osName == "unix" or osName == "posix"
+
+	assert(stdSystem.win32 == win32)
+	assert(stdSystem.linux == linux)
+	assert(stdSystem.macos == macos)
+	assert(stdSystem.unix == unix)
+end
```
aatxe: feels like not a particularly useful test since it just duplicates exactly the definition of the code, i'd honestly say we shouldn't bother with this sort of test at all.

vrn-sn:
@annieetang if you do this in the PR description, the issue will be automatically closed when this merges: https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue.

tests/system.test.luau
Vighnesh-V: We also have a builtin testing library at @std/test which you can use for this, but I sort of agree. I think a way to write this test might be writing some C++ in the `Lute.Test` target that invokes cliMain in such a way that runs this file. Then it grabs either stdout / some variable written to by this test and checks that it matches the "os" returned by the C+++ executable.

lute/std/libs/system.luau
```diff
@@ -0,0 +1,32 @@
...
+-- stdlib for `@lute/system`
+-- Provides re-exports of `@lute/system` and boolean properties for OS detection
+
+local system = require("@lute/system")
+
+local osName = string.lower(system.os)
+
+local win32 = osName == "windows" or osName == "windows_nt"
+local linux = osName == "linux"
+local macos = osName == "darwin" or osName == "macos"
```
annieetang: I think it might be just `darwin` and `windows_nt` but i'm not super sure

annieetang: actually i'll just change it to those two for now?

tests/system.test.luau
annieetang: added! big thanks to varun 🤩

lute/runtime/src/runtime.cpp
```diff
@@ -334,7 +334,7 @@ static void luteopen_libs(lua_State* L)
     }
 }
 
-lua_State* setupState(Runtime& runtime, void (*doBeforeSandbox)(lua_State*))
+lua_State* setupState(Runtime& runtime, std::function<void(lua_State*)> doBeforeSandbox)
```
Vighnesh-V, line 337: Good call on making this a std function!

lute/cli/include/lute/climain.h
```diff
@@ -1,7 +1,10 @@
 #pragma once
+#include <functional>
+#include <string>
 
 struct lua_State;
 struct Runtime;
 
-lua_State* setupCliState(Runtime& runtime);
+lua_State* setupCliState(Runtime& runtime, std::function<void(lua_State*)> testInit = nullptr);
```
Vighnesh-V: Can we change the `testInit` argument to `preSandboxInit`? Since there are probably use cases for this beyond just testing.

lute/cli/src/climain.cpp
```diff
@@ -27,9 +29,6 @@
 #include <string>
 #include <vector>
 
-static int program_argc = 0;
-static char** program_argv = nullptr;
```
Vighnesh-V, line 31: Nice - good call on removing these

Vighnesh-V (changes requested):
Left some small comments. Can we also clang-tidy this PR?

Vighnesh-V (approved):

vrn-sn:
I think we missed splitting `cliruntimefixture.h` into a header and a source file. It's mostly okay, but I think we should still do it. There's a `static` function in there that will be duplicated each time the header is included.

PR #417, opened 2025-10-10
infra: Set up the ability to write .test.luau files that get automatically executed by our CI

### LINKED ISSUE #410: Create standard Luau tests and a GitHub Actions job to run them on CI
Similar to Lute.Test, which currently just runs C++ tests. Could either make a new CI job for this or just have Lute.Test run Luau tests from its main.cpp's `main` function. Not sure if defining `main` would prevent the C++ tests from running, but it's easy enough to figure something out that works.

Vighnesh-V:
This implements something like what is described here: https://github.com/luau-lang/lute/issues/410 , which allows us to write test cases in luau using the std/test library. We can build lute for all the different platforms we support and then run the lute tests as part of a separate pass.

Test Failure documented here: https://github.com/luau-lang/lute/actions/runs/18421725097/job/52496749349?pr=417

Resolves #410.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,12 +30,25 @@ jobs:
       - uses: actions/checkout@v4
 
       - name: Setup and Build Lute
+        id: build_lute
         uses: ./.github/actions/setup-and-build
         with:
           target: Lute.CLI
           config: debug
           options: ${{ matrix.options }}
           token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Remove .luaurc file
+        run: rm .luaurc
+
+      - name: Run Luau Tests
+        shell: bash
+        run: |
+          if [ "${{ runner.os }}" == "Windows" ]; then
+            find tests -name "*.test.luau" | xargs -I {} ${{ steps.build_lute.outputs.exe_path }} run {}
+          else
+            find tests -name "*.test.luau" | xargs -I {} ${{ steps.build_lute.outputs.exe_path }} run {}
+          fi
 
   test:
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -56,21 +56,54 @@ jobs:
...
+        with:
+          target: Lute.CLI
+          config: debug
+          options: ${{ matrix.options }}
+          token: ${{ secrets.GITHUB_TOKEN }}
+      - name: Remove .luaurc file
+        run: rm .luaurc
+
+      - name: Run Tests
+        run: find tests -name "*.test.luau" | xargs -I {} ${{ steps.build_tests.outputs.exe_path }} run {}
```
Vighnesh-V: When we are able to support a `lute test` command, this can be turned into `lute test`

.github/workflows/ci.yml
```diff
@@ -56,21 +56,54 @@ jobs:
...
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Setup Luau - Lute Unit Tests
+        id: build_tests
+        uses: ./.github/actions/setup-and-build
+        with:
+          target: Lute.CLI
```
vrn-sn: the `build` job builds the exact same artifact above, can we use a workflow artifact to avoid the shared work?
https://docs.github.com/en/actions/tutorials/store-and-share-data

maybe it makes sense to build into the logic of `setup-and-build` since it's a helper action anyway

.github/workflows/ci.yml
```diff
@@ -89,7 +122,7 @@ jobs:
   aggregator:
     name: Gated Commits
     runs-on: ubuntu-latest
-    needs: [build, test, check-format]
+    needs: [build, test, luau-lute-test, check-format]
```
vrn-sn: also need to update the aggregator's success logic under `run`

.github/workflows/ci.yml
vrn-sn: i.e. we could add another job that simply builds and stores the artifact, and have both these jobs `needs:` it

vrn-sn (approved):
Tiny nit: could we rename the build and test jobs now that both are actually doing some form of testing (and update their names in the aggregator as well).

Issue #251, opened 2025-05-02
path library


JohnnyMorganz:
We should expose a portable mechanism for working with file system paths, similar to Python os.path / pathlib or node path

ghostnaps:
I'm most familiar with the NodeJS Path API so I'll just toss in my wishlist of what I'd love to see out of this:
1. [path.basename](https://nodejs.org/docs/v22.16.0/api/path.html#pathbasenamepath-suffix)
1. [path.dirname](https://nodejs.org/docs/v22.16.0/api/path.html#pathdirnamepath)
1. [path.extname](https://nodejs.org/docs/v22.16.0/api/path.html#pathextnamepath)
1. [path.join](https://nodejs.org/docs/v22.16.0/api/path.html#pathjoinpaths)
1. [path.normalize](https://nodejs.org/docs/v22.16.0/api/path.html#pathnormalizepath)
1. [path.resolve](https://nodejs.org/docs/v22.16.0/api/path.html#pathresolvepaths)

I would also find it useful to have an equivalent of `__dirname` so I can construct paths relative to the current file.

annieetang:
Closing since @skberkeley addressed everything!

PR #427, opened 2025-10-14
stdlib: filepath library 

skanosue:
I've implemented a filepath library loosely inspired by `node.path` and Python's `pathlib` to address #251. It consists of two sublibraries specialized for win32 and posix paths, which are each called based on the system. These sublibraries can also be specifically accessed in the case that some poor soul on a Unix machine wants to manipulate DOS filepaths. I've tried my best to match the behavior to the `node.path` library.

The tests are in an outdated format; I'll update them once @~Vighnesh-V's changes are in.


lute/std/libs/path/win32.luau
```diff
@@ -0,0 +1,271 @@
...
+function isabsolute(path: WindowsPathLike): boolean
+	if typeof(path) == "string" then
+		path = parse(path)
+	end
+	path = path :: WindowsPath
+	return path.isAbsolute
+end
+
+-- In place extends path with addend
+function _joinHelper(path: WindowsPath, addend: WindowsPathLike): ()
```
Vighnesh-V: `local function joinHelper`

lute/std/libs/path/win32.luau
```diff
@@ -0,0 +1,271 @@
+export type WindowsPath = {
+	parts: { string },
+	isAbsolute: boolean,
+	isUNCPath: boolean,
+	driveLetter: string?,
+}
```
Vighnesh-V: @aatxe Is the convention luacase for fields on table types?

lute/std/libs/path/win32.luau
```diff
@@ -0,0 +1,271 @@
...
+			return "\\\\" .. parts
+		elseif path.driveLetter ~= nil then
+			return table.concat({ path.driveLetter, ":", (if path.isAbsolute then "\\" else ""), parts })
+		else
+			return parts
+		end
+	end
+end
+
+function isabsolute(path: WindowsPathLike): boolean
```
Vighnesh-V: `local function`

lute/std/libs/path/posix.luau
```diff
@@ -0,0 +1,204 @@
...
+	end
+
+	return {
+		parts = parts,
+		isAbsolute = isAbs,
+	}
+end
+
+function resolve(parts: { PosixPathLike }): string
+	error("TODO")
```
Vighnesh-V: Do you need to fill this in?

lute/std/libs/path/posix.luau
```diff
@@ -0,0 +1,204 @@
...
+function isabsolute(path: PosixPathLike): boolean
+	if typeof(path) == "string" then
+		path = parse(path)
+	end
+	path = path :: PosixPath
+	return path.isAbsolute
+end
+
+-- In place extends path with addend
+function _joinHelper(path: PosixPath, addend: PosixPathLike): ()
```
Vighnesh-V: joinHelper

Vighnesh-V (dismissed):
High level style feedback mostly - the actual code here looks good, and this is such a nice utility to have :)

Style conventions:
1. `local function` over `function`. Anything that's publicly visible will be something you'll explicitly export.
2. Field name are luacase (I think)
3. helper methods that are won't be exported can be `camelCase`.

Once my testing pr makes it in, we'll also have structural table diffing, so that might be useful for the tests you've written :)

lute/std/libs/path/posix.luau
Vighnesh-V: just saw the PR description

Vighnesh-V:
Q1: Can you use system.cwd to get the current directory you need for now? Or will that just be in the lute/lute/std/libs/path/... dir always, instead of the directory from which you are running?

Vighnesh-V:
Q2: Can we/should we do something clever with operator overloading and the  `//` or `..` or `/` operator to provide a first class experience for path construction? We could do something clever like implement the div metamethod for this struct.

E.g.:

```
local result = path.home
result = result /  "somedir" / "another" / "test.luau"
```

skanosue:
> Q1: Can you use system.cwd to get the current directory you need for now? Or will that just be in the lute/lute/std/libs/path/... dir always, instead of the directory from which you are running?

I don't think `system.cwd` exists? We need extra C++ work to implement it

lute/std/libs/path/win32.luau
skanosue: We appear to be doing snake case lol: https://github.com/luau-lang/lute/blob/7b4e269ba4652442900875f1085150569d38388b/definitions/fs.luau#L15

skanosue:
> Q2: Can we/should we do something clever with operator overloading and the `//` or `..` or `/` operator to provide a first class experience for path construction? We could do something clever like implement the div metamethod for this struct.

Sure, I can do that! I'll model it after [Python's operator overload](https://docs.python.org/3/library/pathlib.html#operators)

lute/std/libs/path/win32.luau
aatxe: all of it should be luacase.

skanosue:
> Sure, I can do that! I'll model it after [Python's operator overload](https://docs.python.org/3/library/pathlib.html#operators)

Nvm, the community does not seem to want this lol

lute/std/libs/path/win32.luau
skanosue: nvm, we are doing lua case

lute/std/libs/path/init.luau
```diff
@@ -0,0 +1,96 @@
+local system = require("@lute/system")
+
+local posix = require("@self/posix")
+local win32 = require("@self/win32")
+
+export type PathObj = posix.PosixPath | win32.WindowsPath
+
+export type PathLike = string | PathObj
```
aatxe:
```suggestion
export type path = posix.PosixPath | win32.WindowsPath

export type pathlike = string | PathObj
```

everything exported in the standard library should be in luacase.

aatxe (changes requested):

lute/std/libs/path/init.luau
```diff
@@ -0,0 +1,96 @@
+local system = require("@lute/system")
+
+local posix = require("@self/posix")
+local win32 = require("@self/win32")
+
+export type pathobj = posix.path | win32.path
```
aatxe:
```suggestion
export type path = posix.path | win32.path
```

I don't think we should be abbreviating in names, and I don't understand what `object` actually adds for us here.

lute/std/libs/path/init.luau
```diff
@@ -0,0 +1,96 @@
...
+
+export type pathlike = string | pathobj
+
+local onWindows = system.os == "windows"
+
+function basename(path: pathlike): string?
```
aatxe:
```suggestion
local function basename(path: pathlike): string?
```

lute/std/libs/path/init.luau
```diff
@@ -0,0 +1,96 @@
...
+	if onWindows then
+		return win32.basename(path :: win32.pathlike)
+	else
+		return posix.basename(path :: posix.pathlike)
+	end
+end
+
+function dirname(path: pathlike): string
```
aatxe:
```suggestion
local function dirname(path: pathlike): string
```

lute/std/libs/path/init.luau
```diff
@@ -0,0 +1,96 @@
...
+	if onWindows then
+		return win32.dirname(path :: win32.pathlike)
+	else
+		return posix.dirname(path :: posix.pathlike)
+	end
+end
+
+function extname(path: pathlike): string
```
aatxe:
```suggestion
local function extname(path: pathlike): string
```

lute/std/libs/path/init.luau
```diff
@@ -0,0 +1,96 @@
...
+	if onWindows then
+		return win32.extname(path :: win32.pathlike)
+	else
+		return posix.extname(path :: posix.pathlike)
+	end
+end
+
+function format(path: pathlike): string
```
aatxe:
```suggestion
local function format(path: pathlike): string
```

lute/std/libs/path/init.luau
```diff
@@ -0,0 +1,96 @@
...
+	if onWindows then
+		return win32.format(path :: win32.pathlike)
+	else
+		return posix.format(path :: posix.pathlike)
+	end
+end
+
+function isabsolute(path: pathlike): boolean
```
aatxe:
```suggestion
local function isabsolute(path: pathlike): boolean
```

lute/std/libs/path/init.luau
```diff
@@ -0,0 +1,96 @@
...
+	if onWindows then
+		return win32.isabsolute(path :: win32.pathlike)
+	else
+		return posix.isabsolute(path :: posix.pathlike)
+	end
+end
+
+function join(...: pathlike): pathobj
```
aatxe:
```suggestion
local function join(...: pathlike): pathobj
```

lute/std/libs/path/init.luau
```diff
@@ -0,0 +1,96 @@
...
+	if onWindows then
+		return win32.join(...)
+	else
+		return posix.join(...)
+	end
+end
+
+function normalize(path: pathlike): pathobj
```
aatxe:
```suggestion
local function normalize(path: pathlike): pathobj
```

lute/std/libs/path/init.luau
```diff
@@ -0,0 +1,96 @@
...
+	if onWindows then
+		return win32.normalize(path :: win32.pathlike)
+	else
+		return posix.normalize(path :: posix.pathlike)
+	end
+end
+
+function parse(path: pathlike): pathobj
```
aatxe:
```suggestion
local function parse(path: pathlike): pathobj
```

lute/std/libs/path/init.luau
```diff
@@ -0,0 +1,96 @@
...
+	if onWindows then
+		return win32.parse(path :: win32.pathlike)
+	else
+		return posix.parse(path :: posix.pathlike)
+	end
+end
+
+function resolve(...: pathlike): string
```
aatxe:
```suggestion
local function resolve(...: pathlike): string
```

lute/std/libs/path/win32.luau
```diff
@@ -0,0 +1,282 @@
+local process = require("@lute/process")
+
+export type pathKind = "unc" | "absolute" | "relative"
```
aatxe:
```suggestion
export type pathkind = "unc" | "absolute" | "relative"
```

lute/std/libs/path/win32.luau
```diff
@@ -0,0 +1,282 @@
...
+
+export type path = {
+	parts: { string },
+	kind: pathKind,
```
aatxe:
```suggestion
	kind: pathkind,
```

aatxe (changes requested):

aatxe (approved):

PR #431, opened 2025-10-16
feat: add `process.execpath` to get the current lute executable

Nerixyz:
I was trying to get the code generation to run through CMake (which would run lute in the build). When integrating this into luthier, I needed a way to identify the current lute executable to pass to CMake.

This adds `execpath` to `@lute/process`, which mirrors Node.js' [process.execPath](https://nodejs.org/docs/latest/api/process.html#processexecpath) to get the lute executable path - this is immune to `exec -a foo lute` having `foo` as `argv[0]`.

[`uv_setup_args`](https://docs.libuv.org/en/v1.x/misc.html#c.uv_setup_args) can allocate on some platforms. That memory should be free'd with [`uv_library_shutdown`](https://docs.libuv.org/en/v1.x/misc.html#c.uv_library_shutdown). I'm not sure where it's appropriate to call this.

Passing around the `argv0` also seems a bit wrong. Maybe this path shouldn't be stored in `Runtime`?

### DIFF
--- a/lute/runtime/src/runtime.cpp
+++ b/lute/runtime/src/runtime.cpp
@@ -25,6 +25,13 @@ static void lua_close_checked(lua_State* L)
 
 static std::string getExecPath(const char* argv0)
 {
+    // FIXME: uv_exepath requires that uv_setup_args(argc, argv) was called.
+    // This function can allocate on some platforms. However, the state is
+    // persisted for the duration of the program and cleaning it up with
+    // uv_library_shutdown will only delete it once. This is a problem in unit
+    // tests that call cliMain multiple times. There, the function would leak
+    // memory. On the three main platforms supported by Lute however, uv doesn't
+    // require uv_setup_args to be called, so it's currently left out.
     char buf[LUTE_PATH_MAX];
     size_t len = sizeof(buf);
     if (uv_exepath(buf, &len) == 0)
--- a/lute/runtime/src/runtime.cpp
+++ b/lute/runtime/src/runtime.cpp
@@ -19,6 +19,13 @@
 
 #include <string>
 #include <assert.h>
+#include <climits> // IWYU pragma: keep
+
+#ifdef PATH_MAX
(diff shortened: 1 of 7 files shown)


lute/runtime/src/runtime.cpp
```diff
@@ -19,16 +19,35 @@
...
     if (L)
         lua_close(L);
 }
 
-Runtime::Runtime()
+static std::string getExecPath(const char* argv0)
+{
+    char buf[LUTE_PATH_MAX];
+    size_t len = sizeof(buf);
+    if (uv_exepath(buf, &len) == 0)
```
aatxe, line 37: If this makes allocations that need to be cleaned up, then I suppose they should be cleaned up either here (if we no longer need any of that memory) or in the destructor for `Runtime` (if we need to hold onto those allocations). It seems like there's no reason to believe, given that you're providing the buffer here, that we'd need that data to keep existing.

Nerixyz: This doesn't allocate - it only writes to the buffer on the stack from which we copy the data into a `std::string`. However, `uv_setup_args`, which is called in `cliMain` can allocate.

aatxe: Same comment applies there then. It's either needed until the end of the program, in which case it should be clean up at the end of the function, or it's not, in which case it should happen right away. If need be, we can introduce a little RAII struct:
```cpp
struct SetupArgs {
    SetupArgs() {
        uv_setup_args(...)
    }
    
    ~SetupArgs() {
        -- clean up here
    }

aatxe (approved):

Nerixyz:
Looks like uv wants `uv_library_shutdown` to be called once per "process", because it uses a local `static`: https://github.com/libuv/libuv/blob/a944c422cca5522073e03710ca7fd08f53218358/src/uv-common.c#L975-L978. So in the tests, the library is shut down only once.
Not sure what's the best way to resolve this. Maybe uv's logic for getting the executable name should be implemented here?

PR #434, opened 2025-10-16
std/time: hide the internal contents of a duration in the type

aatxe:
This tweaks the implementation of `std/time` to hide the actual data inside a duration from external uses. This PR implements the information hiding just by using types here, so the internal view of all the functions casts to an internal type that lets it access the data, but the external API only presents the view where that data is hidden. We could instead use closures for information hiding here, and likely deciding one or the other as a systemic approach to data hiding is the right call for Lute. This PR at least gives us one way to look at how we approach information hiding.

### DIFF
--- a/.luaurc
+++ b/.luaurc
@@ -1,8 +1,9 @@
 {
     "languageMode": "strict",
+    "lint": { "*": true, "LocalShadow": false },
     "aliases": {
         "batteries": "./batteries",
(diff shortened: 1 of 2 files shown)


aatxe:
Leaving this here, but closing it because I think I'm convinced that we want to just do `_foo` for now, and not worry about "proper" encapsulation to any degree.

PR #436, opened 2025-10-16
Reserve stack space for creating thread refs in `getRefForThread` itself

vrn-sn:
Follow-up to #432. This was an alternative listed on that PR and is a better solution since `getRefForThread` is also called from other places in the codebase.

### DIFF
--- a/lute/runtime/src/ref.cpp
+++ b/lute/runtime/src/ref.cpp
@@ -23,7 +23,7 @@ void Ref::push(lua_State* L) const
 
 std::shared_ptr<Ref> getRefForThread(lua_State* L)
 {
-    lua_checkstack(L, 1);
+    lua_rawcheckstack(L, 1);
     lua_pushthread(L);
     std::shared_ptr<Ref> ref = std::make_shared<Ref>(L, -1);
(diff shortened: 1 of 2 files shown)


lute/runtime/src/ref.cpp
```diff
@@ -23,6 +23,7 @@ void Ref::push(lua_State* L) const
 
 std::shared_ptr<Ref> getRefForThread(lua_State* L)
 {
+    lua_rawcheckstack(L, 1);
```
vrn-sn, line 26: @vegorov-rbx, I figured it was okay to use `lua_rawcheckstack` here since we `lua_pop` it immediately, but please let me know if my understanding of this API is incorrect. I didn't want to call `lua_checkstack` and handle errors in this internal API; I think it's better to just crash if we are OOM. Am I understanding the behaviors of `lua_checkstack` and `lua_rawcheckstack` correctly?

vrn-sn: We discussed offline and agreed this is fine.

vegorov-rbx (approved):

PR #437, opened 2025-10-17
Feature: add support for user input in `@std/io`

### LINKED ISSUE #360: Add user input support to the standard library
Currently, the Lute standard library does not provide a way to get input from the user interactively (e.g., reading from stdin or prompting for user input during script execution).

**Feature Request:**
Please consider adding functions or utilities to the standard library that enable interactive user input. This could include:
- Reading strings, numbers, or other types from stdin
- Prompting the user and collecting their response during script runtime
- Handling basic validation or error cases for input

**Use Cases:**
- Scripts that need to interactively configure behavior
- Educational examples or REPL-style usage
- CLI apps requiring user choices, confirmations, or data entry

**Current Workarounds:**
- Parsing CLI arguments (not interactive)
- Reading from files (not interactive)

Thank you for considering this enhancement.

annieetang:
### Problem
Lute did not support user input yet.

### Solution
We introduce user input support to the Lute and the Lute standard library! 
Users can use the `input()` function found in`@std/io` to process and store user input from command line, piped input, or file redirection. See examples below and in `examples/user_input.luau`.
Functionality was implemented using `libuv`'s library and can be found in `lute/io`

#### Scripts

Get user input
```lua
local io = require("@std/io")

local input = io.input()
print(input)
```

Get user input with prompt
```lua
local io = require("@std/io")

local name = io.input("Please enter your name: ")
print(name)
```

#### Ways to execute the script

Command line
```bash
$ lute user_input.luau
> [will wait for user input in command line]
```

Piped input
```bash
$ echo hello | lute user_input.luau
```

Input redirected from file
```bash
$ lute user_input.luau < input_text.txt
```

### Future Work
- Tests
- `stdout` and `stderr` functionality and other I/O utilities.
- Currently, providing a prompt to the user with `.input()` will produce a new line between said prompt and the user input, this will most likely be resolved once `stdout` is implemented

Thank you @vrn-sn for massive help (esp with `libuv`) 🙏

Resolves #360


lute/io/include/lute/io.h
```diff
@@ -0,0 +1,22 @@
...
+
+// open the library as a standard global luau library
+int luaopen_io(lua_State* L);
+// open the library as a table on top of the stack
+int luteopen_io(lua_State* L);
+
+namespace io
+{
+
+int input(lua_State* L);
```
annieetang: hm maybe this in `lute/io` should be renamed to `stdin`?

Vighnesh-V: stdin is the file descriptor name - I think `io.read/write` could be a good alternative name

lute/io/src/io.cpp
```diff
@@ -0,0 +1,141 @@
...
+#include "lua.h"
+#include "lualib.h"
+
+
+namespace io
+{
+
+struct IOHandle
+{
+    Luau::Variant<uv_pipe_t, uv_tty_t> streamVariant;
```
Vighnesh-V, line 17: Some options here - asking you since I actually don't know the details.
The libuv docs say that `uv_pipe_t/uv_tty_t` are both 'subtypes' of `uv_stream_t`. Can we replace this variant with just that - is there a reason you care about which kind of stream this is?

Option 2 - can this be generic? Then you can instantiate the `IOHandle<uv_pipe_t>` and name it with a `using` statement.

The reason I ask is I don't see where you use the streamVariant in this structure, besides storage.

lute/io/src/io.cpp
```diff
@@ -0,0 +1,141 @@
...
+    handle->resumeToken = getResumeToken(L);
+    handle->L = L;
+    handle->self = handle;
+
+    uv_handle_type ht = uv_guess_handle(fileno(stdin));
+    if (ht == UV_TTY)
+    {
+        uv_tty_t& tty = handle->streamVariant.emplace<uv_tty_t>();
+        handle->stream = (uv_stream_t*)&tty;
+        uv_tty_init(handle->loop, (uv_tty_t*)&tty, fileno(stdin), 0);
```
Vighnesh-V: What happens if this call fails?

lute/io/src/io.cpp
```diff
@@ -0,0 +1,141 @@
...
+    {
+        fprintf(stderr, "Process pipe buffer allocation failed!\n");
+    }
+    static_cast<IOHandle*>(handle->data)->buf = buf;
+}
+
+// IOHandle is closed immediately after one read since we don't need a long running stream for this API.
+static void onTtyRead(uv_stream_t* stream, ssize_t nread, const uv_buf_t* buf)
+{
+    IOHandle* handle = static_cast<IOHandle*>(stream->data);
```
Vighnesh-V, line 56: Is it possible that handle is nullptr - if it is, then the following code might crash right?

lute/io/src/io.cpp
```diff
@@ -0,0 +1,141 @@
...
+namespace io
+{
+
+struct IOHandle
+{
+    Luau::Variant<uv_pipe_t, uv_tty_t> streamVariant;
+    uv_stream_t* stream = nullptr;
+    uv_loop_t* loop = nullptr;
+    ResumeToken resumeToken;
+    lua_State* L = nullptr;
```
Vighnesh-V: Is this ever used?

lute/io/src/io.cpp
```diff
@@ -0,0 +1,141 @@
...
+                lua_pushlstring(L, data.c_str(), data.size());
+                return 1;
+            }
+        );
+    }
+    else if (nread < 0)
+    {
+        handle->resumeToken->fail("Error reading from stdin");
+    }
+    else if (nread == UV_EOF)
```
Vighnesh-V: I think this is dead code since UV_EOF is always negative right?

lute/io/src/io.cpp
```diff
@@ -0,0 +1,141 @@
...
+            [data = std::string(buf->base, nread)](lua_State* L) -> int
+            {
+                lua_pushlstring(L, data.c_str(), data.size());
+                return 1;
+            }
+        );
+    }
+    else if (nread < 0)
+    {
+        handle->resumeToken->fail("Error reading from stdin");
```
Vighnesh-V: I think you can also do `uv_strerror(nread)` here to get the actual error code, and that might also include `UV_EOF`.

lute/io/src/io.cpp
Vighnesh-V: I think we'll want to signal an error here - something like failed to init tty.

lute/io/src/io.cpp
```diff
@@ -0,0 +1,141 @@
...
+        uv_tty_t& tty = handle->streamVariant.emplace<uv_tty_t>();
+        handle->stream = (uv_stream_t*)&tty;
+        uv_tty_init(handle->loop, (uv_tty_t*)&tty, fileno(stdin), 0);
+        handle->stream->data = handle.get();
+    }
+    else if (ht == UV_NAMED_PIPE || ht == UV_FILE)
+    {
+        uv_pipe_t& pipe = handle->streamVariant.emplace<uv_pipe_t>();
+        handle->stream = (uv_stream_t*)&pipe;
+        uv_pipe_init(handle->loop, &pipe, 0);
```
Vighnesh-V: ditto!

lute/io/src/io.cpp
```diff
@@ -0,0 +1,141 @@
...
+    handle->loop = uv_default_loop();
+    handle->resumeToken = getResumeToken(L);
+    handle->L = L;
+    handle->self = handle;
+
+    uv_handle_type ht = uv_guess_handle(fileno(stdin));
+    if (ht == UV_TTY)
+    {
+        uv_tty_t& tty = handle->streamVariant.emplace<uv_tty_t>();
+        handle->stream = (uv_stream_t*)&tty;
```
Vighnesh-V: ok I think I see now. We need to handle piping, file redirects differently from someone typing in a terminal. I think what you can do is get rid of the `stream` field, and replace it, in IOHandle with the following:
```
      uv_stream_t* getStream() {
          if (auto* tty = streamVariant.get_if<uv_tty_t>())
              return (uv_stream_t*)tty;
          if (auto* pipe = streamVariant.get_if<uv_pipe_t>())
              return (uv_stream_t*)pipe;
          return nullptr;
      }
```

lute/io/src/io.cpp
```diff
@@ -0,0 +1,141 @@
...
+#include "lualib.h"
+
+
+namespace io
+{
+
+struct IOHandle
+{
+    Luau::Variant<uv_pipe_t, uv_tty_t> streamVariant;
+    uv_stream_t* stream = nullptr;
```
Vighnesh-V: I think we can get rid of this member and replace it with the getter method for streamVariant.

lute/io/src/io.cpp
Vighnesh-V: I'm resolving this since I think I figured this out later on - I don't think this can be generic, and Luau::Variant seems reasonable here.

lute/io/src/io.cpp
vrn-sn: It returns a status code like other libuv APIs, so we can check if it's negative and report the stringified error if so.

lute/io/src/io.cpp
vrn-sn: Shouldn't be possible, and if it happens I think we should crash; it would mean we set something up wrong on our end.

lute/io/src/io.cpp
annieetang: ah thanks, forgot to remove!

lute/io/src/io.cpp
```diff
@@ -0,0 +1,141 @@
...
+    auto handle = std::make_shared<IOHandle>();
+    handle->loop = uv_default_loop();
+    handle->resumeToken = getResumeToken(L);
+    handle->self = handle;
+
+    uv_handle_type ht = uv_guess_handle(fileno(stdin));
+    if (ht == UV_TTY)
+    {
+        uv_tty_t& tty = handle->streamVariant.emplace<uv_tty_t>();
+        uv_tty_init(handle->loop, (uv_tty_t*)&tty, fileno(stdin), 0);
```
aatxe:
```suggestion
        uv_tty_init(handle->loop, static_cast<uv_tty_t*>(&tty), fileno(stdin), 0);
```

lute/io/src/io.cpp
```diff
@@ -0,0 +1,141 @@
...
+        if (auto* pipe = streamVariant.get_if<uv_pipe_t>())
+            return (uv_stream_t*)pipe;
+        return nullptr;
+    }
+};
+
+static void allocBuffer(uv_handle_t* handle, size_t suggestedSize, uv_buf_t* buf)
+{
+    buf->base = (char*)malloc(suggestedSize);
+    buf->len = buf->base ? suggestedSize : 0;
```
aatxe: We shouldn't do allocations this way. We're doing C++, even if the libuv APIs are C. If the IOHandle owns the buffer, then we should have the backing allocation owned by IOHandle, e.g. it should have a `std::vector<char>` or a `std::unique_ptr<std::vector<char>>` and calls consuming the memory as a `uv_buf_t` should temporarily construct the `uv_buf_t` in reference to that backing allocation. There should not be `malloc` and `free` in Lute's code. If it's anywhere else, that is a bug to be fixed as well.

lute/io/src/io.cpp
```diff
@@ -0,0 +1,141 @@
...
+    }
+    else if (ht == UV_NAMED_PIPE || ht == UV_FILE)
+    {
+        uv_pipe_t& pipe = handle->streamVariant.emplace<uv_pipe_t>();
+        uv_pipe_init(handle->loop, &pipe, 0);
+        uv_pipe_open(&pipe, fileno(stdin));
+    }
+    else
+    {
+        handle->resumeToken->fail("Unsupported stdin type");
```
vrn-sn: We can just exit here with `luaL_error` instead of scheduling an error.

lute/io/src/io.cpp
vrn-sn: We should use `luaL_error` here too, something like
```suggestion
        if (int status = uv_tty_init(handle->loop, (uv_tty_t*)&tty, fileno(stdin), 0); status < 0)
            luaL_error(uv_strerror(status));
```

lute/io/src/io.cpp
```diff
@@ -94,18 +86,28 @@ int read(lua_State* L)
     if (ht == UV_TTY)
     {
         uv_tty_t& tty = handle->streamVariant.emplace<uv_tty_t>();
-        uv_tty_init(handle->loop, (uv_tty_t*)&tty, fileno(stdin), 0);
+        int result = uv_tty_init(handle->loop, static_cast<uv_tty_t*>(&tty), fileno(stdin), 0);
+        if (result < 0)
+        {
+            luaL_error(L, "Failed to initialize TTY: %s", uv_strerror(result));
+            return -1;
```
vrn-sn: We don't need all these `return -1;`s, `luaL_error` will throw a runtime error that's caught by the Luau VM. You can just call it and assume we'll exit the function there.

lute/io/src/io.cpp
```diff
@@ -0,0 +1,134 @@
...
+    if (ht == UV_TTY)
+    {
+        uv_tty_t& tty = handle->streamVariant.emplace<uv_tty_t>();
+        if (int status = uv_tty_init(handle->loop, (uv_tty_t*)&tty, fileno(stdin), 0); status < 0)
+            luaL_error(L, "Failed to initialize TTY: %s", uv_strerror(status));
+    }
+    else if (ht == UV_NAMED_PIPE || ht == UV_FILE)
+    {
+        uv_pipe_t& pipe = handle->streamVariant.emplace<uv_pipe_t>();
+        if (int status = uv_pipe_init(handle->loop, static_cast<uv_pipe_t*>(&pipe), 0); status < 0)
```
Vighnesh-V: While this is both correct and very clever, it is a little hard to reason about because 0 is falsy, positive and negative are both truthy. Can we extract the `uv_pipe_init` and just check `status < 0` in the condition?

Vighnesh-V: ditto for the other cases we've done this.

Vighnesh-V (changes requested):
This looks great! I have one more comment that I think should be addressed before we can merge this in.

lute/io/src/io.cpp
annieetang: haha ya sounds good ty!

vrn-sn: @Vighnesh-V I don't see how extracting it changes anything, or is this more for readability? I don't see how 0 being falsy affects anything here

Vighnesh-V: readability. It's weird to me because >= 0 represents 'success', and less than represents 'failure'. However, > 0 is truthy and 0 is falsy, so it wasn't immediately clear to me with how that behaved with the multiline if statement (the semantics is that status gets bound and you just jump to the new line), since without the `;`, defaults to the truthy/falsiness of `status`. I think for clarity it's just easier to extract the `status` declaration. You are correct - this doesn't actually change anything, the code behaves correctly.

Vighnesh-V (approved):

lute/io/src/io.cpp
```diff
@@ -0,0 +1,136 @@
...
+
+struct IOHandle
+{
+    Luau::Variant<uv_pipe_t, uv_tty_t> streamVariant;
+    uv_loop_t* loop = nullptr;
+    ResumeToken resumeToken;
+    std::shared_ptr<IOHandle> self;
+    std::vector<char> buffer;
+
+    ~IOHandle() {}
```
aatxe:
```suggestion
```

PR #438, opened 2025-10-17
[Infra] Enable AddressSanitizer in CI builds

vrn-sn:
As our tests become more comprehensive, we should be able to catch subtle memory errors with this. ASan is intentionally enabled only in CI, not in the release job. Windows is left out for now because of added complexity.


CMakeLists.txt
```diff
@@ -114,6 +114,14 @@ else()
     list(APPEND LUTE_OPTIONS -Werror) # Warnings are errors
 endif()
 
+option(LUTE_ENABLE_ASAN "Enable AddressSanitizer" OFF)
+if (LUTE_ENABLE_ASAN)
+    if(NOT MSVC)
+        add_compile_options(-fsanitize=address -fno-omit-frame-pointer)
```
hgoldstein, line 120: Why do we need `-fno-omit-frame-pointer`?

hgoldstein (approved):
A couple comments:
* UBSan should be free as well right? I remember running into a bug that you suggested might be catchable with UBSan 😄 
* What would we be missing to enable MSan?
* Should we cut a tracking issue for MSVC? It seems like there's ASan support even if we had trouble implementing it for this PR?

Vighnesh-V:
I think this is also a great opportunity to look into TSan! (Obv can happen in a different PR).

CMakeLists.txt
vrn-sn: Helps sanitizers extract more accurate stack trace information by just following `%rbp`, especially if we ever need an optimized build with a sanitizer enabled. Optimized builds will often omit this stack frame pointer and reuse the register as a general purpose register, so we override that here.

PR #441, opened 2025-10-18
Enable UBSan in CI

vrn-sn:
This is a follow-up to #438. For the purposes of CI, we don't need finer-grained control over which sanitizers are turned on, so both are bundled under the same flag. If we do need finer control, we can change this in the future.


Vighnesh-V:
@vrn-sn Should we just make an issue to track this and close this now?

vrn-sn:
I can try to tackle it today, I think the main issue is that enabling UBSan seems to trigger the stack frame size limit for `boringssl` on MacOS. If I can just find a way to disable it for that library, we can enable this.

Vighnesh-V (approved):

PR #442, opened 2025-10-18
Break bidirectional dependencies between Lute.Runtime and every runtime library

vrn-sn:
### Problem

`Lute.Runtime` is our lowest-level library, and each of our runtime libraries like `Lute.Crypto`, `Lute.Fs`, etc. correctly links against it. However, `Lute.Runtime` also currently links against each of these auxiliary libraries simply to register them. This means that `Lute.Runtime` cannot currently be built without also building each of the auxiliary libraries.

### Solution

This solution is modeled after the Luau CLI's design, where a given instance of a Luau VM can have new globals like `require` injected in, but the registration doesn't occur in `Luau.VM` itself. Instead, it happens in `Luau.Repl.CLI`. Similarly, we inject in these `@lute/*` registrations in `Lute.CLI.lib` instead of in the runtime code itself.

This also makes it easier to reason about the runtime as we work to move to an "entirely managed by libuv" model.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -16,6 +16,14 @@ project(
     LANGUAGES CXX C
 )
 
+option(LUTE_ENABLE_ASAN "Enable AddressSanitizer" OFF)
+if (LUTE_ENABLE_ASAN)
+    if(NOT MSVC)
+        add_compile_options(-fsanitize=address -fno-omit-frame-pointer)
+        add_link_options(-fsanitize=address)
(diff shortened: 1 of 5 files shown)


vrn-sn:
Also moved some sanitizer stuff around to pass CI. It looked like certain Luau static libraries weren't being instrumented, and it was causing false positive ASAN errors.

aatxe (approved):

PR #444, opened 2025-10-18
Add ``errors`` assertion

wmccrthy:
This is a super common assertion and something I wanted to use in a project  I was working on so figure'd I would add it.

Not sure what the larger vision is for testing API, but hopefully this fits somewhere.

Can also add an optional `expectedError` param

### DIFF
--- a/lute/std/libs/assert.luau
+++ b/lute/std/libs/assert.luau
@@ -12,6 +12,13 @@ function assertions.neq<T>(lhs: T, rhs: T)
 	end
 end
 
+function assertions.throwserror(callback: () -> ())
+	local success = pcall(callback) -- callback shudn't return anything, but type solver wouldn't shut up
+	if success then
+		error({ msg = `{callback} did not throw error.`})
+	end
+end
+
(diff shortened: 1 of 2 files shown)


lute/std/libs/assert.luau
```diff
@@ -12,6 +12,13 @@ function assertions.neq<T>(lhs: T, rhs: T)
 	end
 end
 
+function assertions.throwserror(callback: () -> ())
```
Vighnesh-V: Small nit: `assertions.errors` or `assertion.throws` instead (i know tableeq is an exception)?

Vighnesh-V (approved):
Left one small nit about the naming convention, but I'll leave it up to you. This is a useful assertion to have for sure!

lute/std/libs/assert.luau
wmccrthy: sure!

aatxe:
GitHub reviews was failing on my phone when I looked at this, but I also wanted it renamed `errors`, so nice. :)

PR #445, opened 2025-10-19
Fix new path lib incorrectly checking OS

checkraisefold:
tin
the path lib literally does not work on windows at all because of this error


vrn-sn:
Could we access this via `@std/system` instead?

checkraisefold:
> Could we access this via `@std/system` instead?

done - I also edited the `test.luau` lib to do this for consistency's sake and because it's the reason I used a relative require at all

vrn-sn (approved):

PR #447, opened 2025-10-19
Refactor `lute setup` to write to luaurc, add std type defs

### LINKED ISSUE #424: Generate or add definitions to luaurc in `lute setup`
We can generate (or add aliases to an existing) luaurc files for the definitions being set up when someone runs `lute setup`.

checkraisefold:
Refactors the `lute setup` command to:
1. Put the regular lute typedefs into a `lute` subdirectory, and add a `std` subdirectory with the standard library files inside
2. Accept a flag to create a new luaurc, or overwrite the aliases of one if it exists, to add the type definition folders. Unfortunate caveat is that the std lib json encoder is not deterministic, so each time it moves the json keys around - this is pretty annoying and creates a lot of diff noise. Not sure if there's much to be done about this

Luthier:
1. Fix a backslash escaping bug for std lib and command C++ file gen that cropped up when I added this
2. Change definition generation to use above structure
3. Use projectRelative more
4. Change order of `generateFilesIfNeeded` calls to fix a bug - CLI commands rely on the generated type def output, so they should be generated after type defs

Also fix the trailing comma in the project luaurc which is typically not valid json

Closes #424

### DIFF
--- a/lute/cli/src/climain.cpp
+++ b/lute/cli/src/climain.cpp
@@ -200,6 +200,7 @@ static void displayHelp(const char* argv0)
     printf("Setup Options:\n");
     printf("  lute setup");
     printf("    Generates type definition files for the language server.\n");
+    printf("      --with-luaurc Adds/overwrites aliases to the type definition files to the luaurc in the working directory.\n");
     printf("\n");
     printf("General Options:\n");
     printf("  -h, --help    Display this usage message.\n");
--- a/lute/cli/src/climain.cpp
+++ b/lute/cli/src/climain.cpp
@@ -200,7 +200,7 @@ static void displayHelp(const char* argv0)
     printf("Setup Options:\n");
     printf("  lute setup");
     printf("    Generates type definition files for the language server.\n");
-    printf("      --with-luaurc Adds/overwrites aliases to the type definition files to the luaurc in the working directory.\n");
+    printf("      --with-luaurc Defines aliases to the type definition files in the working directory's luaurc file.\n");
     printf("\n");
     printf("General Options:\n");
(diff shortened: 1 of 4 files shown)


aatxe:
Technically, luaurc format is JSON5-ish, not actually JSON, so that's why trailing comma works, but it's fine.

checkraisefold:
> Technically, luaurc format is JSON5-ish, not actually JSON, so that's why trailing comma works, but it's fine.

mainly I encountered this when testing this command, was using the lute .luaurc as a test bench and the json parser errors when it encounters the trailing comma
"JSON5-ish" is scary sounding

aatxe:
Yeah, I mean, the specifics of the JSON spec are really quite terrible, that's why the luaurc file is way more permissive (and also it's written to be parsed by an extremely janky system that just... hacks Luau's parser for tables into this state)

lute/cli/src/climain.cpp
```diff
@@ -200,6 +200,7 @@ static void displayHelp(const char* argv0)
     printf("Setup Options:\n");
     printf("  lute setup");
     printf("    Generates type definition files for the language server.\n");
+    printf("      --with-luaurc Adds/overwrites aliases to the type definition files to the luaurc in the working directory.\n");
```
aatxe:
```suggestion
    printf("      --with-luaurc Defines aliases to the type definition files in the working directory's luaurc file.\n");
```

aatxe (commented):
Slight help doc tweak, but otherwise LGTM.

Vighnesh-V (approved):
Lets gooo! I've been wanting this for a while :)

PR #450, opened 2025-10-20
feature: Create a Virtual File System implementation that can restore code from a bytecode bundle.

Vighnesh-V:
This PR implements a 'bundle' virtual file system, for navigating requires and loading bytecode from an executable compiled with `lute compile`.

### DIFF
--- a/lute/require/include/lute/bundlevfs.h
+++ b/lute/require/include/lute/bundlevfs.h
@@ -2,15 +2,16 @@
 
 #include "lute/modulepath.h"
 
+#include "Luau/DenseHash.h"
+
+#include <functional>
 #include <optional>
 #include <string>
-#include <map>
 
 class BundleVfs
 {
 public:
-    BundleVfs() = default;
-    BundleVfs(const std::map<std::string, std::string>* bundleMap);
+    BundleVfs(const Luau::DenseHashMap<std::string, std::string>& bundleMap);
 
     NavigationStatus resetToPath(const std::string& path);
 
@@ -25,6 +26,6 @@ public:
     std::optional<std::string> getConfig() const;
 
 private:
-    const std::map<std::string, std::string>* filePathToBytecode = nullptr;
+    const Luau::DenseHashMap<std::string, std::string>& filePathToBytecode;
     std::optional<ModulePath> modulePath;
 };
--- a/lute/require/include/lute/bundlevfs.h
+++ b/lute/require/include/lute/bundlevfs.h
@@ -1,0 +1,30 @@
+#pragma once
+
+#include "lute/modulepath.h"
+
+#include <optional>
+#include <string>
+#include <map>
+
+class BundleVfs
+{
+public:
+    BundleVfs() = default;
+    BundleVfs(const std::map<std::string, std::string>* bundleMap);
+
+    NavigationStatus resetToPath(const std::string& path);
+
+    NavigationStatus toParent();
+    NavigationStatus toChild(const std::string& name);
+
+    bool isModulePresent() const;
+    std::string getIdentifier() const;
+    std::optional<std::string> getContents(const std::string& path) const;
+
+    bool isConfigPresent() const;
+    std::optional<std::string> getConfig() const;
+
+private:
+    const std::map<std::string, std::string>* filePathToBytecode = nullptr;
+    std::optional<ModulePath> modulePath;
+};
--- a/lute/require/include/lute/bundlevfs.h
+++ b/lute/require/include/lute/bundlevfs.h
@@ -11,7 +11,7 @@
 class BundleVfs
 {
 public:
-    BundleVfs(const Luau::DenseHashMap<std::string, std::string>& bundleMap);
+    BundleVfs(Luau::DenseHashMap<std::string, std::string> bundleMap);
 
     NavigationStatus resetToPath(const std::string& path);
 
@@ -26,6 +26,6 @@ public:
     std::optional<std::string> getConfig() const;
 
 private:
-    const Luau::DenseHashMap<std::string, std::string>& filePathToBytecode;
+    const Luau::DenseHashMap<std::string, std::string> filePathToBytecode;
     std::optional<ModulePath> modulePath;
 };
--- a/lute/require/src/bundlevfs.cpp
+++ b/lute/require/src/bundlevfs.cpp
@@ -4,10 +4,10 @@
 
 #include <string>
 
-static const std::string BUNDLE_PREFIX = "@bundle/";
+constexpr std::string_view kBundlePrefix = "@bundle/";
 
-BundleVfs::BundleVfs(const Luau::DenseHashMap<std::string, std::string>& bundleMap)
-    : filePathToBytecode(bundleMap)
+BundleVfs::BundleVfs(Luau::DenseHashMap<std::string, std::string> bundleMap)
+    : filePathToBytecode(std::move(bundleMap))
 {
 }
 
@@ -15,8 +15,8 @@ static bool isBundleModule(const Luau::DenseHashMap<std::string, std::string>& b
 {
     // Strip @bundle/ prefix if present
     std::string lookupPath = path;
-    if (path.find(BUNDLE_PREFIX) == 0)
-        lookupPath = path.substr(BUNDLE_PREFIX.size());
+    if (path.rfind(kBundlePrefix, 0) == 0)
+        lookupPath = path.substr(kBundlePrefix.size());
 
     // Check direct file match
     if (bundleMap.find(lookupPath) != nullptr)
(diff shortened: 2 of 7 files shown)


lute/require/include/lute/bundlevfs.h
```diff
@@ -0,0 +1,30 @@
+#pragma once
+
+#include "lute/modulepath.h"
+
+#include <optional>
+#include <string>
+#include <map>
```
vrn-sn:
```suggestion
#include "Luau/DenseHash.h"
```
Is it possible to replace `std::map` with Luau's dense hash map?

lute/require/src/bundlevfs.cpp
```diff
@@ -0,0 +1,174 @@
...
+    if (bundleMap->find(lookupPath + "/init.luau") != bundleMap->end())
+        return true;
+
+    // Check for init.lua in directory
+    if (bundleMap->find(lookupPath + "/init.lua") != bundleMap->end())
+        return true;
+
+    return false;
+}
+[[maybe_unused]]
```
vrn-sn:
```suggestion
```
Do we need this?

lute/require/src/bundlevfs.cpp
```diff
@@ -0,0 +1,174 @@
...
+static bool isBundleDirectory(const std::map<std::string, std::string>* bundleMap, const std::string& path)
```
vrn-sn: Can we take the map as a reference instead? We can also remove the initial `nullptr` check.

lute/require/src/bundlevfs.cpp
```diff
@@ -0,0 +1,174 @@
...
+        return false;
+
+    // Strip @bundle/ prefix if present
+    std::string lookupPath = path;
+    std::string bundlePrefix = "@bundle/";
+    if (path.find(bundlePrefix) == 0)
+        lookupPath = path.substr(bundlePrefix.size());
+
+    // Check direct file match
+    if (bundleMap->find(lookupPath) != bundleMap->end())
```
vrn-sn: I believe in `ModulePath`'s implementation, we already do all this "append suffix and check" logic before calling `isAFile`. I don't think this function itself needs to be aware of appending file extensions at all.

lute/require/src/bundlevfs.cpp
```diff
@@ -0,0 +1,174 @@
...
+    if (!bundleMap)
+        return false;
+
+    // Strip @bundle/ prefix if present
+    std::string lookupPath = path;
+    std::string bundlePrefix = "@bundle/";
+    if (path.find(bundlePrefix) == 0)
+        lookupPath = path.substr(bundlePrefix.size());
+
+    // A directory exists if any file in the bundle starts with this path followed by a slash
```
vrn-sn, line 37: qq: are we not embedding directories explicitly? For example, `@std` embedding represented directories as a "file" containing `#directory`.

lute/require/src/bundlevfs.cpp
```diff
@@ -0,0 +1,174 @@
...
+}
+
+static bool isBundleModule(const std::map<std::string, std::string>* bundleMap, const std::string& path)
+{
+    if (!bundleMap)
+        return false;
+
+    // Strip @bundle/ prefix if present
+    std::string lookupPath = path;
+    std::string bundlePrefix = "@bundle/";
```
vrn-sn: Might be worth throwing `bundlePrefix` at the top of the file somewhere as a constant since it shows up in a few different functions.

lute/require/src/bundlevfs.cpp
```diff
@@ -0,0 +1,174 @@
...
+
+std::string BundleVfs::getIdentifier() const
+{
+    LUAU_ASSERT(modulePath);
+    ResolvedRealPath result = modulePath->getRealPath();
+    LUAU_ASSERT(result.status == NavigationStatus::Success);
+    return result.realPath;
+}
+
+std::optional<std::string> BundleVfs::getContents(const std::string& path) const
```
vrn-sn, line 104: `getContents` is only called if `isModulePresent` returned true, so we can pretty safely just access your `filePathToBytecode` map with maybe some asserts to make sure things are in a sane state.

lute/require/src/bundlevfs.cpp
```diff
@@ -0,0 +1,174 @@
...
+    if (path.find(bundlePrefix) == 0)
+        lookupPath = path.substr(bundlePrefix.size());
+
+    // Try direct lookup
+    auto it = filePathToBytecode->find(lookupPath);
+    if (it != filePathToBytecode->end())
+        return it->second;
+
+    // Try with .luau extension
+    it = filePathToBytecode->find(lookupPath + ".luau");
```
vrn-sn: Similar to an earlier comment, we shouldn't need to do all this file extension stuff here. `getContents` is called with the `loadname` provided by `BundleVfs::getIdentifier`. If I'm reading this right, your implementation should already return a `loadname` with a file extension (by using `ModulePath` under the hood), so we don't need to redo all that here.

lute/require/src/bundlevfs.cpp
Vighnesh-V: Nah - we embed files by the snapshot of their paths at time of compilation - the directory structure is encoded as part of the key (file path). I think this is fine, just because at snapshot time, static require tracing should successfully trace all the possible requirable paths.

lute/require/src/bundlevfs.cpp
Vighnesh-V: good point - deleted the additional extension checking from this.

lute/require/src/bundlevfs.cpp
```diff
@@ -0,0 +1,127 @@
+#include "lute/bundlevfs.h"
+
+#include "Luau/Common.h"
+
+#include <string>
+
+static const std::string BUNDLE_PREFIX = "@bundle/";
```
vrn-sn:
```suggestion
#include "Luau/Common.h"
#include "Luau/DenseHash.h"

#include <string>
#include <string_view>

constexpr std::string_view kBundlePrefix = "@bundle/";
```
to match existing constant style in other places in the code

lute/require/src/bundlevfs.cpp
```diff
@@ -0,0 +1,127 @@
...
+
+std::optional<std::string> BundleVfs::getContents(const std::string& path) const
+{
+    // Strip @bundle/ prefix if present
+    std::string lookupPath = path;
+    if (path.find(BUNDLE_PREFIX) == 0)
+        lookupPath = path.substr(BUNDLE_PREFIX.size());
+
+    // Try direct lookup
+    const auto* value = filePathToBytecode.find(lookupPath);
```
vrn-sn:
```suggestion
    const std::string* value = filePathToBytecode.find(lookupPath);
```
Just for clarity

lute/require/src/bundlevfs.cpp
```diff
@@ -0,0 +1,127 @@
...
+#include "Luau/Common.h"
+
+#include <string>
+
+static const std::string BUNDLE_PREFIX = "@bundle/";
+
+BundleVfs::BundleVfs(const Luau::DenseHashMap<std::string, std::string>& bundleMap)
+    : filePathToBytecode(bundleMap)
+{
+}
```
vrn-sn:
```suggestion
BundleVfs::BundleVfs(Luau::DenseHashMap<std::string, std::string> bundleMap)
    : filePathToBytecode(std::move(bundleMap))
{
}
```
I think it makes sense for BundleVfs to own this? (Unrelated to `isBundleModule` and `isBundleDirectory` taking `bundleMap` as a reference.) Would also need to change the member's type to a value instead of a reference.

lute/require/src/bundlevfs.cpp
```diff
@@ -0,0 +1,127 @@
...
+        return true;
+
+    return false;
+}
+
+static bool isBundleDirectory(const Luau::DenseHashMap<std::string, std::string>& bundleMap, const std::string& path)
+{
+    // Strip @bundle/ prefix if present
+    std::string lookupPath = path;
+    if (path.find(BUNDLE_PREFIX) == 0)
```
vrn-sn: We don't have C++20's `starts_with` API, but we can improve this to a O(1) check for now. I think as future TODO, we could just have some string utilities exposed for this type of thing.
```suggestion
    if (path.rfind(BUNDLE_PREFIX, 0) == 0)
```
(same for other places in this file where we do `string.find(prefix) == 0`)

lute/require/src/bundlevfs.cpp
```diff
@@ -0,0 +1,127 @@
+#include "lute/bundlevfs.h"
+
+#include "Luau/Common.h"
```
vrn-sn, line 3:
```suggestion
#include "Luau/Common.h"
#include "Luau/DenseHash.h"

#include <string>
#include <string_view>
```

vrn-sn (approved):
Approving with one final comment that might have been missed in the previous round! Feel free to merge when ready.

PR #452, opened 2025-10-20
feature: Implement a Static Require Tracer in order to bundle files correctly when bytecode compiling them.

Vighnesh-V:
This PR implement a static require tracer for Luau as a prerequisite to enabling multi file bytecode compilation. This module statically resolves requires relative to a given entry module, effectively snapshotting the state of the filesystem for embedding in a lute executable. This module graph will be embedded in the lute executable when `lute compile` is invoked and will let the bundle virtual file system [PR](https://github.com/luau-lang/lute/pull/450) correctly navigate the directory structure to resolve `require` calls and match up those paths to the pre-compiled bytecode embedded in the bundle.

### DIFF
--- a/lute/cli/src/staticrequires.cpp
+++ b/lute/cli/src/staticrequires.cpp
@@ -6,10 +6,7 @@
 #include "Luau/FileUtils.h"
 
 #include <cstdio>
-#include <filesystem>
 #include <queue>
-
-namespace fs = std::filesystem;
 
 // AST visitor to extract require() calls
 class RequireExtractor : public Luau::AstVisitor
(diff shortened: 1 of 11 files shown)


aatxe (approved):

lute/cli/src/staticrequires.cpp
```diff
@@ -0,0 +1,199 @@
...
+
+// AST visitor to extract require() calls
+class RequireExtractor : public Luau::AstVisitor
+{
+public:
+    std::vector<std::string> requirePaths;
+
+    bool visit(Luau::AstExprCall* call) override
+    {
+        if (auto global = call->func->as<Luau::AstExprGlobal>())
```
vrn-sn, line 21: In the Luau repo, we do some very basic heuristics that allow us to also support things like this:
```luau
-- AST groups are collapsed
local res = require(("./some/path"))

-- Basic local tracking is supported
local path = "./some/other/path"
local res2 = require(path)
```

Don't necessarily need to support this now, but just wanted to mention it in case we need something more robust in the future.

lute/cli/src/staticrequires.cpp
```diff
@@ -0,0 +1,199 @@
...
+    RequireExtractor extractor;
+    result.root->visit(&extractor);
+
+    return extractor.requirePaths;
+}
+
+std::optional<std::string> StaticRequireTracer::resolveRequire(const std::string& requirer, const std::string& required)
+{
+    // Get the directory containing the requiring file (relative to rootDirectory)
+    std::string requirerDir;
```
vrn-sn, line 118: If I'm remembering correctly, we do have `getParentPath` in `FileUtils`. Wonder if it accomplishes the same thing here?

lute/cli/src/staticrequires.cpp
```diff
@@ -0,0 +1,199 @@
...
+        isFileFunc,
+        isDir
+    );
+
+    if (!modulePath)
+        return std::nullopt;
+
+    // Navigate according to the require path
+    // Split require path by '/' and navigate
+    std::string reqPath = required;
```
vrn-sn, line 154: This function is the part that should ultimately be replaced with our static require resolver, but just wanted to point you to `Navigator::navigateThroughPath`. There's a nice trick we can do using `std::string_view`s that prevents a new `std::string` allocation per while-loop iteration.

vrn-sn (commented):
Sorry that I couldn't review this earlier, but it looks good! I left some comments, but none are really requesting changes—just things to think about as we keep working on this.

PR #454, opened 2025-10-20
Port codemod into lute

skanosue:
This PR copies the [codemod implementation](https://github.com/luau-lang/codemod) mostly as-is into lute, invokable via `lute transform`. The main tweak I made was to load migrations by the path passed via the CLI, rather than using `require`, so that migration paths don't have to be passed relative to the codemod source code.


tests/transform.test.luau
```diff
@@ -0,0 +1,78 @@
...
+		local transformeePath = path.format(path.join(transformTestDir, "transformee.luau"))
+		local transformeeHandle = fs.open(transformeePath, "w+")
+		fs.write(transformeeHandle, TRANSFORMEE_CONTENT)
+		fs.close(transformeeHandle)
+
+		-- Do
+		-- Run the transformer on the transformee
+		local result = process.run({ lutePath, "transform", transformerDest, transformeePath }, {
+			cwd = process.cwd(),
+		})
```
Vighnesh-V, line 44: We really need a way of intercepting io handles directly on a process (not about this PR, just more generally)

tests/transform.test.luau
```diff
@@ -0,0 +1,78 @@
...
+local lutePath = path.format(
+	path.join(
+		"build",
+		if system.macos then "xcode" elseif system.win32 then "vs2022" else "",
+		"debug",
+		"lute",
+		"cli",
+		"lute"
+	)
+)
```
Vighnesh-V: instead of hard coding this, can we get this directly from the cli args / process api? I think Varun has made some changes to how this gets passed down, so you might be able to get the path of the `lute` executable via : https://github.com/luau-lang/lute/pull/458

lute/std/libs/syntax/query.luau
```diff
@@ -0,0 +1,27 @@
...
+		return true
+	end
+
+	visitor.visitBlock(block, finder)
+
+	return nodes
+end
+
+return {
+	selectExpression = selectExpression,
```
Vighnesh-V: I think this should be luacase.

Vighnesh-V: also, maybe `selectif` or `filterblock`? Reason I suggest this is that the function doesn't selectExpressions - it maps expressions to whatever you'd like + filters them.
```
query.selectif(block, function (e) return e.hasprop() end) : { bool }
```

lute/cli/commands/transform/lib/strings.luau
```diff
@@ -0,0 +1,36 @@
+local strings = {}
+
+function strings.startsWith(str: string, prefix: string): boolean
```
Vighnesh-V: startswith/removeprefix/endswith/removesuffix/...

lute/cli/commands/transform/lib/path.luau
```diff
@@ -0,0 +1,55 @@
+local strings = require("./strings")
```
Vighnesh-V: luacase here as well.

lute/cli/commands/transform/lib/arguments.luau
```diff
@@ -0,0 +1,69 @@
...
+	--- Prevent making changes to files
+	dryRun: boolean,
+	--- Path to the migration module to execute
+	migrationPath: string,
+	--- Configuration flags to pass to the migration
+	migrationOptions: { [string]: string },
+	--- List of file paths to process
+	--- Note: no directory traversal or filtering has been applied on this list
+	filePaths: { string },
+}
```
Vighnesh-V: exported type members gotta be luacase too :)

lute/cli/commands/transform/lib/arguments.luau
```diff
@@ -0,0 +1,69 @@
...
+
+local function parse(arguments: { string }): Config
+	-- TODO: replace with proper CLI system
```
Vighnesh-V, line 16: We have argparsing working rn - probably worth re-writing in terms of that.

Vighnesh-V (changes requested):
Sorry, the requests do be a bit repetitive :( lots of luacasing to do + we have a cli parser - we should augment that if it's missing anything or use that rn.

tests/transform.test.luau
aatxe: tfw "process API is hard" and nobody believes me, and then they add a really simplistic one, and we constantly run into the limitations of it lol

lute/std/libs/syntax/query.luau
aatxe: `selectif` is a great name for this.

tests/transform.test.luau
skanosue: yeah blessup, didn't know that existed

lute/cli/commands/transform/lib/arguments.luau
skanosue: i don't think this needs to be exported at all actually

lute/cli/commands/transform/lib/arguments.luau
Vighnesh-V: retracting this comment - cli parsing is currently not in std, and so we can't use it for now.

lute/cli/commands/transform/init.luau
```diff
@@ -0,0 +1,168 @@
...
+			source = source,
+			options = options,
+		}
+
+		-- TODO: should we wrap in pcall? For now we don't do this to preserve stack trace
+		local result = migration.transform(ctx)
+		-- local success, result = pcall(transform, ctx)
+		-- if not success then
+		-- 	error(`Failed to apply to {path}: {result}`)
+		-- end
```
Vighnesh-V: Probably good idea to implement this or delete the comment!

Vighnesh-V (approved):

PR #455, opened 2025-10-20
Call `uv_setup_args` at `lute` and `lute-tests` entry points

### LINKED ISSUE #451: Call `uv_setup_args` before `uv_exepath`.
```
    // FIXME: uv_exepath requires that uv_setup_args(argc, argv) was called.
    // This function can allocate on some platforms. However, the state is
    // persisted for the duration of the program and cleaning it up with
    // uv_library_shutdown will only delete it once. This is a problem in unit
    // tests that call cliMain multiple times. There, the function would leak
    // memory. On the three main platforms supported by Lute however, uv doesn't
    // require uv_setup_args to be called, so it's currently left out.
```

From a comment in the execpath implementation. We should do `uv_setup_args` in a way where it runs in the entry point, and where `uv_library_shutdown` only gets called once. This probably involves tweaking some of the way our test harness is setup as well.

vrn-sn:
At the entry points of `lute` and `lute-tests`, we now call `uv_setup_args` with the new `UvGlobalState` RAII wrapper.

It's still the case that unit tests that call `cliMain` don't individually call `uv_set_args`, but this is probably the behavior we want anyway. Calling `cliMain` directly in these tests allows us to inject in custom arguments for Lute runtime code, with the caveat that libuv's "introspective" APIs like `uv_exepath` will still return globally registered information. This is fine because our unit tests shouldn't need to test functionality at the `libuv`-level anyway.

The types of APIs that `uv_setup_args` enables aren't APIs that make sense for us to inject in unit tests.

Resolves #451.

### DIFF
--- a/lute/cli/CMakeLists.txt
+++ b/lute/cli/CMakeLists.txt
@@ -26,10 +26,12 @@ target_sources(Lute.CLI.lib PRIVATE
     include/lute/climain.h
     include/lute/compile.h
     include/lute/tc.h
+    include/lute/uvstate.h
 
     src/climain.cpp
     src/compile.cpp
     src/tc.cpp
+    src/uvstate.cpp
 )
 
 target_compile_features(Lute.CLI.lib PUBLIC cxx_std_17)
--- a/lute/cli/include/lute/uvstate.h
+++ b/lute/cli/include/lute/uvstate.h
@@ -1,0 +1,12 @@
+#pragma once
+
+struct UvGlobalState
+{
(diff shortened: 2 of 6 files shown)


vrn-sn:
The PR linked below (#458) is follow-up work to this PR, which cleans up all the `argv[0]` logic we have scattered throughout Lute.CLI.

Vighnesh-V (approved):

aatxe (approved):

PR #457, opened 2025-10-20
fix: Fix typecheck errors in std library

Vighnesh-V:
There are at least a handful of type errors in the standard library. This PR tries to fix them.

### DIFF
--- a/lute/std/libs/syntax/printer.luau
+++ b/lute/std/libs/syntax/printer.luau
@@ -71,6 +71,21 @@ local function printInterpolatedString(expr: luau.AstExprInterpString): string
 	return result
 end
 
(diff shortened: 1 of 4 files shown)


lute/std/libs/syntax/printer.luau
```diff
@@ -111,7 +126,7 @@ local function printVisitor()
 	end
 
 	printer.visitTypeString = function(node: luau.AstTypeSingletonString)
-		write(printString(node))
```
hgoldstein, line 114: I think you can just change `printString` to take a union

Vighnesh-V: fixed :)

aatxe (approved):

PR #458, opened 2025-10-21
Move `getExecPath` logic to Lute.Process and clean up all uses of `argv[0]` in Lute.CLI

vrn-sn:
Follow-up work to #455. We shouldn't need to store global state like the executable name in `Runtime` itself. The only Luau-facing API that needs that value is the new `process.execpath` API, so the implementation is moved to `process.cpp`.

While refactoring some files after removing `argv0` from the `Runtime` constructor, I realized that we pass around `argv[0]` in many places where we don't need to. I've removed it entirely from helper functions like `displayCheckHelp`.

There are only two functions that actually need the executable name: `checkForAppendedBytecode` and `compileScript`. Rather than passing it in, we instead call into `Lute.Process`'s new `getExecPath` API to retrieve it.

### DIFF
--- a/lute/cli/src/compile.cpp
+++ b/lute/cli/src/compile.cpp
@@ -1,17 +1,32 @@
 #include "lute/compile.h"
 
 #include "lute/options.h"
+#include "lute/process.h"
 #include "uv.h"
 
 #include <fstream>
+#include <string>
 
 const char MAGIC_FLAG[] = "LUTEBYTE";
 const size_t MAGIC_FLAG_SIZE = sizeof(MAGIC_FLAG) - 1;
 const size_t BYTECODE_SIZE_FIELD_SIZE = sizeof(uint64_t);
 
-AppendedBytecodeResult checkForAppendedBytecode(const std::string& executablePath)
+AppendedBytecodeResult checkForAppendedBytecode()
 {
     AppendedBytecodeResult result;
+
+    std::string executablePath;
+    {
+        std::string error;
+        std::optional<std::string> execPathOpt = process::getExecPath(&error);
+        if (!execPathOpt)
+        {
+            fprintf(stderr, "Could not get path to executable: %s\n", error.c_str());
+            return result;
+        }
+        executablePath = *execPathOpt;
+    }
+
     std::ifstream exeFile(executablePath, std::ios::binary | std::ios::ate);
     if (!exeFile)
     {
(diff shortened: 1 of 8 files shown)


lute/cli/src/compile.cpp
```diff
@@ -54,8 +69,21 @@ AppendedBytecodeResult checkForAppendedBytecode(const std::string& executablePat
     return result;
 }
 
-int compileScript(const std::string& inputFilePath, const std::string& outputFilePath, const std::string& currentExecutablePath)
+int compileScript(const std::string& inputFilePath, const std::string& outputFilePath)
 {
+    std::string currentExecutablePath;
+    ;
```
aatxe:
```suggestion
```

vrn-sn:
Hold on... the diff didn't collapse correctly. Let me try a rebase.

vrn-sn:
Fixed now.

lute/process/src/process.cpp
```diff
@@ -474,9 +485,36 @@ int cwd(lua_State* L)
...
+    // Executable path is not expected to change during process lifetime, so we
+    // can safely cache it after the first retrieval.
+    static std::optional<std::string> cachedPath = std::nullopt;
+    if (cachedPath)
+        return *cachedPath;
+
+    char buf[LUTE_PATH_MAX];
+    size_t len = sizeof(buf);
+
+    if (int status = uv_exepath(buf, &len); status < 0)
```
aatxe, line 499: Does uv_exepath specify the would-be length if it fails? If so, can we remove the predefined 8kb max and just use a std::vector and resize it if need be?

aatxe (commented):
Looks good to me, besides wanting to know about the allocation bit.

vrn-sn:
@Vighnesh-V can you check if this works as expected with your bundling work?

Vighnesh-V:
@vrn-sn It'll probably work. Can you do the following test? In a file called main.luau
```
print("Hello from main")
```

lute compile main.luau, 
```
./main
```

If that works, it'll work with my PR :)

lute/process/src/process.cpp
vrn-sn: Documentation just says this:
> Gets the executable path. You must call uv_setup_args before calling this function.

Looking at the implementation, I don't see it ever returning `UV_ENOBUFS`. Looks like it just truncates the path, so not much to work with there. It'll pass if the length is too short.

vrn-sn:
@Vighnesh-V tested and works as expected.

lute/process/src/process.cpp
Vighnesh-V: I think in the uv docs, a different method,
`uv_cwd` does set the size of the second argument if there's not enough space and returns `UV_ENOBUFS`. I'd wager this does the same thing.

vrn-sn: It should, but I don't think it does. Made this issue: https://github.com/libuv/libuv/issues/4911.

aatxe (approved):

PR #461, opened 2025-10-21
stdlib: loadbypath

skanosue:
Implementation of `loadbypath` in user land using `luau.load`. Addresses #321 .


lute/std/libs/path/posix.luau
```diff
@@ -105,15 +105,23 @@ local function joinHelper(path: path, addend: pathlike): ()
...
-	local path = posix.parse(parts[1])
+	local path: path
+	if typeof(parts[1]) == "string" then
+		path = posix.parse(parts[1])
+	else
+		path = {
+			parts = table.clone((parts[1] :: path).parts),
+			absolute = (parts[1] :: path).absolute,
+		}
+	end
```
aatxe:
```suggestion
	local path = if typeof(parts[1]) == "string" then
		posix.parse(parts[1])
	else
		{
			parts = table.clone((parts[1] :: path).parts),
			absolute = (parts[1] :: path).absolute,
		}
	end
```

lute/std/libs/path/win32.luau
```diff
@@ -149,7 +149,16 @@ function win32.join(...: pathlike): path
...
+	local path: path
+	if typeof(parts[1]) == "string" then
+		path = win32.parse(parts[1])
+	else
+		path = {
+			parts = table.clone((parts[1] :: path).parts),
+			kind = (parts[1] :: path).kind,
+			driveLetter = (parts[1] :: path).driveLetter,
+		}
+	end
```
aatxe:
```suggestion
	local path = if typeof(parts[1]) == "string" then
		win32.parse(parts[1])
	else
		{
			parts = table.clone((parts[1] :: path).parts),
			kind = (parts[1] :: path).kind,
			driveLetter = (parts[1] :: path).driveLetter,
		}
	end
```

Vighnesh-V:
Also will resolve https://github.com/luau-lang/lute/issues/463

lute/std/libs/luau.luau
```diff
@@ -0,0 +1,183 @@
...
+
+export type AstTypePackGeneric = luteLuau.AstTypePackGeneric
+
+export type AstTypePackVariadic = luteLuau.AstTypePackVariadic
+
+export type AstTypePack = luteLuau.AstTypePack
+
+export type ParseResult = luteLuau.ParseResult
+
+export type Bytecode = luteLuau.Bytecode
```
aatxe, line 153: all of these should become luacase ultimately, but it's okay if that doesn't happen in this PR.

lute/luau/src/luau.cpp
```diff
@@ -2657,12 +2657,10 @@ int compile_luau(lua_State* L)
 int load_luau(lua_State* L)
 {
     const std::string* bytecode_string = static_cast<std::string*>(luaL_checkudata(L, 1, COMPILE_RESULT_TYPE));
-    const char* path = luaL_optlstring(L, 2, "=luau.load", nullptr);
-    std::string chunk_name = path;
-    if (chunk_name != "=luau.load")
-        chunk_name.insert(0, "@");
+    const char* path = luaL_checkstring(L, 2);
```
vrn-sn: Can't highlight this to give a suggestion for some reason.
1. We should rename `path` to `chunkname` since at this point it's no longer a file path.
2. nit: `bytecode_string` -> `bytecode` or `bytecodeString` (I don't know why this C++ code chose to use snake case, but we should just change it while we're here).

lute/luau/src/luau.cpp
```diff
@@ -2657,12 +2657,10 @@ int compile_luau(lua_State* L)
...
+    int envIndex = lua_isnoneornil(L, 3) ? 0 : 3;
 
-    luau_load(L, chunk_name.c_str(), bytecode_string->c_str(), bytecode_string->length(), lua_gettop(L) > 2 ? 3 : 0);
+    luau_load(L, path, bytecode_string->c_str(), bytecode_string->length(), envIndex);
```
vrn-sn:
```suggestion
    if (luau_load(L, path, bytecode_string->c_str(), bytecode_string->length(), envIndex) != 0)
        luaL_error(L, "could not load bytecode: %s", lua_tostring(L, -1));
```
Sorry, some of this stuff isn't stuff that you missed, it's just been broken and I'm realizing it now.

vrn-sn: Alternatively, just this:
```suggestion
    if (luau_load(L, path, bytecode_string->c_str(), bytecode_string->length(), envIndex) != 0)
        lua_error(L); // luau_load pushed the error message onto the stack
```

Vighnesh-V:
I don't want to be annoying lol, but I think i'm going to ask that we separate the path.join fix since it's affecting at least a few lute users and we should deliver that faster (it's also uncontroversial).

vrn-sn (approved):
Looks good, I would just be sure to rename the PR title and description with the new API name.

PR #462, opened 2025-10-21
stdlib: add `std/fs` and `std/process` to handle file paths as luau `path` and `pathlike` types instead of `string`

### LINKED ISSUE #433: Update the `fs` library to interact with path objects
Update the `fs` library methods to interact with the userdata represenation of filepaths being added as part of #427

annieetang:
### Problem
Some Lute `fs` and `process` functions have been processing file paths as `string` types while other places in Lute have been using `path` or `pathlike` for file path handling/processing. We want to add support for `path` and `pathlike` file paths to `fs` and `process` functions.

### Solution

We're introducing `fs` and `process` to the Lute standard library as `@std/fs` and `@std/process` that act as wrappers around the existing Lute `fs` and `process` files where we can pass in `pathlike` type paths in `@std/fs` and return `path` type paths in `@std/process` instead of needing to parse and format each call.

There's also a refactor of the `path/` directory that separates out type definitions into a separate file outside of the `init`. This was partially due to a previous attempt of handling path types and running into cyclic dependencies due to the nature of the `process.luau` file, but seemed reasonable to keep around for potential future use.

Added tests and found some minor bugs in `process.cpp` that is also fixed in this pr

Closes https://github.com/luau-lang/lute/pull/462


definitions/process.luau
```diff
@@ -1,3 +1,5 @@
+local pathtypes = require("@std/path/types")
```
annieetang: using `require("@std/path/types")` instead of `require("@std/path")` only in this file because of a cyclic dependency between `process.luau` and both `posix/init.luau` and `win32.luau` files
but other files should be able to just use `require("@std/path")` and the exported types from there (like before the refactor)

annieetang: tried to implement it like this so it doesn't break everyone's usages of the `path` and `pathlike` types (if there are other usages)

annieetang: nvm, talked with ariel and gonna have the definition files use the types files too

lute/std/libs/fs.luau
```diff
@@ -0,0 +1,109 @@
+--!strict
+-- @std/fs
+-- stdlib for `@lute/fs`
+-- Provides file system utility functions that handles file paths as path types instead of strings
+
+local fs = require("@lute/fs")
+local stdPath = require("@std/path")
```
aatxe:
```suggestion
local path = require("@std/path")
```

annieetang: o the function args almost all use `path: pathlike` as the variable - should I change those to something else?

aatxe: you can just leave it then, i suppose. i don't dislike shadowing, but it's definitely a disputed thing.

lute/std/libs/path/posix/types.luau
```diff
@@ -0,0 +1,8 @@
+export type path = {
+	parts: { string },
+	absolute: boolean,
+}
+
+export type pathlike = string | path
+
+return { "successfully exported posix path types" }
```
skanosue: is this needed?

lute/std/libs/path/win32/types.luau
```diff
@@ -0,0 +1,11 @@
...
+
+export type path = {
+	parts: { string },
+	kind: pathkind,
+	driveLetter: string?,
+}
+
+export type pathlike = string | path
+
+return { "successfully exported win32 path types" }
```
skanosue: ditto

lute/std/libs/path/types.luau
```diff
@@ -0,0 +1,7 @@
+local posixtypes = require("@std/path/posix/types")
+local win32types = require("@std/path/win32/types")
+
+export type path = posixtypes.path | win32types.path
+export type pathlike = string | path
+
+return { "successfully exported path types" }
```
skanosue: ditto

lute/std/libs/path/init.luau
```diff
@@ -3,9 +3,10 @@ local system = require("@std/system")
 local posix = require("@self/posix")
 local win32 = require("@self/win32")
 
-export type path = posix.path | win32.path
+local types = require("@self/types")
```
skanosue: nit: this technically shadows the builtin types library

lute/std/libs/fs.luau
```diff
@@ -0,0 +1,109 @@
...
+type HandleMode = fs.HandleMode
+type FileHandle = fs.FileHandle
+type FileType = fs.FileType
+type FileMetadata = fs.FileMetadata
+type DirectoryEntry = fs.DirectoryEntry
+type WatchHandle = fs.WatchHandle
+type WatchEvent = fs.WatchEvent
+type pathlike = stdPath.pathlike
+
+function open(path: pathlike, mode: HandleMode?): FileHandle
```
skanosue: these should all be `local function`

lute/std/libs/fs.luau
```diff
@@ -0,0 +1,109 @@
...
+local fs = require("@lute/fs")
+local stdPath = require("@std/path")
+
+type HandleMode = fs.HandleMode
+type FileHandle = fs.FileHandle
+type FileType = fs.FileType
+type FileMetadata = fs.FileMetadata
+type DirectoryEntry = fs.DirectoryEntry
+type WatchHandle = fs.WatchHandle
+type WatchEvent = fs.WatchEvent
```
skanosue: we should luacase and export these

lute/std/libs/process.luau
```diff
@@ -0,0 +1,42 @@
...
+-- stdlib for `@lute/process`
+-- Provides process-related utility functions that handles file paths as path types instead of strings
+
+local process = require("@lute/process")
+local stdPath = require("@std/path")
+
+type StdioKind = process.StdioKind
+type ProcessRunOptions = process.ProcessRunOptions
+type ProcessResult = process.ProcessResult
+type path = stdPath.path
```
skanosue: luacase

lute/std/libs/process.luau
```diff
@@ -0,0 +1,42 @@
...
+type pathlike = stdPath.pathlike
+
+function homedir(): path
```
skanosue: local function these

lute/std/libs/fs.luau
```diff
@@ -0,0 +1,109 @@
...
+type pathlike = stdPath.pathlike
+
+function open(path: pathlike, mode: HandleMode?): FileHandle
+	return fs.open(stdPath.format(path), mode)
```
skanosue: now might not be the best time, but one day we should probably just rewrite `fs` so the C++ implementation handles path objs as well, i can see a workflow where we're constantly parsing/formatting paths under the hood

lute/std/libs/process.luau
```diff
@@ -0,0 +1,42 @@
...
+-- @std/process
+-- stdlib for `@lute/process`
+-- Provides process-related utility functions that handles file paths as path types instead of strings
+
+local process = require("@lute/process")
+local pathlib = require("@std/path")
+
+type StdioKind = process.StdioKind
+type ProcessRunOptions = process.ProcessRunOptions
+type ProcessResult = process.ProcessResult
```
skanosue: luacase + export pls

skanosue (dismissed):

lute/std/libs/fs.luau
Vighnesh-V: types are BactrianCase right?

lute/std/libs/path/win32/types.luau
aatxe: just `return {}` is fine.

lute/std/libs/path/posix/types.luau
annieetang: i don't think so anymore, i was trying to fix a test failing and forgot to undo this

lute/std/libs/fs.luau
aatxe: all the builtin ones in the definitions were already PascalCase, but yeah, we should make all of the exported type surface be luacase

lute/std/libs/fs.luau
annieetang: I was talking with @vrn-sn yesterday about maybe how requiring `@std/path` or the luau path type in the C++ layer seems a bit weird? since the path obj is technically a higher level than the stuff that's happening in `fs.cpp`? varun, pls correct me if I'm wrong tho

lute/std/libs/path/init.luau
annieetang: yea, how else shld I rename it... `pathtypes`?

lute/std/libs/fs.luau
skanosue: it is a little weird, but i think there's potentially a performance issue here, because even a reasonably simple workflow like:
```
local p = path.join('.', 'tmp')
local handle = fs.open(p, 'w+')
fs.write(handle, 'foobar')
fs.close(handle)
...
fs.remove(p)
```
will result in two calls to `path.format()`

tests/stdfs.test.luau
```diff
@@ -0,0 +1,172 @@
...
+local path = require("@std/path")
+local fs = require("@std/fs")
+
+local tmpdir = "tmp"
+
+test.suite("FsSuite", function(suite)
+	suite:case("open_read_write_close_and_stat", function(assert)
+		if not fs.exists(tmpdir) then
+			fs.mkdir(tmpdir)
+		end
```
skberkeley, line 10: use `suite:beforeeach`/`suite:aftereach`/`suite:beforeall`/`suite:afterall` if you can

tests/stdfs.test.luau
```diff
@@ -0,0 +1,172 @@
...
+
+		local file = path.join(tmpdir, "file.txt")
+		assert.eq(fs.exists(file), false)
+
+		local h = fs.open(file, "w+")
+		fs.write(h, "abc")
+		fs.close(h)
+
+		local m = fs.stat(file)
+		assert.eq(m ~= nil, true)
```
skberkeley:
```suggestion
		assert.neq(m, nil)
```

tests/stdfs.test.luau
```diff
@@ -0,0 +1,172 @@
...
+	end)
+
+	suite:case("cleanup_tmpdir", function(assert)
+		if fs.exists(tmpdir) then
+			local entries = fs.listdir(tmpdir)
+			if #entries == 0 then
+				fs.rmdir(tmpdir)
+			end
+		end
+	end)
```
skberkeley: we should try not to write tests in a way that assume they run in a certain order

tests/stdprocess.test.luau
```diff
@@ -0,0 +1,46 @@
...
+test.suite("ProcessSuite", function(suite)
+	suite:case("homedir_and_cwd_and_execpath", function(assert)
+		local hd = process.homedir()
+		assert.eq(hd ~= nil, true)
+
+		local cwd = process.cwd()
+		assert.eq(cwd ~= nil, true)
+
+		local ex = process.execpath()
+		assert.eq(ex ~= nil, true)
```
skberkeley: use `assert.neq`.

tests/stdprocess.test.luau
```diff
@@ -0,0 +1,46 @@
...
+	end)
+
+	suite:case("run_echo_array", function(assert)
+		local r = process.run({ "echo", "hello-from-lute" })
+		assert.eq(r.exitcode, 0)
+		assert.eq(type(r.stdout) == "string", true)
```
skberkeley: we can also check `assert.eq(r.stdout, "hello-from-lute\n")`

tests/stdprocess.test.luau
```diff
@@ -0,0 +1,46 @@
...
+		local r = process.run({ "sh", "-c", "exit 41" })
+		assert.eq(r.exitcode, 41)
+		assert.eq(r.ok, false)
+	end)
+
+	-- uncomment to test process.exit (will terminate the test runner)
+	-- suite:case("process_exit", function(assert)
+	--     local ok, err = pcall(function() process.exit(0) end)
+	--     assert.eq(ok, false)
+	-- end)
```
skberkeley: dunno that this is worth leaving around

skberkeley (changes requested):
source changes look good!

tests/stdprocess.test.luau
annieetang: code coverage 🥲

tests/stdprocess.test.luau
Vighnesh-V: you can also do `assert.tableeq(r, {exitcode = 0, r.stdout = "actual output"})` :)

tests/stdfs.test.luau
```diff
@@ -0,0 +1,140 @@
...
+
+test.suite("FsSuite", function(suite)
+	suite:beforeall(function()
+		if not fs.exists(tmpdir) then
+			fs.mkdir(tmpdir)
+		end
+	end)
+
+	suite:afterall(function()
+		fs.rmdir(tmpdir)
```
skberkeley, line 14: we should only delete it if it didn't exist before

tests/stdfs.test.luau
```diff
@@ -0,0 +1,140 @@
...
+	end)
+
+	suite:case("open_read_write_close_and_stat", function(assert)
+		local file = path.join(tmpdir, "file.txt")
+		assert.eq(fs.exists(file), false)
+
+		local h = fs.open(file, "w+")
```
skberkeley, line 20: i think we should check `assert.eq(fs.exists(file), true)` here

tests/stdfs.test.luau
```diff
@@ -0,0 +1,140 @@
...
+		fs.write(h, "abc")
+		fs.close(h)
+
+		local m = fs.stat(file)
+		assert.eq(m.size, 3)
+
+		local hr = fs.open(file, "r")
+		local got = fs.read(hr)
+		assert.eq(got, "abc")
```
skberkeley: nit
```suggestion
		assert.eq(fs.read(hr), "abc")
```

tests/stdfs.test.luau
```diff
@@ -0,0 +1,140 @@
...
+		assert.eq(fs.type(tmpdir), "dir")
+
+		fs.remove(file)
+	end)
+
+	suite:case("mkdir_listdir_rmdir", function(assert)
+		local dir = path.join(tmpdir, "subdir")
+		if not fs.exists(dir) then
+			fs.mkdir(dir)
+		end
```
skberkeley, line 61: are you expecting dir to ever exist already?

tests/stdfs.test.luau
```diff
@@ -0,0 +1,140 @@
...
+
+		assert.eq(fs.exists(dir), true)
+		assert.eq(fs.type(dir), "dir")
+
+		local entries = fs.listdir(tmpdir)
+		local found = false
+		for _, e in ipairs(entries) do
```
skberkeley: i don't think you need `ipairs`

tests/stdfs.test.luau
```diff
@@ -0,0 +1,140 @@
...
+			if e.name == "subdir" then
+				found = true
+			end
+		end
+		assert.eq(found, true)
+		fs.rmdir(dir)
```
skberkeley, line 74: can you check that dir is gone afterwards?

tests/stdfs.test.luau
```diff
@@ -0,0 +1,140 @@
...
+			if tostring(filename):find("watched.txt") then
+				triggered = true
+			end
+		end)
+
+		fs.writestringtofile(watched, "x")
+
+		local start = os.time()
+		while not triggered and os.time() - start < 2 do
+		end
```
skberkeley, line 127: i hate it, but ig we don't have a better way to do it?

skberkeley (changes requested):
just a few fs comments

tests/stdfs.test.luau
annieetang: oh it's `false` because if the file exist before we `fs.open` then we shouldn't be trying to run `write` calls on an existing file we didn't create, but can I just leave it as is and fix it when I refactor to use `system.tmpdir`

skberkeley: oh sorry, i meant checking that the file exists _after_ we call open

annieetang:
updated most comments, left the dir stuff to be fixed when I update to use `system.tempdir` if that's ok!

skberkeley (approved):

Vighnesh-V (approved):

PR #466, opened 2025-10-22
runtime: implement `system.tmpdir` to provide access to the temporary files directory.

aatxe:
Varying OSes provide different preferred locations for temporary files, and libuv provides an API by which we can get at that. We should make that available so that people can build cross-platform code that interacts with these directories safely.

### DIFF
--- a/lute/system/src/system.cpp
+++ b/lute/system/src/system.cpp
@@ -123,6 +123,30 @@ int lua_uptime(lua_State* L)
 
     return 1;
 }
+
+int lua_tmpdir(lua_State* L)
+{
+    size_t size = 255;
+    std::string tmpdir;
+    tmpdir.reserve(size);
+
+    int res = uv_os_tmpdir(tmpdir.data(), &size);
+    if (res == UV_ENOBUFS)
+    {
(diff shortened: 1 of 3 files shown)


annieetang (approved):
tysm!!

lute/system/src/system.cpp
```diff
@@ -123,6 +123,30 @@ int lua_uptime(lua_State* L)
...
+        tmpdir.reserve(size); // libuv updates the size to what's required
+        res = uv_os_tmpdir(tmpdir.data(), &size);
+    }
+
+    if (res != 0)
+    {
+        luaL_error(L, "libuv error: %s", uv_strerror(res));
+    }
+
+    lua_pushstring(L, tmpdir.c_str());
```
SPY, line 145: I would suggest to use `lua_pushlstring` for cases when a string's length is known.

aatxe: Ah, good point, we should correct this across a lot of code. Thanks!

aatxe: (I know this because I copy-pasted the code from another function in the first place that has the same mistake)

PR #468, opened 2025-10-23
std/json: distinguish more explicitly between array and object

aatxe:
We got some useful feedback from @vocksel about not being able to get a clean refinement between array and object with the existing types setup, and the correct solution to this problem here is for us to leverage another newproxy as a key to distinguish between objects and arrays. We've provided some additional functions to the interface of `json` both to construct `object`s and to test against values being `object`s or `array`s. There's also some clean up in here for json as a library generally, including correcting its casing now that it's a standard library and including `object` and `array` as part of the exported types as well.

### DIFF
--- a/lute/std/libs/json.luau
+++ b/lute/std/libs/json.luau
@@ -1,12 +1,15 @@
 local json = {
 	--- Not actually a nil value, a newproxy stand-in for a null value since Luau has no actual representation of `null`
-	NULL = newproxy() :: nil,
+	null = newproxy() :: nil,
 }
 
-type JSONPrimitive = nil | number | string | boolean
-type Object = { [string]: Value }
-type Array = { Value }
-export type Value = JSONPrimitive | Array | Object
+type jsonprimitive = nil | number | string | boolean
+type object = { [string]: value }
+type array = { [number]: value }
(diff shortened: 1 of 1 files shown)


Vighnesh-V (approved):

lute/std/libs/json.luau
```diff
@@ -1,12 +1,14 @@
...
 	--- Not actually a nil value, a newproxy stand-in for a null value since Luau has no actual representation of `null`
-	NULL = newproxy() :: nil,
+	null = newproxy() :: nil,
 }
 
-type JSONPrimitive = nil | number | string | boolean
-type Object = { [string]: Value }
-type Array = { Value }
-export type Value = JSONPrimitive | Array | Object
+export type object = { [string]: value }
```
Vighnesh-V, line 6: I thought exported types were PascalCase

vrn-sn: No, we standardized this in the CONTRIBUTING.md:
> All Luau APIs exposed publically by Lute in both the runtime and the standard library must have identifiers written in luacase. This means module names, function names, table fields, exported types, etc. should be written in all lowercase, and should be named ideally with one or two words succinctly.

PR #469, opened 2025-10-23
Port AST serializer tests to new runner

skberkeley:
This needed to happen sooner or later, and I preferred sooner because the old tests were spitting out a bunch of output which makes it hard to debug CI


Vighnesh-V (approved):
very nice 🚀

PR #472, opened 2025-10-23
Skanosue/codemod queries

skberkeley:


lute/std/libs/syntax/query.luau
```diff
@@ -0,0 +1,50 @@
+local luau = require("@std/luau")
```
Vighnesh-V: I'm guessing you're splitting the codemod PR up? Can we add a selectRequires query as well?

lute/std/libs/syntax/query.luau
```diff
@@ -0,0 +1,50 @@
...
+	:filter(function(call)
+		local firstArg = call.arguments[1].node
+		return firstArg.tag == "indexname" and firstArg.index.text == "OldComponent"
+	end)
+
+local replacements = select(p, function(call: luau.AstExprCall)
+	return call.arguments[1].node :: luau.AstExprIndexName
+end):replace(function(node)
+	return "NewComponent"
+end)
```
wmccrthy:
```suggestion
end)

return {
	newQuery = newQuery,
	select = select,
}

-- I think a really nice addition to this would be having a 'Path' like class that wraps nodes internally (within the query implementation
-- this class would have methods for:
-- 	replace
-- 	remove
-- 	insertAfter
-- 	insertBefore
-- which makes adding those methods on the query object much easier and more intuitive IMO
```

PR #473, opened 2025-10-23
stdlib: wrap tmpdir result in path

skberkeley:
Wrap `system.tmpdir` in path to make our stdlib interfaces more consistent

### DIFF
--- a/lute/std/libs/system.luau
+++ b/lute/std/libs/system.luau
@@ -5,12 +5,19 @@
 
 local system = require("@lute/system")
 
+local path = (require)("@std/path") -- trickery to avoid circular dependency
+local pathTypes = require("@std/path/types")
+
 local osName = string.lower(system.os)
 
 local win32 = osName == "windows_nt"
 local linux = osName == "linux"
 local macos = osName == "darwin"
 local unix = macos or linux or osName == "unix" or osName == "posix"
+
+local function tmpdir(): pathTypes.path
+	return path.parse(system.tmpdir())
+end
 
 return table.freeze({
 	os = system.os,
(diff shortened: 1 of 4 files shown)


lute/std/libs/system.luau
```diff
@@ -5,13 +5,20 @@
...
 
+local path = (require)("@std/path") -- trickery to avoid circular dependency
+local pathTypes = require("@std/path/types")
+
 local osName = string.lower(system.os)
 
 local win32 = osName == "windows_nt"
 local linux = osName == "linux"
 local macos = osName == "darwin"
 local unix = macos or linux or osName == "unix" or osName == "posix"
```
vrn-sn: If we can move this logic to a submodule (e.g. `@std/system/platforminfo`), we can break the circular dependency:
1. `@std/system` requires `@self/platforminfo` so that it can continue to export these booleans.
2. `@std/system` requires `@std/path` so that we can define `tmpdir`.
3. `@std/path` requires `@std/system/platforminfo` so that it can do its platform-specific logic.

lute/std/libs/system.luau
```diff
@@ -5,13 +5,20 @@
 
 local system = require("@lute/system")
 
+local path = (require)("@std/path") -- trickery to avoid circular dependency
+local pathTypes = require("@std/path/types")
```
vrn-sn: If we break the circular dependency here, do we still need to get the types like this, or can we access them through `path`?

lute/std/libs/system.luau
vrn-sn: Don't have to solve it exactly like that, but typically an edit-time circular dependency that isn't present at runtime means that we can split things up into tighter-scoped modules.

aatxe: Yeah, this is a good idea.

lute/std/libs/system.luau
skberkeley: yeah pathtypes is because of the circular import

vrn-sn (approved):

PR #478, opened 2025-10-24
Fix multiple runtime APIs that were incorrectly using `std::string::reserve`

vrn-sn:
### Problem

Multiple runtime APIs incorrectly were using `std::string::reserve` where they should have been using `std::string::resize`. This led to UB that seemed to behave fine at runtime, but the introduction of #467 caused all of these APIs to break.

### Solution

We introduce `uvutils::getStringFromUv` as a utility API to help reduce errors when interacting with libuv. Internally, this handles all the buffer logic and simply returns a value or an error to the caller.

In fixing this, I also discovered two issues with our CI:
1. One of the checks in `path.posix.test.luau` is flipped but was mistakenly passing because of the previously broken `process.cwd()` API.
2. Our Luau tests don't even run on Windows! 😞

I've fixed (1) and have made a separate issue for (2): #483.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,8 +43,7 @@ jobs:
         run: rm .luaurc
 
       - name: Run Luau Tests
-        shell: bash
-        run: find tests -name "*.test.luau" | xargs -I {} $(echo '${{ steps.build_lute.outputs.exe_path }}' | tr '\\' '/') run {}
+        run: find tests -name "*.test.luau" | xargs -I {} ${{ steps.build_lute.outputs.exe_path }} run {}
 
   run-lute-cxx-unittests:
     runs-on: ${{ matrix.os }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,6 +43,7 @@ jobs:
         run: rm .luaurc
 
       - name: Run Luau Tests
+        shell: bash
         run: find tests -name "*.test.luau" | xargs -I {} ${{ steps.build_lute.outputs.exe_path }} run {}
 
   run-lute-cxx-unittests:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,8 +3,8 @@ name: CI
 on:
   push:
     branches: ["primary"]
-  pull_request:
-    branches: ["primary"]
+  # pull_request:
+  #   branches: ["primary"]
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,7 +44,7 @@ jobs:
 
       - name: Run Luau Tests
         shell: bash
-        run: find tests -name "*.test.luau" | xargs -I {} ${{ steps.build_lute.outputs.exe_path }} run {}
+        run: find tests -name "*.test.luau" | xargs -I {} $(echo '${{ steps.build_lute.outputs.exe_path }}' | tr '\\' '/') run {}
 
   run-lute-cxx-unittests:
     runs-on: ${{ matrix.os }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,8 +3,8 @@ name: CI
 on:
   push:
     branches: ["primary"]
-  # pull_request:
-  #   branches: ["primary"]
+  pull_request:
+    branches: ["primary"]
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
--- a/lute/runtime/include/lute/uvutils.h
+++ b/lute/runtime/include/lute/uvutils.h
@@ -1,0 +1,20 @@
+#pragma once
+
+#include <cstddef>
+#include <string>
+
+namespace uvutils
+{
+
+struct StringResult
+{
+    int status = 0;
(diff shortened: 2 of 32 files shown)


.github/workflows/ci.yml
```diff
@@ -43,7 +43,8 @@ jobs:
         run: rm .luaurc
 
       - name: Run Luau Tests
-        run: find tests -name "*.test.luau" | xargs -I {} ${{ steps.build_lute.outputs.exe_path }} run {}
+        shell: bash
```
Vighnesh-V: does this just use git bash on windows?

lute/runtime/include/lute/uvutils.h
```diff
@@ -0,0 +1,20 @@
...
+namespace uvutils
+{
+
+struct StringResult
+{
+    int status = 0;
+    std::string value;
+};
+
+typedef int (*BufferWriter)(char* buffer, size_t* size);
```
Vighnesh-V, line 19: nit: Let's prefer std::function over C style function pointers - we've had to do this refactor at least a couple times in the codebase.

lute/runtime/src/uvutils.cpp
```diff
@@ -0,0 +1,29 @@
...
+        // `size` now contains the required size
+        buffer.resize(size);
+        writeStatus = bufferWriter(buffer.data(), &size);
+    }
+
+    if (writeStatus < 0)
+        return StringResult{writeStatus, ""};
+
+    buffer.resize(size);
+    return StringResult{writeStatus, std::move(buffer)};
```
Vighnesh-V: This is kind of weird because it mixes C style code with C++ style. You're writing this in C style and using it C style, but giving it a C++ style interface.

E.g. in `system.cpp`:
```
    uvutils::StringResult result = uvutils::getStringFromUv(uv_os_gethostname);
    if (result.status < 0)
        luaL_error(L, "libuv error: %s", uv_strerror(result.status));
```

If you want to do C style:
```
int getStringFromUv(BufferWriter bufferWriter, size_t initialBufferSize, std:string& outBuf)
```

If you want to do C++ style and don't care about the cause of the error, just knowing that there is an error:
```
std::optional<std::string> getStringFromUv(BufferWriter bufferWriter, size_t initialBufferSize)
```
If you do care about the cause of the error, you could maybe do Luau::Variant

Vighnesh-V (changes requested):
Requesting some small changes, but this broadly looks good to go.

lute/runtime/src/uvutils.cpp
Vighnesh-V: I made this mistake a bunch in the bytecode compilation PR and I need to go clear that up :)

.github/workflows/ci.yml
vrn-sn: Yeah, I think the `find` command was silently not doing anything before and somehow passing in CI, despite running nothing.

lute/runtime/include/lute/uvutils.h
vrn-sn: This is one where I don't actually think we should. The idea is for `BufferWriter` to represent all those libuv functions that write to a buffer, and those will always use a pure C interface. (i.e. we will never pass a C++ lambda that captures anything to this, so I don't think we need the extra space for `std::function`).

lute/runtime/src/uvutils.cpp
vrn-sn: I can replace with something using `Luau::Variant`. I was just trying to hide away the C-style buffer logic but didn't really attempt to hide all the C-style stuff that shows up with using libuv (but I can do that).

vrn-sn:
Currently also trying to figure out why Windows tests are failing now that I'm trying to re-enable them.

Edit: giving up on this for now and reverting these Windows-related changes.

lute/runtime/src/uvutils.cpp
Vighnesh-V: I think from the way this is being used, returning a std::optional is probably the right move!

vrn-sn: We do care about the cause of the error though since it's often reported to Luau.

lute/runtime/include/lute/uvutils.h
Vighnesh-V: That makes sense, I'm convinced.

lute/runtime/src/uvutils.cpp
vrn-sn: I switched this out with `Luau::Variant` if you want to take a look. We still need the error message for reporting, so I've made the return type `Luau::Variant<std::string, UvError>`.

lute/system/src/system.cpp
```diff
@@ -88,24 +91,12 @@ int lua_totalmemory(lua_State* L)
...
-        res = uv_os_gethostname(hostname.data(), &sz);
-    }
-
-    if (res != 0)
-    {
-        luaL_error(L, "libuv error: %s", uv_strerror(res));
-    }
-
-    lua_pushlstring(L, hostname.c_str(), hostname.size());
+    Luau::Variant<std::string, uvutils::UvError> result = uvutils::getStringFromUv(uv_os_gethostname);
```
Vighnesh-V: nit: auto here for readability (and in the other places where this is done).

Vighnesh-V (approved):
Some style comments, but not blocking.

PR #484, opened 2025-10-27
feature: Introduce a multifile payload format for embedding into the lute binary

Vighnesh-V:
This PR is an incremental step towards supporting multi-file bytecode compilation. This PR defines a `LuteExePayload`, for encoding and decoding the multi file bytecode bundles that will be appended to the `lute` binary when shipping executables.

Once this PR is in, we can start writing the glue code that will let us produce a lute executable with multiple .luau files embedded.

### DIFF
--- a/lute/cli/include/lute/compile.h
+++ b/lute/cli/include/lute/compile.h
@@ -1,6 +1,9 @@
 #pragma once
 
+#include "Luau/DenseHash.h"
 #include "Luau/FileUtils.h"
+
+#include <string_view>
 
 struct AppendedBytecodeResult
 {
@@ -11,3 +14,56 @@ struct AppendedBytecodeResult
 AppendedBytecodeResult checkForAppendedBytecode();
 
 int compileScript(const std::string& inputFilePath, const std::string& outputFilePath);
+
+struct LuteDecodeResult;
+
+struct LuteEncodeResult
+{
+    std::string payload;
+    size_t bytesWritten = 0;
+    size_t compressedPayloadSizeBytes = 0;
+    size_t uncompressedPayloadSizeBytes = 0;
+};
+
+/**
+ * Represents a bundle of compiled Luau files ready for injection.
+ *
+ * Uncompressed bundle format (before compression):
+ *   For each file:
+ *     [path_length: uint32_t]
+ *     [path_string: char[path_length]]
+ *     [bytecode_size: uint64_t]
+ *     [bytecode_data: byte[bytecode_size]]
+ *
+ * Injectable payload format (after compression, what goes into executable):
+ *   [compressed_bundle_data: byte[compressed_size]]
+ *   [compressed_size: uint64_t]
+ *   [uncompressed_size: uint64_t]
+ *   [num_files: uint32_t]
+ *   [entry_point_path_string: char[entry_point_path_length]]
+ *   [entry_point_path_length: uint32_t]
+
+ */
+struct LuteExePayload
+{
+    LuteExePayload() = default;
+    void add(const std::string& luauFilePath);
+
+    std::optional<LuteEncodeResult> encode();
+    static std::optional<LuteDecodeResult> decode(const std::string_view binary);
+
+    std::string entryPointPath;
+    Luau::DenseHashMap<std::string, std::string> filePathToBytecode{""}; // path -> bytecode
+
+private:
+    bool parseFromDecompressedBundle(std::string_view decompressedBundle);
+    std::vector<std::string> filePaths;
+};
+
+struct LuteDecodeResult
+{
+    LuteExePayload payload;
+    size_t bytesRead = 0;
+    size_t compressedPayloadSizeBytes = 0;
+    size_t uncompressedPayloadSizeBytes = 0;
+};
--- a/lute/cli/src/compile.cpp
+++ b/lute/cli/src/compile.cpp
@@ -159,9 +159,7 @@ void LuteExePayload::add(const std::string& luauFilePath)
 {
     // First file added becomes the entry point
     if (filePaths.empty())
-    {
         entryPointPath = luauFilePath;
-    }
 
     filePaths.push_back(luauFilePath);
 }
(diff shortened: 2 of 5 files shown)


lute/cli/include/lute/compile.h
```diff
@@ -11,3 +14,57 @@ struct AppendedBytecodeResult
...
+};
+
+struct LuteDecodeResult
+{
+    LuteExePayload payload;
+    size_t bytesRead = 0;
+    size_t compressedPayloadSizeBytes = 0;
+    size_t uncompressedPayloadSizeBytes = 0;
+};
+
```
annieetang: could we move this up to where it's first defined in L18?

lute/cli/include/lute/compile.h
```diff
@@ -11,3 +14,57 @@ struct AppendedBytecodeResult
...
+ *   [entry_point_path_length: uint32_t]
+
+ */
+struct LuteExePayload
+{
+    LuteExePayload() = default;
+    void add(const std::string& luauFilePath);
+
+    std::optional<LuteEncodeResult> encode();
+    static std::optional<LuteDecodeResult> decode(const std::string_view binary);
```
annieetang, line 53: does `decode` need to be `static`?

lute/cli/src/compile.cpp
```diff
@@ -146,3 +153,274 @@ int compileScript(const std::string& inputFilePath, const std::string& outputFil
...
     return 0;
 }
+
+void LuteExePayload::add(const std::string& luauFilePath)
+{
+    // First file added becomes the entry point
+    if (filePaths.empty())
+    {
+        entryPointPath = luauFilePath;
+    }
```
annieetang: nit: could we get rid of the `{}` here and L172-174? and some other places in `decode()`

lute/cli/src/compile.cpp
```diff
@@ -146,3 +153,274 @@ int compileScript(const std::string& inputFilePath, const std::string& outputFil
...
+    int compressResult = compress2(
+        compressedData.data(),           // destination buffer
+        &compressedSize,                 // in/out: buffer size / actual compressed size
+        reinterpret_cast<const Bytef*>(uncompressedBundle.data()),  // source data
+        uncompressedSize,                // source size
+        Z_BEST_COMPRESSION               // compression level (9)
+    );
+
+    if (compressResult != Z_OK)
+        return std::nullopt;
```
annieetang, line 234: should we error in the cases where we failed to compress? or is returning null fine?

lute/cli/src/compile.cpp
```diff
@@ -146,3 +153,274 @@ int compileScript(const std::string& inputFilePath, const std::string& outputFil
...
+    size_t pos = binary.size() - MAGIC_FLAG_SIZE;
+
+    // Read entry_point_path_length
+    if (pos < sizeof(uint32_t))
+    {
+        return std::nullopt;
+    }
+    pos -= sizeof(uint32_t);
+    uint32_t entryPointPathLength;
+    memcpy(&entryPointPathLength, binary.data() + pos, sizeof(uint32_t));
```
annieetang: is there a way we can pull this out into a helper function or lambda? the next few lines look like it's generally the same checks for pos and memcpy things?

lute/cli/src/compile.cpp
```diff
@@ -146,3 +153,274 @@ int compileScript(const std::string& inputFilePath, const std::string& outputFil
...
+        {
+            fprintf(stderr, "Invalid bundle: incomplete bytecode data\n");
+            return false;
+        }
+
+        std::string bytecode(decompressedBundle.data() + offset, bytecodeSize);
+        offset += bytecodeSize;
+
+        // Store in map
+        filePathToBytecode[filePath] = bytecode;
```
annieetang, line 449: can we use `.emplace()` here instead?

Vighnesh-V: DenseHash doesn't actually implement an emplace, so not in this case :)

lute/cli/include/lute/compile.h
Vighnesh-V: That forward declaration is intentional. Note that `decode` returns a `LuteDecodeResult` with contains a `LuteExePayload` member by value. In order to do that, the compiler has to have the full definition of `LuteExePayload` up front, otherwise it can't work out the sizes of the struct. That's why we have to move the definition of `LuteDecodeResult` to after `LuteExePayload`, since that way it can see the full definition of `LuteExePayload`. Because `LuteDecodeResult` is inside of an optional, it's okay to forward declare it and reference it inside of `LuteExePayload`.

lute/cli/src/compile.cpp
Vighnesh-V: hmm, so returning `nullopt` in this case is good enough for knowing whether or not there is an error -e.g. you were able to encode / decode. I suppose we can printf to stderr to let the user know exactly what went wrong? Does that seem reasonable as a compromise?

lute/cli/src/compile.cpp
Vighnesh-V: can do!

lute/cli/include/lute/compile.h
Vighnesh-V: I think it does - the intention here is to have `std::optional` represent the error case when you're unable to decode. If decoding succeeds, we also want to return some statistics about compression + the executable payload. In this case, you don't actually need a LuteExePayload instance to decode - the decode method should construct one for you imo, which lends itself to a static implementation. If this was a member method, then it might look something like:
```
LuteExePayload p;
optional<results> r = p.decode(binary)
if (r)
{
   ...
}
```

as opposed to:
```
if (auto result = LuteExePayload::decode(binary))
{
   ... handle success.
}
```

lute/cli/src/compile.cpp
annieetang: ooh my bad!

lute/cli/src/compile.cpp
annieetang: ya sounds good with me! I saw you use printf in other places so maybe?

lute/cli/src/compile.cpp
Vighnesh-V: I pulled this out into a lambda.

annieetang (approved):
ltgm!

PR #485, opened 2025-10-27
stdlib: fix memory leak in luau.compile

skberkeley:
Fixed a memory leak which was causing GC for my other changes to fail. Specifically, there was a `std::string` which wasn't being freed because it was being moved into a userdata object, so I attached a destructor to it. I also got rid of the tag which was being applied to the userdata object being constructed because it wasn't being checked, and we were already applying a type-specific metatable to it. Thanks @~vrn-sn for helping me come up the fix!

Also added a missing export for `system.tmpdir`.

### DIFF
--- a/lute/luau/src/luau.cpp
+++ b/lute/luau/src/luau.cpp
@@ -2649,7 +2649,7 @@ int compile_luau(lua_State* L)
         sizeof(std::string),
         [](void* ptr)
         {
-            static_cast<std::string*>(ptr)->~basic_string();
+            static_cast<std::string*>(ptr)->~std::basic_string();
         }
     ));
 
--- a/lute/luau/src/luau.cpp
+++ b/lute/luau/src/luau.cpp
@@ -2644,7 +2644,14 @@ int compile_luau(lua_State* L)
 
     std::string bytecode = Luau::compile(std::string(source, source_size), opts);
 
-    std::string* userdata = static_cast<std::string*>(lua_newuserdatatagged(L, sizeof(std::string), kCompilerResultTag));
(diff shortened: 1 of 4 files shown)


Vighnesh-V (approved):

lute/luau/src/luau.cpp
```diff
@@ -2644,7 +2644,14 @@ int compile_luau(lua_State* L)
 
     std::string bytecode = Luau::compile(std::string(source, source_size), opts);
 
-    std::string* userdata = static_cast<std::string*>(lua_newuserdatatagged(L, sizeof(std::string), kCompilerResultTag));
+    std::string* userdata = static_cast<std::string*>(lua_newuserdatadtor(
+        L,
+        sizeof(std::string),
+        [](void* ptr)
+        {
+            static_cast<std::string*>(ptr)->~basic_string();
```
vrn-sn:
```suggestion
            static_cast<std::string*>(ptr)->~std::basic_string();
```
suggestion—unless you already tried this and it doesn't compile for some C++ reason (wouldn't be surprised) 😄

vrn-sn (approved):

lute/luau/src/luau.cpp
vrn-sn: Guess I was wrong, wonder if ~std::string() works or maybe std::~string(). Might be worth testing locally. My main concern is about basic_string's destructor not covering everything that string allocates. Another option is maybe ~string().

skberkeley: looks like `std::string::~string()` works

PR #488, opened 2025-10-28
Lute transform can use stdlib loadbypath

skberkeley:
Fixes failing lute transform unit test

### DIFF
--- a/lute/cli/commands/transform/init.luau
+++ b/lute/cli/commands/transform/init.luau
@@ -1,28 +1,13 @@
 local fs = require("@lute/fs")
-local luau = require("@lute/luau")
+local luau = require("@std/luau")
 local pathLib = require("@std/path")
 
 local arguments = require("@self/lib/arguments")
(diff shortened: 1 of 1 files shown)


Vighnesh-V (approved):

lute/cli/commands/transform/init.luau
```diff
@@ -1,28 +1,13 @@
 local fs = require("@lute/fs")
-local luau = require("@lute/luau")
+local luau = require("@std/luau")
 local pathLib = require("@std/path")
 
 local arguments = require("@self/lib/arguments")
 local files = require("@self/lib/files")
 local types = require("@self/lib/types")
 
-local function exhaustiveMatch(value: never): never
```
JohnnyMorganz, line 9: Was removing this function intentional? I still see it referenced later in the file

skberkeley: no definitely a mistake, i'll add it back rn

PR #490, opened 2025-10-28
refactor: Add some utilities for reading and writing Lute Executables in a more systematic fashion

Vighnesh-V:
This PR adds a `LuteExecutable` wrapper, that handles extracting the `LuteExePayload` out of a binary and also creating a valid `lute` executable with the payload injected in correctly. This will set us up to be able to do cross compilation a bit easier, by separating the `payload` from the actual act of manipulating an executable file.

### DIFF
--- a/lute/cli/include/lute/compile.h
+++ b/lute/cli/include/lute/compile.h
@@ -67,3 +67,27 @@ struct LuteDecodeResult
     size_t compressedPayloadSizeBytes = 0;
     size_t uncompressedPayloadSizeBytes = 0;
 };
+
+/**
+ * Manages creating and reading Lute executables with embedded bytecode bundles.
+ *
+ * Binary format (from end of file, reading backward):
+ *   [lute runtime executable's bytes]
+ *   [compressed bundle data]
+ *   [compressed_size: uint64_t]
+ *   [uncompressed_size: uint64_t]
+ *   [num_files: uint32_t]
+ *   [entry_point_path_string: char[entry_point_path_length]]
+ *   [entry_point_path_length: uint32_t]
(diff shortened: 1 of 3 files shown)


lute/cli/include/lute/compile.h
```diff
@@ -67,3 +67,27 @@ struct LuteDecodeResult
...
+ * Binary format (from end of file, reading backward):
+ *   [lute runtime executable's bytes]
+ *   [compressed bundle data]
+ *   [compressed_size: uint64_t]
+ *   [uncompressed_size: uint64_t]
+ *   [num_files: uint32_t]
+ *   [entry_point_path_string: char[entry_point_path_length]]
+ *   [entry_point_path_length: uint32_t]
+ *   [MAGIC_FLAG: "LUTEBYTE" (8 bytes)]
+ */
```
aatxe, line 83: Is there a good reason that this binary format is backwards? It really feels like it should just be `LUTEBYTE` followed by the rest of the data in order, and we should scan to the magic flag and then read forward from there. This is what I would _expect_, so I think there should be a clear reason why the format deviates from that.

Vighnesh-V: Yeah! The reason is that seeking to the very end of the file and looking for the `LUTEBYTE` magic flag is faster than looking for it somewhere in the middle of the binary. Another nice reason is that if LUTEBYTE somehow showed up in the binary before the actual payload, it's possible that we might fail to decode correctly.

Vighnesh-V: I actually found a bug with my implementation - I was reading the file into memory before checking the LUTEBYTE at the end, when we can just seek to the end and check LUTEBYTE before reading this into memory..

aatxe (approved):

PR #491, opened 2025-10-29
@std/fs: replace mkdir with createdirectory to handle pathlike objects and create intermediate directories

### LINKED ISSUE #413: Wrapped mkdir that allows `-p` arguments.
Related to new fs api functions - https://github.com/luau-lang/lute/issues/308

Right now there is an implementation already in the lute setup commands that looks like:
```
function mkdirdashp(home: string, subpath: string)
	local subdir = home
	for _, part in string.split(subpath, "/") do
		subdir = subdir .. "/" .. part
		if not fs.exists(subdir) then
			fs.mkdir(subdir)
		end
	end
end
```
we should bring this in as a standard lib file system function with the api:
```
fileio.mkdir : string -> ()/success/failure
```

annieetang:
### Problem

Resolves https://github.com/luau-lang/lute/issues/413

We're currently using `fs.mkdir` for all use cases without the option of making intermediate parent folders, and the `mkdirdashp` function in lute setup commands can be moved to the `std/fs` library for broader use cases.

### Solution

We replace `fs.mkdir` in the standard library with `fs.createdirectory` (a more user friendly name!) that takes in a `path: pathlike` argument as well as an `options` type to specify if the user wants to make intermediate parent folders with `makeparents: boolean` 
For example: `fs.createdirectory(tmpdir, { makeparents = true })` (see more in `examples/create_directory.luau`)

We also update the lute setup commands file to use `fs.createdirectory()` and add/update relevant tests


lute/std/libs/fs.luau
```diff
@@ -88,6 +91,26 @@ local function readasync(filepath: pathlike): string
 	return fs.readasync(pathlib.format(filepath))
 end
 
+local function createdirectory(path: pathlike, options: createdirectoryoptions): ()
```
Vighnesh-V: These options should be optional! When they're nil, this should just do mkdir under the hood.

examples/create_directory.luau
```diff
@@ -0,0 +1,5 @@
+local fs = require("@std/fs")
+local path = require("@std/path")
+
+local directory = path.parse("example/new_directory/subdir")
```
aatxe: maybe make this make the directory in tmpdir?

lute/cli/commands/setup/init.luau
```diff
@@ -50,11 +36,12 @@ if not table.find(args, "--with-luaurc") then
 end
 
 local luauRcPath = `{cwd}/.luaurc`
+local luauRcPathPathlike = path.parse(luauRcPath)
```
aatxe: do paths not have an `__tostring`? if they do (which they should), we can just parse it for the path instead of having two locals here

tests/std/fs.test.luau
```diff
@@ -133,6 +133,25 @@ test.suite("FsSuite", function(suite)
 			fs.remove(watched)
 		end
 	end)
+
+	suite:case("createdirectory_makeparents_true", function(assert)
+		local nestedDir = path.join(tmpdir, "nested", "directory")
+		fs.createdirectory(nestedDir, { makeparents = true })
+		assert.eq(fs.exists(nestedDir), true)
+
+		-- TODO: create a fs.removedirectory with option to remove parents as well?
```
aatxe: Not sure that this would really make sense (the remove parents option). How would you know when to stop? Do we just remove the whole disk?

annieetang: haha yea that's why I didn't implement it yet cuz I realized that might not be a great option to have...

examples/create_directory.luau
annieetang: ah this was just an example folder but I can pop it in a tmpdir

lute/cli/commands/setup/init.luau
annieetang: asked sora and he said there isn't yet but I'll add it!

Vighnesh-V (dismissed):

lute/std/libs/fs.luau
```diff
@@ -15,6 +15,10 @@ export type watchhandle = fs.WatchHandle
 export type watchevent = fs.WatchEvent
 export type pathlike = pathlib.pathlike
 
+export type createdirectoryoptions = {
+	makeparents: boolean,
```
aatxe: each individual option should probably be optional

aatxe (approved):

PR #493, opened 2025-10-29
refactor: rename exprifelse field names

skberkeley:
These field names are [misleading](https://home.sandiego.edu/~baber/logic/conditionals), so rename them to match what we have in our C++ Ast types. Also, stop exporting AstExprIfElseIfs, because it's more like a utility type rather than an actual AST node.

### DIFF
--- a/definitions/luau.luau
+++ b/definitions/luau.luau
@@ -233,7 +233,7 @@ export type AstStatElseIf = {
 	elseifkeyword: Token<"elseif">,
 	condition: AstExpr,
 	thenkeyword: Token<"then">,
-	consequent: AstStatBlock,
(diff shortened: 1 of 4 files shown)


hgoldstein (changes requested):
Are we about to break the flag clipper?

skberkeley:
> Are we about to break the flag clipper?

I'll consult Wyatt

skberkeley:
> Are we about to break the flag clipper?

Wyatt has already pinned the lute version for consumer repos, so nothing will break immediately. At some point I'll help him port stuff over when he unpins lute

hgoldstein (approved):
This is a breaking change: we will need to update any downstream consumers but this LGTM.

PR #494, opened 2025-10-29
lute transform: add exhaustive match back

skberkeley:
`exhaustiveMatch` got accidentally deleted in a previous PR. Also took the chance to update the transform test since we have `tempDir` now. The migration processing code is already written in a way that `exhaustiveMatch` is never actually reachable right now, but it's a good check to have in case we expand the migration API in the future.


annieetang (approved):

annieetang:
is there something we can do to avoid accidentally deleting functions... code coverage metrics?

update: synced offline!

PR #495, opened 2025-10-29
restructure std tests

skberkeley:
I moved some tests around to match the source folder structure so it's more obvious where tests should go when they're added in the future. Open to any feedback!


tests/loadbypath.test.luau
```diff
@@ -32,7 +32,7 @@ test.suite("Lute CLI", function(suite)
 		end
 	end)
 
-	suite:case("help1", function(check)
```
annieetang, line 35: 💀

tests/cli/compile.test.luau
```diff
@@ -23,7 +23,7 @@ local COMPILEE_CONTENTS = [[return "Hello world"]]
 local lutePath = process.execpath()
 local tmpDir = system.tmpdir()
 
-test.suite("Lute CLI", function(suite)
+test.suite("Lute CLI Compile", function(suite)
 	suite:case("help1", function(check)
```
annieetang: could u also change this case name haha

vrn-sn (approved):

PR #496, opened 2025-10-29
Add `luau.resolverequire` API

vrn-sn:
Open to bikeshedding about naming and usage patterns for this API. Just wanted to get a PR up with the core logic since many people have asked for this.

As of now, `luau.staticrequire` allows us to resolve a given require path (`path`) relative to a given file (`fromchunkname`), returning the absolute file path of the target module. Because this is built on `Luau.RequireNavigator`, aliases defined in `.luaurc` files are also handled correctly, and we'll be able to trivially add support for `.config.luau` files as well since that's being added to `Luau.RequireNavigator` in an upcoming Luau release.


vrn-sn:
Example usage right now:
```
static_require_test
├── .luaurc
├── dependency.luau
└── entry.luau
```

.luaurc
```json
{
    "aliases": {
        "dep": "./dependency"
    }
}
```

dependency.luau
```luau
return {}
```

entry.luau
```luau
local luau = require("@lute/luau")
print(`@dep resolves to: {luau.staticrequire("@dep", "@./entry.luau")}`)
```

Running entry.luau with Lute gives this:
```zsh
> lute entry.luau
@dep resolves to: /my/absolute/lute/path/static_require_test/dependency.luau
```

definitions/luau.luau
```diff
@@ -580,4 +580,8 @@ function luau.load(bytecode: Bytecode, chunkname: string, env: { [any]: any }?):
 	error("not implemented")
 end
 
+function luau.staticrequire(path: string, fromchunkname: string): string
```
Vighnesh-V: One thought is that `fromchunkname` is a very internal implementation detail. I would love a version of this that looks like:
```
luau.staticrequire(requirer: pathlike, | string, from: (pathlike | string)?) : pathlike
 --- if from is nil, default to the current file
 --- if this is non-nil, append `@` to it and perform the require
end
```

in addition to the regular `chunkname` version of the api.

Vighnesh-V: If I think about how an ordinary user might use this, I could see it being unintuitive to add the @:
```
luau.staticrequire("@dep", "./entry.luau")`
```

Vighnesh-V:
General comment about naming - it doesn't look like this actually performs the require - it really just returns a path right. Maybe: `resolverequirepath` might be a better name?

definitions/luau.luau
vrn-sn: I'd imagine that would be exposed as a helper in `@std/luau` but would still wrap this `@lute/luau` API.

aatxe:
I'd want to call this either `resolverequire` or, if we anticipate some more functions for require-as-library-functionality in the future, `luau.require.resolve`

vrn-sn:
I've renamed the API to `luau.resolverequire` and have also exposed a C++ version in `resolverequire.h` called `resolveRequire`:

```cpp
std::optional<std::string> resolveRequire(std::string requirePath, std::string requirerChunkname, std::string* error);
int resolverequire_luau(lua_State* L);
```

The Luau API (`resolverequire_luau`) now just calls into the C++ API.

@annieetang and @Vighnesh-V, I believe the C++ API should be useful for your work with extracting module return types and bundling.

aatxe (approved):

PR #498, opened 2025-10-30
Feature: lute test subcommand

Vighnesh-V:
This PR introduces a skeleton for the `lute test` subcommand. In future PR's I'll build out the functionality described in the help message, e.g. `--list, --filter, --reporter = ...`

### DIFF
--- a/tests/cli/test.test.luau
+++ b/tests/cli/test.test.luau
@@ -1,7 +1,8 @@
-local process = require("@lute/process")
+local path = require("@std/path")
+local process = require("@std/process")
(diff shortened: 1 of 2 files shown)


tests/cli/test.test.luau
```diff
@@ -0,0 +1,25 @@
+local process = require("@lute/process")
```
skberkeley:
```suggestion
local process = require("@std/process")
```
i think we want to move away from using `@lute` if we can?

skberkeley (approved):
thanks for adding tests!

PR #499, opened 2025-10-30
Feature: Fetch all .spec.luau files / .test.luau files

Vighnesh-V:
This PR updates the `lute test` subcommand to find all .spec.luau / .test.luau files in a given directory, defaulting to ./test


lute/cli/commands/test/init.luau
```diff
@@ -39,6 +41,12 @@ local function main(...: string)
 			return
 		end
 	end
+
+	local searchpath = #args > 0 and args or { "./tests" }
```
skberkeley:
```suggestion
	local searchpath = if #args > 0 then args else { "./tests" }
```

lute/cli/commands/test/init.luau
```diff
@@ -39,6 +41,12 @@ local function main(...: string)
...
+
+	local files = finder.findtestfiles(searchpath)
```
skberkeley, line 47: running tests coming later?

tests/cli/test.test.luau
```diff
@@ -21,6 +21,15 @@ test.suite("lute test", function(suite)
...
 		assert.eq(result.exitcode, 0)
 	end)
+
+	suite:case("lute test discovers test files in custom directory", function(assert)
+		-- Run lute test with custom directory path
+		local result = process.run({ lutePath, "test", "tests/cli/discovery" })
+
+		-- Check that it discovers test files (.test.luau and .spec.luau)
+		assert.eq(result.exitcode, 0)
+		assert.neq(result.stdout:find("Found 2 test files", 1, true), nil)
```
skberkeley, line 31: i would personally prefer that we put files which are the subject of a test in `system.tmpdir()`

lute/cli/commands/test/finder.luau
```diff
@@ -0,0 +1,54 @@
+local ps = require("@std/process")
+local fs = require("@std/fs")
+local path = require("@std/path")
+local string = require("@std/string")
+
+local function extend<T>(tbl: { T }, other: { T })
+	for _, entry in other do
+		table.insert(tbl, entry)
+	end
+end
```
skberkeley: thoughts on putting this in `@std/table`? maybe a question for `proj-lute`

lute/cli/commands/test/finder.luau
```diff
@@ -0,0 +1,54 @@
...
+
+local function findtestfiles(paths: { string }): { path.path }
+	local files = {}
+
+	for _, filepath in paths do
+		local pathObj = path.parse(filepath)
+
+		if not path.isabsolute(pathObj) then
+			pathObj = path.join(path.parse(ps.cwd()), pathObj)
```
skberkeley: put `ps.cwd()` in a local outside the for loop? also the `@std` version of `cwd` will already return a path object

lute/cli/commands/test/finder.luau
```diff
@@ -0,0 +1,54 @@
...
+		end
+
+		filepath = path.format(pathObj)
+
+		if fs.type(filepath) == "dir" then
```
skberkeley: `fs` should accept path objects

lute/cli/commands/test/finder.luau
```diff
@@ -0,0 +1,54 @@
...
+end
+
+local function istestfile(filename: string): boolean
+	return string.endswith(filename, ".test.luau") or string.endswith(filename, ".spec.luau")
+end
+
+local function findtestfilesrec(directory: path.path): { path.path }
+	local results = {}
+
+	for _, entry in fs.listdir(path.format(directory)) do
```
skberkeley:
```suggestion
	for _, entry in fs.listdir(directory) do
```

skberkeley (changes requested):
mostly path related nits

aatxe (approved):

lute/cli/commands/test/finder.luau
aatxe: yeah, that would be good. it'd be okay to refactor later too.

lute/cli/commands/test/init.luau
Vighnesh-V: Yup. I'll put that up separately since it's a slightly more involved change.

tests/cli/test.test.luau
Vighnesh-V: I see these more as part of a test fixture, so I think it's reasonable to have on disk in the repo. They aren't temporary files at all - the point is to check that `lute test` can discover files in this directory. Once I build this out a bit more, this will be pretty useful for writing tests that expect failures etc..

skberkeley (approved):

lute/cli/commands/test/finder.luau
Vighnesh-V: I opened a PR to do this here https://github.com/luau-lang/lute/pull/501 :)

PR #500, opened 2025-10-30
stdlib: add `__tostring` to paths

annieetang:
### Problem

Paths didn't have a default way of parsing between a `path` or `pathlike` and a string other than using `.format`

### Solution

We add `__tostring` to paths so users can now just `print(path)` or `tostring(path)` to get the string version of a path

part of https://github.com/luau-lang/lute/pull/491!

### DIFF
--- a/lute/std/libs/path/init.luau
+++ b/lute/std/libs/path/init.luau
@@ -1,11 +1,11 @@
-local pathlib = {}
-
 local platforminfo = require("@std/system/platforminfo")
 
 local posix = require("@self/posix")
 local win32 = require("@self/win32")
 
 local pathtypes = require("@self/types")
+
+local pathlib = {}
 
 type PathInterface = {
 	__tostring: () -> string,
@@ -82,10 +82,8 @@ end
 
 function pathlib.resolve(...: pathlike): path
 	if onWindows then
-		-- return win32.resolve(...)
 		return setmetatable(win32.resolve(...), pathlib)
 	else
-		-- return posix.resolve(...)
 		return setmetatable(posix.resolve(...), pathlib)
 	end
 end
--- a/lute/std/libs/path/init.luau
+++ b/lute/std/libs/path/init.luau
@@ -7,12 +7,8 @@ local pathtypes = require("@self/types")
 
 local pathlib = {}
 
-type PathInterface = {
-	__tostring: () -> string,
-}
-
-export type path = setmetatable<pathtypes.path, PathInterface>
-export type pathlike = setmetatable<pathtypes.pathlike, PathInterface>
+export type path = pathtypes.path
+export type pathlike = pathtypes.pathlike
 
 local onWindows = platforminfo.win32
 
@@ -90,13 +86,5 @@ end
 
 pathlib.win32 = win32
 pathlib.posix = posix
-
-function pathlib:__tostring(): string
-	if onWindows then
-		return win32.format(self :: path)
-	else
-		return posix.format(self :: path)
-	end
-end
 
 return table.freeze(pathlib)
--- a/lute/std/libs/path/init.luau
+++ b/lute/std/libs/path/init.luau
@@ -54,33 +54,33 @@ end
 
 function pathlib.join(...: pathlike): path
 	if onWindows then
-		return setmetatable(win32.join(...), pathlib)
+		return win32.join(...)
 	else
-		return setmetatable(posix.join(...), pathlib)
+		return posix.join(...)
 	end
 end
 
 function pathlib.normalize(path: pathlike): path
 	if onWindows then
-		return setmetatable(win32.normalize(path :: win32.pathlike), pathlib)
+		return win32.normalize(path :: win32.pathlike)
 	else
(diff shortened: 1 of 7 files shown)


lute/std/libs/path/posix/init.luau
```diff
@@ -1,8 +1,12 @@
...
 local posixtypes = require("@self/types")
 
-export type path = posixtypes.path
-export type pathlike = posixtypes.pathlike
+type PathInterface = {
+	__tostring: () -> string,
+}
+
+export type path = setmetatable<posixtypes.path, PathInterface>
+export type pathlike = setmetatable<posixtypes.pathlike, PathInterface>
```
skberkeley: should this go in the `types` file?

lute/std/libs/path/posix/init.luau
```diff
@@ -121,7 +125,7 @@ function posix.join(...: pathlike): path
 		joinHelper(path, parts[i])
 	end
 
-	return path
+	return setmetatable(path, posix)
```
skberkeley: i think this `setmetatable` call should go above, where we construct the path literal

lute/std/libs/path/posix/init.luau
```diff
@@ -151,10 +155,10 @@ function posix.normalize(path: pathlike): path
 		end
 	end
 
-	return {
+	return setmetatable({
 		parts = newParts,
 		absolute = path.absolute,
-	}
+	}, posix)
```
skberkeley: should the metatable be a separate table, not `posix` itself? it feels like there's a conflation happening between the library of functions and the path object itself

lute/std/libs/path/win32/init.luau
```diff
@@ -1,9 +1,13 @@
...
 local win32types = require("@self/types")
 
-export type pathkind = win32types.pathkind
-export type path = win32types.path
-export type pathlike = win32types.pathlike
+type PathInterface = {
+	__tostring: () -> string,
+}
+
+export type pathkind = setmetatable<win32types.pathkind, PathInterface>
```
skberkeley: why does pathkind need a metatable?

lute/std/libs/path/init.luau
```diff
@@ -7,8 +7,12 @@ local pathtypes = require("@self/types")
...
 local pathlib = {}
 
-export type path = pathtypes.path
-export type pathlike = pathtypes.pathlike
+type PathInterface = {
+	__tostring: () -> string,
+}
+
+export type path = setmetatable<pathtypes.path, PathInterface>
+export type pathlike = setmetatable<pathtypes.pathlike, PathInterface>
```
skberkeley: does path not already have a metatable with `__tostring` being a union of `posix.path` and `win32.path`?

tests/std/path/path.posix.test.luau
```diff
@@ -535,4 +535,18 @@ test.suite("PathPosixResolveSuite", function(suite)
 	end)
 end)
 
+test.suite("PathPosixToStringSuite", function(suite)
+	suite:case("toString_posix_absolute_path", function(assert)
+		local pathobj: posix.path = posix.parse("/home/user/documents/file.txt")
```
skberkeley: nit: store path literals in a variable

tests/std/path/path.win32.test.luau
```diff
@@ -685,4 +685,18 @@ test.suite("PathWin32ResolveSuite", function(suite)
 	end)
 end)
 
+test.suite("PathWin32ToStringSuite", function(suite)
+	suite:case("toString_windows_absolute_path", function(assert)
+		local pathobj: win32.path = win32.parse("C:\\Users\\username\\Documents\\file.txt")
```
skberkeley: nit: store path literals in a variable

lute/std/libs/path/init.luau
annieetang: oops true thanks

lute/std/libs/path/posix/init.luau
annieetang: do you mean exporting `PathInterface` in `types.luau` and calling it as `posixtypes.PathInterface`?

lute/std/libs/path/init.luau
```diff
@@ -54,33 +54,33 @@ end
 
 function pathlib.join(...: pathlike): path
 	if onWindows then
-		return win32.join(...)
+		return setmetatable(win32.join(...), pathlib)
```
aatxe: all of these are setting the metatable of the path to contain _way_ more than the interface of `__tostring` provides. why are we doing this?

aatxe: it also looks like the changes to the platform specific path types already provides the necessary metatable, so we can just do nothing here in the top-level `@std/path`

aatxe (changes requested):

lute/std/libs/path/init.luau
annieetang: oh wait I totally forgot to remove these after sora's change, thanks for the catch!

aatxe (approved):

PR #501, opened 2025-10-30
stdlib: add table.extend to the lute standard library

Vighnesh-V:
This adds a method to the standard library that takes an arraylike table and extends it by another arraylike table.


annieetang:
could you also update the files.luau file to use extend from std?

Vighnesh-V:
@annieetang Good catch, I've just done that.

annieetang (approved):
🚢

PR #503, opened 2025-10-30
stdlib: expose test environment from test library

Vighnesh-V:
This exposes a method in std/test to grab the registered test cases out of the internal environment.

### DIFF
--- a/lute/std/libs/test.luau
+++ b/lute/std/libs/test.luau
@@ -128,7 +128,7 @@ function test.suite(name: string, registerFn: (TestSuite) -> ())
 end
 
 -- get all registered tests without running them (internal)
-function test._getregistered()
+function test._registered()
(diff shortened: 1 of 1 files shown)


lute/std/libs/test.luau
```diff
@@ -127,6 +127,11 @@ function test.suite(name: string, registerFn: (TestSuite) -> ())
 	table.insert(env.suites, suite)
 end
 
+-- get all registered tests without running them (internal)
+function test._getregistered()
```
```suggestion
function test._registered()
```

Does the `get` add anything here? feels very Java, Luau APIs are just the thing, e.g. `string.byte` vs `String.getByte`

Vighnesh-V: Nah, it really doesn't. I'll clip it.

aatxe (approved):

PR #505, opened 2025-10-30
feature: list all discovered tests

Vighnesh-V:
This PR adds support for discovering and eventually running test cases. Using the loadbypath api, we can load all the discovered test files and eventually run them (future PR). This PR just implements a simple list all test cases command, much like `doctest`.

In order to populate the test environment, I've overridden `require` with one that intercepts calls to "@std/test" and returns the version of std/test require'd by the `test` subcommand itself. This ensures that loading modules registers tests in a way that makes them visible by the subcommand.

### DIFF
--- a/tests/cli/test.test.luau
+++ b/tests/cli/test.test.luau
@@ -3,9 +3,18 @@ local path = require("@std/path")
 local process = require("@std/process")
 local system = require("@std/system")
 local test = require("@std/test")
-
+local fs = require("@std/fs")
 local lutePath = path.format(process.execpath())
 local tmpDir = system.tmpdir()
+
+local expected = [[Anonymous:
+	Passing
+ExampleSuite:
+	should pass
+SmokeSuite:
+	make_fail
+	make_pass
+]]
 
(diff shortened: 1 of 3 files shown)


tests/cli/test.test.luau
```diff
@@ -27,11 +36,11 @@ test.suite("lute test", function(suite)
 
 	suite:case("lute test discovers test files in custom directory", function(assert)
 		-- Run lute test with custom directory path
-		local result = process.run({ lutePath, "test", "tests/cli/discovery" })
+		local result = process.run({ lutePath, "test", "--list", "tests/cli/discovery" })
 
 		-- Check that it discovers test files (.test.luau and .spec.luau)
 		assert.eq(result.exitcode, 0)
-		assert.neq(result.stdout:find("Found 2 test files", 1, true), nil)
+		assert.eq(expected, result.stdout)
```
Vighnesh-V, line 43: Alas, another place where process redirection / snapshot testing would be amazing :)

vrn-sn (approved):

PR #506, opened 2025-10-30
stdlib: Reworking AST queries

skberkeley:
Initial reworking of AST queries to try to make authoring codemods a bit more ergonomic. APIs are still in flux, so open to suggestions!

Follow up work still needed:
- Reworking `query:replace` so programmers don't need to pass replacement maps around


examples/query.luau
```diff
@@ -0,0 +1,25 @@
...
+	:filter(function(call)
+		local calledFunc = call.func
+		return calledFunc.tag == "global" and calledFunc.name.text == "require"
+	end)
+	:filter(function(call)
+		local firstArg = call.arguments[1].node
+		return firstArg.tag == "indexname" and firstArg.index.text == "OldComponent"
+	end)
+
+-- Can we improve the type solver here so we don't need to add this annotation?
```
hgoldstein, line 18: Which annotation are you talking about?

lute/std/libs/syntax/query.luau
```diff
@@ -3,28 +3,163 @@
...
+	| { tag: "prepend", content: string }
+	| { tag: "replace", content: string }
+	| { tag: "remove" }
 
-	finder.visitExpression = function(node)
-		local result = predicate(node)
-		if result then
-			table.insert(nodes, result)
+type replacement = string | node | { delta }
+type replacements = { [node]: replacement }
```
hgoldstein: Should types here be capitalized?

hgoldstein (commented):
Would you be able to add an end-to-end test? It feels weird to see all of this infrastructure without being able to _actually_ run a codemod.

lute/std/libs/syntax/query.luau
skberkeley: I think the latest in lute naming conventions is to luacase types too (can you confirm @aatxe ?)

examples/query.luau
skberkeley: On the line right after:
Image

definitions/luau.luau
```diff
@@ -211,7 +212,7 @@ export type AstExprIfElse = {
 	elseexpr: AstExpr,
 }
 
-export type AstExpr = { kind: "expr" } & (
+export type AstExpr = { kind: "expr"? } & (
```
hgoldstein: Why is this optional?

tests/syntax/query.test.luau
```diff
@@ -0,0 +1,166 @@
...
+
+test.suite("AST Query", function(suite)
+	suite:case("filter", function(assert)
+		local testQuery = mockQuery({
+			createNumberExpr(-2),
+			createNumberExpr(-1),
+			createNumberExpr(0),
+			createNumberExpr(1),
+			createNumberExpr(2),
+		})
```
hgoldstein: Why not just parse this as normal Luau code?

lute/std/libs/syntax/query.luau
```diff
@@ -3,28 +3,163 @@
...
+		visitTypeIntersection = visit,
+		visitTypeArray = visit,
+		visitTypeTable = visit,
+		visitTypeFunction = visit,
+
+		visitTypePackExplicit = visit,
+		visitTypePackGeneric = visit,
+		visitTypePackVariadic = visit,
+
+		visitToken = function(n)
```
wmccrthy, line 131: I think we should use `visit` here too. I've found it's often useful to query specifically for tokens, since it's a straightforward way to search for and replace exact text

lute/std/libs/syntax/utils.luau
```diff
@@ -0,0 +1,229 @@
...
+end
+
+function utils.isTypePackGeneric(n: luau.AstNode): luau.AstTypePackGeneric?
+	return if n.kind == "typepack" and n.tag == "generic" then n else nil
+end
+
+function utils.isTypePack(n: luau.AstNode): luau.AstTypePack?
+	return if n.kind == "typepack" then n else nil
+end
+
```
wmccrthy, line 228: same vein as my other comment, but maybe worth adding `isToken`
```suggestion

function utils.isToken(n: luau.AstNode): luau.Token?
	return if n.istoken then n else nil
end

```

lute/std/libs/syntax/query.luau
```diff
@@ -3,28 +3,163 @@
...
+
+export type query<T = node> = {
+	nodes: { T },
+	filter: (self: query<T>, pred: (T) -> boolean) -> query<T>,
+	replace: (self: query<T>, repl: (T) -> replacement?) -> replacements,
+	map: <U>(self: query<T>, fn: (T) -> U?) -> query<U>,
+}
+
+local queryLib = {}
+
```
wmccrthy, line 22: spoken abt offline, but `forEach` would be nice
```suggestion

function queryLib.forEach<T>(self: query<T>, callback: (T) -> ())
	for _, node in self.nodes do
		callback(node)
	end
end

```

definitions/luau.luau
skberkeley: This not being optional means checks like `AstExprConstantNil <: AstExpr` doesn't subtype, but I think the better thing to do might be to add a `kind` field to each variant too

lute/std/libs/syntax/query.luau
Vighnesh-V: exported types should be luacase as far as I can tell, but this is alright if it's not being exported.

definitions/luau.luau
hgoldstein: I think so, this also means that:
```
if e.kind then
end
```
... would exclude those fields, which seems incorrect.

lute/std/libs/syntax/query.luau
skberkeley: I originally did use `visit` here, but the trouble is that we have `AstExprs` like `AstExprConstantNumber` which are technically tokens too, which means that the selection function gets called twice on them, so we end up overcounting in a sense. What do you think about building out a `selectToken` in the future which only selects tokens?

lute/std/libs/syntax/query.luau
skberkeley: it'll be exported later, so i'll keep it as is!

lute/std/libs/syntax/query.luau
wmccrthy: Ah I gotcha, that makes sense. I think 'selectToken' would be a very nice addition then.

skberkeley:
Added AST replacement reserialization! I've added unit tests for reserialization stuff, as well as a basic sanity test for a codemod using the new APIs and @wmccrthy has volunteered to write a more complete suite of E2E tests

tests/std/syntax/printer.test.luau
```diff
@@ -0,0 +1,236 @@
...
+        ]]
+
+		local ast = parser.parse(source)
+		local replacements = query.select(ast, syntaxUtils.isExprConstantString):replace(function(s)
+			return if s.text == "World" then "1 + 2" else nil
+		end)
+
+		local result = printer.printstatement(ast, replacements)
+
+		local expected = "            local name = 1 + 2" -- TODO: preserve trivia so trailing newline is preserved
```
hgoldstein, line 21: That's not a `TODO`, that's an important part of the contract for dealing with CSTs. Also what happened to the newline at the beginning of the string?

tests/syntax/query.test.luau
hgoldstein: That's not what I meant. Why isn't this:
```luau
local testFile = [[
    local numbers = {0, 1, 2, 3, 4}
]]
```

Ditto for the rest of this file.

definitions/luau.luau
```diff
@@ -104,13 +99,15 @@ export type AstExprIndexName = {
...
 export type AstExprIndexExpr = {
+	kind: "expr",
 	tag: "index",
 	expression: AstExpr,
 	openbrackets: Token<"[">,
 	index: AstExpr,
 	closebrackets: Token<"]">,
 }
 
+-- I don't like that we have this type; a lot of its data should live in the parents, which should then have `AstStatBlock` as its body
```
hgoldstein, line 110: If this is how the AST is structured, then we match it, especially for a CST.

examples/query_transformer.luau
```diff
@@ -0,0 +1,24 @@
...
+	local replacements = q:filter(function(bin)
+		return bin.operator.text == "~="
+			and bin.lhsoperand.tag == "local"
+			and bin.rhsoperand.tag == "local"
+			and bin.lhsoperand.token.text == bin.rhsoperand.token.text
+	end):replace(function(bin)
+		return `math.isnan({(bin.lhsoperand :: luau.AstExprLocal).token.text})`
+	end)
+
+	return printer.printfile(parsedResult, replacements)
```
hgoldstein: Why is this not one long expression like:
```
query.select(parsedResult.root, syntax_utils.isExprBinary)
    :filter(function(bin)
        return bin.operator.text == "~="
            and bin.lhsoperand.tag == "local"
            and bin.rhsoperand.tag == "local"
            and bin.lhsoperand.token.text == bin.rhsoperand.token.text
    end)
    :replace(function(bin)
        return `math.isnan({(bin.lhsoperand :: luau.AstExprLocal).token.text})`
    end)
    :print()
```

lute/std/libs/syntax/printer.luau
```diff
@@ -2,157 +2,292 @@
...
+			return visitor.visittype(replacement, self) -- LUAUFIX: refine replacement to be AstType
+		elseif replacement.kind == "typepack" then
+			return visitor.visittypepack(replacement, self) -- LUAUFIX: refine replacement to be AstTypePack
+		elseif replacement.kind == "local" then
+			error("TODO: Add visit function to visit library")
+		elseif replacement.kind == "attribute" then
+			error("TODO: Add visit function to visit library")
+		else
+			error(`Unknown node kind in replacement: {replacement.kind}`)
+		end
```
hgoldstein: This feels bizarre: why do we need to match on individual node kinds here? Why do we have random holes?

lute/std/libs/syntax/printer.luau
```diff
@@ -2,157 +2,292 @@
...
+		for _, delta in replacement do
+			if delta.tag == "replace" then
+				self:printReplacement(node, delta.content)
+			end
+		end
+		for _, delta in replacement do
+			if delta.tag == "append" then
+				self:printReplacement(node, delta.content)
+			end
+		end
```
hgoldstein: Can we do something more constrained here? It feels weird if we have a single node and there are _multiple_ replacements. How about the rule for now is that you get one `delta`?

lute/std/libs/syntax/query.luau
```diff
@@ -3,28 +3,173 @@
...
+	elseif nodeKind == "expr" then
+		visitor.visitexpression(ast :: luau.AstExpr, selectVisitor) -- LUAUFIX: code too complex to typecheck
+	elseif nodeKind == "type" then
+		visitor.visittype(ast :: luau.AstType, selectVisitor)
+	elseif nodeKind == "typepack" then
+		visitor.visittypepack(ast :: luau.AstTypePack, selectVisitor)
+	elseif nodeKind == "token" then
+		visitor.visittoken(ast :: luau.Token, selectVisitor)
+	else
+		error("Unsupported node kind: " .. tostring(nodeKind))
```
hgoldstein: Could this be exhaustive?

lute/std/libs/syntax/types.luau
```diff
@@ -0,0 +1,15 @@
+local luau = require("@std/luau")
+
+type node = luau.AstNode
+
+export type removal = { tag: "remove" }
+
+export type delta =
+	| { tag: "append", content: string | node }
+	| { tag: "prepend", content: string | node }
+	| { tag: "replace", content: string | node }
```
hgoldstein: The content for deltas should _always_ be `string`.

hgoldstein (changes requested):
See comments.

examples/query.luau
```diff
@@ -0,0 +1,25 @@
...
+local p = query
+	.select(ast, utils.isExprCall)
+	:filter(function(call)
+		local calledFunc = call.func
+		return calledFunc.tag == "global" and calledFunc.name.text == "require"
+	end)
+	:filter(function(call)
+		local firstArg = call.arguments[1].node
+		return firstArg.tag == "indexname" and firstArg.index.text == "OldComponent"
+	end)
```
Vighnesh-V, line 16: this is pretty sick.

Vighnesh-V (approved):
This is dope - it looks good to me

tests/syntax/query.test.luau
skberkeley: oh i see what you mean
i wanted to be able to unit test the other functions in the query library without using `select`. i could parse `local numbers = {0, 1, 2, 3, 4}` as a whole and wrangle the `AstExprConstantNumbers` out of the AST, but this seems more concise for the same effect?

tests/std/syntax/printer.test.luau
skberkeley: I agree, i'm definitely not trying to claim that this PR makes ast querying/reserialization feature complete, but in the interest of not letting this PR grow indefinitely, I was planning on handling trivia reserialization in a follow up

> Also what happened to the newline at the beginning of the string?

There never was a newline:

```luau
local source = [[
print(`Hello {name}`)
]]

print(string.byte(source, 1))
```
```
% lute scratch.luau 
112
```

definitions/luau.luau
skberkeley: It's not how the AST or CST are structured in the C++ implementation, do you think it's worth refactoring sooner rather than later?

examples/query_transformer.luau
skberkeley: oops yeah, i can make that change! i think i was trying to debug autocomplete at the time

lute/std/libs/syntax/printer.luau
skberkeley: You're right, I'll add a general `visit` function to the visitor library, it shouldn't be on consumers to check the node kind

lute/std/libs/syntax/printer.luau
skberkeley: Sure, I suppose codemod authors can always just specify longer strings in their deltas, so the expressivity isn't reduced

lute/std/libs/syntax/query.luau
skberkeley: Yuh, it should be once I add a general visit function to the visitor library

lute/std/libs/syntax/types.luau
skberkeley: is there no world where codemod writers might benefit from being able to construct an AST node object from scratch and using it as a delta content?

definitions/luau.luau
hgoldstein: Why is it like this then? We should probably refactor it _soon_ but out of scope for this PR.

lute/std/libs/syntax/types.luau
hgoldstein: _Potentially_, but deltas are meant to be lower level components.

examples/query_transformer.luau
hgoldstein: Did you update this file?

lute/std/libs/syntax/printer.luau
```diff
@@ -2,157 +2,262 @@
...
 
-type PrintVisitor = visitor.Visitor & {
-	result: buffer,
-	cursor: number,
-}
+local function printReplacement(self: PrintVisitor, node: luau.AstNode, replacement: types.replacement)
+	if typeof(replacement) == "string" then
+		self:write(replacement)
+	elseif replacement.kind ~= nil then -- LUAUFIX: type solver should know replacement might have a kind field
+		-- TODO: preserve trivia
```
hgoldstein, line 90: Is this still `TODO`?

tests/syntax/query.test.luau
hgoldstein: > i wanted to be able to unit test the other functions in the query library without using `select`

That doesn't seem like a meaningful test to me: it's just visual noise.

definitions/luau.luau
skberkeley: I hate to be that guy, but these type definitions were written before I started working on this stuff

lute/std/libs/syntax/query.luau
```diff
@@ -3,28 +3,168 @@
...
+	}
+end
+
+function queryLib.forEach<T>(self: query<T>, callback: (T) -> ()): ()
+	for _, node in self.nodes do
+		callback(node)
+	end
+end
+
+function queryLib.select<T>(ast: node, fn: (node) -> (T | { T })?): query<T>
```
hgoldstein: Can we not have select flat-map? Select should just be an entrypoint, we can have map and flat-map as separate functions.

examples/query_transformer.luau
```diff
@@ -0,0 +1,23 @@
+local luau = require("@std/luau")
+local parser = require("@std/syntax/parser")
+local query = require("@std/syntax/query")
+local syntax_utils = require("@std/syntax/utils")
+
+local function transform(ctx)
+	local parsedResult = parser.parsefile(ctx.source)
+
+	return query
+		.select(parsedResult.root, syntax_utils.isExprBinary)
```
hgoldstein: Can this be named something different for now? We talked about this offline, that `select` really ought to be a method on `query`, but the new solver doesn't quite do the right thing here yet. Maybe name this `query.selectFromRoot`

lute/std/libs/syntax/query.luau
```diff
@@ -3,28 +3,168 @@
...
+			else
+				table.insert(nodes, selected)
+			end
+		end
 		return true
 	end
 
-	visitor.visitblock(block, finder)
+	local selectVisitor: visitor.Visitor = {
+		visitBlock = visit,
```
hgoldstein, line 80: Maybe we're missing a building block here: kind of annoying that you have to make this literal with a ton of methods that almost all do the same thing?

tests/std/syntax/query.test.luau
```diff
@@ -0,0 +1,148 @@
...
+			sum = sum + n.value
+		end)
+
+		assert.eq(sum, 3) -- 0 + 1 + 2 = 3
+	end)
+
+	suite:case("select_nums", function(assert)
+		local ast = luau.parse("print(1 + 2)")
+		local queryResult: query.query<luau.AstExprConstantNumber> = query.select(
+			ast.root :: luau.AstNode,
```
hgoldstein: why do we need this type cast?

tests/std/syntax/query.test.luau
```diff
@@ -0,0 +1,148 @@
...
+				end
+				return nil
+			end
+		)
+
+		assert.eq(#queryResult.nodes, 2)
+		local num: luau.AstExpr = queryResult.nodes[1]
+		assert.eq(num.kind, "expr")
+		assert.eq(num.tag, "number")
+		assert.eq((num :: luau.AstExprConstantNumber).value, 1)
```
hgoldstein: What about this cast?

hgoldstein (changes requested):

lute/std/libs/syntax/query.luau
skberkeley: Done! Also got rid of it from `map`

examples/query_transformer.luau
skberkeley: done!

lute/std/libs/syntax/query.luau
skberkeley: yeah i'll add a `visitor.createVisitor(defaultVisit: (AstNode -> boolean)?)` in a followup

tests/std/syntax/query.test.luau
skberkeley: Solver complains that AstStatBlock doesn't have location? field, so it doesn't subtype against AstNode

tests/std/syntax/query.test.luau
skberkeley: the `assert.eq` calls above don't refine `num` to `AstExprConstantNumber`

lute/std/libs/syntax/query.luau
```diff
@@ -3,28 +3,159 @@
...
+		end
+	end
+	return {
+		nodes = newNodes,
+		filter = self.filter,
+		replace = self.replace,
+		map = self.map,
+		forEach = self.forEach,
+		__root = self.__root,
+	}
```
hgoldstein: This is to get around the recursive type restriction yeah?

lute/std/libs/syntax/query.luau
```diff
@@ -3,28 +3,159 @@
...
---- Selects a list of expressions from the provided block that match the provided predicate.
---- Useful for defining declarative-based programs, where the matched nodes are mutated for the migration
-function query.selectif<T>(block: luau.AstStatBlock, predicate: (luau.AstExpr) -> T?): { T }
-	local nodes = {}
-	local finder = visitor.create()
+export type query<T = node> = {
+	nodes: { T },
+	filter: (self: query<T>, pred: (T) -> boolean) -> query<T>,
+	replace: (self: query<T>, repl: (T) -> types.replacement?) -> types.replacements,
+	map: <U>(self: query<T>, fn: (T) -> (U | { U })?) -> query<U>, -- recursion violation reported
```
hgoldstein: This signature is wrong now I think

hgoldstein (changes requested):
Just the signature being out of date

lute/std/libs/syntax/query.luau
skberkeley: Actually I think this can just be in-place, since the recursive type restriction doesn't impact the runtime

lute/std/libs/syntax/query.luau
skberkeley: nice catch

lute/std/libs/syntax/query.luau
```diff
@@ -3,28 +3,152 @@
...
-	local nodes = {}
-	local finder = visitor.create()
+export type query<T = node> = {
+	nodes: { T },
+	filter: (self: query<T>, pred: (T) -> boolean) -> query<T>,
+	replace: (self: query<T>, repl: (T) -> types.replacement?) -> types.replacements,
+	map: <U>(self: query<T>, fn: (T) -> U?) -> query<U>, -- recursion violation reported
+	forEach: (self: query<T>, callback: (T) -> ()) -> (),
+	__root: node,
+} -- TODO: Add a select: (self: query<T>, fn: (node) -> U?) -> query<U> method and a flatMap: <U>(self: query<T>, fn: (T) -> { U }?) -> query<U> method
```
hgoldstein: nit: `flatMap` is `(T) -> { U }`: the `nil` is redundant given that `nil` is the same as `{}`

hgoldstein (approved):

lute/std/libs/syntax/query.luau
skberkeley: oh tru

examples/query_transformer.luau
```diff
@@ -0,0 +1,20 @@
...
+		:filter(function(bin)
+			return bin.operator.text == "~="
+				and bin.lhsoperand.tag == "local"
+				and bin.rhsoperand.tag == "local"
+				and bin.lhsoperand.token.text == bin.rhsoperand.token.text
+		end)
+		:replace(function(bin)
+			return `math.isnan({(bin.lhsoperand :: luau.AstExprLocal).token.text})`
+		end)
+		:print()
```
wmccrthy, line 17: does this work? doesn't look like there is a print method on the `query` class

skberkeley: yup, it does! the `replacements` type returned by `replace()` has a `print` method

wmccrthy: oh nice, missed that 🔥

PR #509, opened 2025-10-31
format all .luau files

Vighnesh-V:
Applies stylua formatting.

### DIFF
--- a/definitions/luau.luau
+++ b/definitions/luau.luau
@@ -134,24 +134,21 @@ export type AstExprAnonymousFunction = {
 	body: AstFunctionBody,
 }
 
-export type AstExprTableItem =
-	| { kind: "list", value: AstExpr, separator: Token<"," | ";">? }
-	| {
-		kind: "record",
-		key: Token<string>,
-		equals: Token<"=">,
-		value: AstExpr,
-		separator: Token<"," | ";">?,
-	}
-	| {
-		kind: "general",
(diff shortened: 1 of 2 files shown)


hgoldstein (approved):

definitions/luau.luau
```diff
@@ -134,24 +134,21 @@ export type AstExprAnonymousFunction = {
...
+	separator: Token<"," | ";">?,
+} | {
+	kind: "general",
+	indexeropen: Token<"[">,
+	key: AstExpr,
+	indexerclose: Token<"]">,
+	equals: Token<"=">,
+	value: AstExpr,
+	separator: Token<"," | ";">?,
+}
```
skberkeley, line 151: what version of stylua do you have? i can't reproduce this with the vscode extension

vrn-sn: Yeah seems like this doesn't agree with CI either. Shouldn't everything already be StyLua'd?

Vighnesh-V: Yeah this is extremely weird. I just ran the stylua extension in the root of my repo.

Vighnesh-V: Lemme bump my stylua version and see.

Vighnesh-V:
yeah, I bumped my stylua version to 2.3.0 and this went away. I was using one distributed by cargo. Closing as the repo is already formatted.

PR #510, opened 2025-10-31
stdlib: add kinds to ast types

skberkeley:
This PR adds kinds to AST types, which lets us do nice things with refinements like:
```luau
if node.kind == "expr" then
    if node.tag == "conditional" then
        node. -- we get autocomplete for things like 'condition' here
    end
end
```

This also required updating the serialization implementation of the parser.

### DIFF
--- a/definitions/luau.luau
+++ b/definitions/luau.luau
@@ -36,6 +36,7 @@ export type Token<Kind = string> = {
 	position: Position,
 	text: Kind,
 	trailingtrivia: { Trivia },
+	istoken: true,
 }
 
 export type Eof = Token<""> & { tag: "eof" }
@@ -44,6 +45,7 @@ export type Pair<T, Separator> = { node: T, separator: Token<Separator>? }
 export type Punctuated<T, Separator = ","> = { Pair<T, Separator> }
 
 export type AstLocal = {
+	kind: "local",
 	name: Token<string>,
 	colon: Token<":">?,
 	annotation: AstType?,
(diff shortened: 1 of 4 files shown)


definitions/luau.luau
```diff
@@ -186,6 +192,7 @@ export type AstExprTypeAssertion = {
 	annotation: AstType,
 }
 
+-- helper type for elseif clauses of an if-else expression, not actually an AstExpr itself
 export type AstExprIfElseIfs = {
```
aatxe: if this isn't an `AstExpr`, it should probably not be called that.

```suggestion
export type AstElseIf = {
```

definitions/luau.luau
```diff
@@ -369,8 +377,8 @@ export type AstStatTypeFunction = {
 	body: AstFunctionBody,
 }
 
-export type AstStat =
-	| AstStatBlock
+export type AstStat = { kind: "stat" } & (
+	AstStatBlock
```
aatxe:
```suggestion
	| AstStatBlock
```
leading union

definitions/luau.luau
```diff
@@ -204,8 +211,8 @@ export type AstExprIfElse = {
 	elseexpr: AstExpr,
 }
 
-export type AstExpr =
-	| AstExprGroup
+export type AstExpr = { kind: "expr" } & (
+	AstExprGroup
```
aatxe:
```suggestion
	| AstExprGroup
```

leading union

aatxe: i feel like stylua will hate this though even though it's the right way to write this kind of stuff

definitions/luau.luau
skberkeley: Does `AstElseIfExpr` work? there's an `AstStat` equivalent, so I feel like there should still be something to distinguish the two

definitions/luau.luau
skberkeley: huh, stylua doesn't actually flag it

aatxe (approved):

definitions/luau.luau
aatxe: yeah, that's good if the two versions are different.

PR #515, opened 2025-11-03
feature: lute compile supports multiple files and static require tracing.

Vighnesh-V:
This PR extends the `lute compile` command, which takes a file and:
- Figures out all the transitive requires for that file
- Creates a bundle of precompiled bytecode
- Encodes the bundle into a payload
- Produces a runnable executable with the `lute` runtime embedded, along with the bundle appended to the end of the binary

This PR explicitly doesn't handle making this work with `.luaurc` or `config.luau` files - this will come in a subsequent PR.


lute/cli/src/climain.cpp
```diff
@@ -25,6 +27,7 @@
 #include "lute/runtime.h"
 #include "lute/system.h"
 #include "lute/task.h"
+#include "lute/staticrequires.h"
```
vrn-sn: alphabetical include

lute/cli/src/climain.cpp
```diff
@@ -90,6 +93,31 @@ static void luteopen_libs(lua_State* L)
...
 }
 
+void* createBundleRequireContext(lua_State* L, Luau::DenseHashMap<std::string, std::string> bundleMap)
+{
+    void* ctx = lua_newuserdatadtor(
+        L,
+        sizeof(RequireCtx),
+        [](void* ptr)
+        {
+            static_cast<RequireCtx*>(ptr)->~RequireCtx();
```
vrn-sn:
```suggestion
            std::destroy_at(static_cast<RequireCtx*>(ptr));
```

lute/cli/src/climain.cpp
```diff
@@ -108,6 +136,27 @@ lua_State* setupCliState(Runtime& runtime, std::function<void(lua_State*)> preSa
...
+{
+    return setupState(
+        runtime,
+        [bundleMap = std::move(bundleMap), preSandboxInit = std::move(preSandboxInit)](lua_State* L)
+        {
+            luteopen_libs(L);
+            if (Luau::CodeGen::isSupported())
+                Luau::CodeGen::create(L);
+
+            luaopen_require(L, requireConfigInit, createBundleRequireContext(L, bundleMap));
```
vrn-sn:
```suggestion
            luaopen_require(L, requireConfigInit, createBundleRequireContext(L, std:move(bundleMap)));
```

lute/cli/src/climain.cpp
```diff
@@ -108,6 +136,27 @@ lua_State* setupCliState(Runtime& runtime, std::function<void(lua_State*)> preSa
     );
 }
 
+lua_State* setupBundleState(
+    Runtime& runtime,
+    Luau::DenseHashMap<std::string, std::string> bundleMap,
+    std::function<void(lua_State*)> preSandboxInit = nullptr
```
vrn-sn: If the bundler doesn't need this, can we just remove it from this function?

vrn-sn (commented):
Is it possible to use the new `resolveRequire` API internally now (or maybe coming in a follow-up)?

lute/cli/src/climain.cpp
Vighnesh-V: removed!

Vighnesh-V:
> Is it possible to use the new `resolveRequire` API internally now (or maybe coming in a follow-up)?

I've gone ahead and implemented the version of static require that uses the resolveRequire api you added, but I opted to stick it in a follow up PR since there was a bunch of additional cleanup work I needed to do.

lute/cli/src/staticrequires.cpp
```diff
@@ -80,7 +78,12 @@ std::vector<std::string> StaticRequireTracer::trace(const std::string& rootDirec
             }
             else
             {
-                fprintf(stderr, "Warning: Could not resolve require('%s') from '%s'\n", req.c_str(), filePath.c_str());
+                // Skip warning for built-in libraries (@std and @lute)
+                bool isBuiltinLibrary = req.find("@std/") == 0 || req.find("@lute/") == 0;
```
vrn-sn: `.rfind("@...", 0)` to avoid scanning whole string

vrn-sn (approved):
Outside of this PR, I think we need a ticket for abstracting away printing to `stdout` in `Lute.CLI.lib`. I'd like if we could write unit tests for these places where we display stuff to the user by just capturing what would have gone to `stdout`/`stderr` and checking it.

lute/cli/include/lute/staticrequires.h
```diff
@@ -9,7 +9,7 @@
 class StaticRequireTracer
 {
 public:
-    StaticRequireTracer();
+    StaticRequireTracer() = default;
```
vrn-sn, line 12: Also, we don't have any other constructors, so won't the default one be generated anyway?

PR #516, opened 2025-11-03
[old] refactor: update `stdlib` module files to follow format in lute `definitions/`

annieetang:
### Problem

The file structure before in`@std` is a little inconsistent when we're writing the modules so it's a bit hard to know how to structure a new addition in our std lib. It also means we can't extend our current doc generator for the lute definitions to the stdlib because the [regex matching](https://github.com/luau-lang/lute/blob/primary/docs/scripts/reference.luau#L17) doesn't detect the different structures in these files. 

### Solution

We update the std lib files to follow the format of the lute `definitions/` library ([example](https://github.com/luau-lang/lute/blob/primary/definitions/task.luau)) to standardize how we write modules and help with adding the std lib to our docs site :) 

The file structure (in `lute/std/libs/...`) is now hopefully something like:

file name: `modulename.luau`
```
--!strict
-- @std/modulename (path to use the module)
-- description of the purpose of the module

local modulename = {}

-- types & export types -- 
type nonExportedType = ...
export type exportedType = ...

-- local functions and/or user-facing functions (only user-facing functions should be added to the returned table) --
local function nonuserfacingfunction(): returntype
    ...
end

function modulename.functiontoreturn(): returntype
    ...
end

-- re-exports if extended from another library and other non-functions --

return table.freeze(modulename)
```

### DIFF
--- a/examples/task-wait.luau
+++ b/examples/task-wait.luau
@@ -1,8 +1,8 @@
 local task = require("@lute/task")
-local time = require("@lute/time")
+local time = require("@std/time")
 
 print(task.wait(1))
-print(task.wait(time.duration.seconds(1)))
+print(task.wait(time.seconds(1)))
 print(task.wait())
 
 -- while true do
--- a/lute/std/libs/syntax/query.luau
+++ b/lute/std/libs/syntax/query.luau
@@ -1,10 +1,15 @@
-local luau = require("@lute/luau")
+--!strict
+-- @std/syntax/query
+-- Provides utility functions for querying Luau AST nodes
 
+local luau = require("@lute/luau")
 local visitor = require("./visitor")
+
+local query = {}
 
 --- Selects a list of expressions from the provided block that match the provided predicate.
 --- Useful for defining declarative-based programs, where the matched nodes are mutated for the migration
-local function selectif<T>(block: luau.AstStatBlock, predicate: (luau.AstExpr) -> T?): { T }
+function query.selectif<T>(block: luau.AstStatBlock, predicate: (luau.AstExpr) -> T?): { T }
 	local nodes = {}
 	local finder = visitor.createVisitor()
 
@@ -22,6 +27,4 @@ local function selectif<T>(block: luau.AstStatBlock, predicate: (luau.AstExpr) -
 	return nodes
 end
 
-return {
-	selectif = selectif,
-}
+return table.freeze(query)
--- a/lute/std/libs/syntax/query.luau
+++ b/lute/std/libs/syntax/query.luau
@@ -2,7 +2,7 @@
 -- @std/syntax/query
 -- Provides utility functions for querying Luau AST nodes
 
-local luau = require("@lute/luau")
+local luau = require("@std/luau")
 local visitor = require("./visitor")
 
 local query = {}
@@ -11,7 +11,7 @@ local query = {}
 --- Useful for defining declarative-based programs, where the matched nodes are mutated for the migration
 function query.selectif<T>(block: luau.AstStatBlock, predicate: (luau.AstExpr) -> T?): { T }
 	local nodes = {}
-	local finder = visitor.createVisitor()
+	local finder = visitor.createvisitor()
 
 	finder.visitExpression = function(node)
 		local result = predicate(node)
@@ -22,7 +22,7 @@ function query.selectif<T>(block: luau.AstStatBlock, predicate: (luau.AstExpr) -
 		return true
 	end
 
-	visitor.visitBlock(block, finder)
+	visitor.visitblock(block, finder)
 
 	return nodes
 end
--- a/lute/std/libs/syntax/visitor.luau
+++ b/lute/std/libs/syntax/visitor.luau
@@ -1,6 +1,10 @@
 --!strict
+-- @std/syntax/visitor
+-- Visitor pattern implementation for traversing Luau AST nodes
 
 local luau = require("@std/luau")
+
+local visitorlib = {}
 
 export type Visitor = {
 	visitBlock: (luau.AstStatBlock) -> boolean,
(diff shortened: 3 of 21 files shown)


lute/std/libs/syntax/visitor.luau
```diff
@@ -877,12 +881,12 @@ local function createVisitor()
...
-	visitTypePack = visitTypePack,
-	visitToken = visitToken,
-}
+visitorlib.createVisitor = createVisitor
+visitorlib.visitBlock = visitBlock
+visitorlib.visitStatement = visitStatement
+visitorlib.visitExpression = visitExpression
+visitorlib.visitType = visitType
+visitorlib.visitTypePack = visitTypePack
+visitorlib.visitToken = visitToken
```
annieetang: workaround to reduce lines changed because this file is long and there's a lot of local usages of these functions too

lute/std/libs/syntax/query.luau
```diff
@@ -1,10 +1,15 @@
-local luau = require("@lute/luau")
+--!strict
+-- @std/syntax/query
+-- Provides utility functions for querying Luau AST nodes
 
+local luau = require("@lute/luau")
```
skberkeley: since you're here, could you update this to be `@std/luau`?

lute/std/libs/syntax/visitor.luau
skberkeley: dym luacasing these too?

skberkeley (approved):
couple of small asks

lute/std/libs/syntax/visitor.luau
annieetang: it wasn't luacased before so I'd also have to change all the files that use those (it's used in @std/libs/syntax/ files, but also would it mess with anyone already using these)

do you think it's worth changing?

lute/std/libs/syntax/query.luau
annieetang: yep tyty

lute/std/libs/syntax/visitor.luau
skberkeley: i think so, this is a change we need to make at some point anyway, and afaik the only external consumer is the codemod users (@~wmccrthy), who have pinned the lute version they use already

tests/cli/loadbypath.test.luau
```diff
@@ -41,7 +41,7 @@ test.suite("Lute CLI Run", function(suite)
 		fs.write(requireeFile, REQUIREE_CONTENTS)
 		fs.close(requireeFile)
 
-		local result = process.run({ lutePath, requirerPath, requireePath })
+		local result = process.run({ tostring(lutePath), tostring(requirerPath), tostring(requireePath) })
```
annieetang: i wasn't able to use `path.join` for this and was getting errors resolving path to string type

examples/task-wait.luau
```diff
@@ -1,8 +1,8 @@
 local task = require("@lute/task")
-local time = require("@lute/time")
+local time = require("@std/time")
 
 print(task.wait(1))
-print(task.wait(time.duration.seconds(1)))
+print(task.wait(time.seconds(1)))
```
aatxe, line 5: This doesn't make any sense. What happened to the duration library?

aatxe (changes requested):
This PR does too many things at once. If you're going to do big sweeping changes, do one at a time per PR, so that they can be reviewed properly as a _sweep_ over everything.

examples/task-wait.luau
annieetang: oh `std/time` exports a `durations` table that's just the duration library that I guess just shortens the path from `time.duration` to just `time`
tho since the time.luau files are confusing I can keep this as `lute/time` 

(resolved in split up pr)

Vighnesh-V:
I'd like to split apart the test.luau library such that `asserts`, `reporters`, `runners`, `types` and the core `testing` framework are all independently exportable. This enables the eventual test runner system to be more modular + makes it easier to make changes. I was initially thinking about a structure something like:
```
std/libs/test/
    init.luau
    types.luau
    assert.luau
    reporter.luau
    runner.luau
```
@annieetang Do you foresee this structure causing problems with the docs site? Is there something I can do to make this easier for the site? Failing all that, I can still manually write documentation right?

tests/cli/loadbypath.test.luau
skberkeley: can you post the errors you get?

annieetang:
> I'd like to split apart the test.luau library such that `asserts`, `reporters`, `runners`, `types` and the core `testing` framework are all independently exportable. This enables the eventual test runner system to be more modular + makes it easier to make changes. I was initially thinking about a structure something like:
> 
> ```
> std/libs/test/
>     init.luau
>     types.luau
>     assert.luau
>     reporter.luau
>     runner.luau
> ```
> 
> @annieetang Do you foresee this structure causing problems with the docs site? Is there something I can do to make this easier for the site? Failing all that, I can still manually write documentation right?

Should be fine! I was gonna extend the doc generator to traverse subfolders, I think as long as the function formats are like what @Nicell commented as `function crypto.password.hash(password: string): buffer` then it should work w the script!

tests/cli/loadbypath.test.luau
annieetang: talked offline (er, on slack) but just need to change `tostring(lutePath)` and everything else can stay like before

annieetang:
> This PR does too many things at once. If you're going to do big sweeping changes, do one at a time per PR, so that they can be reviewed properly as a _sweep_ over everything.

Would splitting it like
- putting everything in a table + exporting only the table
- luacasing the exported functions
- updating use cases from @lute to @std 

be reasonable? tho the biggest chunk is really the first one and the other two are smaller

annieetang:
closing now that the split up PRs are all merged 🫡

PR #518, opened 2025-11-04
stdlib: lute test shouldn't point to xpcall for error msg

skberkeley:
While working on tests for something else, I noticed that in cases where a test case fails for reasons other than an assertion, the testing harness reports the source of the error as an `xpcall` call in the lute's test implementation. After discussing with @~Vighnesh-V, I've implemented a small workaround where we look lower down in the stack for the source of the error, so that we can actually point to the test case causing the failure in failure reporting.

### DIFF
--- a/lute/std/libs/test.luau
+++ b/lute/std/libs/test.luau
@@ -57,8 +57,8 @@ end
 -- Test Reporting
 type FailedTest = {
 	test: string, -- name of the test case that failed
-	assertion: string, -- name of the assertion that failed (e.g., "assert.eq")
-	error: string, -- the error message from the assertion
+	assertion: string?, -- name of the assertion that failed (e.g., "assert.eq")
+	error: string?, -- the error message from the assertion
 	filename: string, -- source file where the failure occurred
 	linenumber: number, -- line number of the failure
 }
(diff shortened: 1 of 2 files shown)


annieetang (approved):
lgtm!

lute/std/libs/test.luau
```diff
@@ -57,8 +57,8 @@ end
 -- Test Reporting
 type FailedTest = {
 	test: string, -- name of the test case that failed
-	assertion: string, -- name of the assertion that failed (e.g., "assert.eq")
-	error: string, -- the error message from the assertion
+	assertion: string?, -- name of the assertion that failed (e.g., "assert.eq")
+	error: string?, -- the error message from the assertion
```
Vighnesh-V, line 61: nit: It might still be useful to leave this non-optional and broaden it from `errors` for assertions to `errors` for `FailedTests`. I think letting the user know something like: `"Runtime error in your test function"` is still valuable.

lute/std/libs/test.luau
```diff
@@ -213,10 +215,14 @@ function test.run()
 			local fileName, lineNumber = debug.info(4, "sl")
 			local assertName = debug.info(3, "n")
 
+			if assertName == "xpcall" then
```
Vighnesh-V, line 218: Something I didn't think about before - what happens if your test function itself contains an xpcall that fails. Would we still report the error at the right spot?

Vighnesh-V (commented):
Just some thoughts I had. This PR looks good, but lemme know if you think the changes I proposed make sense or not.

lute/std/libs/test.luau
Vighnesh-V: Oh better yet, can we report something like:
```
MyTestCase:
    Failed with: Runtime error in {test} at line {line}
```
as the error message?

lute/std/libs/test.luau
skberkeley: I'll test this, but I think so because of the level we pull the `assertName` out of

PR #519, opened 2025-11-04
Added stacktraces to test runner output

Gravemind1142:
This PR updates the test runner to include stack traces in the output for failed tests. This will make it easier to debug both the tests and the source code.

### DIFF
--- a/lute/std/libs/test.luau
+++ b/lute/std/libs/test.luau
@@ -147,7 +147,7 @@ function test.run()
 	-- Error handler for beforeall hook
 	local function beforeallErrorHandler(suite: TestSuite)
 		return function(err)
-			local stacktrace = debug.traceback(message)
+			local stacktrace = debug.traceback(tostring(err))
 			-- If beforeall fails, skip all tests in the suite
 			total += #suite.cases
 			failed += #suite.cases
@@ -169,7 +169,7 @@ function test.run()
 	-- Error handler for beforeeach hook
 	local function beforeeachErrorHandler(testFullName: string)
 		return function(err)
-			local stacktrace = debug.traceback(message)
+			local stacktrace = debug.traceback(tostring(err))
 			failed += 1
 			local failedtest: FailedTest = {
 				test = testFullName,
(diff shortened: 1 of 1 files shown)


lute/std/libs/test.luau
```diff
@@ -61,6 +61,7 @@ type FailedTest = {
 	error: string, -- the error message from the assertion
 	filename: string, -- source file where the failure occurred
 	linenumber: number, -- line number of the failure
+	stacktrace: string, -- full stacktrace at the point of failure
```
Vighnesh-V, line 64: If you look at `tests/cli/test/test.test.luau` (sorry for horrific naming lol), can you add a test for this? Moreover is it possible to trim the stacktrace? E.g. show the last 5 frames before the crash?

Vighnesh-V (changes requested):
Thanks! This is really nice to have - can you add a test like I suggested as well as look into if it's possible to trim the full stacktrace on display?

Gravemind1142:
> look into if it's possible to trim the full stacktrace on display?

@Vighnesh-V 

Most popular test runners have a way to pass configuration to the unit test runner for things like stacktrace verbosity in CI vs during development. We don't have that today, and forcing engs to use "only truncated" in the meantime seems too restrictive, so I think it would be best to align on how we want configuration for test.luau to work first.

Vighnesh-V:
@Gravemind1142 That's reasonable - we can include the full stacktrace for now and include the verbosity as a toggle in `lute test`. Do you mind making an issue for this in the github and assigning it to me?

lute/std/libs/test.luau
```diff
@@ -74,6 +75,10 @@ local function printFailedTest(failed: FailedTest)
 	print(`❌ {failed.test}`)
 	print(`   {failed.filename}:{failed.linenumber}`)
 	print(`   	{failed.assertion}: {failed.error}`)
+	if failed.stacktrace and failed.stacktrace ~= "" then
```
aatxe, line 78:
```suggestion
	if failed.stacktrace ~= "" then
```

if the `stacktrace` is always present and just empty string when not set, we don't need to check if it's present

Vighnesh-V:
Added in https://github.com/luau-lang/lute/pull/651.

PR #520, opened 2025-11-04
refactor: stdlib/syntax file formats

annieetang:
This one is refactoring modules in `std/libs/syntax` and its use cases

### DIFF
--- a/examples/transformer.luau
+++ b/examples/transformer.luau
@@ -6,7 +6,7 @@ local luau = require("@lute/luau")
 function transformation(ctx)
 	local parsedResult = parser.parsefile(ctx.source)
 
-	local v = visitor.createvisitor()
+	local v = visitor.create()
 
 	v.visitBinary = function(node: luau.AstExprBinary)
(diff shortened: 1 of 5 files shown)


Vighnesh-V:
Just to clarify, this change is:
a) Making it so that the docs generator can interpret the format of the stdlib to produce documentation
b) Making style changes e.g. luacasing things?

Vighnesh-V (approved):

examples/transformer.luau
```diff
@@ -6,7 +6,7 @@ local luau = require("@lute/luau")
 function transformation(ctx)
 	local parsedResult = parser.parsefile(ctx.source)
 
-	local v = visitor.createVisitor()
+	local v = visitor.createvisitor()
```
aatxe: Surely this can just be called `create` since it's already under the visitor library.

aatxe (changes requested):
one change, but otherwise LGTM

aatxe (approved):

PR #521, opened 2025-11-04
refactor: stdlib fs, io, process, and system file formats

annieetang:
this one is for `fs`, `io`, `process`, and `system` in `std/libs` so the docs generator can interpret the format to produce documentation


skberkeley (approved):

lute/std/libs/system/init.luau
```diff
@@ -7,26 +7,27 @@ local system = require("@lute/system")
 local path = require("@std/path")
 local platforminfo = require("@self/platforminfo")
```
aatxe: probably this should just be called `platform`? `platforminfo` feels like A Lot to me

aatxe (approved):

aatxe (approved):

PR #523, opened 2025-11-04
refactor: stdlib string, table, task, vector file formats

annieetang:
this one is for string, table, task, vector in std/libs so the docs generator can interpret the format to produce documentation


skberkeley (approved):

lute/std/libs/string.luau
```diff
@@ -1,3 +1,8 @@
+--!strict
+-- @std/string
+-- stdlib for the built-in string library in Luau
+-- Provides additional utility functions for string operations
+
 local strings = {}
```
aatxe: honestly, let's rename `@std/string` to `@std/stringext`, `@std/vector` to `@std/vectorext`, and `@std/table` to `@std/tableext` and get rid of the re-exports from the builtin string and table libraries. they eliminate fast calls today, and that sucks, even if my preference _is_ that every library would be the singular name of the type. i talked with both slava and ilya, and it just doesn't seem likely that we'll make this stuff not a large performance hit anytime soon.

aatxe: would also accept `@std/strings`, `@std/vectors`, `@std/tables`, i guess, but that feels a little more odd?

aatxe (changes requested):

lute/std/libs/string.luau
annieetang: can do `-ext` and removing the re-exports from string and table!

lute/std/libs/stringext.luau
```diff
@@ -0,0 +1,41 @@
+--!strict
+-- @std/stringext
+-- stdlib for an extension of the built-in string library in Luau
+-- Provides additional utility functions for string operations but does not include re-exports
```
aatxe:
```suggestion
-- Provides additional utility functions for string operations
```

No need to specify that it doesn't include re-exports.

lute/std/libs/tableext.luau
```diff
@@ -0,0 +1,49 @@
+--!strict
+-- @std/tableext
+-- stdlib for an extension of the built-in table library in Luau
+-- Provides additional utility functions for table operations but does not include re-exports
```
```suggestion
-- Provides additional utility functions for table operations
```

No need to specify that it doesn't provide re-exports

lute/std/libs/vectorext.luau
```diff
@@ -0,0 +1,20 @@
+--!strict
+-- @std/vectorext
+-- stdlib for an extension of the built-in vector library in Luau
+-- Provides additional utility functions for vector operations but does not include re-exports
```
aatxe:
```suggestion
-- Provides additional utility functions for vector operations
```

No need to specify that it doesn't provide re-exports.

aatxe (approved):

PR #525, opened 2025-11-04
stdlib: tweak error msg in tests with runtime errors

skberkeley:
Follow on work to #518 to tweak the error message for tests that fail with runtime errors


lute/std/libs/test.luau
```diff
@@ -57,8 +57,8 @@ end
 -- Test Reporting
 type FailedTest = {
 	test: string, -- name of the test case that failed
-	assertion: string?, -- name of the assertion that failed (e.g., "assert.eq")
-	error: string?, -- the error message from the assertion
+	assertion: string?, -- if an assertion failed, name of the assertion that failed (e.g., "assert.eq")
+	assertion_error: string?, -- the error message to report with a failed assertion
```
Vighnesh-V: I think just leaving this as `error` is fine.

Vighnesh-V (approved):

lute/std/libs/test.luau
skberkeley: discussed offline, will keep as `assertion_error` for now as this has been observed to be always `nil` when a runtime error (not an assertion failure) causes a test to fail

aatxe:
```suggestion
	message: string?, -- the error message to report with a failed assertion
```

maybe? this should still be in luacase, not snake_case, and just calling it the message is much more clear to me

skanosue: done!

aatxe (approved):

PR #527, opened 2025-11-05
refactor: StaticRequireTracer should be reworked to use the resolveRequire api

Vighnesh-V:
This PR makes use of the resolverequire api exposed in https://github.com/luau-lang/lute/pull/496 to perform static require tracing. This will handle static resolution of aliases for free during compilation - a followup PR will actually handle loading bundled files when they are required via aliases.

Because the `resolverequire` code operates in terms of absolute paths, all static require tracing is now done in terms of absolute paths as well. This has some knock on consequences for how creating Lute Payloads works, which requires me to refactor this in the same PR.

Prior, the `LuteExePayload::add(file path)` method used to take a relative path to the source file being embedded, and all embedded files would be resolved relative to this path. This path was used as both a key in the bundle, as well as the filepath to the script. For example:
```
lute compile tests/staticrequires/src/main.luau
```
would embed:
```
tests/staticrequires/src/main.luau
tests/staticrequires/src/utils.luau
tests/staticrequires/src/deps/dep1.luau
tests/staticrequires/src/deps/dep2.luau
```

However, static require tracing now strips off the longest common prefix of paths, in order to a) save some bytes and b) avoid exposing the directory structure any more than necessary.

This means that the bundle will look like:
```
main.luau
utils.luau
deps/dep1.luau
deps/dep2.luau
```

However, passing just these as keys is insufficient information for the `LuteExePayload` to be able to find and compile these scripts. That's why the `add` method now takes both a `bundleName` e.g. `main.luau` and a filepath, e.g. `/path/to/main.luau`.

### DIFF
--- a/lute/cli/include/lute/compile.h
+++ b/lute/cli/include/lute/compile.h
@@ -1,8 +1,9 @@
 #pragma once
 
+#include "lute/reporter.h"
+
 #include "Luau/DenseHash.h"
 #include "Luau/FileUtils.h"
-#include "lute/reporter.h"
 
 #include <string_view>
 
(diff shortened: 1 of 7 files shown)


vrn-sn (approved):
We'll need to merge in `primary` and update all the `printf`s in the code. (And probably clang-format because of the new rules.)

Vighnesh-V:
Merged primary, updated the printfs (also converted some existing printf's to use the reporter as well), ran clang-format..

PR #528, opened 2025-11-05
refactor: Split up std test

Vighnesh-V:
In order to make the components of std/test more independently requirable, I've split it up. I'm going to make the luacasing pass in a follow up PR [here](https://github.com/luau-lang/lute/pull/529)


lute/std/libs/assert.luau
```diff
@@ -1,7 +1,3 @@
---!strict
--- @std/assert
--- Standard assertion library for Luau
-
```
annieetang, line 4: sadge
(do we not want to have header comments in stdlib files)

annieetang:
coud we perhaps have the style structure in std/libs/tests/ to follow https://github.com/luau-lang/lute/pull/516 so these can make it into the docs generator and it's consistent styling

Vighnesh-V:
> coud we perhaps have the style structure in std/libs/tests/ to follow #516 so these can make it into the docs generator and it's consistent styling

Yup, very doable, I'll make that change right now.

lute/std/libs/assert.luau
Vighnesh-V: just updated the relevant files.

lute/std/libs/test/reporter.luau
```diff
@@ -0,0 +1,43 @@
...
+	end
+
+	print(string.rep("-", 50))
+	print(`Total:  {result.total}`)
+	print(`Passed: {result.passed} ✓`)
+	print(`Failed: {result.failed} ✗`)
+	print(string.rep("=", 50) .. "\n")
+end
+
+return table.freeze({ simple = simpleReporter })
```
annieetang: can we also pull this into a table defined at the top as `local reporter = {}` and just return `table.freeze(reporter)` 😁 then do `reporter.simple` instead of `local function simpleReporter(...)`

lute/std/libs/test/runner.luau
```diff
@@ -0,0 +1,187 @@
...
+
+	return {
+		failures = failures,
+		total = total,
+		failed = failed,
+		passed = passed,
+	}
+end
+
+return table.freeze({ run = run })
```
annieetang: ditto w comment in `reporter.luau`!

and same about naming, if it could be `local runner = {}` and `table.freeze(runner)` with functions like `function runner.run`

Vighnesh-V: fixed!

lute/std/libs/test/reporter.luau
Vighnesh-V: fixed!

annieetang (approved):
lgtm!

PR #531, opened 2025-11-05
API: luau.typeofmodule v1

annieetang:
### Problem

As part of an effort to bring type-directed documentation generation to Luau, we don't have an automated way of getting the return type of a Luau module, and this is a broader missing feature as well in Luau for being able to analyze the exact type of a module.

### Solution

We introduce an API in `@lute/luau` that allows users to get the return type of a module path by calling `luau.typeofmodule()` and passing in the module path as a string. We create a `Luau::Frontend` to analyze the and resolve the Module, including resolving requires to return the full return type of the specified module! For ease of reviewing, this API temporarily returns the `string` representation while we bikeshed a Luau data structure representation of the `Type` that's returned. 

As written in `examples/get_module_return_type.luau` an example usage would be retrieving the return type of `@std/task`, which when we run 

```luau
local moduleTypeString = luau.typeofmodule("./lute/std/libs/task.luau")
```

We currently return a string representation of the module type:

```
{
    await: (t: {
        co: thread,
        result: any,
        success: boolean?
    }) -> any,
    awaitall: (...{
        co: thread,
        result: any,
        success: boolean?
    }) -> (...any),
    create: <a..., b...>(f: (a...) -> (b...), ...any) -> {
        co: thread,
        result: any,
        success: boolean?
    }
}
```

Which helps us understand that the `task` library has the following functions `await`, `awaitall`, and `create` with their respective full return types.

This is all built on top of @wmccrthy's solid boiler plate of the API (https://github.com/luau-lang/lute/pull/375) and @vrn-sn 's help with the resolve requires!!

### Future Work

- We definitely want to return a data structure representation of the Type instead of the string
- We also want to represent types with aliases instead of returning the exact type (something like how our `definitions/` files look with the exported types and functions).
- If the file we're getting the type of ever has errors, then the result will include the errors because our Frontend mode is set to `NoCheck`, so there will have to be some work done on resolving the errors in the module files I think
- Alias resolving isn't fully working yet with `@std/` directly passed in to the `typemodule` call (see comments below), but will be addressed shortly


lute/luau/src/luau.cpp
```diff
@@ -2525,6 +2535,123 @@ struct AstSerialize : public Luau::AstVisitor
...
+
+            return {{*absolutePath}};
+        }
+
+        return std::nullopt;
+    }
+};
+
+// Based on CliConfigResolver in Analyze.cpp.
+struct ModuleConfigResolver : Luau::ConfigResolver
```
annieetang: This entire `ConfigResolver` is copied from `Analyze.cpp`, lmk if there's anything unnecessary but I think we do need to handle both the `Luau::kConfigName` and `Luau::kLuauConfigName` which are the `.luaurc` files as well as `.config.luau`

lute/luau/CMakeLists.txt
```diff
@@ -10,5 +10,5 @@ target_sources(Lute.Luau PRIVATE
 
 target_compile_features(Lute.Luau PUBLIC cxx_std_17)
 target_include_directories(Lute.Luau PUBLIC "include" ${LIBUV_INCLUDE_DIR})
-target_link_libraries(Lute.Luau PRIVATE Lute.Require Lute.Runtime uv_a Luau.Analysis Luau.Ast Luau.Compiler Luau.Require Luau.VM)
+target_link_libraries(Lute.Luau PRIVATE Lute.Require Lute.Runtime uv_a Luau.Analysis Luau.Ast Luau.Compiler Luau.CLI.lib Luau.Require Luau.VM)
```
skberkeley, line 17: what is this for?

lute/luau/src/luau.cpp
```diff
@@ -20,6 +27,9 @@
 #include <iterator>
 #include <memory>
 #include <string>
+#include <map>
+#include <optional>
+#include <string>
```
skberkeley, line 24: why are we re-importing string? can you also alpha-sort each section of includes?

lute/luau/src/luau.cpp
```diff
@@ -2684,6 +2811,41 @@ int load_luau(lua_State* L)
...
+    Luau::FrontendOptions fopts;
+    fopts.retainFullTypeGraphs = true;
+
+    Luau::Frontend frontend(&fileResolver, &configResolver, fopts);
+    Luau::registerBuiltinGlobals(frontend, frontend.globals);
+    Luau::freeze(frontend.globals.globalTypes);
+
+    frontend.check(modulePath);
+
+    Luau::ModulePtr module = frontend.moduleResolver.getModule(modulePath);
```
skberkeley: dunno if this matters, but `module` is a c++20 keyword, so maybe worth using a different name?

lute/luau/include/lute/luau.h
```diff
@@ -21,12 +21,15 @@ int compile_luau(lua_State* L);
...
 
+int luau_typeofmodule(lua_State* L);
+
 static const luaL_Reg lib[] = {
     {"parse", luau_parse},
     {"parseexpr", luau_parseexpr},
     {"compile", compile_luau},
     {"load", load_luau},
     {"resolverequire", resolverequire_luau},
+    {"typeofmodule", luau_typeofmodule},
```
annieetang: what's the naming convention? I see `luau_xxx` and also `xxx_luau`

lute/luau/src/luau.cpp
```diff
@@ -2525,6 +2535,123 @@ struct AstSerialize : public Luau::AstVisitor
...
+        Luau::SourceCode::Type sourceType;
+        std::optional<std::string> source = std::nullopt;
+
+        source = readFile(name);
+        sourceType = Luau::SourceCode::Module;
+
+        if (!source)
+            return std::nullopt;
+
+        return Luau::SourceCode{*source, sourceType};
```
skberkeley:
```suggestion
        if (std::optional<std::string> source = readFile(name))
             return Luau::SourceCode{*source, Luau::SourceCode::Module};

        return std::nullopt;
```

lute/luau/src/luau.cpp
```diff
@@ -2525,6 +2535,123 @@ struct AstSerialize : public Luau::AstVisitor
...
+
+    // We are currently resolving modules and requires only, and will add support for Roblox globals / types in a subsequent PR.
+    std::optional<Luau::ModuleInfo> resolveModule(const Luau::ModuleInfo* context, Luau::AstExpr* node) override
+    {
+        if (auto expr = node->as<Luau::AstExprConstantString>())
+        {
+            std::string requirePath(expr->value.data, expr->value.size);
+
+            std::string error;
+            std::string requirerChunkname = "@" + context->name;
```
annieetang: `resolveRequire`'s `chunkname` param needs `@` at the front, @vrn-sn did we want to make a wrapper for this in c++ or just for luau (i.e., is this fine)

skberkeley (commented):
a few comments, i'll leave it to someone who knows the resolve logic better to +1

lute/luau/CMakeLists.txt
annieetang: uh I think it's for the Luau::Frontend stuff but I should double check

lute/std/libs/luau.luau
```diff
@@ -199,4 +199,10 @@ function luau.loadbypath(requirePath: path.pathlike): any
 	return luau.load(migrationBytecode, requirePathStr)()
 end
 
+function luau.typeofmodule(modulepath: path.pathlike): string
+	local modulePathStr = if typeof(modulepath) == "string" then modulepath else path.format(modulepath)
```
Vighnesh-V, line 203: path.format should already take a pathlike no?

Vighnesh-V: Do we need to normalize as well here?

lute/luau/src/luau.cpp
```diff
@@ -2684,6 +2811,41 @@ int load_luau(lua_State* L)
     return 1;
 }
 
+int luau_typeofmodule(lua_State* L)
+{
+    std::string modulePath = luaL_checkstring(L, 1);
+
+    ModuleFileResolver fileResolver;
+    ModuleConfigResolver configResolver(Luau::Mode::NoCheck);
```
Vighnesh-V: Should this be using the https://github.com/luau-lang/lute/blob/5b5d0b69a0e58a623d8e36f85a424f285bc50908/lute/cli/src/tc.cpp#L75 ?

lute/luau/src/luau.cpp
```diff
@@ -2525,6 +2535,123 @@ struct AstSerialize : public Luau::AstVisitor
     }
 };
 
+// Based on CliFileResolver in Analyze.cpp.
+struct ModuleFileResolver : Luau::FileResolver
```
Vighnesh-V: We should extract this out into it's own file and write some tests for it.

lute/luau/src/luau.cpp
Vighnesh-V: Same here re: testing.

lute/luau/CMakeLists.txt
Vighnesh-V: This shouldn't pull in Luau.CLI.lib at all. That library is only for the command line specific chunk of lute. If you need Frontend, this already pulls in Luau.Analysis as a dependency.

definitions/luau.luau
```diff
@@ -598,4 +598,9 @@ function luau.resolverequire(path: string, fromchunkname: string): string
 	error("not implemented")
 end
 
+-- For now, we return a string representation of the module's type, but we will expand it to some Luau data structure representation of Type (similar to the AST types) in a subsequent PR.
+function luau.typeofmodule(modulepath: string): string
```
Vighnesh-V, line 602: the definition in `luau.luau` said this takes a pathlike? Should the type here still be string?

lute/luau/include/lute/luau.h
Vighnesh-V: Not sure. Suffix luau appears more frequently, so maybe lets do that. Can be a separate task to make these more consistent.

lute/luau/src/luau.cpp
Vighnesh-V: `clang-format -i /path/to/your/file` will probably work.

lute/luau/src/luau.cpp
Vighnesh-V: I think this is basically a subset of https://github.com/luau-lang/lute/blob/5b5d0b69a0e58a623d8e36f85a424f285bc50908/lute/cli/src/tc.cpp#L75 - can you extract that into a file that both cli/tc and this file can access?

lute/luau/src/luau.cpp
Vighnesh-V: this is fine - prepending @ is expected.

Vighnesh-V (changes requested):
Tests needed + we should move the file resolvers into their own files that are accessible elsewhere. Stuff in lute/ shouldn't need the `Lute.CLI.lib` include.

lute/luau/src/luau.cpp
vrn-sn: I am going to eventually make a wrapper for this in `Lute.Require`, but it's fine for this.

lute/luau/CMakeLists.txt
vrn-sn: It's because of `Luau/FileUtils.h` (@Vighnesh-V: note this is Luau.CLI.lib, not Lute.CLI.lib). I think it's fine—we also use it for Lute.Require's `FileVfs` implementation.

lute/std/libs/luau.luau
skberkeley: i think path.format will be confused if we have aliases in the path?

definitions/luau.luau
skberkeley: i think this is fine? all our functions at the @lute definition level take strings instead of pathlikes

lute/luau/CMakeLists.txt
Vighnesh-V: whew ok. i misread this then, my bad @annieetang !

definitions/luau.luau
Vighnesh-V: cool, that checks out.

lute/luau/src/luau.cpp
annieetang: oH woah I did not know we had that thanks!

lute/luau/CMakeLists.txt
annieetang: nw thanks y'all for clarifying !

lute/luau/src/luau.cpp
```diff
@@ -14,10 +14,14 @@
...
 #include "lualib.h"
 
+#include "lute/configresolver.h"
+#include "lute/moduleresolver.h"
+
 #include <cstddef>
 #include <cstring>
 #include <iterator>
 #include <memory>
+#include <optional>
```
Vighnesh-V: Do you still need this import?

Vighnesh-V (approved):
Looks good - thank you for adding this api!

Left one non blocking comment about trimming a potentially unused include <optional> in `luau.cpp`, but otherwise good to go.

lute/luau/src/luau.cpp
annieetang: nope thanks for the catch!

vrn-sn (approved):
Only note is that the example code and the test case use `./lute/std/libs/task.luau` as their file, but this is slightly brittle if we were to change things in that directory. Maybe the test can create a temporary file, and the example can just reference a file in its directory instead (maybe we can create a subfolder for the example).

annieetang:
> Only note is that the example code and the test case use `./lute/std/libs/task.luau` as their file, but this is slightly brittle if we were to change things in that directory. Maybe the test can create a temporary file, and the example can just reference a file in its directory instead (maybe we can create a subfolder for the example).

Oh yes ty, Vighnesh had also mentioned that to me about maybe doing a cwd + /path/to/std/task? Wdyt

vrn-sn:
I'd say to not tie it to `task.luau` and just make a file for both use cases: the example and the test. Right now, every change to `task.luau` will require a change in the test and in the example file's comment, which just adds unnecessary work.

annieetang:
> I'd say to not tie it to `task.luau` and just make a file for both use cases: the example and the test. Right now, every change to `task.luau` will require a change in the test and in the example file's comment, which just adds unnecessary work.

sg, will make a tmpdir and a super simple source code

tests/std/luau.test.luau
```diff
@@ -1,9 +1,33 @@
...
+				trimmed = stringext.trim("  hello world  "),
+			}
+		]]
+		)
+		fs.close(testFile)
+
+		local moduleTypeString = luau.typeofmodule(testPath)
 
 		-- Basic check for now until the data structure is created
 		assert.neq(moduleTypeString, "")
```
vrn-sn: I think we can hardcode the result for now since it's tied to the string you wrote to the file.

Vighnesh-V (approved):

annieetang:
Talked with Vighnesh and the test case using `require("@std/stringext")` wasn't actually working properly because `resolveRequire` API just creates a `FileVfsContext` but we might need `StdlibVFS` context. Gonna leave this as a follow up task and talk more with @vrn-sn on the right way to make this happen. For now, I'm just having the test case not use `require` if that's ok!

vrn-sn:
I think it's fine to not use `require` in the test case. It's not hard to add support for the other VFS's, but it's hard to expose in a good way. The current API returns absolute filepaths, but a more general API would need to sensibly represent embedded VFS nodes as well (which is why I left it out of scope when implementing `resolveRequire`).

examples/get_module_return_type.luau
```diff
@@ -0,0 +1,25 @@
+local luau = require("@std/luau")
+
+local moduleTypeString = luau.typeofmodule("./lute/std/libs/task.luau")
```
vrn-sn: Test case looks good, but I think we can still improve this example in the same way. I think I'd left a comment about it earlier. We could maybe just make this example a folder with two files: this file and an example module whose type we extract.

annieetang: Sounds good I'll take a look. I'm also planning to have more comprehensive tests once the data structure is thought out and it'll be easier to actually test, so I wanted to keep the first test case here pretty simple (also bc the next thing I'm working on is the data structure haha)

vrn-sn: I think we're talking about different things... I just mean the example in the `examples` folder.

annieetang: ohh lol myb sg!

tests/std/syntax/parser.test.luau
```diff
@@ -125,6 +125,9 @@ test.suite("Parser", function(suite)
 	suite:case("roundtrippableAst", function(assert)
 		local function visitDirectory(directory: string)
 			for _, entry in fs.listdir(directory) do
+				if entry.type ~= "file" then
+					continue
+				end
```
annieetang, line 130: had to add this check because I added a subfolder in `examples/`

vrn-sn (approved):

PR #532, opened 2025-11-06
Refactor: move std/time.duration into proper submodule

### LINKED ISSUE #526: Refactor: move std/time.duration into proper submodule
follow up PR for https://github.com/luau-lang/lute/pull/524#discussion_r2492086478

annieetang:
Resolves https://github.com/luau-lang/lute/issues/526 and should be the last refactor before I can get vitepress doc generation working for stdlib!

Also added tests for `@std/time`

### DIFF
--- a/lute/std/libs/time/init.luau
+++ b/lute/std/libs/time/init.luau
@@ -1,0 +1,12 @@
+--!strict
+-- @std/time
+-- Provides utility functions for time
+
+local duration = require("@self/duration")
+local time = {}
+
+export type duration = duration.duration
+
+time.duration = duration
+
+return table.freeze(time)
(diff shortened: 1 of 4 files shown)


lute/std/libs/time/init.luau
```diff
@@ -0,0 +1,12 @@
+--!strict
+-- @std/time
+-- Provides utility functions for time
+
+local duration = require("@self/duration")
+local time = {}
+
+export type duration = duration.duration
```
annieetang, line 8: we do want to export the duration type in `std/time` right?

aatxe: yeah, definitely.

aatxe (approved):

PR #533, opened 2025-11-06
refactor: Suppress printing to stdout in test cases by logging to a reporter

### LINKED ISSUE #511: Test output being polluted with prints intended for CLI
As of the time of writing, this is our `Lute.Test` output:
```
[doctest] doctest version is "2.4.12"
[doctest] run with "--help" for options
Decode failed: LUTEBYTE magic flag not found
Decode failed: Binary too small (4 bytes) to contain valid payload
Encode failed: Could not read file '/Users/vsaini/git/lute/tests/src/this_file_does_not_exist.luau'
Encode failed: No files added to payload
Decode failed: Expected 99 files but parsed 1
Error: File './tests/src/require/without_config/nested/init' does not exist.
Error: File '/Users/vsaini/git/lute/tests/src/require/without_config/nested/init' does not exist.
===============================================================================
[doctest] test cases:  24 |  24 passed | 0 failed | 0 skipped
[doctest] assertions: 211 | 211 passed | 0 failed |
[doctest] Status: SUCCESS!
```

Our code in `Lute.CLI.lib` shouldn't be directly printing to the terminal.

Vighnesh-V:
This PR adds a reporter to the Lute.CLI.Lib library. Right now, our tests just log directly to stderr / stdout and we don't have any control over this or ability to intercept and test these calls. I've introduced `LuteReporter` which has `TestReporter` and `CliReporter` as subclasses, which store reported strings / logs to stderr/stdout respectively. 

This lets us access the errors logged out in tests as well as suppresses random logging in our tests.

Fixes: https://github.com/luau-lang/lute/issues/511

### DIFF
--- a/lute/cli/CMakeLists.txt
+++ b/lute/cli/CMakeLists.txt
@@ -26,6 +26,7 @@ target_sources(Lute.CLI.lib PRIVATE
     include/lute/climain.h
     include/lute/compile.h
     include/lute/luauflags.h
+    include/lute/reporter.h
     include/lute/staticrequires.h
     include/lute/tc.h
     include/lute/uvstate.h
@@ -46,6 +47,9 @@ target_compile_options(Lute.CLI.lib PRIVATE ${LUTE_OPTIONS})
 add_executable(Lute.CLI)
 
 target_sources(Lute.CLI PRIVATE
+    include/lute/clireporter.h
+
+    src/clireporter.cpp
     src/main.cpp
 )
 
--- a/lute/cli/include/lute/reporter.h
+++ b/lute/cli/include/lute/reporter.h
@@ -1,0 +1,53 @@
+#pragma once
+
+#include <cstdio>
+#include <string>
+#include <vector>
+
+// Interface for reporting output and errors from CLI operations
+class LuteReporter
+{
+public:
+    virtual ~LuteReporter() = default;
+
+    // Report an error message (typically to stderr)
+    virtual void reportError(const std::string& message) = 0;
+
+    // Report normal output (typically to stdout)
+    virtual void reportOutput(const std::string& message) = 0;
+
+    // Variadic template version for formatted error messages using printf-style format specifiers
+    template<typename... Args>
+    void formatError(const char* format, Args&&... args)
+    {
+        reportError(formatMessage(format, std::forward<Args>(args)...));
+    }
+
+    // Variadic template version for formatted output messages using printf-style format specifiers
+    template<typename... Args>
+    void formatOutput(const char* format, Args&&... args)
+    {
+        reportOutput(formatMessage(format, std::forward<Args>(args)...));
+    }
+
+private:
+    // Format implementation using snprintf
+    template<typename... Args>
+    static std::string formatMessage(const char* format, Args&&... args)
+    {
+        // Calculate required buffer size
+        int size = std::snprintf(nullptr, 0, format, args...);
+
+        if (size > 0)
+        {
+            // Allocate buffer and format the string
+            std::vector<char> buffer(size + 1);
+            std::snprintf(buffer.data(), buffer.size(), format, args...);
+
+            return std::string(buffer.data(), size);
+        }
+
+        // Fallback if formatting fails
+        return std::string(format);
+    }
+};
--- a/lute/cli/include/lute/staticrequires.h
+++ b/lute/cli/include/lute/staticrequires.h
@@ -2,6 +2,8 @@
 
 #include "Luau/DenseHash.h"
 
+#include "lute/reporter.h"
+
 #include <optional>
 #include <string>
 #include <vector>
@@ -9,7 +11,7 @@
 class StaticRequireTracer
 {
 public:
-    StaticRequireTracer() = default;
+    StaticRequireTracer(LuteReporter& reporter);
 
     // Trace dependencies starting from an entry point file
     // rootDirectory: Base directory for resolving all requires
@@ -37,4 +39,6 @@ private:
 
     // Resolve a require path relative to the requiring file
     std::optional<std::string> resolveRequire(const std::string& requirer, const std::string& required);
+
+    LuteReporter& reporter;
 };
--- a/lute/cli/include/lute/tc.h
+++ b/lute/cli/include/lute/tc.h
@@ -4,4 +4,6 @@
 #include "Luau/FileUtils.h"
 #include "Luau/Frontend.h"
 
-int typecheck(const std::vector<std::string>& sourceFiles);
+#include "lute/reporter.h"
+
+int typecheck(const std::vector<std::string>& sourceFiles, LuteReporter& reporter);
--- a/lute/cli/src/climain.cpp
+++ b/lute/cli/src/climain.cpp
@@ -13,6 +13,7 @@
 #include "lute/options.h"
 #include "lute/process.h"
 #include "lute/ref.h"
+#include "lute/reporter.h"
 #include "lute/require.h"
 #include "lute/runtime.h"
 #include "lute/staticrequires.h"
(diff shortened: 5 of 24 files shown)


vrn-sn:
I think there's some compile errors right now, could you check those out?

Vighnesh-V:
I think ubuntu runners are a little more picky about includes - this passed on macos locally.  I was missing a <memory> include.

lute/cli/include/lute/reporter.h
```diff
@@ -0,0 +1,53 @@
...
+    template<typename... Args>
+    void formatOutput(const char* format, Args&&... args)
+    {
+        reportOutput(formatMessage(format, std::forward<Args>(args)...));
+    }
+
+private:
+    // Format implementation using snprintf
+    template<typename... Args>
+    static std::string formatMessage(const char* format, Args&&... args)
```
vrn-sn, line 36: Is Luau/StringUtils.h useful for this at all? It should be exposed under Luau.Common now.

Vighnesh-V: The distinction between the two is that this one will give you compile type type checking. However, it doesn't actually matter because the other one will also warn you, by virtue of having all the `-Wformat*` warnings turned on.

If I use the Luau implementation, I'll need to rewrite this one in terms of va_list, which will look a bit weird, since you cannot forward C style variadics without turning them into a va_list (there's no interop between template variadics and c style variadics sadly):
```
 void LuteReporter::formatError(const char* format, ...)
  {
      std::string result;
      va_list args;
      va_start(args, format);
      Luau::vformatAppend(result, format, args);
      va_end(args);
      reportError(result);
  }
```

Vighnesh-V: I'm happy to do either or btw - I'll leave the decision up to you!

vrn-sn: I think let's leave what you have, the `va_list`–templates stuff sounds nasty.

lute/cli/include/lute/staticrequires.h
```diff
@@ -2,14 +2,16 @@
 
 #include "Luau/DenseHash.h"
 
+#include "reporter.h"
```
vrn-sn:
```suggestion
#include "lute/reporter.h"
```

lute/cli/include/lute/tc.h
```diff
@@ -4,4 +4,6 @@
 #include "Luau/Frontend.h"
 #include "Luau/FileUtils.h"
 
-int typecheck(const std::vector<std::string>& sourceFiles);
+#include "reporter.h"
```
vrn-sn:
```suggestion
#include "lute/reporter.h"
```

tests/src/stdsystem.test.cpp
```diff
@@ -21,20 +21,24 @@ TEST_CASE_FIXTURE(CliRuntimeFixture, "std_system_os_matches_host_os")
...
     std::string os = getHostOS();
-
-    auto checkBool = [&](const std::string& field, bool expected) {
-        runCode("local system = require(\"@std/system\")\n"
-                "capture(system." + field + ")\n");
-        std::string output = getCapturedOutput();
+    auto checkBool = [this](const std::string& field, bool expected)
+    {
+        runCode(
+            "local system = require(\"@std/system\")\n"
```
vrn-sn: Can we just make this a raw string? Kind of bizarre to read after the auto-formatting.

tests/src/require.test.cpp
```diff
@@ -7,16 +7,33 @@
...
+    {
+    }
+
+    TestReporter& getReporter()
+    {
+        return *reporter;
+    }
+
+private:
+    std::unique_ptr<TestReporter> reporter;
```
vrn-sn: Looks like we're doing this for each of our test fixtures. Could we instead just move this into a base class like `ReporterFixture` that each of these publicly inherit from? We don't need the constructors of each of these fixtures to concern themselves with the reporter, and this member and `getReporter` can be moved into the base class as well.

lute/cli/src/climain.cpp
```diff
@@ -33,6 +34,8 @@
 #include "lute/version.h"
 #include "lute/vm.h"
 
+#include "clireporter.h"
```
vrn-sn: This file shouldn't need `clireporter.h` right?
```suggestion
```

lute/cli/include/lute/reporter.h
```diff
@@ -0,0 +1,53 @@
...
+    {
+        // Calculate required buffer size
+        int size = std::snprintf(nullptr, 0, format, std::forward<Args>(args)...);
```
vrn-sn:
```suggestion
        int size = std::snprintf(nullptr, 0, format, args...);
```

lute/cli/include/lute/reporter.h
```diff
@@ -0,0 +1,53 @@
...
+
+        if (size > 0)
+        {
+            // Allocate buffer and format the string
+            std::vector<char> buffer(size + 1);
+            std::snprintf(buffer.data(), buffer.size(), format, std::forward<Args>(args)...);
```
vrn-sn:
```suggestion
            std::snprintf(buffer.data(), buffer.size(), format, args...);
```
I think forwarding twice isn't what we want here, plus `snprintf` won't really benefit from move semantics.

vrn-sn (changes requested):

vrn-sn:
Also, don't we need to make CMakeLists.txt changes? (How is this compiling...)

Edit: it's compiling because the new files are being included into existing source files, but we should split stuff into .h/.cpp files and add everything to CMakeLists.txt.

tests/src/cliruntimefixture.cpp
```diff
@@ -5,37 +5,47 @@
...
-        lua_settable(L, LUA_REGISTRYINDEX);
-        lua_pushcfunction(L, capture, "capture");
-        lua_setglobal(L, "capture");
-    });
+    L = setupCliState(
+        *runtime,
+        [rep = reporter.get()](lua_State* L)
+        {
+            lua_pushlightuserdata(L, (void*)rep);
+            lua_setglobal(L, "reporter");
```
vrn-sn:
```suggestion
            lua_setfield(L, LUA_REGISTRYINDEX, "reporter");
```

tests/src/cliruntimefixture.cpp
```diff
@@ -5,37 +5,47 @@
...
+    {
+        // It's a light userdata, proceed to retrieve
+        void* rep = lua_touserdata(L, -1); // Get it from the top of the stack
+        TestReporter* reporter = static_cast<TestReporter*>(rep);
+        reporter->reportOutput(str);
+    }
+    else
+    {
+        throw std::runtime_error("Expected reporter to be a light userdata");
+    }
```
vrn-sn:
```suggestion
    lua_getfield(L, LUA_REGISTRYINDEX, "reporter");
    TestReporter* reporter = static_cast<TestReporter*>(lua_touserdata(L, -1));
    reporter->reportOutput(str);
```
`lua_touserdata` will return `nullptr` if it's not the correct type, so this will segfault if we've set it up wrong. We shouldn't need to do runtime checks for this since the Luau API guarantees it will work if we've set it up correctly in the `setupCliState` callback.

vrn-sn (changes requested):

lute/cli/src/climain.cpp
Vighnesh-V: yup, I've tweaked this.

tests/src/cliruntimefixture.cpp
Vighnesh-V: fixed!

tests/src/cliruntimefixture.cpp
Vighnesh-V: fixed!

tests/src/compile.test.cpp
```diff
@@ -1,21 +1,29 @@
...
 
+#include "doctest.h"
 #include <cstring>
 #include <fstream>
+#include <memory>
 #include <string>
 #include <vector>
 
-TEST_CASE("lutepayload_single_file_roundtrip")
+struct CompileFixture : public LuteFixture
```
vrn-sn: We can just use LuteFixture if this doesn't need anything new?

tests/src/require.test.cpp
```diff
@@ -14,9 +15,13 @@ TEST_CASE_FIXTURE(CliRuntimeFixture, "require_exists")
     CHECK(!lua_isnil(L, -1));
 }
 
-TEST_CASE("require_modules")
+struct RequireFixture : public LuteFixture
```
vrn-sn: This could also just be deleted if LuteFixture is sufficient.

tests/src/staticrequires.test.cpp
```diff
@@ -4,16 +4,23 @@
...
 
+#include "lutefixture.h"
+
 #include <algorithm>
+#include <memory>
 #include <string>
 #include <vector>
 
-TEST_CASE("staticrequiretracer_simple_dependencies")
+struct StaticRequireTracerFixture : public LuteFixture
```
vrn-sn: Same here, I guess CliRuntimeFixture's the only one that really needs to specialize LuteFixture for now

tests/src/stdsystem.test.cpp
```diff
@@ -1,7 +1,7 @@
 #include "doctest.h"
 
 #include "cliruntimefixture.h"
-
+#include <iostream>
```
vrn-sn: Might have missed it, but is this used?

lute/cli/CMakeLists.txt
```diff
@@ -24,13 +24,16 @@ add_library(Lute.CLI.lib STATIC)
...
+    include/lute/clireporter.h
     include/lute/compile.h
     include/lute/luauflags.h
+    include/lute/reporter.h
     include/lute/staticrequires.h
     include/lute/tc.h
     include/lute/uvstate.h
 
     src/climain.cpp
+    src/clireporter.cpp
```
vrn-sn: This and its .h file belong under `Lute.CLI`.

tests/src/stdsystem.test.cpp
```diff
@@ -21,20 +21,25 @@ TEST_CASE_FIXTURE(CliRuntimeFixture, "std_system_os_matches_host_os")
...
-    auto checkBool = [&](const std::string& field, bool expected) {
-        runCode("local system = require(\"@std/system\")\n"
-                "capture(system." + field + ")\n");
-        std::string output = getCapturedOutput();
+    auto checkBool = [this](const std::string& field, bool expected)
+    {
+        std::string code = R"(
+            local system = require("@std/system")
+            report(system.)" + field + R"()
+        )";
```
vrn-sn: Sorry, I didn't realize that `field` was being added in the middle of this; my raw string suggestion is also confusing. Maybe `Luau::format` would be useful here?
```suggestion
        std::string code = Luau::format(R"(
            local system = require("@std/system")
            report(system.%s)
        )", field.c_str());
```

vrn-sn (approved):
Might be worth running clang-format one more time just to be safe since I merged in some new clang-format rules earlier today.

PR #534, opened 2025-11-06
docs: Add basic standard library docs to lute.luau.org

annieetang:
Docs docs docs

Extending existing regex parsing for `definitions/` to our standard library modules too! 

Helps users have better visibility into using `@std/`

This is a super rough/basic extension and it's not great (and the vitepress site pages were manually added to the config) but I think it's a good temporary step for improving our docs, especially as we're reaching Release A soon.

These are some of the generated pages:
image
image
image

### DIFF
--- a/docs/scripts/reference.luau
+++ b/docs/scripts/reference.luau
@@ -1,10 +1,11 @@
-local fs = require("@lute/fs")
-local process = require("@lute/process")
+local fs = require("@std/fs")
+local process = require("@std/process")
+local system = require("@std/system")
 
 local function projectRelative(...)
 	local cwd = process.cwd()
-	local pathSep = if string.find(string.lower(require("@lute/system").os), "windows") then "\\" else "/"
-	return table.concat({ cwd, "..", ... }, pathSep)
+	local sep = string.find(string.lower(system.os), "windows") and "\\" or "/"
+	return table.concat({ tostring(cwd), ... }, sep)
 end
 
 local function extractPropertySignature(module: string, line: string): (string?, string?)
@@ -26,6 +27,8 @@ local function extractPropertySignature(module: string, line: string): (string?,
 	return nil, nil
 end
 
+-- docs for definitions/*.luau --
+
 local function parseDefinitionFile(module: string, filePath: string): { { name: string, signature: string } }
 	local content = fs.readfiletostring(filePath)
 	local lines = string.split(content, "\n")
@@ -45,7 +48,10 @@ local function parseDefinitionFile(module: string, filePath: string): { { name: 
 	return definitions
 end
 
-local function generateMarkdown(moduleName: string, definitions: { { name: string, signature: string } }): string
+local function generateDefinitionsMarkdown(
+	moduleName: string,
+	definitions: { { name: string, signature: string } }
+): string
 	local lines = {
 		"# " .. moduleName,
 		"",
@@ -60,7 +66,7 @@ local function generateMarkdown(moduleName: string, definitions: { { name: strin
 	end)
 
 	for _, def in definitions do
-		table.insert(lines, "## " .. def.name)
+		table.insert(lines, "## " .. moduleName .. "." .. def.name)
 		table.insert(lines, "```luau")
 		table.insert(lines, def.signature)
 		table.insert(lines, "```")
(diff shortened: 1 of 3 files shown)


docs/scripts/reference.luau
```diff
@@ -1,10 +1,11 @@
...
-local process = require("@lute/process")
+local fs = require("@std/fs")
+local process = require("@std/process")
+local system = require("@std/system")
 
 local function projectRelative(...)
 	local cwd = process.cwd()
-	local pathSep = if string.find(string.lower(require("@lute/system").os), "windows") then "\\" else "/"
-	return table.concat({ cwd, "..", ... }, pathSep)
+	local sep = string.find(string.lower(system.os), "windows") and "\\" or "/"
```
aatxe:
```suggestion
	local sep = if string.find(string.lower(system.os), "windows") then "\\" else "/"
```

Do not use and/or for if expressions.

docs/scripts/reference.luau
```diff
@@ -60,7 +66,7 @@ local function generateMarkdown(moduleName: string, definitions: { { name: strin
 	end)
 
 	for _, def in definitions do
-		table.insert(lines, "## " .. def.name)
+		table.insert(lines, "## " .. moduleName .. "." .. def.name)
```
Vighnesh-V: if it's more convenient you can also rewrite this as:
```
`##{moduleName}.{def.name}`
```

docs/scripts/reference.luau
```diff
@@ -1,10 +1,11 @@
...
-local process = require("@lute/process")
+local fs = require("@std/fs")
+local process = require("@std/process")
+local system = require("@std/system")
 
 local function projectRelative(...)
 	local cwd = process.cwd()
-	local pathSep = if string.find(string.lower(require("@lute/system").os), "windows") then "\\" else "/"
-	return table.concat({ cwd, "..", ... }, pathSep)
+	local sep = if string.find(string.lower(system.os), "windows") then "\\" else "/"
```
aatxe:
```suggestion
	local sep = if system.win32 then "\\" else "/"
```

docs/scripts/reference.luau
```diff
@@ -70,28 +76,134 @@ local function generateMarkdown(moduleName: string, definitions: { { name: strin
...
+	local referencePath = projectRelative("lute/std/libs/")
+	local relativePath = string.sub(stdlibFilePath, #referencePath + 1)
+	relativePath = string.gsub(relativePath, "%.luau$", "")
+	local modulePath = string.gsub(relativePath, "/init$", "")
+
+	local lines = {
+		"# " .. moduleName,
+		"",
+		"```luau",
+		"local " .. moduleName .. ' = require("@std/' .. modulePath .. '")',
```
aatxe: Could also be an interpolated string, but doesn't have to be.

docs/scripts/reference.luau
```diff
@@ -70,28 +76,134 @@ local function generateMarkdown(moduleName: string, definitions: { { name: strin
...
+	definitions: { { name: string, signature: string } },
+	stdlibFilePath: string
+): string
+	local referencePath = projectRelative("lute/std/libs/")
+	local relativePath = string.sub(stdlibFilePath, #referencePath + 1)
+	relativePath = string.gsub(relativePath, "%.luau$", "")
+	local modulePath = string.gsub(relativePath, "/init$", "")
+
+	local lines = {
+		"# " .. moduleName,
```
aatxe:
```suggestion
		`# {moduleName}`,
```

docs/scripts/reference.luau
```diff
@@ -70,28 +76,134 @@ local function generateMarkdown(moduleName: string, definitions: { { name: strin
...
+end
+
+local function processStdlibDir(sourceDir: string, referenceDir: string)
+	pcall(fs.createdirectory, referenceDir)
+
+	local entries = fs.listdir(sourceDir)
+
+	for _, entry in entries do
+		local sourcePath = sourceDir .. "/" .. entry.name
+		local referencePath = referenceDir .. "/" .. entry.name
```
aatxe:
```suggestion
		local sourcePath = path.join(sourceDir, entry.name)
		local referencePath = path.join(referenceDir, entry.name)
```

docs/scripts/reference.luau
aatxe: but also, you should just be able to kill this entirely if you just use the path library everywhere which would be better.

aatxe (changes requested):

aatxe (approved):

PR #540, opened 2025-11-07
Fix infinite recursion in ignore

wmccrthy:
`ignore` was recently ported over from a different repo, that used an adhoc `path` implementation. This repo's `path.dirname` returned a string or nil, but lute's `path.dirname` always returns a string. 

On various occasions in ``ignore`` ([ one example here](https://github.com/luau-lang/lute/blob/primary/lute/cli/commands/transform/lib/ignore.luau#L226-L227)), we gate continued processing behind checks that `path.dirname(arbitrary)` exists. These are all invalid checks now and in some of these cases, we recur leading to death by stack overflow.

This PR fixes aforementioned checks, using `path.basename` (returns string | nil if at root) to determine if the current path is the root.

I found this bug when encountering a stack overflow error while trying to use `lute transform`:
```
_readIgnoreRecursive
... (+19979 frames)
@cli/transform/lib/ignore.luau:227 function _readIgnoreRecursive
@cli/transform/lib/ignore.luau:227 function _readIgnoreRecursive
@cli/transform/lib/ignore.luau:227 function _readIgnoreRecursive
@cli/transform/lib/ignore.luau:227 function _readIgnoreRecursive
@cli/transform/lib/ignore.luau:227 function _readIgnoreRecursive
@cli/transform/lib/ignore.luau:214 function isIgnoredDirectory
@cli/transform/lib/files.luau:16 function traverseDirectoryRecursive
@cli/transform/lib/files.luau:52 function getSourceFiles
@cli/transform/init.luau:139 function main
@cli/transform/init.luau:153
```

I rebuilt lute with these changes and `transform` worked as expected.

### DIFF
--- a/lute/cli/commands/transform/lib/ignore.luau
+++ b/lute/cli/commands/transform/lib/ignore.luau
@@ -143,7 +143,8 @@ end
 local function parseGitignore(filepath: path.path): GitignoreData
 	local contents = fs.readfiletostring(path.format(filepath))
 	local parentDirectory = if path.basename(filepath) then path.dirname(filepath) else nil
-	return parseGitignoreContents(assert(parentDirectory), contents)
+	assert(parentDirectory)
+	return parseGitignoreContents(parentDirectory, contents)
 end
 
 local function matchesGlob(glob: Glob, filepath: string, isDirectory: boolean): boolean
--- a/lute/cli/commands/transform/lib/ignore.luau
+++ b/lute/cli/commands/transform/lib/ignore.luau
@@ -142,7 +142,7 @@ end
 
 local function parseGitignore(filepath: path.path): GitignoreData
 	local contents = fs.readfiletostring(path.format(filepath))
-	local parentDirectory = if path.basename(filepath) then path.dirname(filepath) else nil
+	local parentDirectory = if #filepath.parts > 0 then path.dirname(filepath) else nil
 	assert(parentDirectory)
 	return parseGitignoreContents(parentDirectory, contents)
 end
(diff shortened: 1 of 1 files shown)


lute/cli/commands/transform/lib/ignore.luau
```diff
@@ -142,7 +142,8 @@ end
 
 local function parseGitignore(filepath: path.path): GitignoreData
 	local contents = fs.readfiletostring(path.format(filepath))
-	return parseGitignoreContents(assert(path.dirname(filepath)), contents)
+	local parentDirectory = if path.basename(filepath) then path.dirname(filepath) else nil
+	return parseGitignoreContents(assert(parentDirectory), contents)
```
Vighnesh-V: nit, can we pull this assertion out?

Vighnesh-V (changes requested):
small change, but otherwise looks good to me.

Vighnesh-V (approved):

lute/cli/commands/transform/lib/ignore.luau
```diff
@@ -142,7 +142,9 @@ end
 
 local function parseGitignore(filepath: path.path): GitignoreData
 	local contents = fs.readfiletostring(path.format(filepath))
-	return parseGitignoreContents(assert(path.dirname(filepath)), contents)
+	local parentDirectory = if path.basename(filepath) then path.dirname(filepath) else nil
```
skberkeley: we have `path.path` objects, so I believe the following would be more performant:
`if #path.parts > 0 then path.dirname(filepath) else nil`

wmccrthy: If `filePath` is always a table here, won't `path.basename` mirror this exact logic (non-nil if `#path.parts > 0` and nil otherwise)? Happy to change but curious what the performance bottleneck would be

skberkeley: yeah, the logic will be the same, but we avoid the table check and passing the string basename around this way

skberkeley (approved):

Nicell (approved):

PR #548, opened 2025-11-10
Remove infinite loop test for .config.luau

### LINKED ISSUE #1246: luau-lsp 1.56.0 hangs when resolving `.config.luau` files that do not complete
Hello!

We've had at least 3 folks on the Luau team encounter a hang with the `luau-lsp` extension indexing files while working on [lute](https://github.com/luau-lang/lute). The workaround we've found is to pin the extension to `1.55`, suggesting that something changed with `1.56.0` to break things.

Image

vrn-sn:
We already have a test for this in the Luau repository, and it's just causing issues for luau-lsp here. If there ever is a bug with this functionality, it needs to be fixed in the Luau repository anyway, so I think it's okay to remove from here.

I was considering replacing it with a test that created the files in memory, but it seems like too much hassle for not much benefit.

Resolves https://github.com/JohnnyMorganz/luau-lsp/issues/1246 on our end.

### DIFF
--- a/tests/src/require.test.cpp
+++ b/tests/src/require.test.cpp
@@ -142,13 +142,6 @@ TEST_CASE("require_modules")
             );
         }
 
-        SUBCASE("config_luau_timeout")
-        {
-            doFailingSubcase(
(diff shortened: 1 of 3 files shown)


annieetang (approved):
ty 🙏

PR #550, opened 2025-11-10
site: fix path in sidebar generation

annieetang:
Sidebar generation is not resolving the paths correctly for the docs

### DIFF
--- a/docs/.vitepress/config.mts
+++ b/docs/.vitepress/config.mts
@@ -20,8 +20,8 @@ export default withSidebar(
   {
     // ============ [ SIDEBAR OPTIONS ] ============
     // ============ [ RESOLVING PATHS ] ============
-    documentRootPath: './',
-    scanStartPath: './reference/',
+    documentRootPath: '/',
+    scanStartPath: 'reference',
     // ============ [ GROUPING ] ============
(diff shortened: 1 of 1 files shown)


docs/.vitepress/config.mts
```diff
@@ -20,8 +20,8 @@ export default withSidebar(
   {
     // ============ [ SIDEBAR OPTIONS ] ============
     // ============ [ RESOLVING PATHS ] ============
-    documentRootPath: './',
-    scanStartPath: './reference/',
+    documentRootPath: '/',
```
skberkeley: wouldn't this point to the system root dir?

annieetang: but the gh workflow runs in the `docs/` repo

annieetang:
ok I figured out how to test locally and specifying the sidebar paths were messing with the original vitepress path that just handles everything in `base: "/"`
so just using the default sidebar paths works

image

skberkeley (approved):

annieetang:
talked with sora and we're thinking of refactoring things inside `docs/` to include a `docs/docs/` where the markdown files that we want to include on the site will all live there, because this current solution means _any_ `.md` file will be included on the site if it's in this `docs/` repo.

Issue #161, opened 2025-04-21
API: Cryptography


ffrostfall:
We could go with `blake3` and `sha256` for starters?

Perhaps something like `local hashed_stuff = <apinamehere>.hash("blake3", buff)`

aatxe:
Going to hijack this to be the issue that it actually should be.

aatxe:
We should expose high-level cryptography primitives from our SSL library. I want to switch to BoringSSL before we do that though because BoringSSL uses verified cryptography primitives.

aatxe:
Besides `digest`, we should expose `hmac` and `pwhash` at the bare minimum for now. `aead` would be the next priority after that.

aatxe:
My philosophy changed a little bit, but I think we can start with digest, pwhash, and secretbox (as an AE API). We can evolve this going forward as needed, but I'm going to call the initial work here complete now.

PR #552, opened 2025-11-10
crypto: implement a secretbox interface for authenticated encryption

aatxe:
The last remaining bit of work to close out #161 was the desire to have some kind of API for authenticated encryption (AE) or authenticated encrypted with additional data (AEAD). I worked a bit on trying to provide the full AEAD interface from boringssl, but it felt clunky and bad to use in Luau. I don't want to commit to the runtime providing a standard interface that is literally reproducing openssl's headers, that's something that can be done later with FFI. Instead, I want the crypto library to feel more in the spirit of libsodium, providing opinionated default ways of solving common cryptography tasks. The secretbox interface is a well-understood metaphor for AE(AD), and one that works quite nicely. To mitigate the likelihood of nonce or key reuse, the interface provided right now seals a message providing the key and nonce as output. We can extend the API in the future with optional arguments for nonce and key as needed.

### DIFF
--- a/lute/crypto/src/crypto.cpp
+++ b/lute/crypto/src/crypto.cpp
@@ -144,14 +144,21 @@ int lua_secretbox_open(lua_State* L)
     lua_getfield(L, 1, "nonce");
     lua_getfield(L, 1, "key");
 
-    BinaryData ciphertext = extractData(L, 2);
-    size_t nonceLength = crypto_secretbox_keybytes();
-    uint8_t* nonce = static_cast<uint8_t*>(luaL_checkbuffer(L, 3, &nonceLength));
-    size_t keyLength = crypto_secretbox_keybytes();
-    uint8_t* key = static_cast<uint8_t*>(luaL_checkbuffer(L, 4, &keyLength));
+    size_t ciphertextLength = 0;
+    uint8_t* ciphertext = static_cast<uint8_t*>(luaL_checkbuffer(L, 2, &ciphertextLength));
 
-    uint8_t* buffer = static_cast<uint8_t*>(lua_newbuffer(L, ciphertext.length - crypto_secretbox_macbytes()));
-    if (crypto_secretbox_open_easy(buffer, static_cast<const uint8_t*>(ciphertext.data), ciphertext.length, nonce, key) != 0)
+    size_t nonceLength = 0;
+    uint8_t* nonce = static_cast<uint8_t*>(luaL_checkbuffer(L, 3, &nonceLength));
+    if (nonceLength != crypto_secretbox_noncebytes())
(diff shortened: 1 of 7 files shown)


lute/crypto/src/crypto.cpp
```diff
@@ -89,13 +91,89 @@ int lua_digest(lua_State* L)
...
+    // read all three of the required fields out of the table
+    lua_getfield(L, 1, "ciphertext");
+    lua_getfield(L, 1, "nonce");
+    lua_getfield(L, 1, "key");
+
+    BinaryData ciphertext = extractData(L, 2);
+    size_t nonceLength = crypto_secretbox_keybytes();
+    uint8_t* nonce = static_cast<uint8_t*>(luaL_checkbuffer(L, 3, &nonceLength));
+    size_t keyLength = crypto_secretbox_keybytes();
+    uint8_t* key = static_cast<uint8_t*>(luaL_checkbuffer(L, 4, &keyLength));
```
vegorov-rbx: This will not be right.
The third parameter of `luaL_checkbuffer` is an 'out' parameter (like in `luaL_checklstring`).

```suggestion
    size_t nonceLength = 0;
    uint8_t* nonce = static_cast<uint8_t*>(luaL_checkbuffer(L, 3, &nonceLength));
    if (nonceLength != crypto_secretbox_noncebytes())
        luaL_error(L, "%s: nonce buffer should be %d bytes", kOpenName, int(crypto_secretbox_noncebytes()));
    size_t keyLength = 0;
    uint8_t* key = static_cast<uint8_t*>(luaL_checkbuffer(L, 4, &keyLength));
    if (keyLength != crypto_secretbox_keybytes())
        luaL_error(L, "%s: key buffer should be %d bytes", kOpenName, int(crypto_secretbox_keybytes()));
```
Of course we can also require 'at least' instead of 'exact'.

(side note, `crypto_secretbox_keybytes` was used in place of `crypto_secretbox_noncebytes` before)

lute/crypto/src/crypto.cpp
```diff
@@ -89,13 +91,89 @@ int lua_digest(lua_State* L)
...
+    int argumentCount = lua_gettop(L);
+    if (argumentCount != 1)
+        luaL_error(L, "%s: expected 1 argument, but got %d", kOpenName, argumentCount);
+
+    // read all three of the required fields out of the table
+    lua_getfield(L, 1, "ciphertext");
+    lua_getfield(L, 1, "nonce");
+    lua_getfield(L, 1, "key");
+
+    BinaryData ciphertext = extractData(L, 2);
```
vegorov-rbx: This allows `ciphertext` to be a string as well.

lute/crypto/src/crypto.cpp
aatxe: good catch!

vegorov-rbx (approved):

PR #553, opened 2025-11-10
cli: update tc.cpp to use Luau resolvers

### LINKED ISSUE #541: Update LuteConfigResolver in `tc.cpp` to use `Luau::LuteConfigResolver`
cleanup from https://github.com/luau-lang/lute/pull/535

annieetang:
resolves https://github.com/luau-lang/lute/issues/541
I also updated the `LuteFileResolver` to extend `Luau::LuteModuleResolver` since that implementation is built on `Luau::FileResolver` and handles requires.

There are already tests in `cli` that call some of these paths... but lmk if there are more we want to add for this PR

### DIFF
--- a/lute/cli/src/tc.cpp
+++ b/lute/cli/src/tc.cpp
@@ -5,6 +5,9 @@
 #include "Luau/PrettyPrinter.h"
 #include "Luau/TypeAttach.h"
 
+#include "lute/configresolver.h"
+#include "lute/moduleresolver.h"
+
 static const std::string kLuteDefinitions = R"LUTE_TYPES(
 -- Net api
 declare net: {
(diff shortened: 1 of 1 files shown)


lute/cli/src/tc.cpp
```diff
@@ -30,7 +33,7 @@ declare function spawn(path: string): any
 
 )LUTE_TYPES";
 
-struct LuteFileResolver : Luau::FileResolver
+struct LuteFileResolver : Luau::LuteModuleResolver
```
vrn-sn, line 36: Question: is there anything stopping us from using `Luau::LuteModuleResolver` directly in this file?

annieetang: `Luau::LuteModuleResolver` doesn't handle stdin right now, but this `LuteFileResolver` does. I can update `Luau::LuteModuleResolver` to also handle stdin in this PR too if you'd like

vrn-sn: I think it's fine in that case.

vrn-sn (approved):

PR #554, opened 2025-11-10
bugfix: windows path api needs to be able to get the drive letter for a path

Vighnesh-V:
There are a number of tests failing on windows runners because the github runners execute tests under the D:\ drive. However, many of our tests assume execution under the C drive. This PR adds a function called `getdrive(path)` to the win32 path api, as well as fixes some of those tests to be drive aware.

### DIFF
--- a/lute/std/libs/path/win32/init.luau
+++ b/lute/std/libs/path/win32/init.luau
@@ -59,7 +59,7 @@ function win32.extname(path: pathlike): string
 	return filename:sub(dotIndex)
 end
 
-function win32.getdrive(path: pathlike): path
+function win32.drive(path: pathlike): path
 	if typeof(path) == "string" then
 		path = win32.parse(path)
 	end
--- a/lute/std/libs/path/win32/init.luau
+++ b/lute/std/libs/path/win32/init.luau
@@ -59,6 +59,24 @@ function win32.extname(path: pathlike): string
 	return filename:sub(dotIndex)
 end
 
+function win32.getdrive(path: pathlike): path
+	if typeof(path) == "string" then
+		path = win32.parse(path)
+	end
+
+	path = path :: path
+
+	if path.driveLetter == nil then
+		error("Path does not have a drive letter: " .. win32.format(path))
+	end
+
+	return setmetatable({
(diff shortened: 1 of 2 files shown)


lute/std/libs/path/win32/init.luau
```diff
@@ -59,6 +59,24 @@ function win32.extname(path: pathlike): string
 	return filename:sub(dotIndex)
 end
 
+function win32.getdrive(path: pathlike): path
```
aatxe:
```suggestion
function win32.drive(path: pathlike): path
```

tests/std/path/path.win32.test.luau
```diff
@@ -655,10 +697,16 @@ test.suite("PathWin32ResolveSuite", function(suite)
 	end)
 
 	suite:case("resolve_drive_relative_path", function(assert)
-		local result = path.win32.resolve("C:Documents\\file.txt")
+		-- Get the drive letter from the current working directory
+		local cwd = process.cwd()
+		local cwdPath = path.win32.parse(cwd)
+		local driveLetter = if system.win32 then path.win32.drive(cwdPath) else "C"
+
+		local testPath = driveLetter .. ":Documents\\file.txt"
```
skberkeley: won't this error on windows?
relatedly, why doesn't `path.win32.drive` return a letter instead of a full path?

Vighnesh-V: 1) yup, this errors - I need to tweak this line.
2) I picked returning a path because it seems kind of in line for the path library. Drive letter is just a field on the path table anyway.

aatxe: part of the reason i said `drive` and not `driveletter` or something was that it returns a path. it feels like the "drive" being conceptually "the root path for that drive" is right.

Vighnesh-V: I've gone ahead and fixed the indexing issue. This should work correctly now, and I've checked that this test now passes CI for windows machines : https://github.com/luau-lang/lute/actions/runs/19249569625/job/55031444805

skberkeley (approved):

PR #555, opened 2025-11-10
bugfix: createdirectory creates the wrong path on windows machines

Vighnesh-V:
The `fs.createdirectory` api creates intermediate folders by incrementally creating each subportion of a path. When the passed in path is an absolute path, on posix systems we were creating a path that corresponds to `/`. However, on Windows systems, we were leaving off the drive letter, which means that these paths were treated as relative. E.g. instead of creating: `C:\\some\absolute\path`, we were creating `C:some\absolute\path`, relative to the `cwd`. This caused a number of file systems tests to fail when creating these directories, because they were being created in wrong place.

### DIFF
--- a/lute/std/libs/fs.luau
+++ b/lute/std/libs/fs.luau
@@ -5,6 +5,7 @@
 
 local fs = require("@lute/fs")
 local pathlib = require("@std/path")
+local sys = require("@std/system")
 
 local fslib = {}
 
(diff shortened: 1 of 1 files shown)


annieetang:
thanks for the catch! dyt you could add a small test for the posix and win32 paths?

Vighnesh-V:
This job should have all the tests in `fs.test.luau` passing:
https://github.com/luau-lang/lute/actions/runs/19251485507/job/55037226773

vrn-sn (approved):

lute/std/libs/fs.luau
```diff
@@ -91,11 +92,18 @@ end
...
+		local parsed = pathlib.parse(path)
+		local parts: { string } = parsed.parts
 
 		local subdir: pathlike = "."
 		if pathlib.isabsolute(path) then
-			subdir = pathlib.format({ absolute = true, parts = {} })
+			if sys.win32 then
+				-- Windows path - get the drive root like "C:\"
+				subdir = pathlib.win32.drive(parsed)
+			else
```
annieetang, line 103: Should we maybe do else if posix here? Just in case there's some crazy case where it's neither

vrn-sn: I think if someone reports an issue with this, we can add it, but I don't even know what a good value for `subdir` would be in that case (if we were to do it now).

annieetang: Yea that's fair! I would suppose an error or something saying we don't support but this is good with me

PR #556, opened 2025-11-11
lute transform: pass parsed ASTs to codemods

skberkeley:
The existing API of `lute transform` requires that codemod authors call the parser themselves on the source code. Almost all codemod writers are going to operate on the AST anyway, so parse it for them and pass it in the context passed to the transforms. I updated `examples/transformer.luau`, which is used in the existing test for `lute transform`.


hgoldstein (approved):
This seems fine

PR #557, opened 2025-11-11
feature: Lute test runs all the tests

Vighnesh-V:
Run all the tests in your project via:
```
lute test
```

Run some of the tests in a directory via:
```
lute test filter/path/directory
```

### DIFF
--- a/tests/cli/test.test.luau
+++ b/tests/cli/test.test.luau
@@ -26,11 +26,11 @@ test.suite("lute test", function(suite)
 		assert.neq(result.stdout:find("Usage:", 1, true), nil)
 	end)
 
-	suite:case("lute test runs successfully", function(assert)
-		-- Run lute test without arguments
-		local result = process.run({ lutePath, "test" })
+	suite:case("lute test runs tests in discovery folder", function(assert)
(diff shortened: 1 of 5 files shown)


tests/cli/test.test.luau
```diff
@@ -44,11 +44,11 @@ test.suite("lute test", function(suite)
 		assert.strcontains(result.stdout, "Usage:")
 	end)
 
-	suite:case("lute test runs successfully", function(assert)
-		-- Run lute test without arguments
-		local result = process.run({ lutePath, "test" })
+	suite:case("lute test runs tests in discovery folder", function(assert)
+		-- Run lute test on tests/cli/discovery folder
```
Vighnesh-V, line 48: if we dont do this, lute test calls itself recursively.

annieetang (approved):
could u make a note about why we aren't using this in CI for now? or a follow up issue for it

Vighnesh-V:
Created a follow up issue for running this in CI here: https://github.com/luau-lang/lute/issues/634. I'm going to spend some time tomorrow digging into this to figure out exactly why the @batteries require fails with `lute test` and update the issue, but for now, we just print an error and skip over that test.

PR #558, opened 2025-11-11
Implements flag arguments to specify test case and test suite to run for `lute test`

Vighnesh-V:
This PR adds support for lute test to run specific tests:

```
lute test --suite SomeSuite
lute test -s SomeSuite -- runs SomeSuite
lute test --case some_case 
lute test -c some_case -- runs some_case

lute test -s Suite -c case -- runs Suite.case
lute test -s "ABC" -c "some test with spaces" -- runs `ABC.some test with spaces`
```

### DIFF
--- a/lute/cli/commands/test/filter.luau
+++ b/lute/cli/commands/test/filter.luau
@@ -17,8 +17,35 @@ local function filtertests(
 		return env
 	end
 
+	-- Filter by both suite and case name
+	if suiteName and caseName then
+		local caseEntry = env.caseindex[caseName]
+		if caseEntry and caseEntry.suites[suiteName] then
+			local suite = env.suiteindex[suiteName]
+			local filteredSuite = {
+				name = suite.name,
+				cases = caseEntry.suites[suiteName],
+				_beforeeach = suite._beforeeach,
+				_beforeall = suite._beforeall,
+				_aftereach = suite._aftereach,
+				_afterall = suite._afterall,
+			}
+			table.insert(filtered.suites, filteredSuite)
+		end
+		return filtered
+	end
+
+	-- Filter by suite name only
+	if suiteName then
+		local suite = env.suiteindex[suiteName]
+		if suite then
+			table.insert(filtered.suites, suite)
+		end
+		return filtered
+	end
+
 	-- Filter by case name only
-	if caseName and not suiteName then
+	if caseName then
 		local caseEntry = env.caseindex[caseName]
 		if caseEntry then
 			-- Add all anonymous tests with this case name
@@ -37,33 +64,6 @@ local function filtertests(
 				}
 				table.insert(filtered.suites, filteredSuite)
 			end
-		end
-		return filtered
-	end
-
-	-- Filter by suite name only
-	if suiteName and not caseName then
-		local suite = env.suiteindex[suiteName]
-		if suite then
-			table.insert(filtered.suites, suite)
-		end
-		return filtered
-	end
-
-	-- Filter by both suite and case name
-	if suiteName and caseName then
-		local caseEntry = env.caseindex[caseName]
-		if caseEntry and caseEntry.suites[suiteName] then
-			local suite = env.suiteindex[suiteName]
-			local filteredSuite = {
-				name = suite.name,
-				cases = caseEntry.suites[suiteName],
-				_beforeeach = suite._beforeeach,
-				_beforeall = suite._beforeall,
-				_aftereach = suite._aftereach,
-				_afterall = suite._afterall,
-			}
-			table.insert(filtered.suites, filteredSuite)
 		end
 		return filtered
 	end
--- a/lute/cli/commands/test/filter.luau
+++ b/lute/cli/commands/test/filter.luau
@@ -1,0 +1,74 @@
+local testtypes = require("@std/test/types")
+
+local function filtertests(
+	env: testtypes.testenvironment,
+	suiteName: string?,
+	caseName: string?
+): testtypes.testenvironment
+	local filtered: testtypes.testenvironment = {
+		anonymous = {},
+		suites = {},
+		suiteindex = {},
+		caseindex = {},
+	}
+
+	-- No filters: return all tests
+	if not suiteName and not caseName then
+		return env
+	end
+
+	-- Filter by case name only
+	if caseName and not suiteName then
+		local caseEntry = env.caseindex[caseName]
+		if caseEntry then
+			-- Add all anonymous tests with this case name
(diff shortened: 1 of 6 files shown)


lute/cli/commands/test/init.luau
```diff
@@ -16,14 +17,10 @@ Run tests discovered in .test.luau and .spec.luau files.
...
   --list                  List all discovered test cases without running them
-  --filter=PATTERN        Run only tests matching PATTERN
-                          Examples:
-                            --filter=SuiteName           (run all tests in suite)
-                            --filter=SuiteName.testname  (run specific test)
-                            --filter=testname            (run anonymous test)
-  --reporter=REPORTER     Choose test reporter (default: rich)
+  -s, --suite=SUITE       Run only tests in the specified suite
+  -c, --case=CASE         Run only test cases matching the specified name
+  --reporter=REPORTER     Choose test reporter (default: simple)
```
skberkeley: should we get rid of this line for now? you don't appear to be parsing the reporter argument at all

lute/cli/commands/test/init.luau
```diff
@@ -102,25 +101,52 @@ local function main(...: string)
 	local args = { ... }
 	local testPaths = {}
 	local isListMode = false
+	local suiteName: string? = nil
+	local caseName: string? = nil
```
skberkeley, line 101: at some point we should probably move this (and transform) onto the cli battery

lute/cli/commands/test/finder.luau
```diff
@@ -8,37 +8,22 @@ local function istestfile(filename: string): boolean
...
+	local function findfiles(p: path.pathlike)
+		local it = fs.walk(path.normalize(p), { recursive = true })
+		local file = it()
+		while file do
+			if fs.type(file) ~= "dir" and istestfile(path.format(file)) then
+				table.insert(files, path.normalize(file))
+			end
+			file = it()
 		end
+	end
```
skberkeley, line 22: why is this a nested function? it's a little less readable

lute/cli/commands/test/filter.luau
```diff
@@ -0,0 +1,74 @@
...
+		caseindex = {},
+	}
+
+	-- No filters: return all tests
+	if not suiteName and not caseName then
+		return env
+	end
+
+	-- Filter by case name only
+	if caseName and not suiteName then
```
skberkeley: these conditions are simpler if you do `suiteName and caseName` first, then the subsequent ifs can just be `if suiteName ...` and `if caseName ...`

lute/std/libs/test/init.luau
```diff
@@ -49,31 +49,52 @@ function TestSuite:afterall(hook: () -> ())
...
 -- register an anonymous test case
 function test.case(name: string, case: Test)
 	local testcase = { name = name, case = case }
 	table.insert(env.anonymous, testcase)
+
+	-- Update caseindex
+	if not env.caseindex[name] then
+		env.caseindex[name] = { anonymous = {}, suites = {} }
+	end
+	table.insert(env.caseindex[name].anonymous, testcase)
```
skberkeley:
```suggestion
	if not env.caseindex[name] then
		env.caseindex[name] = { anonymous = { testcase }, suites = {} }
	end
```

lute/std/libs/test/init.luau
```diff
@@ -49,31 +49,52 @@ function TestSuite:afterall(hook: () -> ())
...
+
+	-- Update caseindex for all cases in this suite
+	for _, testcase in suite.cases do
+		if not env.caseindex[testcase.name] then
+			env.caseindex[testcase.name] = { anonymous = {}, suites = {} }
+		end
+		if not env.caseindex[testcase.name].suites[name] then
+			env.caseindex[testcase.name].suites[name] = {}
+		end
+		table.insert(env.caseindex[testcase.name].suites[name], testcase)
```
skberkeley, line 87:
```suggestion
		if not env.caseindex[testcase.name] then
			env.caseindex[testcase.name] = { anonymous = { testcase }, suites = {} }
		end
		if not env.caseindex[testcase.name].suites[name] then
			env.caseindex[testcase.name].suites[name] = {}
		end
```

lute/cli/commands/test/finder.luau
Vighnesh-V: Nested because the command supports multiple test directories for discovery - e.g dir1 dir2 etc. So I'm basically doing a `map findfiles dirs` and concatenating all the results together.

lute/std/libs/test/init.luau
Vighnesh-V: I think you still need this line: `table.insert(env.caseindex[testcase.name].suites[name], testcase)`.
This isn't correct because this maps testcase names -> suitenames -> testcases with that name in those suites. If neither of:

```
if not env.caseindex[testcase.name] then
...
if not env.caseindex[testcase.name].suites[name] then
```
end up being true, you still need to add the test case to `env.caseindex[testcase.name].suites` (in this model, it's totally okay for a suite to have many test cases with the same name).

lute/cli/commands/test/filter.luau
```diff
@@ -0,0 +1,74 @@
...
+				_aftereach = suite._aftereach,
+				_afterall = suite._afterall,
+			}
+			table.insert(filtered.suites, filteredSuite)
+		end
+		return filtered
+	end
+
+	-- Filter by suite name only
+	if suiteName then
```
skberkeley: i think these need to be `elseif`s now, otherwise we'll over count when we specify `suiteName` + caseName`?

lute/cli/commands/test/filter.luau
```diff
@@ -0,0 +1,74 @@
...
+			local suite = env.suiteindex[suiteName]
+			local filteredSuite = {
+				name = suite.name,
+				cases = caseEntry.suites[suiteName],
+				_beforeeach = suite._beforeeach,
+				_beforeall = suite._beforeall,
+				_aftereach = suite._aftereach,
+				_afterall = suite._afterall,
+			}
+			table.insert(filtered.suites, filteredSuite)
```
skberkeley: wrap in `if suite`?

skberkeley (changes requested):
just two quick comments

lute/cli/commands/test/filter.luau
Vighnesh-V: To clarify, you're suggesting:
```
if suite name and case name then
elseif suitename
...
elseif casename
...
end

return filtered
```
right?

skberkeley: yup!

skberkeley (approved):

PR #559, opened 2025-11-11
perf: base64 decoder with buffer-based LUT and stricter validation

zoon:
## Summary
- Replace table-based decoding LUT with buffer for faster lookups
- Switch from file-scope `--!native` to function-level `@native` attributes
- Implement RFC 4648-compliant padding handling (strip instead of count)
- Add input validation to reject invalid base64 characters
- Validate non-zero padding bits per RFC 4648 section 3.5
- Optimize decode loop using u32 writes

## Test coverage
- RFC 4648 test vectors (encode/decode)
- Invalid character detection
- Invalid padding/length cases

## Performance
```text
> luau --codegen -O2
-- Output size 1 KB
   old base64.decode     10.61 us/iter
-> new base64.decode      7.31 us/iter
-- Output size 1024 KB
   old base64.decode     10.51 ms/iter
-> new base64.decode      6.80 ms/iter

> luau -O2
-- Output size 1 KB
   old base64.decode     45.48 us/iter
-> new base64.decode     36.32 us/iter
-- Output size 1024 KB
   old base64.decode     47.37 ms/iter
-> new base64.decode     34.48 ms/iter
```

### DIFF
--- a/batteries/base64.luau
+++ b/batteries/base64.luau
@@ -112,7 +112,7 @@ local function decode(input_buffer: buffer): buffer
 			error("Invalid base64 input", 2)
 		end
 		read_offset += 4
-		-- u32 BE: [B1, B2, B3, 0] = b1<<26|b2<<20|b3<<14|b4<<8, trailing 0 will be ovewriten next iteration
+		-- u32 BE: [B1, B2, B3, 0] = b1<<26|b2<<20|b3<<14|b4<<8, trailing 0 will be overwritten next iteration
 		buffer.writeu32(output, write_offset, bit32.byteswap(b1 * 0x4000000 + b2 * 0x100000 + b3 * 0x4000 + b4 * 0x100))
 		write_offset += 3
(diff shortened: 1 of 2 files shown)


batteries/base64.luau
```diff
@@ -1,10 +1,12 @@
---!native
 --!optimize 2
 --!strict
 
+-- assume little-endian system
+assert(string.pack("=i2", 1) == "\1\0", "system must be little-endian")
```
bjcscat: This isn't required, buffers are always LE

bjcscat (approved):
Looks good. Buffer performance is generally worse than table accesses for non-native environments but given that most environments where I assume this will be running will have support for NCG I think its fine to make this change.

batteries/base64.luau
zoon: thank you, fixed

aatxe (approved):

PR #560, opened 2025-11-11
bugfix: remaining failing windows tests

Vighnesh-V:
This PR fixes the remaining windows tests that fail due to platform specific new line conventions. Inside of process.cpp, we now convert CRLF line endings to LF endings automatically.

Finally, some of tasks display diverging behaviour on git bash vs command prompt. I've update the job to run tests on `cmd` in windows and `bash` on posix systems.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,6 +5,7 @@ on:
     branches: ["primary"]
   pull_request:
     branches: ["primary"]
+  workflow_dispatch:
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
@@ -43,7 +44,8 @@ jobs:
         run: rm .luaurc && mv .luaurc.ci .luaurc
 
       - name: Run Luau Tests
-        run: find tests -name "*.test.luau" | xargs -I {} ${{ steps.build_lute.outputs.exe_path }} run {}
+        shell: bash
+        run: find tests -name "*.test.luau" | xargs -I {} $(echo '${{ steps.build_lute.outputs.exe_path }}' | tr '\\' '/') run {}
 
   run-lute-cxx-unittests:
     runs-on: ${{ matrix.os }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,7 +5,6 @@ on:
     branches: ["primary"]
   pull_request:
     branches: ["primary"]
-  workflow_dispatch:
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -5,6 +5,7 @@ on:
     branches: ["primary"]
   pull_request:
     branches: ["primary"]
+  workflow_dispatch:
```
aatxe:
```suggestion
```
Don't think you ended up using this.

tests/std/process.test.luau
```diff
@@ -20,7 +20,12 @@ test.suite("ProcessSuite", function(suite)
 	end)
 
 	suite:case("run_shell_and_cwd", function(assert)
-		local r = process.run({ "pwd" }, { cwd = "/" })
+		local rootdir: string = "/"
+		if system.win32 then
+			local root = pathlib.win32.drive(process.cwd())
+			rootdir = pathlib.format(root)
+		end
```
aatxe, line 27: Arguably, this suggests we should have `system.rootdir()`, but I think the version of that we'd get from libuv probably doesn't do the same thing as this (i.e. I think it'd give you the drive letter of your system install on Windows, not the drive letter of your working directory, but who knows, maybe it does!)

aatxe (approved):

lute/process/src/process.cpp
```diff
@@ -26,6 +26,15 @@
 namespace process
 {
 
+void convertCRLFtoLF(std::string& str)
+{
+      size_t pos = 0;
```
vrn-sn: I believe this can be implemented as a linear-time scan by writing in-place and resizing the string to the number of bytes written.

tests/std/process.test.luau
Vighnesh-V: yeah this would have to live in system I think. I don't think libuv has that method. I can add this to system in a followup.

Vighnesh-V: uv_os_homedir returns the result of %userprofile% or $HOME depending on which platform you're on. Maybe we could expose `system.currentroot` and `system.execroot` and `system.homeroot` depending on which path you care about?

lute/process/src/process.cpp
Vighnesh-V: yeah you right. I've tweaked this to run in linear time.

Vighnesh-V: I explicitly didn't use the std::erase pattern because I only want to convert ocurrences of `\r\n` to `\n`, not just delete all occurrences of `\r`

vrn-sn: Yeah, I was thinking basically what you have now.

vrn-sn (approved):

Issue #551, opened 2025-11-10
Optional + Union type is incorrectly treated as whitespace trivia in luau parser


Gravemind1142:
When parsing:
```luau
local function test(): number? | string
	return 1
end
```
The union token `|` is treated as whitespace trivia.

## Reproduction steps
Run this code snippet:
```luau
local luau = require("@lute/luau")

local code = [[
local function test(): number? | string
	return 1
end
]]

local ast = luau.parse(code)
-- Print every token (and trivia) present in the return type annotation of `test`
local function isToken(value: any): boolean
	return typeof(value) == "table" and value ~= nil and value.istoken == true
end

local function addTriviaPiece(pieces, triv)
	table.insert(pieces, {
		kind = "trivia",
		tag = triv.tag,
		text = triv.text,
		line = triv.location.begin.line,
		column = triv.location.begin.column,
	})
end

local function addTokenPieces(pieces, token, seenTrivia)
	-- leading trivia
	if token.leadingtrivia then
		for _, triv in ipairs(token.leadingtrivia) do
			local key = `L:{triv.location.begin.line}:{triv.location.begin.column}:{triv.tag}:{triv.text}`
			if not seenTrivia[key] then
				addTriviaPiece(pieces, triv)
				seenTrivia[key] = true
			end
		end
	end
	-- the token itself
	table.insert(pieces, {
		kind = "token",
		text = token.text,
		line = token.position.line,
		column = token.position.column,
	})
	-- trailing trivia
	if token.trailingtrivia then
		for _, triv in ipairs(token.trailingtrivia) do
			local key = `T:{triv.location.begin.line}:{triv.location.begin.column}:{triv.tag}:{triv.text}`
			if not seenTrivia[key] then
				addTriviaPiece(pieces, triv)
				seenTrivia[key] = true
			end
		end
	end
end

-- Generic traversal to collect tokens under a subtree (AstType/AstTypePack, etc.)
local function collectTokens(node, pieces, seenTrivia)
	if node == nil then
		return
	end
	if isToken(node) then
		addTokenPieces(pieces, node, seenTrivia)
		return
	end
	if typeof(node) ~= "table" then
		return
	end
	-- Skip plain location tables to avoid unnecessary traversal
	if node.begin and node["end"] then
		return
	end
	for key, value in pairs(node) do
		if key ~= "location" and key ~= "leadingtrivia" and key ~= "trailingtrivia" then
			if isToken(value) then
[...]

PR #563, opened 2025-11-11
lute parse: fix serialization of union type separators

skberkeley:
This PR fixes a bug where the `|` separator was not being correctly serialized when it followed an optional option of a union. For example, the `|` in the union type `number? | string` was not being correctly serialized, and was instead being consumed as the preceding trivia of the `string` option. We now _always_ try to consume the separator if serializing an optional option which isn't the last option of the union.

Added a test checking this functionality.

Addresses #551.

### DIFF
--- a/tests/std/syntax/parser.test.luau
+++ b/tests/std/syntax/parser.test.luau
@@ -969,7 +969,6 @@ test.suite("parse types", function(suite)
 		assert.eq(pair1.node.tag, "reference")
 		assert.eq(pair1.separator, nil)
 
-		-- Wack that optional types are two whole Ast nodes, but it's how it is on the C++ side too
 		local pair2 = unionType.types[2]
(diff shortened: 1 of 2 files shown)


tests/std/syntax/parser.test.luau
```diff
@@ -956,6 +956,30 @@ test.suite("parse types", function(suite)
...
+
+		local unionType = typeAlias.type :: luau.AstTypeUnion
+		assert.eq(unionType.leading, nil)
+		assert.eq(#unionType.types, 3)
+
+		local pair1 = unionType.types[1]
+		assert.eq(pair1.node.tag, "reference")
+		assert.eq(pair1.separator, nil)
+
+		-- Wack that optional types are two whole Ast nodes, but it's how it is on the C++ side too
```
aatxe:
```suggestion
```

aatxe (approved):

PR #565, opened 2025-11-12
lute std: Move AST-related functions and types from @std/luau to @std/syntax

skberkeley:
We previously had redundant parse functions across @std/syntax/parser and @std/luau. I talked this over with @~aatxe and we came to the conclusion that @std/syntax content should be AST-related, while @std/luau is reserved for runtime-related content.


lute/std/libs/syntax/init.luau
```diff
@@ -0,0 +1,161 @@
+local types = require("@self/types")
+
```
aatxe, line 3: I want this module to reexport `parser.parse` and `parser.parseexpr` at minimum, so we can do `syntax.parse` in some of the other code.

aatxe (changes requested):
Overall fine, but one change.

aatxe (approved):

PR #566, opened 2025-11-12
fix: fs.open should handle optional mode

annieetang:
per our [docs](https://lute.luau.org/reference/definitions/fs.html#fs-open), the `mode` argument in `fs.open` is an optional argument, but `fs.open(path)` was erroring with `invalid argument #2 to 'open' (string expected, got nil)` when mode is not specified. We fix `fs.cpp` by defaulting the mode to read only if it's not specified through the `mode: HandleMode` in either `@lute/fs` or `@std/fs`.

### DIFF
--- a/tests/std/fs.test.luau
+++ b/tests/std/fs.test.luau
@@ -164,6 +164,8 @@ test.suite("FsSuite", function(suite)
 		local contents = fs.read(h2)
 		assert.eq(contents, "created file")
 		fs.close(h2)
+
+		fs.remove(file)
 	end)
 end)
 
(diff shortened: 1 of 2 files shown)


tests/std/fs.test.luau
```diff
@@ -152,6 +152,19 @@ test.suite("FsSuite", function(suite)
...
+
+		local h1 = fs.open(file, "w+")
+		fs.write(h1, "created file")
+		fs.close(h1)
+
+		local h2 = fs.open(file) -- with no mode specified, should default to "r"
+		local contents = fs.read(h2)
+		assert.eq(contents, "created file")
+		fs.close(h2)
+	end)
```
skberkeley, line 169: nit: we should delete the file at the end or it'll be visible to the other tests

skberkeley (approved):

tests/std/fs.test.luau
annieetang: ty!

PR #571, opened 2025-11-12
crypto: include the ability to separately generate a key before sealing

aatxe:
The initial `secretbox` API provided opinionatedly generates keys and nonces on your behalf in `seal`. I believe this makes sense for one-off situations, but there's also definitely situations where people would reasonably want to reuse the key across a series of messages being sent with authenticated encryption. We should clearly make that possible.

### DIFF
--- a/lute/crypto/src/crypto.cpp
+++ b/lute/crypto/src/crypto.cpp
@@ -98,12 +98,26 @@ int lua_digest(lua_State* L)
     return 1;
 }
 
-// seal(message: string | buffer): { ciphertext: buffer, key: buffer, nonce: buffer }
+
+// keygen(): buffer
+int lua_secretbox_keygen(lua_State* L)
+{
+    int argumentCount = lua_gettop(L);
+    if (argumentCount != 0)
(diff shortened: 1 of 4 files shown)


lute/crypto/src/crypto.cpp
```diff
@@ -113,11 +126,25 @@ int lua_secretbox_seal(lua_State* L)
     uint8_t* buffer = static_cast<uint8_t*>(lua_newbuffer(L, crypto_secretbox_macbytes() + message.length));
     lua_setfield(L, -2, kCiphertextField);
 
-    // key
-    uint8_t* key = static_cast<uint8_t*>(lua_newbuffer(L, crypto_secretbox_keybytes()));
-    crypto_secretbox_keygen(key);
-    lua_setfield(L, -2, kKeyField);
+    uint8_t* key;
+    // user-provided a key
```
vegorov-rbx: Is this correct?

aatxe: removed the hyphen, oops

vegorov-rbx (approved):

PR #573, opened 2025-11-13
bugfix: lute check supports ~ in luaurc paths

Vighnesh-V:
Users have been having issues with the `.luaurc` generated by the `lute setup` command. This sets certain aliases up in a path with `~` in them but the require rfc didn't support navigating to aliases with `~` in them.

This PR fixes this by allowing the FileVFS to navigate to aliases that represent paths with `~` in them.

Annoyingly, I had to use the filesystem header to handle the test, because it sets up a directory under `~` called `lute_test_special` and writes a file there - if yall have suggestions for how to fix this, please let me know.

### DIFF
--- a/tests/src/require.test.cpp
+++ b/tests/src/require.test.cpp
@@ -1,8 +1,14 @@
 #include "lute/climain.h"
+#include "lute/uvutils.h"
 
 #include "Luau/FileUtils.h"
 
 #include "lua.h"
+
+#include "uv.h"
+
+#include <filesystem>
+#include <fstream>
 
 #include "cliruntimefixture.h"
 #include "doctest.h"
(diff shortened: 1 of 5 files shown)


tests/src/require.test.cpp
```diff
@@ -223,3 +229,36 @@ TEST_CASE_FIXTURE(LuteFixture, "require_by_string_semantics_in_cli")
...
+    // Create test directory and test file
+    std::filesystem::path testDir = std::filesystem::path(homeDir) / "lute_test_special";
+    std::filesystem::path testFile = testDir / "foo.luau";
+    std::filesystem::create_directories(testDir);
+
+    // Write test module file
+    std::ofstream file(testFile);
+    REQUIRE(file.is_open());
+    file << "return { foo = \"bar\" }\n";
+    file.close();
```
aatxe, line 251: would probably be preferable to do the `fopen` vs `_wfopen` stuff that `Luau/FileUtils` does instead of this, which we should have some Lute FileUtils that do that, but it's in a test suite, so like... I'll reluctantly accept `std::filesystem` for now.

aatxe (approved):

tests/src/require.test.cpp
Vighnesh-V: I know, I was really debating doing the version of this without filesystem header, but the test was so much easier to write this way, and it was kind of hard to justify because this is testing code. I'll introduce some utilities for this in a follow up.

Issue #308, opened 2025-05-30
More fs APIs: `fs.copy`, `fs.exists`, recursive `rmdir`.


ffrostfall:
These are all fairly essential file system operations which Lute is lacking in. While you can *kind of* do workarounds, we should really be providing runtime-level support for these operations.

vrn-sn:
I'd also like something analogous to Python's `os.walk` (`luthier.luau` has a somewhat hacky, hand-implemented version right now for parity with `luthier.py`).

deviaze:
`fs.exists` should probably have a TOCTOU like Rust's does, since it shouldn't be relied upon in security-critical stuff

aatxe:
`fs.stat` or similar for file size and other metadata

checkraisefold:
- [x] fs.stat
- [x] fs.exists
- [x] fs.copy
- [x] fs.link
- [x] fs.symlink
- [x] fs.watch
- [x] fs.rmdir recursive
- [x] fs.walk (?)

annieetang:
Closing now that we have all requested `fs` APIs in the comments and list! Feel free to re-open or create a new issue if there are more API requests

PR #574, opened 2025-11-13
bugfix: lute setup should destroy the typedefs folder

Vighnesh-V:
Lute setup had a bug whereby it would write new typedefs into the typedefs folder, but sometimes these could conflict with stale typedefs, causing ambiguous requires. For example, moving "posix.luau" into `posix/init.luau` but keeping `posix.luau` around will lead to an ambiguous require.

To do this, lute setup will call a new standard library method - `fs.removedirectory` with the `{ recursive = true}` option.


aatxe (approved):

annieetang:
addresses "recursive `rmdir`" part of https://github.com/luau-lang/lute/issues/308

Issue #575, opened 2025-11-13
Build failure in Release mode


zoon:
Compiler's -Wmaybe-uninitialized (enabled by -O3) detects
```
error: 'status' may be used uninitialized [-Werror=maybe-uninitialized]
  128 |     if (status == NavigationStatus::NotFound)
```
https://github.com/luau-lang/lute/blob/caf674a9d6f3bc4398b86bcdfd11f62f3b661bf9/lute/require/src/requirevfs.cpp#L105

PR #576, opened 2025-11-13
bugfix: initialize status variable

zoon:
Fixes Release build failure due to uninitialized variable warning with -O3 optimization level. The status variable wasn't assigned when vfsType == VFSType::Lute before use.

https://github.com/luau-lang/lute/issues/575


aatxe (approved):
I love when the compiler is not good at control flow analysis.

PR #577, opened 2025-11-13
Support token selection in selectFromRoot

wmccrthy:
There is currently no way to query for generic `Token` nodes via the `query` library.

Tokens are a little annoying to work with ``visitor``'s ``visitToken`` is called for each of the following types:
	- `AstExprConstantNil` (extension of token)
	- `AstExprConstantNumber` (extension of token)
	- `AstExprConstantBool` (extension of token)
	- `AstExprConstantString` (extension of token)
	- `Token`

In a full pass of `visitor`, each of the token extension types (`AstExprConstant...`) is additionally visited by their respective `visitExprConstant...` function.

Thus, if we were to override `selectVisitor`'s `visitToken` function with our generic `visit` function (calls `fn` on node, selects node if result is not nil), we would end up double counting (selecting same node twice) these `AstExprConstant...` nodes.

This PR adds a utility `isBaseToken` to distinguish between `AstExprConstant...` nodes and the rudimentary `Token` node. We then use this utility in `selectFromRoot`'s `selectVisitor` to avoid double counting the `AstExprConstant...` nodes.

Also, fixes some small bugs along the way:
- `utils.isToken` was performing an improper check; the `Token` type does not have a `kind` field
- `utils.isExprConstantBool` was performing an improper check; `AstExprConstantBool.kind` is always "boolean" not "bool"


skberkeley (approved):
good stuff!

PR #580, opened 2025-11-13
stdlib: Printing AST replacements preserves trivia

skberkeley:
Implemented trivia preservation for AST node replacements by grabbing the leftmost and rightmost trivia of the node being replaced. Along the way, I discovered that trivia is often bound to the EOF token, rather than AST nodes, so I had to make a change to make sure we pass that EOF token around inside of query. To increase the chances that this token is present, I want to encourage codemod authors to use the parse function which returns `ParseResult` (which includes EOF), so I renamed `parse(...) -> AstStatBlock` to `parseBlock`, and `parseFile(...) -> ParseResult` to `parse`.

The bulk of this PR is the machinery for getting that trivia, as well as the tests making sure we correctly reserialize trivia for every AST node. While writing those tests, I also found a bunch of other bugs.


lute/std/libs/syntax/parser.luau
```diff
@@ -8,15 +8,15 @@ local types = require("./types")
...
+function parser.parseblock(source: string): types.AstStatBlock
 	return luau.parse(source).root
 end
 
 function parser.parseexpr(source: string): types.AstExpr
 	return luau.parseexpr(source)
 end
 
-function parser.parsefile(source: string): types.ParseResult
+function parser.parse(source: string): types.ParseResult
```
aatxe, line 19: parsefile was a misnomer here anyway, right? the input here is still the string of the source, not the path.

aatxe (approved):

hgoldstein (changes requested):
Can we have this be a query over tokens instead? That way it's more immediately obvious that we're getting the leftmost trivia.

hgoldstein (changes requested):
Can you add some tests with some intra expression comments? e.g.:
```
return foobar(
  foo, -- what about this
  bar, -- and this
  baz, -- and this??
)
``` 

If I replace `bar` with `quxx` I should still see all the trivia.

lute/std/libs/syntax/utils/trivia.luau
wmccrthy, line 1: since `visitor` visits in order (will visit the left-most token first, and right-most token last), i think we can simplify these. i think using `q.nodes[1]` for leftmost `q.nodes[#q.nodes]` for rightmost will work

lute/std/libs/syntax/utils/trivia.luau
```diff
@@ -0,0 +1,52 @@
+local query = require("../query")
+local types = require("../types")
+
+local retrieverLib = {}
+
+function retrieverLib.leftmosttrivia(n: types.AstNode): { types.Trivia }
+	local q: query.query<types.Token> = query.selectFromRoot(n, function(n)
+		return if n.istoken ~= nil and n.istoken then n else nil
+	end)
```
wmccrthy, line 9:
```suggestion
	local q: query.query<types.Token> = query.selectFromRoot(n, utils.isToken)
```

lute/std/libs/syntax/utils/trivia.luau
```diff
@@ -0,0 +1,52 @@
...
+		end
+	end
+
+	return if leftmostToken ~= nil then leftmostToken.leadingtrivia else {}
+end
+
+function retrieverLib.rightmosttrivia(n: types.AstNode): { types.Trivia }
+	local q: query.query<types.Token> = query.selectFromRoot(n, function(n)
+		return if n.istoken ~= nil and n.istoken then n else nil
+	end)
```
wmccrthy, line 32:
```suggestion
	local q: query.query<types.Token> = query.selectFromRoot(n, utils.isToken)
```

lute/std/libs/syntax/utils/trivia.luau
skberkeley: i originally had a non-query implementation which basically did recursive descent to fetch these trivia instead of using a query. i talked it over with Hunter, and we decided to go with the current approach, where we have a higher guarantee of correctness (it's easier to read few lines of query code) at the cost of potential performance. what you're suggesting is similar since we would need to audit the visitor code to check that all of it is in fact in order, and would be making these helper more brittle by predicating them on an implementation detail of another library (which we admittedly do control)

lute/std/libs/syntax/utils/trivia.luau
skberkeley: it's a good idea, but this introduces a cyclic dependency

lute/std/libs/syntax/utils/trivia.luau
hgoldstein: I don't think the contract for the left most / right most tokens is ever going to change. If we ship this and realize it's too slow in practice, we can always update the implementation.

Alternatively yeah: if we said our visitors _must_ be inorder, then what you suggest works.

hgoldstein (approved):

PR #583, opened 2025-11-14
stdlib: fs.walk

annieetang:
We previously did not have a way to recursively iterate through a directory and its subdirectories. This PR introduces `fs.walk(path, options?)` that allows us to walk through a directory, with an optional boolean to specify recursive or not.

API: 
```luau
export type walkdirectoryoptions = {
	recursive: boolean?,
}

function fslib.walk(path: pathlike, options: walkdirectoryoptions?): () -> pathlike?
```

Example:
```luau
local fs = require("@std/fs")
local iterator = fslib.walk("some/directory", { recursive = true })
local path = iterator()
while path do
	print(path)
	path = iterator()
end
```

See `examples/walk_directory.luau` as well.

Note: Because `for` loops don't support yielding generalized iterators yet, we cannot use `fs.walk` as
```
for path in fs.walk(...) do
    ...
end
```


test1/test.txt
```diff
@@ -0,0 +1 @@
+hi
```
vrn-sn: hi!

lute/std/libs/fs.luau
```diff
@@ -26,6 +26,10 @@ export type removedirectoryoptions = {
 	recursive: boolean?,
 }
 
+export type walkdirectoryoptions = {
```
skberkeley:
```suggestion
export type walkoptions = {
```
we exposed `walk`, not `walkdirectory`

lute/std/libs/fs.luau
```diff
@@ -141,4 +145,35 @@ function fslib.removedirectory(path: pathlike, options: removedirectoryoptions?)
 	end
 end
 
+local function walkdir(path: pathlike, options: walkdirectoryoptions?): () -> pathlike?
```
skberkeley:
```suggestion
local function walkdir(path: pathlike, options: walkdirectoryoptions?): () -> path?
```

in general, library functions should return `path`s and accept `pathlike`s

lute/std/libs/fs.luau
```diff
@@ -141,4 +145,35 @@ function fslib.removedirectory(path: pathlike, options: removedirectoryoptions?)
...
+	local queue = { path }
+	local recursive = options and options.recursive or false
```
skberkeley:
```suggestion
	local recursive = if options ~= nil then options.recursive else false
```
style nit

lute/std/libs/fs.luau
```diff
@@ -141,4 +145,35 @@ function fslib.removedirectory(path: pathlike, options: removedirectoryoptions?)
...
+		while #queue > 0 do
+			local current = table.remove(queue, 1)
+			if not current then
+				return nil
+			end
+
+			if fslib.type(current) == "dir" and recursive then
+				local entries = fslib.listdir(current)
+				for _, entry in entries do
+					local fullPath = pathlib.format(pathlib.join(current, entry.name))
```
skberkeley: why do we need to format?

tests/std/fs.test.luau
```diff
@@ -225,6 +225,30 @@ test.suite("FsSuite", function(suite)
...
+
+	suite:case("walk_directory_recursive", function(assert)
+		local baseDir = path.join(tmpdir, "walktest")
+		local subDir = path.join(baseDir, "subdir")
+		fs.createdirectory(subDir, { makeparents = true })
+
+		local file1 = path.join(baseDir, "file1.txt")
+		local file2 = path.join(subDir, "file2.txt")
+		fs.writestringtofile(file1, "file1")
+		fs.writestringtofile(file2, "file2")
```
skberkeley: if you're gonna write to the files, we should check their contents too

skberkeley (changes requested):

lute/std/libs/fs.luau
```diff
@@ -141,4 +145,34 @@ function fslib.removedirectory(path: pathlike, options: removedirectoryoptions?)
 	end
 end
 
+local function walkdir(path: pathlike, options: walkoptions?): () -> path?
+	local queue = { path }
```
skberkeley:
```suggestion
	local queue = { pathlib.parse(path) }
```
sorry, didn't catch this on the first read thru

tests/std/fs.test.luau
```diff
@@ -225,6 +225,48 @@ test.suite("FsSuite", function(suite)
...
+
+		for _, expected in expectedPaths do
+			local found = false
+			for _, foundPath in foundFiles do
+				if foundPath == expected then
+					found = true
+					break
+				end
+			end
+			assert.eq(found, true, `Expected path {expected} not found in walked paths`)
```
skberkeley: would be nice if this worked, but `assert.eq` doesn't accept a message rn

skberkeley (changes requested):

skberkeley (approved):

annieetang:
addresses part of https://github.com/luau-lang/lute/issues/308

PR #585, opened 2025-11-14
E2E Tests for Query

wmccrthy:


tests/std/syntax/e2eCases/roactToReact/init.luau
```diff
@@ -0,0 +1,280 @@
...
+	cases = require("./roactToReact/cases"),
+}
+
+--[[
+
+High level notes:
+
+- upwards navigation is rly useful; in this codemod, it'd be a lot easier / more performant to query for each function call (generally), update all those that are migrated accordingly (if we have upwards nav
+  we can get the statement by walking up tree from the call node). unfortunately, bc some of the migrations require the whole call statement (i.e local var = call),
+  we cannot do this and have to query for statements for those transforms that require the whole statement context, and separately query for calls for those transforms that only need the call node
```
hgoldstein: Just cut https://github.com/luau-lang/lute/issues/593: I agree.

tests/std/syntax/e2eCases/roactToReact/init.luau
```diff
@@ -0,0 +1,280 @@
...
+
+- I'm not sure of a good way around this one, just a general performance concern, it's tricky to vary operations between node world / string world.
+  When operating on a node, it is not uncommon to want to re-use parts of that node in your transformed output; currently, this will require:
+  	- for string transform: parsing the node into a string, and interpolating into your returned replacement
+	- for node transform: cloning node so it's mutable, updating accordingly (even more impractical, would never do this, just pointing out you CAN)
```
hgoldstein: I think we should be biasing towards string transforms. Most codemods are going to be one-and-done, not multi-pass.

tests/std/syntax/e2eCases/utils.luau
```diff
@@ -0,0 +1,50 @@
...
+)
+	t:selectFromRoot(block :: any, utils.isExprIndexName)
+		:filter(function(node: luau.AstExprIndexName)
+			-- get root expression; determine if it's oldLib
+			return #query
+					.selectFromRoot(node.expression :: any, utils.isBaseToken)
+					:filter(function(node: luau.Token)
+						return node.text == libAlias
+					end).nodes > 0
+		end)
```
hgoldstein: There's nothing stopping you from doing this in one step, you can do something like:
```suggestion
    t:selectFromRoot(block :: any, function (node: luau.AstNode)
        if node.tag != 'index_name' then
            return nil
        end
        if #query.selectFromRoot(node.expression :: any, utils.isBaseToken)
            :filter(function(node: luau.Token)
                return node.text == libAlias
            end).nodes == 0
        then
            return nil
        end
        return node
    end)

```

hgoldstein: Might be nice to add an `exists` helper that can early exit instead of doing `filter` and then `.nodes`

tests/std/syntax/e2eCases/roactToReact/init.luau
wmccrthy: I agree, writing the string transforms was much nicer for this. I guess my point here is that i'm curious if there's a better way to keep parts of existing nodes than using `printer`

tests/std/syntax/e2eCases/utils.luau
wmccrthy: yea, good idea

tests/std/syntax/e2eCases/types.luau
```diff
@@ -0,0 +1,12 @@
...
+export type Transform = (luau.AstStatBlock) -> string
+
+export type Case = { initial: string, expected: string }
+
+export type e2e = {
+	transform: Transform,
+	cases: { Case },
+}
+
+return table.freeze({})
```
hgoldstein, line 12: Is this required?

tests/std/syntax/e2equery.test.luau
```diff
@@ -0,0 +1,59 @@
...
+				fs.writestringtofile(resultPath, result)
+				fs.writestringtofile(initialPath, initialSrc)
+
+				local expectedResult = fs.readfiletostring(expectedPath)
+
+				assert.eq(result, expectedResult)
+			end
+		end)
+	end
+end)
```
hgoldstein, line 69: You should check out `fs.walk` if it's available: I think that's going to be more consistent for your use cases.

tests/std/syntax/e2equery.test.luau
```diff
@@ -0,0 +1,59 @@
...
+
+				formattedTransform(initialPath, expectedPath, transformPath)
+
+				local result = fs.readfiletostring(initialPath)
+				fs.writestringtofile(resultPath, result)
+				fs.writestringtofile(initialPath, initialSrc)
+
+				local expectedResult = fs.readfiletostring(expectedPath)
+
+				assert.eq(result, expectedResult)
```
hgoldstein: 1. We already spoke offline about changing the structure here.
2. You should think about what happens if a given test _doesn't_ conform to the expected structure.

tests/std/syntax/e2eCases/types.luau
wmccrthy: i believe we have to return some value, or requiring the module will error

wmccrthy: doesn't need to be frozen tho 🤷

tests/std/syntax/e2equery.test.luau
wmccrthy: On one hand, we could use git diff via a `process`, but it feels kinda weird to not use an in-house solution. Maybe write a very basic diffing system that compares the output vs snapshot line-by-line?

tests/std/syntax/e2equery.test.luau
```diff
@@ -0,0 +1,67 @@
...
+
+local lutePath = process.execpath()
+
+local allCasesPath = path.format(path.join(".", "tests", "std", "syntax", "e2eCases"))
+
+local function formattedTransform(initialPath: string, expectedPath: string, transformPath: string)
+	stylua(initialPath, expectedPath)
+	process.run({ lutePath, "transform", transformPath, initialPath })
+	stylua(initialPath)
+end
```
hgoldstein: * Why are we running `stylua` on the initial path twice?
* Should these all be `path`s?

tests/std/syntax/e2eCases/apiSurfaceChange/init.luau
```diff
@@ -0,0 +1,49 @@
...
+		:replace(function(node: luau.AstExpr)
+			return (argReplacements[argNodeToFuncName[node]] :: (luau.AstExpr) -> string)(node)
+		end)
+end
+
+local function transform(ctx)
+	local parseresult = ctx.parseresult
+	local block = parseresult.root
+	local libAccessReplacements = qUtils.transformLibraryAccess(block, "sameLibrary", function(expr, index: luau.Token)
+		return `{printer.printnode(expr :: any)}.{apiMap[index.text]}`
```
hgoldstein: Why do you need to cast to `any` here?

tests/std/syntax/e2equery.test.luau
```diff
@@ -0,0 +1,67 @@
...
+		local transformName = k
+		local transformPath = path.format(path.join(allCasesPath, transformName, "init.luau"))
+		local transformCasesPath = path.format(path.join(allCasesPath, transformName, "cases"))
+
+		assert(fs.exists(transformPath), `{transformPath} does not exist.`)
+
+		-- for each case, in .../cases/caseN/...luau
+		-- 		move intial.luau to .../cases/caseN.luau and expected.luau to .../cases/caseN.luau.expect
+		-- 		output result at .../cases/caseN.luau.result
+		--
```
hgoldstein: Is this needed?

tests/std/syntax/e2eCases/apiSurfaceChange/init.luau
wmccrthy: remnant of the old solver's type error, can remove now 👍

tests/std/syntax/e2equery.test.luau
wmccrthy: Oops, yea, we can just stylua everything after the `transform` call. 
I believe all the args should be strings. We pass `path.format` return values into `formattedTransform` calls, which are always `string`

tests/std/syntax/e2equery.test.luau
```diff
@@ -0,0 +1,65 @@
...
+local lutePath = process.execpath()
+
+local allCasesPath = path.format(path.join(".", "tests", "std", "syntax", "e2eCases"))
+
+local function formattedTransform(initialPath: string, expectedPath: string, transformPath: string, resultPath: string)
+	process.run({ lutePath, "transform", "--output", resultPath, transformPath, initialPath })
+	stylua(resultPath, expectedPath)
+end
+
+test.suite("Query and Transform e2e", function(suite: any)
```
hgoldstein: Does this still need to be `any`?

tests/std/syntax/e2equery.test.luau
hgoldstein: Bumping this: I think you want `fs.walk` here.

This definitely _looks_ like this can be abstracted as well.

tests/std/syntax/e2eCases/roactToReact/init.luau
```diff
@@ -0,0 +1,271 @@
...
+	local mountStatements = query.findallfromroot(block, function(node)
+		if not (queryUtils.isStatLocal(node) or queryUtils.isStatAssign(node)) then
+			return nil :: any
+		end
+
+		local isMount = query.findallfromroot(node, queryUtils.isExprCall):exists(function(n)
+			if n.func.tag == "indexname" and n.func.index.text == "mount" then
+				return true
+			end
+			return false
```
hgoldstein: Please do not write `if X then return true else return false end`
```suggestion
			return n.func.tag == "indexname" and n.func.index.text == "mount"
```

tests/std/syntax/e2eCases/roactToReact/init.luau
```diff
@@ -0,0 +1,271 @@
...
+local printer = require("@std/syntax/printer")
+local tableext = require("@std/tableext")
+local queryUtils = require("@std/syntax/utils")
+
+local function mount(block: luau.AstStatBlock): types.replacements
+	-- select all statements with mount calls
+	local mountStatements = query.findallfromroot(block, function(node)
+		if not (queryUtils.isStatLocal(node) or queryUtils.isStatAssign(node)) then
+			return nil :: any
+		end
```
hgoldstein: Are tag checks here not sufficient?

tests/std/syntax/e2eCases/roactToReact/init.luau
```diff
@@ -0,0 +1,271 @@
...
+		return if isMount then node else nil
+	end)
+
+	if #mountStatements.nodes == 0 then
+		return {}
+	end
+
+	return mountStatements:replace(function(mountStatement: luau.AstStatLocal | luau.AstStatAssign)
+		-- get mounted variable
+		local mountedVar = printer.printnode(mountStatement.variables[1].node :: any)
```
hgoldstein: Why does `printnode` need an `any` cast?

tests/std/syntax/e2eCases/roactToReact/init.luau
```diff
@@ -0,0 +1,271 @@
...
+			keyArg = '"ReactRoot"'
+		end
+
+		local mountSetup =
+			`local root = ReactRoblox.createRoot(Instance.new("Folder"))\nlocal actualParent = {parentArg}\nlocal actualKey = {keyArg}`
+		local mount = ([[ReactRoblox.act(function()
+	root:render(ReactRoblox.createPortal({ [actualKey] = %s } :: any, actualParent))
+end)
+local %s = { root = root, parent = actualParent, key = actualKey }]]):format(elementArg, mountedVar)
+		return mountSetup .. "\n" .. mount
```
hgoldstein: Can you pick a style here? I think my preference would be for something like:
```luau
local template = ([[
local root = ReactRoblox.createRoot(Instance.new("Folder"))
local actualParent = %s
local actualKey = %s
ReactRoblox.act(function()
    root:render(ReactRoblox.createPortal({ [actualKey] = %s } :: any, actualParent))
end)
local %s = { root = root, parent = actualParent, key = actualKey }]])
]]):format(...)
```

tests/std/syntax/e2eCases/roactToReact/init.luau
```diff
@@ -0,0 +1,271 @@
...
+	local parseresult = ctx.parseresult
+	local ast = parseresult.root
+	local needsReactRequire, needsReactRobloxRequire = false, false
+
+	-- convert statements
+	local fullStatementRepl = mount(ast)
+
+	-- convert calls
+	local callRepl = query
+		.findallfromroot(ast, queryUtils.isExprCall)
```
hgoldstein, line 171: cc @skberkeley: maybe we should move the query utils to `query`; feels kinda silly that you have to import both.

tests/std/syntax/e2eCases/roactToReact/init.luau
```diff
@@ -0,0 +1,271 @@
...
+	local addedRequiresRepl = {}
+	if needsReactRequire or needsReactRobloxRequire then
+		local roactRequire = utils.getRequireStatements(ast, "Roact").nodes[1]
+		local requireCall = roactRequire.values[1].node :: luau.AstExprCall
+		local requiredModule = requireCall.arguments[1].node :: luau.AstExprIndexName
+		if needsReactRequire and needsReactRobloxRequire then
+			-- get PackagesRoot
+			local packagesRoot = printer.printnode(requiredModule.expression :: any)
+			addedRequiresRepl[roactRequire] =
+				`local React = require({packagesRoot}.React)\nlocal ReactRoblox = require({packagesRoot}.ReactRoblox)\n`
```
hgoldstein: This is another spot where it really hurts readability to have the newline in the middle of the format string.

tests/std/syntax/e2eCases/roactToReact/init.luau
```diff
@@ -0,0 +1,271 @@
...
+			addedRequiresRepl = tableext.combine(requireArgRepl, requiredVarRepl)
+		end
+	end
+
+	return tableext.combine(
+		fullStatementRepl,
+		callRepl,
+		libIndexNameRepl,
+		libIndexExprRepl,
+		addedRequiresRepl :: types.replacements
```
hgoldstein: what happened here?

tests/std/syntax/e2equery.test.luau
wmccrthy: nope

tests/std/syntax/e2eCases/roactToReact/init.luau
wmccrthy: Yea, we can use a single multi-line string for sure. The only nit with the format you provided is that we'll have an extra newline at the start of the string; not a big deal tho

tests/std/syntax/e2eCases/roactToReact/init.luau
wmccrthy: I'm getting all sorts of type errors unless I return any. It seems to be inferring the return type of the function here HAS to match the type of the first return (i.e `nil`, in this case).

tests/std/syntax/e2eCases/roactToReact/init.luau
wmccrthy: remnant of the old solver, can remove

tests/std/syntax/e2eCases/roactToReact/init.luau
wmccrthy: Getting: `TypeError: No valid instantiation could be inferred for generic type parameter K.` on `tableext.combine` unless I cast this last value to `replacements`. Even seeing the error if I cast each individual declaration for these variables to `replacements`

tests/std/syntax/e2equery.test.luau
wmccrthy: What's the main difference you're envisioning? The logic has to be kinda specific since the test cases are all in a flat dir, i.e:
- we walk the directory with all cases; we find `case1.luau`. We need to match this specifically too `case1.luau.result`, `case1.luau.expect`. I'm not sure walking makes this any easier?

I've reformatted to use `fs.walk`, but lmk if there's a different structure you're looking for.

tests/std/syntax/e2eCases/roactToReact/init.luau
hgoldstein: Ohhhh, yes that's a really fucking annoying thing we've been trying to fix, I would recommend you add a return type annotation instead (for now).

tests/std/syntax/e2equery.test.luau
hgoldstein: You could imagine for a larger golden test you have a _lot_ of different subdirectories with different kinds of tests, and maybe different settings for each test.

tests/std/syntax/e2eCases/roactToReact/init.luau
```diff
@@ -0,0 +1,270 @@
...
+	local ast = parseresult.root
+	local needsReactRequire, needsReactRobloxRequire = false, false
+
+	-- convert statements
+	local fullStatementRepl: types.replacements = mount(ast)
+
+	-- convert calls
+	local callRepl: types.replacements = query
+		.findallfromroot(ast, queryUtils.isExprCall)
+		:filter(function(node: luau.AstExprCall)
```
hgoldstein: Is this annotation needed?

tests/std/syntax/e2equery.test.luau
```diff
@@ -0,0 +1,69 @@
...
+end
+
+local function getCases(casesDir: path.path)
+	local cases = fs.listdirectory(path.format(casesDir))
+	return tableext.filter(cases, function(case)
+		return case.type == "file" and path.extname(case.name) == ".luau"
+	end)
+end
+
+local transformsDir = "./tests/std/syntax/e2eCases"
```
hgoldstein: This feels a bit manual: is there any way for us to get this programmatically?

tests/std/syntax/e2equery.test.luau
```diff
@@ -0,0 +1,69 @@
...
+test.suite("E2E Query Transforms", function(suite: any)
```
hgoldstein: Is this `any` still needed?

tests/std/syntax/e2equery.test.luau
```diff
@@ -0,0 +1,69 @@
...
+	local casesWalker = fs.walk(transformsDir, { recursive = true })
+	local curFile = casesWalker()
+
+	while curFile do
+		if path.basename(curFile) == "init.luau" then
+			local transformDir = path.dirname(curFile)
+			local transformName = path.basename(transformDir)
+			local transormFile = path.format(curFile)
```
hgoldstein: spelling: `transformFile`

hgoldstein (dismissed):
See comments, and I believe we should be using all lower case for the path names, so `e2ecases` and `apisurfacechange` instead of `e2eCases` and `apiSurfaceChange` respectively (yes I see there's a folder that's currently violating that).

tests/std/syntax/e2eCases/roactToReact/init.luau
wmccrthy: looks like it is not

tests/std/syntax/e2equery.test.luau
wmccrthy: nope

tests/std/syntax/e2equery.test.luau
wmccrthy: Easiest way I know of is with the `debug` API but not sure if that's the best way. Do you know of any other solutions?

tests/std/syntax/e2equery.test.luau
```diff
@@ -0,0 +1,71 @@
...
+					)
+				end
+			end)
+		end
+
+		curFile = casesWalker()
+	end
+end)
+
+test.run()
```
Vighnesh-V: You should delete this!

Vighnesh-V (changes requested):

PR #588, opened 2025-11-14
stdlib: luacase some ast node fields

skberkeley:
A while back, I went through and luacased the AST node fields. It looks like I missed a few on that pass, so fixing them here.


vrn-sn (approved):

annieetang (approved):
🤠

PR #589, opened 2025-11-14
docs: support doc comments in lute doc gen

annieetang:
‼️ disclaimer - this code is pretty gross and is 100% just an intermediate step in the Luau doc gen plan while I'm still working on the implementation. This entire file should be deleted when we have implementations for parsing a module's return type and generating docs from that instead of this line-by-line pattern matching 😃👍 ‼️

Adding support for adding all sorts of doc comments to functions that are captured by the doc gen script for https://lute.luau.org/!

Only detects comments that are right above the function signature, but you can write 
- [single line comments](https://create.roblox.com/docs/luau/comments#single-line-comments) `---` or
- [block comments](https://create.roblox.com/docs/luau/comments#block-comments) with `--[=[ ]=]`
^this is because we don't have a standardized way of writing doc comments yet in Luau, so I'm just supporting these two for now

There is a mock module file and output documentation file in `docs/test/` that you can see what the doc gen script `reference.luau` generates! They're the `test_module.luau` and `test_module.md` files.

### DIFF
--- a/docs/scripts/reference.luau
+++ b/docs/scripts/reference.luau
@@ -1,53 +1,145 @@
 local fs = require("@std/fs")
 local process = require("@std/process")
 local path = require("@std/path")
+local stringext = require("@std/stringext")
 
-local function extractPropertySignature(module: string, line: string): (string?, string?)
-	-- Match property declarations like: process.env = {} :: { [string]: string }
+-- extractPropertySignature:
+-- module: string
+-- line: string (current line)
+-- pendingDoc: {string} (accumulated doc lines so far)
+-- inBlock: boolean (are we currently inside a block comment?)
+-- blockEq: string? (the "=" sequence used for block comment delimiting, e.g. "=" for --[=[  ; nil when not inside)
+--
+-- returns: name?, signature?, doc?, newPendingDoc, newInBlock, newBlockEq
+local function extractPropertySignature(
+	module: string,
+	line: string,
+	pendingDoc: { string },
+	inBlock: boolean,
+	blockEq: string?
+): (string?, string?, string?, { string }, boolean, string?)
+	local name, signature, doc = nil, nil, nil
+
+	-- 1) If we're *not* in a block, detect block start that may include = signs:
+	if not inBlock then
+		local eq = string.match(line, "^%s*%-%-%[(=*)%[")
+		if eq ~= nil then
+			-- start of block comment, capture any text after the opening on the same line
+			inBlock = true
+			blockEq = eq -- remember the = count for closing pattern
+
+			-- capture part after the opening delimiter, if any
+			local after = string.match(line, "^%s*%-%-%[" .. eq .. "%[(.*)$")
+			if after then
+				after = stringext.trim(after)
+				if after ~= "" then
+					table.insert(pendingDoc, after)
+				end
+			end
+
+			-- nothing else to do this line
+			return nil, nil, nil, pendingDoc, inBlock, blockEq
+		end
+	end
+
+	-- 2) If we're inside a block, accumulate until we find the matching close
+	if inBlock then
+		local eq = blockEq or ""
+		-- find the closing pattern "]]", or "]=]" etc. that matches eq
+		local closingPattern = "%]" .. eq .. "%]"
+		local before = string.match(line, "^(.-)" .. closingPattern)
+		if before then
+			-- closing found on this line
+			before = stringext.trim(before)
+			if before ~= "" then
+				table.insert(pendingDoc, before)
+			end
+
+			-- there might be trailing comment markers like "]]" on the same line;
+			-- we've already captured inner content, so we just exit block mode.
+			inBlock = false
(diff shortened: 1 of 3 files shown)


annieetang:
added a test file for visualization and I also combined a lot of the helper functions between the stdlib and lute (the stuff in `definitions/`) modules to make the code a lil easier to understand hopefully

docs/scripts/reference.luau
```diff
@@ -1,59 +1,152 @@
...
+local whitespacePattern = "^%s*$"
+
+-- extractPropertySignature: (this function runs on one line at a time so we need to keep some states for block comments)
+-- module (module name)
+-- line (current line)
+-- pendingDoc (accumulated doc lines so far)
+-- inBlock (are we currently inside a block comment?)
+-- blockEq (the "=" sequence used for block comment delimiting, e.g. "=" for --[=[  ; nil when not inside)
+--
+-- returns: name?, signature?, doc?, newPendingDoc, newInBlock, newBlockEq
```
aatxe: This should really be a table. multiple return over three is a bad code smell.

aatxe (changes requested):

docs/scripts/reference.luau
```diff
@@ -1,59 +1,187 @@
...
+				after = stringext.trim(after)
+				if after ~= "" then
+					table.insert(pendingDoc, after)
+				end
+			end
+
+			return {
+				name = nil,
+				signature = nil,
+				doc = nil,
```
aatxe:
```suggestion
```

optional fields are optional

docs/scripts/reference.luau
```diff
@@ -1,59 +1,187 @@
...
+			local trimmed = stringext.trim(line)
+			if trimmed ~= "" then
+				table.insert(pendingDoc, trimmed)
+			end
+		end
+
+		return {
+			name = nil,
+			signature = nil,
+			doc = nil,
```
aatxe:
```suggestion
```

docs/scripts/reference.luau
```diff
@@ -1,59 +1,187 @@
...
+	local triple = string.match(line, tripleDashCommentPattern)
+	if triple then
+		local txt = stringext.trim(triple)
+		if txt ~= "" then
+			table.insert(pendingDoc, txt)
+		end
+		return {
+			name = nil,
+			signature = nil,
+			doc = nil,
```
aatxe:
```suggestion
```

docs/scripts/reference.luau
```diff
@@ -1,59 +1,187 @@
...
+	end
+
+	-- 4) Match property declarations like: process.env = {} :: { [string]: string }
 	local propName, propType = string.match(line, `^{module}%.([%w_]+)%s*=%s*.-::%s*(.+)$`)
 	if propName and propType then
-		return propName, propType
+		return {
+			name = propName,
+			signature = propType,
+			doc = (#pendingDoc > 0) and table.concat(pendingDoc, "\n\n") or nil,
```
aatxe:
```suggestion
			doc = if #pendingDoc > 0 then table.concat(pendingDoc, "\n\n") else nil,
```

docs/scripts/reference.luau
```diff
@@ -1,59 +1,187 @@
...
-	local content = fs.readfiletostring(filePath)
-	local lines = string.split(content, "\n")
-	local definitions = {}
-
-	for _, line in lines do
-		local name, signature = extractPropertySignature(tostring(module), line)
+		return {
+			name = fullName,
+			signature = `{params} -> {returnType}`,
+			doc = (#pendingDoc > 0) and table.concat(pendingDoc, "\n\n") or nil,
```
aatxe:
```suggestion
			doc = if #pendingDoc > 0 then table.concat(pendingDoc, "\n\n") else nil,
```

docs/scripts/reference.luau
```diff
@@ -1,59 +1,187 @@
...
+	if not string.match(line, whitespacePattern) then
+		pendingDoc = {}
 	end
 
-	return definitions
+	-- nothing found on this line
+	return {
+		name = nil,
+		signature = nil,
+		doc = nil,
```
aatxe:
```suggestion
```

docs/scripts/reference.luau
```diff
@@ -1,59 +1,187 @@
...
+	inBlock: boolean,
+	blockEq: string?
+): {
+	name: string?,
+	signature: string?,
+	doc: string?,
+	pendingDoc: { string },
+	inBlock: boolean,
+	blockEq: string?,
+}
```
aatxe: I'd pull this into a type definition like
```luau
type PropertySignature = {
	name: string?,
	signature: string?,
	doc: string?,
	pendingDoc: { string },
	inBlock: boolean,
	blockEq: string?,
}
```

aatxe (changes requested):

aatxe (approved):

PR #591, opened 2025-11-17
stdlib: query:findall and query:flatmap

skberkeley:
Follow up work to AST querying:
- Select provides the same functionality as `selectFromRoot`, but rooted on each node contained in the query.
- flatMap allows programmers to return array-like tables in their lambdas, which are then flattened and collected to form the next collection of nodes.

flatMap also allowed me to write an additional unit test which revealed a bug with how token replacements were being handled.

### DIFF
--- a/lute/std/libs/syntax/query.luau
+++ b/lute/std/libs/syntax/query.luau
@@ -2,6 +2,7 @@
 -- @std/syntax/query
 -- Provides utility functions for querying Luau AST nodes
 
+local tableext = require("@std/tableext")
 local types = require("./types")
 local visitor = require("./visitor")
 local utils = require("./utils")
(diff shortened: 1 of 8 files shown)


lute/std/libs/syntax/query.luau
```diff
@@ -86,6 +87,40 @@ function queryLib.selectFromRoot<T>(ast: types.ParseResult | node, fn: (node) ->
 		return true
 	end
 
+	return selectVisitor
+end
+
+function queryLib.select<T, U>(self: query<T>, fn: (node) -> U?): query<U>
```
aatxe:
```suggestion
function queryLib.findall<T, U>(self: query<T>, fn: (node) -> U?): query<U>
```

lute/std/libs/syntax/query.luau
```diff
@@ -86,6 +87,40 @@ function queryLib.selectFromRoot<T>(ast: types.ParseResult | node, fn: (node) ->
...
+		local mapped = fn(node)
+		if #mapped > 0 then
+			tableext.extend(newNodes, mapped)
+		end
+	end
+	self.nodes = newNodes
+	return self
+end
+
+function queryLib.selectFromRoot<T>(ast: types.ParseResult | node, fn: (node) -> T?): query<T>
```
aatxe:
```suggestion
function queryLib.findAllFromRoot<T>(ast: types.ParseResult | node, fn: (node) -> T?): query<T>
```

aatxe (commented):
I left some comments locally, but I think that probably they're better addressed by tweaking something about the query API altogether. You do not, and probably shouldn't have, the API of the query _inside_ the same top-level API if you want to distinguish between "operations that create queries" and "operations that interact with queries." This is exactly where OO is applicable.

aatxe (approved):
we can rename `query.findallfromroot` to `query.findall` when we separate the two libraries as alluded to in my previous review comment, but this is still an improvement

PR #594, opened 2025-11-18
std/syntax: `visitor.visit` should support table items (`luau.AstExprTableItem`)

wmccrthy:
`visitor.visit` will currently fallback to `exhaustiveMatch` and error if passed an `AstExprTableItem` node.
This causes `printnode` to fail if passed a `replacements` map containing any `AstExprTableItem` node, preventing table entries from being replaced.

You can minimally repro this with:

```luau
local printer = require("@std/syntax/printer")
local parser = require("@std/syntax/parser")

local demo = [[

local tbl = {
    a = 1,
}

]]

local ast = parser.parseblock(demo)

local replacements = {}

local tbl = query.selectFromRoot(ast, utils.isExprTable):forEach(function(n)
    for i, entry in n.entries do
        replacements[entry] = `a{i} = "hello",`
    end
end)

print(printer.printnode(ast, replacements))
```

This PR ensures `visitor.visit` supports table entries and that we can replace them in codemods

### DIFF
--- a/definitions/luau.luau
+++ b/definitions/luau.luau
@@ -135,7 +135,7 @@ export type AstExprAnonymousFunction = {
 }
 
 -- helper types for items contained in an AstExprTable, not actually AstExprs themselves
-export type AstExprTableItemList = { kind: "list", value: AstExpr, separator: Token<"," | ";">? }
+export type AstExprTableItemList = { kind: "list", value: AstExpr, separator: Token<"," | ";">?, istableitem: true }
 
 export type AstExprTableItemRecord = {
 	kind: "record",
@@ -143,6 +143,7 @@ export type AstExprTableItemRecord = {
 	equals: Token<"=">,
 	value: AstExpr,
 	separator: Token<"," | ";">?,
+	istableitem: true,
 }
 
 export type AstExprTableItemGeneral = {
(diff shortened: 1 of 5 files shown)


lute/std/libs/syntax/utils/init.luau
```diff
@@ -58,6 +58,10 @@ function utils.isExprTable(n: types.AstNode): types.AstExprTable?
 	return if n.kind == "expr" and n.tag == "table" then n else nil
 end
 
+function utils.isExprTableItem(n: types.AstNode): types.AstExprTableItem?
+	return if n.kind == "list" or n.kind == "record" or n.kind == "general" then n else nil
+end
+
```
skberkeley, line 64: can we add an `istableitem` field to each of these?

skberkeley (changes requested):

lute/std/libs/syntax/utils/init.luau
skberkeley: *to each type of table item

definitions/luau.luau
```diff
@@ -155,7 +155,9 @@ export type AstExprTableItemGeneral = {
 	separator: Token<"," | ";">?,
 }
 
-export type AstExprTableItem = AstExprTableItemList | AstExprTableItemRecord | AstExprTableItemGeneral
+export type AstExprTableItem =
+	{ istableitem: true }
+	& (AstExprTableItemList | AstExprTableItemRecord | AstExprTableItemGeneral)
```
skberkeley: i think things will play nicer with the new solver if we add `istableitem` to each component type of the union

definitions/luau.luau
skberkeley, line 166: this should probably be a `Punctuated` but that's irrelevant to this PR

skberkeley (approved):
just the thing about the type definitions, and it should be good to go. thanks for adding this fix!

PR #595, opened 2025-11-18
lute/luau: implement a `span` userdata to replace serializing location and position as tables.

### LINKED ISSUE #572: Represent `luau.Location` in a more compact form
For the `luau.parse` API, every `Token` has a `Location`, which itself is a table with two members that represents a beginning and end position for the token. Each position itself is a table with two numbers, `begin` and `end`. Reportedly, allocating these tables is a significant chunk of time taken when serializing ASTs for userland. We should have some sort of compact representation that's faster to serialize.

One option is representing locations as a buffer with a convenient API for accessing the different parts of a `Location`, for example:
```luau
--- location.luau

-- Preferably this would be opaque
export type Location = buffer

local exports = {}

function exports.begin_line(b)
    return buffer.readu32(b, 0)
end

function exports.end_line(b)
    return buffer.readu32(b, 1)
end

-- etc ...
```

Another option is creating a bespoke `extern` type, e.g.:
```luau
-- location.luau

declare extern type Position with
    line: number
    col: number
end

declare extern type Location with
    begin: Position
    ["end"]: Position
end

```

aatxe:
Resolves #572 by replacing the location and position tables with a single userdata, `span`, that should be defined as:
```luau
declare extern type span with
    beginline: number,
    begincolumn: number,
    endline: number,
    endcolumn: number
end
```

This includes a constructor `span.create` that takes a table so that you can use punned syntax to construct a span easily, e.g.
```luau
local location = span.create {
    beginline = 5,
    begincolumn = 6,
    endline = 5,
    endcolumn = 7,
}
```

My informal measurements showed a 30-40% speedup, but we should probably try to put together more formal benchmarks for parsing in a follow-up work before we do any direct optimization of parsing.


lute/luau/include/lute/luau.h
```diff
@@ -13,6 +13,10 @@ int luteopen_luau(lua_State* L);
 namespace luau
 {
 
+static const char kSpanType[] = "span";
+static const char kCompileResultType[] = "CompileResult";
+static const char kSpanCreateName[] = "spanCreate";
```
skberkeley: do we need this at all? i only see one usage of it

lute/std/libs/syntax/init.luau
```diff
@@ -1,8 +1,15 @@
...
-export type Position = types.Position
-export type Location = types.Location
+local luau = require("@lute/luau")
+
+export type span = types.span
+
+local span = table.freeze({
+	create = function(tbl: { beginline: number, begincolumn: number, endline: number, endcolumn: number }): span
+		return luau.span.create(tbl)
+	end,
```
skberkeley: does this not work?
```suggestion
	create = luau.span.create
```

skberkeley (approved):

lute/std/libs/syntax/init.luau
aatxe: I think that is less likely to work with @annieetang's current doc stuff, but also it probably doesn't work either way now _anyway_, so it's fine, yeah!

lute/luau/include/lute/luau.h
aatxe:
```suggestion
static const char kSpanCreateName[] = "span.create";
```

aatxe: it's a code style/consistency thing. I want to do some clean up of this in general at some point. Basically, add a bit of structure to how we define all of these module headers so that they're done consistently.

PR #596, opened 2025-11-18
fix: exclude `docs/test/` from doc gen and fix file naming

annieetang:
`test/` folder was being published to the site and the filenames were the internal table variable names (i.e. `printerlib`) instead of the intended module names (i.e. `printer`)

### DIFF
--- a/docs/scripts/reference.luau
+++ b/docs/scripts/reference.luau
@@ -254,10 +254,9 @@ local function processDirectory(
 
 			print(`Processing {moduleName}...`)
 
-			local moduleAlias = moduleName
-			if requireLibraryAlias == "std" then
-				moduleAlias = STDLIB_MODULE_ALIASES[moduleName] or moduleName
-			end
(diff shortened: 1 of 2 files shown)


docs/scripts/reference.luau
```diff
@@ -254,11 +254,12 @@ local function processDirectory(
 
 			print(`Processing {moduleName}...`)
 
+			local moduleAlias = moduleName
 			if requireLibraryAlias == "std" then
-				moduleName = STDLIB_MODULE_ALIASES[moduleName] or moduleName
+				moduleAlias = STDLIB_MODULE_ALIASES[moduleName] or moduleName
 			end
```
hgoldstein:
```suggestion
            local moduleAlias = if requireLibraryAlias == "std" 
                then STDLIB_MODULE_ALIASES[moduleName] or moduleName
                else moduleName

end
```

annieetang: tysm

aatxe (approved):

PR #597, opened 2025-11-19
First version of lute lint

skberkeley:
Initial implementation of `lute lint`. It currently supports exactly one rule and one source file, with multiple rules and source files coming in a future PR. By default, it pretty prints reported violations (barring tab shenanigans).

Some notes:
- I had to copy `cli.luau` over from batteries because `@batteries` isn't embedded with lute commands.
- I added a `@commands` alias so that the lint rules I wrote could have access to lint-related types. Longer term, we should probably figure out a better place to put those types.


.luaurc
```diff
@@ -3,6 +3,7 @@
     "aliases": {
         "batteries": "./batteries",
         "std": "./lute/std/libs",
-        "lute": "./definitions"
+        "lute": "./definitions",
+        "commands": "./lute/cli/commands"
     }
-}
+}
```
aatxe:
```suggestion
}

```

examples/lints/divide_by_zero.luau
```diff
@@ -0,0 +1,38 @@
...
+local function lint(ast: syntax.AstStatBlock, sourcepath: path.path): { lintTypes.LintViolation }
+	local violations = {}
+
+	query
+		.findallfromroot(ast, utils.isExprBinary)
+		:filter(function(bin)
+			return bin.operator.text == "/" or bin.operator.text == "//" or bin.operator.text == "%"
+		end)
+		:filter(function(bin)
+			return utils.isExprConstantNumber(bin.rhsoperand) ~= nil
```
hgoldstein: Why is this a `utils` call and not a tag check?

examples/lints/divide_by_zero.luau
```diff
@@ -0,0 +1,38 @@
...
+		end)
+		:foreach(function(n: syntax.AstExprBinary)
+			table.insert(violations, { -- LUAUFIX: severity isn't inferred as a singleton, so table.insert is mad
+				lintname = name,
+				location = n.location,
+				message = message,
+				severity = "warning",
+				sourcepath = sourcepath,
+			})
+		end)
```
hgoldstein: I wonder if it's worthwhile adding some sort of "map to table" function like:
```luau
query:maptotable(function(n: syntax.AstExprBinary): lintTypes.Violation
    return {
        lintname = name,
		location = n.location,
		message = message,
		severity = "warning",
		sourcepath = sourcepath,
    }
end) -- has `{ lintTypes.Violation }`
```

hgoldstein (commented):
Looks like you have some failing tests

examples/lints/divide_by_zero.luau
skberkeley: i can make it a tag check! they're semantically the same although the tag checking facilitates refining (add type predicates soon?)

examples/lints/divide_by_zero.luau
skberkeley: yeah that seems reasonable to me; otherwise this pattern will pop up in every single lint written

skberkeley: *every single lint where the violation is scoped to a single node

examples/lints/almost_swapped.luau
```diff
@@ -0,0 +1,101 @@
...
+						endline = nextStat.location.endline,
+						endcolumn = nextStat.location.endcolumn,
+					}),
+					message = message,
+					severity = "warning",
+					sourcepath = sourcepath,
+				})
+			end
+		end
+	end)
```
hgoldstein: should this be `maptoarray`?

lute/cli/commands/lint/types.luau
```diff
@@ -0,0 +1,19 @@
...
+	location: syntax.span,
+	message: string,
+	severity: severity,
+	sourcepath: path.path,
+} -- TODO: add suggested fix
+
+export type LintRule = {
+	lint: (ast: syntax.AstStatBlock, sourcepath: path.path) -> { LintViolation },
+	name: string,
+}
```
hgoldstein, line 17: Let's make these `read`

lute/cli/commands/cli.luau
```diff
@@ -0,0 +1,175 @@
+local cli = {}
```
hgoldstein, line 1: @aatxe: did we come to a conclusion on bundling these with `lute`?

examples/lints/almost_swapped.luau
skberkeley: it's more awkward to because the violations span multiple nodes

lute/cli/commands/lint/types.luau
skberkeley: done!

examples/lints/almost_swapped.luau
```diff
@@ -0,0 +1,101 @@
...
+		return false
+	end
+end
+
+-- Report instances of attempted swaps like:
+-- a = b; b = a
+local function lint(ast: syntax.AstStatBlock, sourcepath: path.path): { lintTypes.LintViolation }
+	local violations = {}
+
+	query.findallfromroot(ast, utils.isStatBlock):foreach(function(block)
```
hgoldstein: This is also my beef with `foreach`. I think it's noticeably less readable than just:
```
local nodes = query.findallfromroot(ast, utils.isStatBlock):nodes()
for _, block in nodes do
  -- ...
end
```

skberkeley: that's a fair point, i can rewrite this to just iterate directly over the nodes

hgoldstein (approved):

PR #603, opened 2025-11-20
stdlib: update tableext by removing unnecessary funcs and types

annieetang:
As part of the API audit/cleanup, we remove `fold`, `foreach`, `array` and `dictionary` from `tableext` and any use cases


skberkeley (approved):
lgtm

PR #605, opened 2025-11-20
stdlib: luacase @std/luau

annieetang:
As part of the API audit/cleanup, we are luacasing!!

### DIFF
--- a/definitions/luau.luau
+++ b/definitions/luau.luau
@@ -600,13 +600,13 @@ function luau.parseexpr(source: string): AstExpr
 	error("not implemented")
 end
 
-export type bytecode = { bytecode: string }
+export type Bytecode = { bytecode: string }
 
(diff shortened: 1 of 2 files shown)


definitions/luau.luau
```diff
@@ -600,13 +600,13 @@ function luau.parseexpr(source: string): AstExpr
 	error("not implemented")
 end
 
-export type Bytecode = { bytecode: string }
+export type bytecode = { bytecode: string }
```
annieetang: do we wanna luacase `@lute/luau` as well?

skberkeley: i would prefer not to right now, but it's probably worth dogfooding

skberkeley (approved):

definitions/luau.luau
annieetang: nw i'll leave it out

PR #606, opened 2025-11-20
stdlib: cleanup @std/fs (renaming/removing APIs)

annieetang:
As part of the API audit/cleanup, we are fixing `@std/fs` APIs.

This is mostly renaming and removing unnecessary functions, gonna leave `fs.watch` to be its separate PR

Parallel changes to `@lute/fs` will be in a subsequent PR after we've agreed on how @std should look like


lute/std/libs/fs.luau
```diff
@@ -10,13 +10,14 @@ local sys = require("@std/system")
 local fslib = {}
 
 export type handlemode = fs.HandleMode
-export type filehandle = fs.FileHandle
+export type file = fs.FileHandle
 export type filetype = fs.FileType
```
annieetang: do we want to rename this from `filetype` to `typ`? I feel like `typ` might not be as intuitive

lute/std/libs/fs.luau
```diff
@@ -10,13 +10,14 @@ local sys = require("@std/system")
...
-export type filemetadata = fs.FileMetadata
+export type metadata = fs.FileMetadata
 export type directoryentry = fs.DirectoryEntry
```
annieetang: ditto above comment on renaming this to `item`, but I'm cool with it if we want to rename from `directoryentry` to `item`

lute/std/libs/fs.luau
skberkeley: the usage would probably look like
```luau
local fs = require('@std/fs')
local t: fs.typ
```
so i tink it's fine?

skberkeley (approved):

lute/std/libs/fs.luau
annieetang: oh ok, same w `directoryentry` as `item` then I guess?

aatxe: i think writing `type` is fine actually for the same qualified import reason, `fs.type` doesn't shadow `type`

lute/std/libs/fs.luau
```diff
@@ -10,13 +10,14 @@ local sys = require("@std/system")
...
-export type filemetadata = fs.FileMetadata
+export type file = fs.FileHandle
+export type type = fs.FileType
+export type metadata = fs.FileMetadata
 export type directoryentry = fs.DirectoryEntry
 export type watchhandle = fs.WatchHandle
 export type watchevent = fs.WatchEvent
-export type pathlike = pathlib.pathlike
+
+type pathlike = pathlib.pathlike
```
aatxe, line 20: an aside, but we should _really_ name this type `path` if it's the one everyone is writing everywhere.

aatxe (approved):

PR #609, opened 2025-11-21
refactor: split process.run and process.shell into separate functions

annieetang:
As part of the API audit/cleanup, we separate `run` and `shell` execution in `@std/process` to be separate functions for more intuitive APIs 😊

Also add `assert.erroreq` function to compare a function's error message with the expected error msg

### DIFF
--- a/definitions/process.luau
+++ b/definitions/process.luau
@@ -1,4 +1,4 @@
-export type StdioKind = "default" | "inherit" | "none" | ""
+export type StdioKind = "default" | "inherit" | "none"
 
 export type ProcessRunOptions = {
 	cwd: string?,
--- a/definitions/process.luau
+++ b/definitions/process.luau
@@ -7,8 +7,8 @@ export type ProcessRunOptions = {
 	env: { [string]: string }?,
 }
 
-export type ProcessShellOptions = {
-	shell: string?,
+export type ProcessSystemOptions = {
+	system: string?,
 	cwd: string?,
 	stdio: StdioKind?,
 
@@ -40,7 +40,7 @@ function process.run(args: { string }, options: ProcessRunOptions?): ProcessResu
 	error("not implemented")
 end
 
-function process.shell(command: string, options: ProcessShellOptions?): ProcessResult
+function process.system(command: string, options: ProcessSystemOptions?): ProcessResult
 	error("not implemented")
 end
 
--- a/definitions/process.luau
+++ b/definitions/process.luau
@@ -1,7 +1,14 @@
 export type StdioKind = "default" | "inherit" | "none" | ""
 
 export type ProcessRunOptions = {
-	shell: (string | boolean)?,
+	cwd: string?,
+	stdio: StdioKind?,
+
+	env: { [string]: string }?,
+}
+
+export type ProcessShellOptions = {
+	shell: string?,
 	cwd: string?,
 	stdio: StdioKind?,
 
@@ -29,7 +36,11 @@ function process.cwd(): string
 	error("not implemented")
 end
 
-function process.run(args: string | { string }, options: ProcessRunOptions?): ProcessResult
+function process.run(args: { string }, options: ProcessRunOptions?): ProcessResult
+	error("not implemented")
+end
+
+function process.shell(command: string, options: ProcessShellOptions?): ProcessResult
 	error("not implemented")
 end
 
--- a/examples/process.luau
+++ b/examples/process.luau
@@ -16,19 +16,19 @@ print(r1.stdout)
 print(r2.stdout)
 print(r3.stdout)
 
-local r4 = process.run("echo Hello, lute!", { shell = true })
+local r4 = process.shell("echo Hello, lute!")
 print(r4.stdout)
 
-local r5 = process.run({ "echo", "$HOME" }, { env = { HOME = "/home/lute" }, shell = true })
+local r5 = process.shell("echo $HOME", { env = { HOME = "/home/lute" } })
 print(r5.stdout)
 
 local r6 = process.run({ "pwd" }, { cwd = "/" })
 print(r6.stdout)
 
-local r7 = process.run("echo $0", { shell = true })
+local r7 = process.shell("echo $0")
 print(r7.stdout)
 
-local r8 = process.run("echo $0", { shell = "/bin/sh" })
+local r8 = process.shell("echo $0", { shell = "/bin/sh" })
 print(r8.stdout)
 
 process.exit(0)
--- a/examples/process.luau
+++ b/examples/process.luau
@@ -1,6 +1,8 @@
 local process = require("@std/process")
 local task = require("@std/task")
 
+process.system("echo hello", { cwd = {} })
+
 local result = process.run({ "echo", "Hello, lute!" })
 
 print(result.exitcode)
--- a/examples/process.luau
+++ b/examples/process.luau
@@ -16,19 +16,19 @@ print(r1.stdout)
 print(r2.stdout)
 print(r3.stdout)
 
-local r4 = process.shell("echo Hello, lute!")
+local r4 = process.system("echo Hello, lute!")
 print(r4.stdout)
 
-local r5 = process.shell("echo $HOME", { env = { HOME = "/home/lute" } })
+local r5 = process.system("echo $HOME", { env = { HOME = "/home/lute" } })
 print(r5.stdout)
 
 local r6 = process.run({ "pwd" }, { cwd = "/" })
 print(r6.stdout)
 
-local r7 = process.shell("echo $0")
+local r7 = process.system("echo $0")
 print(r7.stdout)
 
-local r8 = process.shell("echo $0", { shell = "/bin/sh" })
+local r8 = process.system("echo $0", { system = "/bin/sh" })
 print(r8.stdout)
 
 process.exit(0)
--- a/lute/process/src/process.cpp
+++ b/lute/process/src/process.cpp
@@ -350,28 +350,69 @@ ProcessOptions parseOptions(lua_State* L, int index)
         return opts;
 
     lua_getfield(L, index, "system");
-    if (lua_isstring(L, -1))
-        opts.customShell = lua_tostring(L, -1);
+    if (!lua_isnil(L, -1))
+    {
+        if (lua_isstring(L, -1))
+        {
+            opts.customShell = lua_tostring(L, -1);
+        }
+        else
+        {
+            luaL_error(L, "process.system option 'system' must be a string");
+        }
+    }
     lua_pop(L, 1);
 
     lua_getfield(L, index, "cwd");
     if (!lua_isnil(L, -1))
-        opts.cwd = lua_tostring(L, -1);
+    {
+        if (lua_isstring(L, -1))
+        {
+            opts.cwd = lua_tostring(L, -1);
+        }
+        else
+        {
+            luaL_error(L, "process option 'cwd' must be a string");
+        }
+    }
     lua_pop(L, 1);
 
     lua_getfield(L, index, "stdio");
-    if (lua_isstring(L, -1))
-        opts.stdioKind = lua_tostring(L, -1);
+    if (!lua_isnil(L, -1))
+    {
+        if (lua_isstring(L, -1))
+        {
+            opts.stdioKind = lua_tostring(L, -1);
+        }
+        else
+        {
+            luaL_error(L, "process option 'stdio' must be a string");
+        }
+    }
     lua_pop(L, 1);
 
     lua_getfield(L, index, "env");
-    if (lua_istable(L, -1))
+    if (!lua_isnil(L, -1))
     {
-        lua_pushnil(L);
-        while (lua_next(L, -2))
+        if (lua_istable(L, -1))
         {
-            opts.env[luaL_checkstring(L, -2)] = luaL_checkstring(L, -1);
-            lua_pop(L, 1);
+            lua_pushnil(L);
+            while (lua_next(L, -2))
+            {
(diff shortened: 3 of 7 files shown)


examples/process.luau
annieetang, line 13: tested on mine and sora's machine but these lines are erroring lol but it seems like a `task` issue but just noting it here

hgoldstein (approved):
LGTM: I realize that on other platforms the "shell" command is called "system," as in `os.system` in Python and `process.system` in Ruby. Might be worth aligning there.

lute/process/src/process.cpp
```diff
@@ -442,6 +334,171 @@ int run(lua_State* L)
...
+        {
+            lua_rawgeti(L, 1, i);
+            args.push_back(lua_tostring(L, -1));
+            lua_pop(L, 1);
+        }
+    }
+    else
+    {
+        args.push_back(lua_tostring(L, 1));
+    }
```
hgoldstein: This should _just_ be checking a table here right? Are we supporting both in each API?

lute/process/src/process.cpp
```diff
@@ -442,6 +334,171 @@ int run(lua_State* L)
...
+        {
+            lua_pushnil(L);
+            while (lua_next(L, -2))
+            {
+                env[luaL_checkstring(L, -2)] = luaL_checkstring(L, -1);
+                lua_pop(L, 1);
+            }
+        }
+        lua_pop(L, 1);
+    }
```
hgoldstein: This seems like it could be shared code.

hgoldstein (changes requested):
Sorry, should have looked at the implementations more closely, I have some comments.

lute/process/src/process.cpp
annieetang: oh wait I see what you mean, thanks for the catch will change bc process.system doesn't support a table

lute/process/src/process.cpp
```diff
@@ -442,6 +342,114 @@ int run(lua_State* L)
...
+        }
+    }
+    else
+    {
+        args.push_back(lua_tostring(L, 1));
+    }
+
+    if (args.empty() || args[0].empty())
+    {
+        luaL_error(L, "process.run requires a non-empty command");
```
hgoldstein: Should this be two different error messages? One for providing an empty set of arguments and one for providing the empty string as the initial argument?

lute/process/src/process.cpp
```diff
@@ -442,6 +342,114 @@ int run(lua_State* L)
...
+        {
+            lua_rawgeti(L, 1, i);
+            args.push_back(lua_tostring(L, -1));
+            lua_pop(L, 1);
+        }
+    }
+    else
+    {
+        args.push_back(lua_tostring(L, 1));
+    }
```
hgoldstein: I think this was from before, where you could pass a `string` or a `{ string }` to this function. Should this just check for a table and otherwise error?

lute/process/src/process.cpp
```diff
@@ -442,6 +342,114 @@ int run(lua_State* L)
...
+    lua_getfield(L, index, "env");
+    if (lua_istable(L, -1))
+    {
+        lua_pushnil(L);
+        while (lua_next(L, -2))
+        {
+            opts.env[luaL_checkstring(L, -2)] = luaL_checkstring(L, -1);
+            lua_pop(L, 1);
+        }
+    }
```
hgoldstein, line 393: What is the lute convention for error handling? e.g. what should I expect if I run:
```luau
process.system("echo hello", { cwd = {} })
```

hgoldstein (changes requested):

lute/process/src/process.cpp
annieetang: ah that is currently not getting caught and we seg fault so I'll add some error handling, thanks!

tests/std/process.test.luau
```diff
@@ -55,14 +71,97 @@ test.suite("ProcessSuite", function(suite)
...
 		assert.tableeq(r, {
 			exitcode = 41,
 			stdout = "",
 			stderr = "",
 			ok = false,
 		})
 	end)
+
+	suite:case("process_run_options_error_cases", function(assert)
+		local ok, err = pcall(function()
```
annieetang: wait nvm give me a sec to use `xpcall`

lute/std/libs/test/assert.luau
```diff
@@ -105,6 +106,19 @@ function assertions.buffereq(lhs: buffer, rhs: buffer)
...
+		error({ msg = `Function did not error as expected.` })
+	end
+
+	local actualErrorMessage = tostring(err)
+	-- pcall appends filename and line number of the error, so we check suffix only
+	if not stringext.hassuffix(actualErrorMessage, expectedErrorMessage) then
+		error({ msg = `Expected error message "{expectedErrorMessage}", but got "{actualErrorMessage}"` })
+	end
+end
+
```
annieetang, line 121: should i move this into a separate pr?

hgoldstein: It's fine to include this

lute/process/src/process.cpp
```diff
@@ -442,6 +342,159 @@ int run(lua_State* L)
     return lua_yield(L, 0);
 }
 
+ProcessOptions parseOptions(lua_State* L, int index)
+{
+    ProcessOptions opts;
+
+    if (!lua_istable(L, index))
+        return opts;
```
hgoldstein: The options here ought to be table or nil yeah? Should we error if you pass in something else?

lute/process/src/process.cpp
```diff
@@ -442,6 +342,159 @@ int run(lua_State* L)
...
+            lua_pushnil(L);
+            while (lua_next(L, -2))
+            {
+                if (!lua_isstring(L, -2) || !lua_isstring(L, -1))
+                {
+                    luaL_error(L, "process option 'env' must be a table of string keys and string values");
+                }
+                else
+                {
+                    opts.env[luaL_checkstring(L, -2)] = luaL_checkstring(L, -1);
```
hgoldstein: Is there a reason we're not using `luaL_checkstring` elsewhere?

lute/process/src/process.cpp
```diff
@@ -442,6 +342,159 @@ int run(lua_State* L)
...
+    if (!lua_isnil(L, -1))
+    {
+        if (lua_isstring(L, -1))
+        {
+            opts.customShell = lua_tostring(L, -1);
+        }
+        else
+        {
+            luaL_error(L, "process.system option 'system' must be a string");
+        }
```
hgoldstein: Do we want to add helpers like this for common Lua -> C++ conversions?

definitions/process.luau
```diff
@@ -1,7 +1,14 @@
 export type StdioKind = "default" | "inherit" | "none" | ""
```
hgoldstein: What does empty string mean here? Is it redundant with nil?

lute/process/src/process.cpp
```diff
@@ -442,6 +342,159 @@ int run(lua_State* L)
...
+        resolvedShell = opts.customShell;
+    }
+
+    std::vector<std::string> args;
+    args.clear();
+    args.emplace_back(resolvedShell);
+    args.emplace_back(shellArg);
+    args.emplace_back(command);
+
+    return executionHelper(L, args, opts);
```
hgoldstein:
```suggestion
    return executionHelper(L, {resolvedShell, shellArg, command}, opts);
```

lute/process/src/process.cpp
```diff
@@ -442,6 +342,159 @@ int run(lua_State* L)
...
+
+    if (args.empty())
+    {
+        luaL_error(L, "process.run requires a non-empty table of arguments");
+        return 0;
+    }
+    if (args[0].empty())
+    {
+        luaL_error(L, "process.run requires a non-empty command as the first argument");
+        return 0;
```
hgoldstein: For my own learning: is returning 0 meaningful here?

hgoldstein (changes requested):
Just some more questions about the implementation

lute/process/src/process.cpp
annieetang: oh that's a good point, I forgot about luaL_checkstring, I'll use that instead of custom error msgs ty

lute/process/src/process.cpp
annieetang: honestly no, I just checked when varun and I were implementing `io.cpp` and we didn't write a return after an error because the luaL_error exits anyways, so I'll also remove it from here

definitions/process.luau
annieetang: thanks for the catch! I suppose it is redundant with nil, so I'll remove :)

lute/process/src/process.cpp
```diff
@@ -442,6 +339,128 @@ int run(lua_State* L)
...
+    {
+        luaL_error(L, "process.run expects a table of arguments as the first parameter");
+    }
+
+    std::vector<std::string> args;
+    int len = lua_objlen(L, 1);
+    for (int i = 1; i <= len; i++)
+    {
+        lua_rawgeti(L, 1, i);
+        args.push_back(lua_tostring(L, -1));
```
hgoldstein: Let's make this `checkstring`: the function signature says we take a `{ string }` so we can hold people to that.

_Some_ dynamic languages will try to be nice for you here and convert anything string-like into a string, but I think that tends to be more of a bane than a boon. We can always revisit this later.

lute/std/libs/test/assert.luau
```diff
@@ -105,6 +106,19 @@ function assertions.buffereq(lhs: buffer, rhs: buffer)
...
+function assertions.erroreq(func: () -> (), expectedErrorMessage: string)
+	local success, err = pcall(func)
+	if success then
+		error({ msg = `Function did not error as expected.` })
+	end
+
+	local actualErrorMessage = tostring(err)
+	-- pcall appends filename and line number of the error, so we check suffix only
+	if not stringext.hassuffix(actualErrorMessage, expectedErrorMessage) then
+		error({ msg = `Expected error message "{expectedErrorMessage}", but got "{actualErrorMessage}"` })
```
hgoldstein: We should reference the suffix aspect, or try to chop the beginning of the error off if possible.

tests/std/process.test.luau
```diff
@@ -55,14 +60,69 @@ test.suite("ProcessSuite", function(suite)
...
 
 	suite:case("run_nonzero_exitcode", function(assert)
-		local r = process.run({ "sh", "-c", "exit 41" })
+		local r = process.system("sh -c 'exit 41'")
 		assert.tableeq(r, {
 			exitcode = 41,
 			stdout = "",
 			stderr = "",
 			ok = false,
 		})
```
hgoldstein: I think it's fine to leave this test as-is.

tests/std/process.test.luau
```diff
@@ -55,14 +60,69 @@ test.suite("ProcessSuite", function(suite)
...
+			process.run({ "echo", "hello" }, { env = "not-a-table" })
+		end, "process option 'env' must be a table")
+
+		assert.erroreq(function()
+			process.run({ "echo", "hello" }, "invalid_option")
+		end, "process options must be a table")
+
+		assert.erroreq(function()
+			process.run({ "echo", "hello" }, "not-a-table")
+		end, "process options must be a table")
```
hgoldstein, line 95: Maybe feedback for Vighnesh more than you. We should consider adding something like:
```luau
suite:casewith("process_run_options_error_cases" function (args, opts, msg)
    assert.erroreq(function()
        process.run(args, opts)
    end, msg)
end, {
    {{}, nil, "process.run requires a non-empty table of arguments"},
    { { "echo", "hello" }, { cwd = {} }, "invalid argument #-1 to 'run' (string expected, got table)" },
    -- ...
})
```
... this is an instance where not having first class tuples _really_ hurts.

lute/process/src/process.cpp
```diff
@@ -442,6 +339,128 @@ int run(lua_State* L)
...
+        lua_pop(L, 1);
+    }
```
hgoldstein, line 413: What happens when someone passes in a non-array table?

hgoldstein (approved):
This is probably fine, I think the only thing that sticks out to me is the error handling for something like:
```
process.run({ huh = 42 })
```
... and the error handling in general for deserializing user provided arguments into C++.

lute/std/libs/test/assert.luau
annieetang: the beginning error has the specific line number attached so I just thought it would be easier to check the suffix and ignore the beginning instead of do some `stringext.removeprefix` thing with pattern matching. what do you mean by referencing the suffix aspect? more than a comment?

lute/process/src/process.cpp
annieetang: `process.run({ huh = 42 })` errors with `process.run requires a non-empty table of arguments`

annieetang: oh but non array table with a string like `process.run({"hi"})` errors with `Failed to spawn process: no such file or directory`

lute/process/src/process.cpp
annieetang: can add this to https://github.com/luau-lang/lute/issues/611

lute/process/src/process.cpp
annieetang: addressed offline and added a test case for some of these and arraylike tables

lute/std/libs/test/assert.luau
annieetang: addressed offline!

annieetang:
will do a deeper look on error handling when working on https://github.com/luau-lang/lute/issues/611

PR #610, opened 2025-11-21
transform: Add output directory optional argument

### LINKED ISSUE #607: `lute transform` should have an `--output` flag
External request: when writing tests for `lute transform` it's convenient to be able to write the output to a separate file.

skberkeley:
Resolves #607 
--output allows specifying an output file and is only supported when a single input file is provided.

### DIFF
--- a/lute/cli/commands/transform/init.luau
+++ b/lute/cli/commands/transform/init.luau
@@ -130,8 +130,12 @@ local function applyMigration(
 		local toWrite = if typeof(result) == "string" then result else printer.printfile(parseresult, result)
 
 		if toWrite == types.DELETION_MARKER then
-			print(`Marking {pathStr} for deletion`)
-			table.insert(deletedPaths, pathStr)
+			if outputDirPath then
+				print(`Skipping writing {pathStr} as it was marked for deletion`)
+			else
+				print(`Marking {pathStr} for deletion`)
+				table.insert(deletedPaths, pathStr)
+			end
 		elseif not dryRun then
 			if outputDirPath then
 				assert(outputFilePath ~= nil)
--- a/lute/cli/commands/transform/init.luau
+++ b/lute/cli/commands/transform/init.luau
@@ -82,7 +82,7 @@ local function applyMigration(
 	paths: { pathLib.path },
 	options: { [string]: string | boolean },
 	dryRun: boolean,
-	outputDir: string?
+	outputPath: string?
 )
 	local deletedPaths = {}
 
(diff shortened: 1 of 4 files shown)


annieetang (commented):
is this creating the output file as the same name as the original file? do you think we could add an option to specify the specific output file name as well? in case users want to specify the file name differently too

skberkeley:
> is this creating the output file as the same name as the original file? do you think we could add an option to specify the specific output file name as well? in case users want to specify the file name differently too

I considered it, but `lute transform` can take in any number of input files, so I opted to just keep the original base file names instead

annieetang (approved):
lgtm, but I'd defer to maybe @hgoldstein for a check to see if it works with the external ask?

lute/cli/commands/transform/init.luau
```diff
@@ -94,6 +94,11 @@ local function applyMigration(
 		migration.initialize(ctx)
 	end
 
+	local outputDirPath = if outputDir then pathLib.parse(outputDir) else nil
+	if outputDirPath and not fs.exists(outputDirPath) then
+		error("Output directory does not exist: " .. pathLib.format(outputDirPath))
+	end
```
hgoldstein: I think it's fine to make the output directory if it doesn't exist (and it's parent directories).

lute/cli/commands/transform/init.luau
```diff
@@ -113,23 +118,30 @@ local function applyMigration(
...
+		local outputFilePath = nil
+		if outputDirPath then
+			local baseName = pathLib.basename(pathObj)
+			if baseName == nil then
+				error(`Could not determine basename for path {pathStr}`)
 			end
-		else
-			local replacements = result :: syntaxTypes.replacements
+			outputFilePath = pathLib.join(outputDirPath, baseName)
+		end
```
hgoldstein: So the behavior here would be:
```bash
$ lute transform codemod/init.luau --output-directory bing/ foo/bar/baz.luau
# We now make `bing/baz.luau`?
```

This feels error prone given that it's _really_ common to have a single filename all over the place: `init.luau`. If you want to do the output directory thing then I would have the filename be:
```
local output = pathlib.join(outputDirPath, pathObj)
```
You'll have to make sure `pathObj` is normalized, if it contains a `..` in it that might get odd.

hgoldstein (changes requested):
Two options here.

You could change this so that if `--output` is used with multiple file inputs we error. I think that's pretty reasonable behavior, and you dodge a lot of annoying questions about the behavior of this for multiple files.

If we're outputting to a directory then there's a couple things. High level, I'd rename this to `--output-directory`. `--output` is pretty consistently for files.

skberkeley:
> Two options here.
> 
> You could change this so that if `--output` is used with multiple file inputs we error. I think that's pretty reasonable behavior, and you dodge a lot of annoying questions about the behavior of this for multiple files.
> 
> If we're outputting to a directory then there's a couple things. High level, I'd rename this to `--output-directory`. `--output` is pretty consistently for files.

Discussed offline, we've decided to restrict `--output` to the case where only a single input file is specified!

hgoldstein (approved):

PR #615, opened 2025-11-24
lute lint: support multiple lint rules

skberkeley:
Adds support for passing multiple lint rules to `lute lint` by supporting either passing a single lint rule file or a directory containing multiple rules. I converted the rule positional argument into an optional one since at some point we'll presumably want to allow calling `lute lint` and getting a bunch of default rules.

### DIFF
--- a/lute/cli/commands/lint/init.luau
+++ b/lute/cli/commands/lint/init.luau
@@ -106,9 +106,12 @@ local function main(...: string)
 	for _, rule in lintRules do
 		tableext.extend(violations, rule.lint(ast, inputFilePath))
 	end
-	table.sort(violations, function(a, b)
-		return syntax.span.cmp(a.location, b.location)
-	end)
+	table.sort(
+		violations,
+		function(a: types.LintViolation, b: types.LintViolation) -- LUAUFIX: bidirecitonal inference should mean that annotations on a and b aren't needed
+		return a.location < b.location
+		end
+	)
 
 	print(`Printing {#violations} violations\n`)
 	printer.printLints(violations, fileContent)
--- a/lute/cli/commands/lint/init.luau
+++ b/lute/cli/commands/lint/init.luau
@@ -16,7 +16,7 @@ Lint the specified Luau file using the specified lint rule(s).
 
 OPTIONS:
   -h, --help              Show this help message
-  -r, --rules [RULE]      Path to a single lint rule or a folder containing lint rules. If a folder is provided, all .luau files will be assumed to be lint rules.
+  -r, --rules [RULE]      Path to a single lint rule or a folder containing lint rules. If a folder is provided, any subfolders containing init.luau files will be treated as modules exporting lint rules, while all other .luau files will be treated as individual lint rules.
 
 PATH:
   Path to the Luau file to be linted.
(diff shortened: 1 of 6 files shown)


lute/std/libs/syntax/init.luau
```diff
@@ -6,6 +6,17 @@ local luau = require("@lute/luau")
...
+		-- if a begins before b, a is less than b
+		-- if beginnings match, compare endings
+		if a.beginline < b.beginline or (a.beginline == b.beginline and a.begincolumn < b.begincolumn) then
+			return true
+		elseif a.beginline == b.beginline and a.begincolumn == b.begincolumn then
+			return a.endline < b.endline or (a.endline == b.endline and a.endcolumn < b.endcolumn)
+		else
+			return false
+		end
+	end,
```
hgoldstein: Please call this function `lessthan`, but can we attach a metatable to `span` instead?

lute/cli/commands/lint/init.luau
```diff
@@ -37,13 +37,36 @@ local function loadLintRule(path: path.pathlike): types.LintRule
 		`{path} must return a table with a 'lint' function property`
 	)
 
-	return loaded :: types.LintRule -- We have to cast because we can't assert that loaded.lint is a more specific functio type
+	return loaded :: types.LintRule -- We have to cast because we can't assert that loaded.lint is a more specific function type
```
hgoldstein: I know this is a comment from before but can you expand on this? It's because we can't assert that `loaded.lint` is specifically, say, a `(foo) -> bar`, only that it's a `function`

lute/cli/commands/lint/init.luau
```diff
@@ -37,13 +37,36 @@ local function loadLintRule(path: path.pathlike): types.LintRule
...
+			end
+
+			curr = walker()
+		end
+
+		return rules
+	else
+		return { loadLintRule(path) }
+	end
 end
```
hgoldstein, line 71: This feels like it could be awkward for lint rules that are defined in multiple files. What do you imagine the directory structure being here? Every lint rule in its own file?

lute/std/libs/fs.luau
```diff
@@ -140,7 +141,7 @@ end
 
 -- Note: for loops do not support yielding generalized iterators, so we cannot use fs.walk as `for path in fs.walk(...) do` directly. A while loop can be used instead.
 function fslib.walk(path: pathlike, options: walkoptions?): () -> path?
-	local queue = { pathlib.parse(path) }
+	local queue = { if typeof(path) == "string" then pathlib.parse(path) else path }
```
hgoldstein: Why do you need this? It looks like `pathlib.parse` can just early exit if it's given something that's already `pathlike`

hgoldstein (changes requested):
Mainly re: `loadLintRules`. We should probably have a slightly more flexible structure there.

lute/std/libs/syntax/init.luau
skberkeley: added a metamethod!

lute/cli/commands/lint/init.luau

lute/cli/commands/lint/init.luau
skberkeley: offlined, i've updated this to try to load lint rules from a module if it sees a `init.luau` file in a directory/subdirectory

lute/std/libs/fs.luau
skberkeley: this is a bit of a microoptimization, i can revert this change

lute/cli/commands/lint/init.luau
```diff
@@ -37,13 +37,43 @@ local function loadLintRule(path: path.pathlike): types.LintRule
...
+				local entries = fs.listdirectory(currentPath)
+				local childPaths = tableext.map(entries, function(entry)
+					return pathLib.format(pathLib.join(currentPathObj, entry.name))
+				end)
+				tableext.extend(queue, childPaths)
+			end
+		elseif pathLib.extname(currentPath) == ".luau" then
+			table.insert(rules, loadLintRule(currentPath))
+		end
+	end
```
hgoldstein, line 68: Does `walk` not do what you need?

hgoldstein (approved):

lute/cli/commands/lint/init.luau
skberkeley: I thought about it, but I think tracking whether we need to skip files if we load from a module would end up being more bug prone than managing the queue ourselves here

PR #616, opened 2025-11-24
stdlib: re-implement `fs.watch` as an iterator

annieetang:
As part of the API audit/cleanup, we reimplement `fs.watch` as an iterator instead of passing in a callback so you can use `fs.watch` on a task cooperatively.

This will then have the issue that `fs.walk` has where for loops do not support yielding generalized iterators, so we can only implement with a while loop for now.

Usage could be something like 

```lua

local watcher = fs.watch(tmpdir)

fs.writestringtofile(watched, "x")

local event
local start = os.clock()

repeat
	event = watcher:next()
	if not event then
		task.wait(0.01)
	end
until event or (os.clock() - start > 2)

if event then
	print("Event detected:", event.change and "change" or "rename")
else
	print("No event detected within the timeout period.")
end

watcher:close()

```

See full examples in `examples/watch_directory.luau` or `fs.test.luau:watch_iterator_on_change`

This PR also fixes a use-after-free in `task.cpp`- see https://github.com/luau-lang/lute/pull/616#issuecomment-3599127135 for details


Vighnesh-V:
Quick comment - can you also add the details of the UAF you fixed? I'm just curious

annieetang:
> Quick comment - can you also add the details of the UAF you fixed? I'm just curious

Yup! there was a UAF because of the `task.wait(0.01)` call ([log here](https://github.com/luau-lang/lute/actions/runs/19832872039/job/56823040680?pr=616)) that seems to be from `task.cpp` `yieldLuaStateFor` when it called `delete yield` in each complete yield call. I'm wondering if something about deleting the yield each time instead of when we close the handler at the end messes up the async calls between the watcher and the task delay? So I left deleting the yield to be in the `uv_close` callback

lute/task/src/task.cpp
```diff
@@ -54,12 +54,15 @@ static void yieldLuaStateFor(lua_State* L, uint64_t milliseconds, bool putDeltaT
                     if (yield->putDeltaTimeOnStack)
                         lua_pushnumber(L, static_cast<double>(uv_now(uv_default_loop()) - yield->startedAtMs) / 1000.0);
 
-                    delete yield;
                     return stackReturnAmount;
                 }
             );
 
-            uv_timer_stop(&yield->uvTimer);
```
Vighnesh-V: I'm not a 100% sure that this is correct. I think what this needs to do is stop the timer and then issue a `uv_close` without the callback _inside_ the `complete` callback, since that gets scheduled later. That's what I had to do to fix the UAF in fs.watch. @vrn-sn what do you think?

Vighnesh-V: I think we should also define a destructor for WaitData and move the logic for stopping the timer and invoking uv_close on the handle in there.

Vighnesh-V (changes requested):
Small change to how we fix the UAF.

lute/task/src/task.cpp
annieetang: hm ok I'll put the UAF into a separate PR

lute/fs/src/fs.cpp
annieetang, line 577: restored this but this means the user doesn't need to explicitly call `watcher:close()` now right? but they still have the option to

Vighnesh-V (approved):

PR #617, opened 2025-11-25
lute lint: Support multiple lintee files

skberkeley:
Update lute lint to support multiple input files by assuming that all non option arguments are paths to be linted. If a directory is passed, we recursively walk the directory for all the files to be linted. We error handle any files that error while linting so we can keep linting other files.

### DIFF
--- a/lute/cli/commands/lint/init.luau
+++ b/lute/cli/commands/lint/init.luau
@@ -19,7 +19,7 @@ OPTIONS:
   -r, --rules [RULE]      Path to a single lint rule or a folder containing lint rules. If a folder is provided, any subfolders containing init.luau files will be treated as modules exporting lint rules, while all other .luau files will be treated as individual lint rules.
 
 PATHS:
-  Path(s) to the Luau file(s) or folders containing Luau files to be linted.
+  Path(s) to the Luau file(s) or folders containing Luau files to be linted. Only files with .luau or .lua extensions will be linted.
 
 EXAMPLES:
   lute lint -r examples/lints/almost_swapped.luau bad_swap.luau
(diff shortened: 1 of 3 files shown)


lute/cli/commands/lint/init.luau
```diff
@@ -92,36 +114,46 @@ local function main(...: string)
...
-	end
-	table.sort(
-		violations,
-		function(a: types.LintViolation, b: types.LintViolation) -- LUAUFIX: bidirecitonal inference should mean that annotations on a and b aren't needed
-			return a.location < b.location
+		local curr = walker()
+		while curr ~= nil do
+			local inputFilePath = pathLib.parse(curr)
+
+			if pathLib.extname(inputFilePath) == ".luau" then
```
hgoldstein: You should extend this to support `.lua` files

tests/cli/lint.test.luau
```diff
@@ -224,6 +229,114 @@ violator.luau:3:1-4:6 ──
...
+		local result = process.run({ lutePath, "lint", "-r", "examples/lints", lintee })
+
+		-- Check
+		assert.eq(result.exitcode, 0)
+
+		local expected = [[
+lintee.luau': @std/syntax/parser.luau:12: parsing failed:
+(0, 0) - (0, 5): Incomplete statement: expected assignment or a function call
+]]
+		assert.neq(result.stdout:find(expected, 1, true), nil)
```
hgoldstein, line 407: We should probably _not_ crash the linter if one file has a parse error.

hgoldstein (changes requested):

tests/cli/lint.test.luau
skberkeley: i think i forgot to include multiple lintee files; i intended to demonstrate that we do gracefully continue

lute/cli/commands/lint/init.luau

hgoldstein (approved):

PR #618, opened 2025-11-25
lute lint: gracefully handle rules that error

skberkeley:
Currently, if a single lint rule errors, it'll exit the whole lint. If we wrap the lint rule calls in `pcall` instead, we can gracefully continue calling the other lint rules too.

### DIFF
--- a/tests/cli/lint.test.luau
+++ b/tests/cli/lint.test.luau
@@ -276,9 +276,8 @@ violator.luau:1:1-2:6 ──
 
 		assert.neq(
 			result.stdout:find(
-				"Error applying lint rule 'AlwaysErrors': ./lints/erroring_rule.luau:4: This rule always errors",
-				1,
-				true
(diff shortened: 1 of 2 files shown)


tests/cli/lint.test.luau
```diff
@@ -224,6 +224,69 @@ violator.luau:3:1-4:6 ──
...
+		-- Run the transformer on the transformee
+		local result = process.run({ lutePath, "lint", "-r", rulesDir, violatorPath })
+
+		-- Check
+		assert.eq(result.exitcode, 0)
+
+		assert.neq(
+			result.stdout:find("warning[almost_swapped]: This looks like a failed attempt to swap.", 1, true),
+			nil
+		)
```
hgoldstein, line 266: A bet a `stringcontains` helper would be warranted.

hgoldstein (approved):

tests/cli/lint.test.luau
skberkeley: yeah maybe i'll write a codemod to port the tests too lol

PR #619, opened 2025-11-25
add assert.strcontains

skberkeley:
what the title says
also updated test files which were using the `assert.neq(str:find(...), nil)` pattern to use the new helper instead

### DIFF
--- a/lute/std/libs/test/assert.luau
+++ b/lute/std/libs/test/assert.luau
@@ -119,6 +119,24 @@ function assertions.erroreq(func: () -> (), expectedErrorMessage: string)
 	end
 end
 
+function assertions.strcontains(haystack: string, needle: string, msg: string?, instances: number?)
+	if instances == nil then
+		instances = 1
+	end
+
+	assert(instances > 0, "instances must be greater than 0") -- LUAUFIX: Type stating should mean this comparison doesn't error.
+
+	if instances == 1 then
+		if haystack:find(needle, 1, true) == nil then
+			error(if msg then msg else `Expected "{haystack}" to contain "{needle}".`)
+		end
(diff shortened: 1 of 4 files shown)


lute/std/libs/test/assert.luau
```diff
@@ -119,6 +119,24 @@ function assertions.erroreq(func: () -> (), expectedErrorMessage: string)
 	end
 end
```
annieetang, line 120: oo I just merged in an `assert.erroreq` that's pretty similar, do you think you could remove that and update those usages to the `assert.stcontains` since it looks like this one covers any sort of outputted string

hgoldstein: I'm not sure what you mean here?

lute/std/libs/test/assert.luau
```diff
@@ -119,6 +119,24 @@ function assertions.erroreq(func: () -> (), expectedErrorMessage: string)
...
 
+function assertions.strcontains(haystack: string, needle: string, msg: string?, instances: number?)
+	if instances == nil then
+		instances = 1
+	end
```
hgoldstein, line 125: I just dug into this and it's a really annoying bug to fix, but you should be able to do:
```luau
instances = instances or 1
```

hgoldstein (approved):

lute/std/libs/test/assert.luau
hgoldstein: I'm not sure what you're referring to?

annieetang: I just meant we might be able to use assert.strcontains in the cases of assert.erroreq

PR #620, opened 2025-11-26
Battery for Text Diffing

wmccrthy:
Adds a `diffText` battery, that implements a myers-diff esque algorithm using the notion of an edit graph. An edit graph can be described as:
- src string along x axis
- destination string along y axis
- each edge in the graph represents a diff operation:
	- (n, n) -> (n+1, n) represents a deletion of src[n+1]
	- (n, n) -> (n, n+1) represents an addition of destination[n+1]
	- (n, n) -> (n+1, n+1) exists only if src[n+1] == destination[n+1]
- The shortest path from (0, 0) -> (len(src), len(destination)) represents the shortest-edit-sequence.

Myers-diff works by greedily following diagonal edges (n,n -> n+1,n+1) but the algorithm is not super intuitive / readable imo.

This implementation stays consistent with the greedy nature but varies slightly in how the graph is traversed. We use a deque to BFS the edit graph in a greedy fashion: whenever a node has a diagonal edge, we push it to the front of the deque, while other edges are always pushed to the back.

Here is what the output of `examples/diffText.luau` looks like:

Screenshot 2025-12-04 at 6 02 45 PM

**Follow up work:** support side-by-side diff printing


skberkeley (commented):
couple quick nits:
- `diffText` should be luacased
- it might be worth starting a `collections` module with `deque` as a submodule? maybe dogfood this in #proj-lute

batteries/collections/deque.luau
skberkeley, line 1: Is there a reason `deque` can't be a wrapper around an array and we use `table.insert` and `table.remove` for pushing/popping?

wmccrthy: I assumed insertion/removal at index 1 of an array will require O(n) time to adjust indices of all other array items. I would imagine `table` has some internal optimizations, but not sure exactly how it works in this case...
If it is O(n) for insertion/removal of index 1,it should be faster to use a linked-list for the deque

wmccrthy: Actually, I think I can simplify this implementation quite a bit by just using two pointers and a map-like table. There is no need for a linked list and all it's overhead

batteries/collections/deque.luau
```diff
@@ -0,0 +1,68 @@
+export type Deque<T> = {
+	values: { [number]: T },
+	pushfront: (self: Deque<T>, value: T) -> (),
+	pushback: (self: Deque<T>, value: T) -> (),
+	popfront: (self: Deque<T>) -> T,
+	popback: (self: Deque<T>) -> T,
+	peekfront: (self: Deque<T>) -> T,
+	peekback: (self: Deque<T>) -> T,
+	len: number,
```
skberkeley: can we add a `__len` metamethod for this?

batteries/collections/deque.luau
skberkeley, line 1: can you use a more object-oriented pattern here? ie point `__index` at a table and stuff

batteries/collections/deque.luau
```diff
@@ -0,0 +1,68 @@
...
+}
+
+local deque = {}
+
+function deque.new<T>(initValue: T?): Deque<T>
+	local self = {} :: Deque<T>
+	self.values = {}
+	local head, tail = 0, if initValue then 0 else -1
```
skberkeley: idt we need to use closures here; it seems cleaner to just add `__head` and `__tail` fields to the object itself, and if people mess with them so be it

batteries/collections/deque.luau
```diff
@@ -0,0 +1,68 @@
+export type Deque<T> = {
+	values: { [number]: T },
+	pushfront: (self: Deque<T>, value: T) -> (),
+	pushback: (self: Deque<T>, value: T) -> (),
+	popfront: (self: Deque<T>) -> T,
+	popback: (self: Deque<T>) -> T,
+	peekfront: (self: Deque<T>) -> T,
+	peekback: (self: Deque<T>) -> T,
```
skberkeley: if you're not erroring on an empty deque, this should return `T?`

batteries/difftext/difftext.luau
```diff
@@ -0,0 +1,193 @@
...
+	b: string,
+	options: {
+		byLine: boolean?,
+	}?
+)
+	if options and options.byLine then
+		return myersdiffbyline(a, b)
+	else
+		return myersdiffbychar(a, b)
+	end
```
skberkeley: my personal preference would be to diff by line as default, but maybe bike shed this in proj-lute

batteries/difftext/myersdiff.luau
```diff
@@ -0,0 +1,193 @@
...
+	if options and options.byLine then
+		return myersdiffbyline(a, b)
+	else
+		return myersdiffbychar(a, b)
+	end
+end
+
+return table.freeze({
+	rawdiff = diff,
+	applydiff = applyDiff,
```
skberkeley: are you using this for golden tests?

batteries/difftext/init.luau
skberkeley, line 1: maybe rename `difftext` to `init` instead of having a file that just exports?

skberkeley (changes requested):

batteries/difftext/myersdiff.luau
wmccrthy: I was planning to just use `resultStr == expectedStr` to check if golden test passes/fails and use `printdiff` to display output in the event of failure

batteries/difftext/init.luau
wmccrthy: I only have `init` bc `printdiff` and `difftext` are in separate files. `printdiff` uses `difftext` so we need `init` to avoid circular dependency had if export everything from `difftext.lua`. Maybe I'll just move the `printdiff` code into init? It's better for readability to keep raw diffing and printing logic separate, but I get your point

wmccrthy: actually, I have better idea

batteries/difftext/myersdiff.luau
skberkeley: is there a reason we need `applydiff` then?

skberkeley: sorry, just realized it's being used elsewhere

batteries/difftext/init.luau
```diff
@@ -0,0 +1,33 @@
+local myersdiff = require("./difftext/myersdiff")
+local printdiff = require("./difftext/printdiff")
+
+export type Diff = myersdiff.Diff
+export type ApplyDiff = (string) -> string
```
skberkeley: nit: luacase

batteries/difftext/init.luau
```diff
@@ -0,0 +1,33 @@
+local myersdiff = require("./difftext/myersdiff")
+local printdiff = require("./difftext/printdiff")
```
skberkeley:
```suggestion
local myersdiff = require("@self/myersdiff")
local printdiff = require("@self/printdiff")
```

tests/batteries/difftext.test.luau
```diff
@@ -0,0 +1,64 @@
...
+
+local function randomString(len: number)
+	local chars = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789 \n\t"
+	local result = {}
+	for i = 1, len do
+		local index = math.random(1, #chars)
+		table.insert(result, chars:sub(index, index))
+	end
+	return table.concat(result)
+end
```
skberkeley: do we really need random tests here? it seems benign here but i really don't want to introduce potential for flaky tests down the line

batteries/difftext/init.luau
```diff
@@ -0,0 +1,33 @@
...
+export type Diff = myersdiff.Diff
+export type ApplyDiff = (string) -> string
+
+local function diff(
+	a: string,
+	b: string,
+	options: {
+		byChar: boolean?,
+	}?
+): (Diff, ApplyDiff)
```
skberkeley: why do we need applydiff too? if it's just for the tests can't we just check the diff structurally?

batteries/difftext/myersdiff.luau
```diff
@@ -0,0 +1,185 @@
...
+	local x, y = node.x, node.y
+	if typeof(a) == "table" and typeof(b) == "table" then
+		return x < #a and y < #b and a[x + 1] == b[y + 1]
+	end
+	a = a :: string
+	b = b :: string
+	return x < #a and y < #b and a:sub(x + 1, x + 1) == b:sub(y + 1, y + 1)
+end
+
+export type DiffOperation = {
```
skberkeley: luacase

batteries/difftext/myersdiff.luau
```diff
@@ -0,0 +1,185 @@
...
+	key: "EQUAL" | "ADD" | "DELETE",
+	text: string,
+	position: number?,
+}
+
+export type Diff = { DiffOperation }
```
skberkeley: luacase

batteries/difftext/myersdiff.luau
```diff
@@ -0,0 +1,185 @@
...
+
+local function myersdiff(a: string | { string }, b: string | { string }): Diff
+	local start = editGraphNode(0, 0)
+	local editGraphDeque = deque.new(start)
+
+	local diff = {} :: Diff
+	local seen = {} :: { [number]: { [number]: true } }
+	-- lua dumb, have to do this weird mapping since we can't use EditGraphNode directly (it will use table memory address equivalence, not value equivalence for hashing)
+	seen[0] = {}
+	seen[0][0] = true
```
skberkeley: what would you do here ideally?

batteries/difftext/myersdiff.luau
```diff
@@ -0,0 +1,185 @@
...
+		if not seen[curNode.x + 1] then
+			seen[curNode.x + 1] = {}
+		end
+
+		-- we've reached bottom right vertex; indicates full edit from a -> b
+		if curNode.x == #a and curNode.y == #b then
+			-- reconstruct diff operations by tracing path from termination node to start
+			while curNode.prev do
+				local diffOperation = edgeToDiff(curNode, a, b)
+				table.insert(diff, 1, diffOperation)
```
skberkeley: this is O(n^2), why not construct diff in reverse and reverse the entire thing afterwards?

batteries/difftext/myersdiff.luau
```diff
@@ -0,0 +1,185 @@
...
+
+	while #editGraphDeque do
+		local curNode = editGraphDeque:popfront()
+
+		if not seen[curNode.x] then
+			seen[curNode.x] = {}
+		end
+		if not seen[curNode.x + 1] then
+			seen[curNode.x + 1] = {}
+		end
```
skberkeley: this feels weird to me; why are we populating the seen set for the node we're moving to, rather than the current one? also is the seen set being used for anything right now?

batteries/difftext/myersdiff.luau
```diff
@@ -0,0 +1,185 @@
...
+}
+
+export type Diff = { DiffOperation }
+
+local function edgeToDiff(curNode: EditGraphNode, a: string | { string }, b: string | { string }): DiffOperation
+	-- maps stepping from curNode.prev -> curNode to a diff operation
+	-- (n, n) -> (n+1, n) == DELETION of a[n]
+	-- (n, n) -> (n, n+1) == ADDITION of b[n]
+	-- (n, n) -> (n+1, n+1) == EQUAL (no diff)
+	local prev = curNode.prev :: EditGraphNode
```
skberkeley: can we return `nil` if `currNode.prev == nil` instead? this isn't making the fullest use of the type solver

batteries/difftext/init.luau
skberkeley: also, can we put collect the type definitions in a `types.luau` file? this is more in line with other parts of lute and also makes it easier to avoid circular dependencies

batteries/difftext/printdiff.luau
```diff
@@ -0,0 +1,112 @@
...
+	end
+	return result
+end
+
+local function printDiff(
+	a: string,
+	b: string,
+	options: {
+		detailed: boolean?,
+		-- sideBySide: boolean?
```
skberkeley: let's not commit commented code
```suggestion
```

batteries/difftext/printdiff.luau
```diff
@@ -0,0 +1,112 @@
...
+			if i <= #diff then
+				result ..= "\n"
+			end
+			i += 1
+		end
+	end
+	return result
+end
+
+local function printDiff(
```
skberkeley: this is kind of a misnomer since it's not actually printing
maybe either actually print the result or rename this to `difftostring`?

batteries/difftext/myersdiff.luau
```diff
@@ -0,0 +1,185 @@
...
+	end
+	a = a :: string
+	b = b :: string
+	return x < #a and y < #b and a:sub(x + 1, x + 1) == b:sub(y + 1, y + 1)
+end
+
+export type DiffOperation = {
+	key: "EQUAL" | "ADD" | "DELETE",
+	text: string,
+	position: number?,
```
skberkeley: is this unused?

batteries/difftext/printdiff.luau
skberkeley: it's a little weird to me that this calls `myersdiff`, maybe it should be passed a `diff` object and just do the work of converting that to a string

batteries/difftext/printdiff.luau
```diff
@@ -0,0 +1,112 @@
...
+	for i, op in diff do
+		result ..= if op.key == "ADD"
+			then richterm.brightGreen(`+ {op.text}`)
+			elseif op.key == "DELETE" then richterm.brightRed(`- {op.text}`)
+			else `{op.text}`
+		if i < #diff then
+			result ..= "\n"
+		end
+	end
+	return result
```
skberkeley: can you build up an array of strings and then return `table.concat(...)` instead?

tests/batteries/difftext.test.luau
skberkeley: if we do keep random tests around we should probably run each test more than once

batteries/difftext/printdiff.luau
skberkeley, line 1: can you add tests for this too?

skberkeley (changes requested):

batteries/difftext/myersdiff.luau
wmccrthy: i'd rather just store `seen[pos]` where pos is the current `x` and `y`. I don't think there's a way to do this with tuples in Lua and storing tables won't work

batteries/difftext/myersdiff.luau
wmccrthy: seen is used to not add previously visited / already queued nodes to the deque. I pre-populate seen[x+1] because it is [checked here](https://github.com/luau-lang/lute/pull/620/files#diff-89fa02d6cc5226545ff808ef6d51a46cc5789cf5b4c1b8b54cd0a512bdd98fd0R98-R106)

batteries/difftext/init.luau
wmccrthy: Yea i suppose we don't need it; my reasoning behind `applyDiff` and using random strings for tests was so that we could cover (in theory) a ton of different cases in a concise way. Though I think it's fine to manually construct a bunch of test cases and check the diff structure

batteries/difftext/printdiff.luau
wmccrthy: Fair. My thinking here is that in most use cases, people won't care about the raw diff result, they just want th nice printed output. So why make them do this:
```
local difftext = require("...")
local printdiff = require("...")
local diff = difftext(strA, strB)
local printedDiff = printdiff(diff)
```

when they can instead do:
```
local printdiff = require("...")
local printedDiff = printdiff(strA, strB)
```

wmccrthy: I agree that the name is misleading though

batteries/difftext/myersdiff.luau
skberkeley: ah i see, you could do `{x}-{y}` lol

wmccrthy: cheeky, i like that tbh, the 2d map is unclear and annoying

batteries/difftext/printdiff.luau
skberkeley: can you rename/luacase this? right now actually printing the diff will look like `print(printdiff(a, b))` which also looks a bit strange

batteries/difftext/myersdiff.luau
```diff
@@ -0,0 +1,132 @@
...
+end
+
+-- true if a[x+1] == b[y+1]; indicates an EQUAL operation (no diff) is valid
+local function hasDiagonalEdge(node: EditGraphNode, a: string | { string }, b: string | { string }): boolean
+	local x, y = node.x, node.y
+	if typeof(a) == "table" and typeof(b) == "table" then
+		return x < #a and y < #b and a[x + 1] == b[y + 1]
+	end
+	a = a :: string
+	b = b :: string
```
skberkeley: can you make this an exhaustive check rather than using typecasts? ie something like
```luau
if typeof(a) == "table" then
    if typeof(b) == "table" then
        ...
    else
        ...
    end
else
    ...
end
```

batteries/difftext/myersdiff.luau
```diff
@@ -0,0 +1,132 @@
...
+		local curNode = editGraphDeque:popfront()
+
+		-- we've reached bottom right vertex; indicates full edit from a -> b
+		if curNode.x == #a and curNode.y == #b then
+			-- reconstruct diff operations by tracing path from termination node to start
+			while curNode.prev do
+				local diffOperation = edgeTodiff(curNode, a, b) :: diffoperation
+				table.insert(diff, diffOperation)
+				curNode = curNode.prev
+			end
```
skberkeley: let's try not to use typecasts if we can; they clobber the information the type system is reporting
we could instead do something like
```luau
local currOp = edgeToDiff(curNode, a, b)
while currOp do
    table.insert(diff, currOp)
    curNode = curNode.prev
    currOp = edgeToDiff(curNode, a, b)
end
```

batteries/difftext/myersdiff.luau
```diff
@@ -0,0 +1,132 @@
...
+		end
+
+		-- add neighbors (diag prioritzed, then deletion, then insertion)
+		if hasDiagonalEdge(curNode, a, b) then
+			local edgePositionStr = `{curNode.x + 1},{curNode.y + 1}`
+			if not seen[edgePositionStr] then
+				seen[edgePositionStr] = true
+				editGraphDeque:pushfront(editGraphNode(curNode.x + 1, curNode.y + 1, curNode))
+			end
+			continue
```
skberkeley, line 100: might be worth commenting that we optimize by avoiding adding the vertical/horizontal edges when we can

batteries/difftext/myersdiff.luau
```diff
@@ -0,0 +1,132 @@
...
+			end
+		end
+	end
+
+	local i, j = 1, #diff
+	while i < j do
+		diff[i], diff[j] = diff[j], diff[i]
+		i = i + 1
+		j = j - 1
+	end
```
skberkeley: do you want to add this to `tableext`?

batteries/difftext/printdiff.luau
```diff
@@ -0,0 +1,106 @@
...
+		end
+	end
+
+	return srcVisual, destVisual
+end
+
+local function printDiffByLineDetailed(a: string, b: string): string
+	local diff = myersdiffbyline(a, b)
+	local i = 1
+	local result = {} :: { string }
```
skberkeley: a nit
```suggestion
	local result: { string } = {}
```

batteries/difftext/printdiff.luau
```diff
@@ -0,0 +1,106 @@
...
+			srcVisual ..= richterm.brightRed(op.text)
+			destVisual ..= richterm.brightGreen(op.text)
+		elseif op.key == "DELETE" then
+			srcVisual ..= richterm.bgRed(op.text)
+		elseif op.key == "ADD" then
+			destVisual ..= richterm.bgGreen(op.text)
+		end
+	end
+
+	return srcVisual, destVisual
```
skberkeley: can we use `table.concat` for `srcVisual` and `destVisual`?

tests/batteries/difftext.test.luau
```diff
@@ -0,0 +1,413 @@
...
+		assert.eq(diffResult[1].text, "line1")
+		assert.eq(diffResult[2].key, "DELETE")
+		assert.eq(diffResult[2].text, "line2")
+		assert.eq(diffResult[3].key, "ADD")
+		assert.eq(diffResult[3].text, "")
+	end)
+end)
+
+test.suite("printdiff", function(suite)
+	-- Helper function to strip ANSI color codes from output for easier testing
```
skberkeley: is it a massive pita to add color codes to the comparison strings?

skberkeley (commented):
minor comments, looking good!

tests/batteries/difftext.test.luau
wmccrthy: Nah should be fine, just rough on the eyes

batteries/difftext/myersdiff.luau
wmccrthy: I'm curious if there's a better way to handle this with the type system. While `a` and `b` can be a `string` or table of strings, we KNOW that a and b will always be of the same type (if a is a string, b is a string, if a is a table, b is a table); which makes doing the exhaustive route redundant (i guess we can just error in paths that won't be reached). Is there another way to enforce this constraint?

batteries/difftext/myersdiff.luau
wmccrthy: sure, pretty common usecase i suppose

wmccrthy: small enough i dont think i need to break out a separate PR for it, wdyt?

wmccrthy: should we differentiate between an in-place reversal and one that returns a new array?

skberkeley: yeah i think this PR is fine 
maybe `reverse<T>(arr: { T }, inplace: boolean?)`

batteries/difftext/myersdiff.luau
```diff
@@ -21,19 +22,26 @@ end
 -- true if a[x+1] == b[y+1]; indicates an EQUAL operation (no diff) is valid
 local function hasDiagonalEdge(node: EditGraphNode, a: string | { string }, b: string | { string }): boolean
 	local x, y = node.x, node.y
-	if typeof(a) == "table" and typeof(b) == "table" then
-		return x < #a and y < #b and a[x + 1] == b[y + 1]
+	if typeof(a) == "table" then
+		if typeof(b) == "table" then
+			return x < #a and y < #b and a[x + 1] == b[y + 1]
+		else
+			error("haDiagonalEdge inputs should have the same type.")
```
skberkeley: typo?

batteries/difftext/myersdiff.luau
```diff
@@ -21,19 +22,26 @@ end
...
+		end
+	elseif typeof(b) == "table" then
+		error("haDiagonalEdge inputs should have the same type.")
```
skberkeley: typo

batteries/difftext/myersdiff.luau
```diff
@@ -21,19 +22,26 @@ end
...
 	end
-	a = a :: string
-	b = b :: string
+
+	a = a
+	b = b
```
skberkeley: do we still need these?

lute/std/libs/tableext.luau
```diff
@@ -56,4 +56,15 @@ function tableext.keys<K, V>(tbl: { [K]: V }): { K }
 	return keys
 end
 
+function tableext.reverse<T>(tbl: { T }, inplace: boolean?)
+	local new = if inplace then tbl else {}
```
skberkeley, line 60: clever!

skberkeley (approved):

batteries/difftext/myersdiff.luau
wmccrthy: nope 🚀

tests/batteries/difftext.test.luau
Vighnesh-V: If you do want to do this, use math.randomseed + also run tests more than once and be careful to check for failure first without an assert, print the seed, and then produce the assert (that way we can repro exactly why stuff failed). If this is being used to assert a property, it's worth it imo.

PR #623, opened 2025-12-01
stdlib: Some fixups to assert functions

skberkeley:
- Update `strcontains` to use coalescing to assign to `instances`. This was a fix suggested by @~hgoldstein in another PR which got automerged over the break.
- Also updates `erroreq` to handle the cases where we actually want to pass stuff to `func`

### DIFF
--- a/lute/std/libs/test/assert.luau
+++ b/lute/std/libs/test/assert.luau
@@ -106,8 +106,8 @@ function assertions.buffereq(lhs: buffer, rhs: buffer)
 	end
 end
 
-function assertions.erroreq(func: () -> (), expectedErrorMessage: string)
-	local success, err = pcall(func)
+function assertions.erroreq<A...>(func: (A...) -> ...unknown, expectedErrorMessage: string, ...: A...)
+	local success, err = pcall(func, ...)
 	if success then
 		error({ msg = `Function did not error as expected.` })
 	end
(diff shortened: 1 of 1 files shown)


annieetang (approved):
ty!

lute/std/libs/test/assert.luau
```diff
@@ -120,18 +120,16 @@ function assertions.erroreq(func: () -> (), expectedErrorMessage: string)
 end
 
 function assertions.strcontains(haystack: string, needle: string, msg: string?, instances: number?)
-	if instances == nil then
-		instances = 1
-	end
+	instances = instances or 1
```
hgoldstein, line 123: For future reference (hopefully not for long) the root issue is that the new solver treats refinements and data flow as effectively disjoint systems. Before (in SSA form):
```luau
if instances_0 == nil then
    instances_1 = 1
end
instances_2 = phi(instances_0, instances_1)
assert(instances_2 > 0)
```
This is true in the sense of data flow, but _not_ in the sense of type flow. Really it ought to be:
```luau
if instances_0 == nil then
    instances_1 = instances_0 :: instances_0 & nil
    instances_2 = 1
else
    instances_3 = instances_0 :: instances_0 & ~nil
end
instances_4 = phi(instances_3, instances_2)
assert(instances_4 > 0)
```

hgoldstein (approved):

lute/std/libs/test/assert.luau
skberkeley: what does `phi` do in the context of types? is it unioning its arguments?

PR #626, opened 2025-12-02
Adds support for luaurc files in executables from `lute compile`

Vighnesh-V:
The `lute compile` command is built on the `resolveRequire` api. This is `.luaurc` aware, which means that we can resolve require statements that have aliases in them. 
However, while running the binary that `lute compile` emits, we aren't able to resolve these aliased requires at runtime, because we don't have access to the `.luaurc`. This PR adds support for this by:

1. Collecting the paths to .luaurc during static require tracing
2. Injecting this configuration into the final executable.
3. Reading the config back out when running the embedded scripts and navigating to directories based on the aliases stored.

Currently doesn't support config.luau.


aatxe:
The description of point 3 confuses me. We are _embedding_ the files that are locally available at that path into the bundle, right?

Vighnesh-V:
> The description of point 3 confuses me. We are _embedding_ the files that are locally available at that path into the bundle, right?

Yup, that's right. What I meant to say is that the `BundleVFS` uses those embedded configurations to resolve aliases. E.g., when the require system asks for "@myalias/foo", it'll use the embedded files to figure out what path "@myalias" means. When expanded, this will point to a file within the bundle as we resolved all these references during require tracing.

lute/require/src/bundlevfs.cpp
```diff
@@ -130,12 +149,37 @@ std::optional<std::string> BundleVfs::getContents(const std::string& path) const
...
 
 std::optional<std::string> BundleVfs::getConfig() const
 {
-    // Currently, we do not support .luaurc files in bundles.
+    LUAU_ASSERT(modulePath);
+
+    // Get the potential config path for the current module path
+    std::string configPath = modulePath->getPotentialConfigPath(".luaurc");
+
+    // Strip @bundle/ prefix if present
```
aatxe, line 175: This feels misleading, you're using the constant where the `/` is _not_ included.

aatxe (commented):
Can you add test where there's two luaurcs embedded as well?

lute/require/src/bundlevfs.cpp
Vighnesh-V: its a mistype :) - in the module path, the prefix is always @bundle, not `@bundle/`. I'll update the comment

Vighnesh-V: Whoops - I was the fool all along. Stripping `@bundle/` is correct, I had swapped the constants. I've updated all uses of `kBundlePrefix` -> `kBundlePrefixPath`, and the test with two luaurc's now works correctly.

Vighnesh-V:
I realized the compile end to end test was also not being invoked correctly. 

The code that retrieves the bundled bytecode from the end of the executable uses the process's exec path - e.g. the `lute test` binary that runs the test. However, in the test, we pass `cliMain(/path/to/separately/compiled/executable)`. This path was getting ignored, so we never extracted the bytecode bundle and executed it. The fix here is to ensure that the code that checks for the bytecode bundle reads from argv[0], so that it's behavior can be tested in unit tests by passing the appropriate path.

The test now correctly exercises the full compilation of `tests/src/staticrequires/main.luau`.

tests/src/staticrequires/.luaurc
```diff
@@ -0,0 +1,5 @@
+{
+    "aliases": {
+        "example": "./dep"
+    }
+}
```
aatxe:
```suggestion
}

```
trailing newline

aatxe (approved):

PR #627, opened 2025-12-02
lute lint: warning suppression comments

skberkeley:
Implement warning suppression by checking if any ancestor nodes contain leading trivia with comments which suppress the relevant rule. This lets us do nice things like suppress in the following case:
```luau
-- lute-lint-ignore(divide_by_zero)
if true then
	local y = 10 / 0
end
```

The story is a bit more complicated with violations which span multiple nodes: We check if there are any suppressing nodes which subsume the violation, _or_ whether the first node completely contained in the violation suppresses the warning. This is roughly similar to what Selene does, and allows suppressing cases like the following:
```luau
-- lute-lint-ignore(almost_swapped)
a = b
b = a
```
We'll need some follow up work to do (file) global warning suppression.

### DIFF
--- a/lute/cli/commands/lint/init.luau
+++ b/lute/cli/commands/lint/init.luau
@@ -113,6 +113,15 @@ local function maybeParseSuppressions(
 	local rules = cache[node]
 	if rules == nil then
 		rules = parseSuppressions(node)
+
+		if node.kind == "stat" and node.tag == "block" then
+			-- The logic here is a little different, since we only want suppression to apply to the first statement
+			-- Store a sentinel value at this node to avoid reparsing suppressions
+			cache[node] = {}
+			-- Parse the rules that this node suppresses and attach them to the first statement of the block
+			cache[node.statements[1]] = rules
+		end
+
 		cache[node] = rules
 	end
 	return rules
@@ -126,23 +135,6 @@ local function violationIsSuppressed(
 	local violationSuppressed = false
 
 	local function nodeIsSuppressed(node: syntax.AstNode): boolean
-		if node.kind == "stat" and node.tag == "block" then
-			-- The logic here is a little different, since we only want suppression to apply to the first statement
-			-- If the block is empty, early exit
-			if #node.statements == 0 then
-				return false
-			end
-
-			if suppressionCache[node] == nil then
-				-- Store a sentinel value at this node to avoid reparsing suppressions
-				suppressionCache[node] = {}
-				-- Parse the rules that this node suppresses and attach them to the first statement of the block
-				suppressionCache[node.statements[1]] = parseSuppressions(node)
-				-- Recurse into the block
-				return true
-			end
-		end
-
 		if syntax.span.subsumes(violation.location, node.location) then
 			-- The first node which is fully contained within a violation can suppress it
 			local rules = maybeParseSuppressions(node, suppressionCache)
@@ -158,11 +150,11 @@ local function violationIsSuppressed(
 			return false
 		end
 
-		local rules = suppressionCache[node]
-		if rules == nil then
-			-- Find all rules that this node suppresses
-			rules = parseSuppressions(node)
-			suppressionCache[node] = rules
+		local rules = maybeParseSuppressions(node, suppressionCache)
+
+		-- AstStatBlock suppressions should only apply to the first statement, not the block itself
+		if node.kind == "stat" and node.tag == "block" then
+			return true
 		end
 
 		if rules[violation.lintname] then
--- a/lute/cli/commands/lint/init.luau
+++ b/lute/cli/commands/lint/init.luau
@@ -106,7 +106,10 @@ local function parseSuppressions(node: syntax.AstNode): { [string]: boolean }
 	return suppressions
 end
 
-local function maybeParseSuppressions(node: syntax.AstNode, cache: { [syntax.AstNode]: { [string]: boolean } }): { [string]: boolean }
+local function maybeParseSuppressions(
+	node: syntax.AstNode,
+	cache: { [syntax.AstNode]: { [string]: boolean } }
+): { [string]: boolean }
 	local rules = cache[node]
 	if rules == nil then
 		rules = parseSuppressions(node)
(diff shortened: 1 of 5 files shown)


lute/cli/commands/lint/init.luau
```diff
@@ -89,6 +91,52 @@ local function loadDefaultRules(): { types.LintRule }
...
+		if syntax.span.subsumes(node.location, violation.location) then
+			local rules = suppressionCache[node]
+			if rules == nil then
+				-- Find all rules that this node suppresses
+				local leadingTrivia = triviaUtils.leftmosttrivia(node)
+				rules = {}
+				for _, trivia in leadingTrivia do
+					if trivia.tag == "blockcomment" or trivia.tag == "comment" then
+						for rule in trivia.text:gmatch("lute%-lint%-ignore%((.+)%)") do
+							print(`Found suppression for rule '{rule}'`)
```
hgoldstein: I would not include the print statement here, though maybe it's time to add a logging library to the batteries

lute/cli/commands/lint/init.luau
```diff
@@ -89,6 +91,52 @@ local function loadDefaultRules(): { types.LintRule }
 	return rules
 end
 
+local function violationIsSuppressed(
+	violation: types.LintViolation,
+	ast: syntax.AstStatBlock,
+	suppressionCache: { [syntax.AstNode]: { string } }
```
hgoldstein: Maybe make this a set instead?

lute/cli/commands/lint/printer.luau
```diff
@@ -37,10 +37,14 @@ local function printLint(lint: types.LintViolation, sourceLines: { string }): ()
 		local gutterSize = math.floor(math.log10(location.endline)) + 3 -- + 2 for padding around line number
 		local blankGutter = string.rep(" ", gutterSize)
 
-		local renderedSourceLines = { ` {location.beginline} │ ╭ {sourceLines[location.beginline]}` }
+		local renderedSourceLines =
+			{ ` {string.format(`%-{gutterSize - 2}d`, location.beginline)} │ ╭ {sourceLines[location.beginline]}` } -- LUAUFIX: the string.format call shouldn't error
```
hgoldstein: Why does it error here?

hgoldstein (changes requested):
Mainly: remove the errant print statement. It's a suppression, it shouldn't say anything.

lute/cli/commands/lint/printer.luau
skberkeley: TypeError: We cannot statically check the type of `string.format` when called with a format string that is not statically known. If you'd like to use an unchecked `string.format` call, you can cast the format string to `any` using `:: any`.

skberkeley: Which seems nonsensical because the first argument is an interpolated string, which we can statically say is a string

lute/cli/commands/lint/init.luau

lute/cli/commands/lint/init.luau
skberkeley: sg, will do

lute/cli/commands/lint/init.luau
```diff
@@ -89,6 +91,51 @@ local function loadDefaultRules(): { types.LintRule }
...
+				-- If the node is suppressed, we don't need to recurse deeper
+				return false
+			else
+				-- A deeper node might suppress this rule
+				return true
+			end
+		else
+			-- Violation doesn't come from this node, so don't recurse deeper
+			return false
+		end
```
hgoldstein: Minor style nit, can we early return here like:
```luau
if not syntax.span.subsumes(node.location, violation.location) then
	-- Violation doesn't come from this node, so don't recurse deeper
	return false
end
-- Rest of the code
```

lute/cli/commands/lint/printer.luau
hgoldstein: That error is correct. The point is that we don't know _what_ the interpolated string is: it could contain a hidden format specifier. For example what if you did:
```luau
local s = "%s"
print(string.format(`%d {s}`, 42))
```
I think we talked about moving this check to be a lint and it's more of an issue in non-managed languages (e.g. string formatting is a stack corruption machine).

tests/cli/lint.test.luau
```diff
@@ -446,6 +446,49 @@ violator.luau:3:1-4:6 ──
...
+		fs.writestringtofile(
+			violatorPath,
+			[[
+-- lute-lint-ignore(divide_by_zero)
+local x = 1 / 0
+
+-- lute-lint-ignore(divide_by_zero)
+if true then
+	local y = 10 / 0
+end
```
hgoldstein, line 462: Can we add tests here that show we still respect the lint when it's in between statements that are suppressed? e.g.:
```
-- lute-lint-ignore(divde_by_zero)
local x = 1 / 0

local y = 1 / 0 -- this should still have a lint violation.
```

hgoldstein (approved):
GTG but please add more tests.

lute/cli/commands/lint/printer.luau
skberkeley: gotcha, i'll remove the comments for now then

lute/cli/commands/lint/init.luau
```diff
@@ -82,13 +84,96 @@ local function loadDefaultRules(): { types.LintRule }
...
+
+		-- AstStatBlock suppressions should only apply to the first statement, not the block itself
+		if node.kind == "stat" and node.tag == "block" then
+			return true
+		end
+
+		if rules[violation.lintname] then
+			violationSuppressed = true
+			-- If the node is suppressed, we don't need to recurse deeper
+			return false
```
hgoldstein, line 163: I don't think this holds for if you have something like:
```luau
-- lute-lint-ignore(foo)
do
    local x = bar
    do
        -- lute-lint-ignore(bar)
        local y = baz
    end
end
```

hgoldstein (changes requested):
Sorry, can you please add a test for the comment I left

lute/cli/commands/lint/init.luau
hgoldstein: nvm I forgot that this occurs per lint rulle

hgoldstein (approved):

lute/cli/commands/lint/init.luau
skberkeley: added a test for nested lint rules!

PR #629, opened 2025-12-02
Infra: opt out of concurrency grouping for CI running on the `primary` branch

vrn-sn:
If two PRs are merged in to `primary` close to each other, our CI badge temporarily switches to a failing icon because our concurrency grouping logic cancels the first commit's CI run (treated as a failure). Instead, we opt out of concurrency grouping using the run's unique `run_id` if the current ref matches the ref pointed at by `primary`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,8 +6,10 @@ on:
   pull_request:
     branches: ["primary"]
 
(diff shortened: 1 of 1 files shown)


vrn-sn:
Tested the old concurrency group logic for PRs with the empty commit above, and it seems like it still works fine.

annieetang (approved):

Nicell (approved):

PR #630, opened 2025-12-02
Add `resolverequire` API to standard library with filepath semantics

vrn-sn:
Internally, we simply need to prepend a `@` to the filepath to create a valid chunkname, but the new API does this automatically to hide the chunkname implementation details from users.


annieetang (approved):
ty!! do you think you could update the use cases too 😅 (I think it's just the `typeofmodule` one, but I can throw up a PR for that if you'd like)

vrn-sn:
Going to do it in a separate PR just so I can merge this right after #629 and test if that PR is working correctly.

annieetang:
oh also I realized the other use case was the C++, not in Luau so nvm!

PR #636, opened 2025-12-03
Deque Battery

wmccrthy:
Adds a `deque` battery serving as the first member of a `collections` module.


batteries/collections/deque.luau
```diff
@@ -0,0 +1,74 @@
...
+	peekback: (self: Deque<T>) -> T?,
+}
+
+local deque = {}
+deque.__len = function<T>(self: Deque<T>)
+	return self.len
+end
+deque.__index = deque
+
+function deque.new<T>(initValue: T?): Deque<T>
```
skberkeley: we should probably have a `dequeLib` table with this as its only member, otherwise `.new` will be suggested as a method on deque objects

batteries/collections/deque.luau
```diff
@@ -0,0 +1,74 @@
+export type Deque<T> = {
+	__head: number,
+	__tail: number,
+	_values: { [number]: T },
+	len: number,
```
skberkeley: nit:
```suggestion
	_head: number,
	_tail: number,
	_values: { [number]: T },
```

batteries/collections/deque.luau
```diff
@@ -0,0 +1,74 @@
...
+deque.__index = deque
+
+function deque.new<T>(initValue: T?): Deque<T>
+	local self = {} :: Deque<T>
+	self._values = {}
+	self.__head, self.__tail = 0, if initValue then 0 else -1
+	self.len = if initValue then 1 else 0
+	if initValue then
+		self._values[self.__head] = initValue
+		self._values[self.__tail] = initValue
```
skberkeley:
```suggestion
		self._values[self.__head] = initValue
```

batteries/collections/deque.luau
skberkeley, line 1: i wonder if the invariant we want to maintain is `tail >= head` if the deque isn't empty; then we can do `#deque = if __tail < __head then 0 else __tail - __head + 1`, and we don't have to worry about managing len too

batteries/collections/deque.luau
```diff
@@ -0,0 +1,74 @@
...
+	local toReturn = self._values[self.__tail]
+	self._values[self.__tail] = nil
+	self.__tail -= 1
+	return toReturn
+end
+
+function deque.peekfront<T>(self: Deque<T>)
+	return self._values[self.__head]
+end
+function deque.peekback<T>(self: Deque<T>)
```
skberkeley, line 66: nit
```suggestion
end

function deque.peekback<T>(self: Deque<T>)
```

tests/batteries/collections/deque.test.luau
```diff
@@ -0,0 +1,120 @@
...
+		assert.eq(popped, 2)
+		assert.eq(deq.len, 0)
+	end)
+
+	suite:case("deque:popback", function(assert)
+		local deq = deque.new(2)
+		local popped = deq:popback()
+		assert.eq(popped, 2)
+		assert.eq(deq.len, 0)
+	end)
```
skberkeley, line 51: can we test popping from a deque with >2 elements?

tests/batteries/deque.test.luau
```diff
@@ -0,0 +1,120 @@
...
+		local deq2 = deque.new(1)
+		assert.eq(deq2.len, 1)
+		assert.eq(deq2:peekfront(), 1)
+		deq2:popfront()
+		assert.eq(deq2.len, 0)
+		assert.eq(deq2:peekfront(), nil)
+		assert.eq(deq2:peekback(), nil)
+	end)
+
+	suite:case("pushing from size 0 -> 1", function(assert)
```
skberkeley: these seem redundant with the previous two tests. is there additional functionality you're trying to test?

tests/batteries/deque.test.luau
```diff
@@ -0,0 +1,120 @@
...
+		assert.eq(deq:peekback(), nil)
+		deq:pushback(1)
+		assert.eq(deq.len, 1)
+		assert.eq(deq:peekfront(), 1)
+		assert.eq(deq:peekback(), 1)
+	end)
+
+	suite:case("popping from empty queue", function(assert)
+		local deq = deque.new()
+		assert.errors(function()
```
skberkeley: nit: we could use `assert.erroreq` to check the message too

tests/batteries/deque.test.luau
```diff
@@ -0,0 +1,120 @@
...
+		local deq = deque.new()
+		deq:pushback(1)
+		deq:pushback(2)
+		deq:pushback(3)
+
+		assert.eq(deq.len, 3)
+		assert.eq(deq:peekfront(), 1)
+		assert.eq(deq:peekback(), 3)
+
+		deq:popfront()
```
skberkeley:
```suggestion
		assert.eq(deq:popfront(), 1)
```

tests/batteries/deque.test.luau
```diff
@@ -0,0 +1,120 @@
...
+		assert.eq(deq.len, 2)
+		assert.eq(deq:peekfront(), 2)
+		assert.eq(deq:peekback(), 3)
+
+		deq:popback()
```
skberkeley:
```suggestion
		assert.eq(deq:popback(), 3)
```

skberkeley (changes requested):

batteries/collections/deque.luau
wmccrthy: good shout

batteries/collections/deque.luau
wmccrthy: I like that idea; I suppose theres no reason why we'd ever need the length of a deque too, should only ever care about the head and tail

batteries/collections/deque.luau
wmccrthy: If we only have a single element, shouldn't the head and tail both point to it?

skberkeley: aren't you just writing initValue to the same slot twice?

tests/batteries/deque.test.luau
wmccrthy: Nah, these are remnants from when I used a linked list and wanted to explicitly check that `next` and `prev` pointers updated correctly; seem entirely irrelevant now

batteries/collections/deque.luau
wmccrthy: oh yea lol, long day 😮‍💨

tests/batteries/collections/deque.test.luau
wmccrthy: should be covered by by the `e2e` case

tests/batteries/collections/deque.test.luau
skberkeley, line 1: can you move this to `test/batteries/collections/deque.test.luau`?

skberkeley (approved):
looks good! just need to move the tests to a collections folder

PR #637, opened 2025-12-03
lute lint: add report directive

skberkeley:
Adds a report directive which allows overriding an ignore directive from an outer scope. See test for example usage


lute/cli/commands/lint/init.luau
```diff
@@ -91,57 +91,61 @@ local function loadDefaultRules(): { types.LintRule }
...
 	end
 
-	return suppressions
+	return directives
 end
 
-local function maybeParseSuppressions(
+local function maybeParseDirectives(
 	node: syntax.AstNode,
 	cache: { [syntax.AstNode]: { [string]: boolean } }
```
hgoldstein, line 120: The types seem totally wrong here now, what's up with that?

skberkeley: I realized there's actually _3_ possible values a rule can map to: `nil`, `true`, and `false`, so we can use the following encoding:
- `nil`: The node doesn't have any directives corresponding to the given rule
- `true`: The node has a directive which suppresses the given rule
- `false`: The node has a directive which _un_suppresses the given rule

hgoldstein (approved):

PR #640, opened 2025-12-03
fix: fs.watch test

### LINKED ISSUE #639: flaky test: fs.test.luau:watch_iterator_on_change
event is sometimes `nil` 

```luau
suite:case("watch_iterator_on_change", function(assert)
	local watchedFileName = "watched.txt"
	local watched = path.join(tmpdir, watchedFileName)
	if fs.exists(watched) then
		fs.remove(watched)
	end

	local watcher = fs.watch(tmpdir)

	fs.writestringtofile(watched, "x")

	local start = os.clock()
	local event

	repeat
		event = watcher:next()
		if not event then
			task.wait(0.01)
		end
	until event or (os.clock() - start > 2)

	assert.neq(event, nil)
	assert.eq(event.change or event.rename, true)

	watcher:close()
	if fs.exists(watched) then
		fs.remove(watched)
	end
end)
```

annieetang:
fixes #639

### DIFF
--- a/tests/std/fs.test.luau
+++ b/tests/std/fs.test.luau
@@ -91,39 +91,32 @@ test.suite("FsSuite", function(suite)
 	end)
 
 	suite:case("watch_iterator_on_change", function(assert)
-		for i = 1, 10 do
-			print("Iteration", i)
-			local watchedFileName = "watched.txt"
-			local watched = path.join(tmpdir, watchedFileName)
-			if fs.exists(watched) then
-				fs.remove(watched)
+		local watchedFileName = "watched.txt"
+		local watched = path.join(tmpdir, watchedFileName)
+		if fs.exists(watched) then
+			fs.remove(watched)
+		end
+
+		local watcher = fs.watch(tmpdir)
+
+		fs.writestringtofile(watched, "x")
+
+		local event
+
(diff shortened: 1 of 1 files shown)


tests/std/fs.test.luau
```diff
@@ -93,23 +95,20 @@ test.suite("FsSuite", function(suite)
...
+		-- Note: we had a timeout for this loop before, but it was leading to a flaky test where event = nil, so we're removing it.
+		-- If this hangs indefinitely at, we'll see this test fail later and revisit.
 		local event
-
 		repeat
 			event = watcher:next()
 			if not event then
 				task.wait(0.01)
 			end
-		until event or (os.clock() - start > 2)
```
skberkeley, line 112: can we print how long this is taking so we can ballpark in case we need to revisit in the future?

tests/std/fs.test.luau
```diff
@@ -1,8 +1,10 @@
 local fs = require("@std/fs")
 local path = require("@std/path")
+local process = require("@std/process")
 local system = require("@std/system")
 local test = require("@std/test")
 local task = require("@lute/task")
+local tasklib = require("@std/task")
```
skberkeley: unused?

skberkeley (changes requested):

Vighnesh-V (approved):

skberkeley (approved):

PR #644, opened 2025-12-04
lute lint: respect gitignores

skberkeley:
Update lute lint to respect .gitignore files using the same infrastructure that `lute transform` already uses. I moved some files so that `lint` wasn't dependent on `transform`


tests/cli/lint.test.luau
```diff
@@ -872,6 +872,57 @@ end
...
+violator2.lua:1:11-17 ──
+   │
+ 1 │ local y = 10 / 0
+   │           ^^^^^^
+   │
+]]
+		assert.strnotcontains(result.stdout, expected)
+
+		-- Teardown
+		fs.removedirectory(modulePath, { recursive = true })
```
hgoldstein, line 970: Would it make sense to add this as a testing utility? I don't think Luau has any feature for scoped operations, though maybe Ariel has a trick up her sleeve. Feels like it's easy to forget and it's just boilerplate.

hgoldstein (approved):

tests/cli/lint.test.luau
skberkeley: you raise a good point, something like `suite:case():after()` would be nice because this teardown doesn't run if any of the asserts above fail

PR #647, opened 2025-12-04
Lute.Require: delete all "fake root" logic and replace with new `to_alias_fallback` Luau.Require API

vrn-sn:
### Problem

The "fake root" hack we use in the `require` virtual filesystem needs to be removed.

1. It's inefficient: it requires us to repeatedly copy an injected `.luaurc` around to require built-in libraries like `@std`.
2. It's difficult to reason about: forces our virtual filesystems to diverge from the structures they're meant to represent.
3. It's semantically incorrect:
    - The presence of a fake root node allows users to `require` "up" (using `../`) into fake filesystem nodes.
    - The existing hack also forces us to inject `.luaurc` files that don't conform to the official spec for `.luaurc` files (e.g. mapping `@std` to `$std` isn't permitted by the spec since aliases must map to paths, but the implementation allows this to avoid being opinionated on what a "valid path" could look like in an abstract context).

### Solution

In [Luau release 0.701](https://github.com/luau-lang/luau/releases/tag/0.701), we introduced a new `to_alias_fallback` API to natively support host-defined aliases, such as Lute's `@std` alias. This PR migrates Lute's require implementations to this new API. This includes `Package::RequireVfs`, which can now automatically redirect aliases to pre-declared dependencies when executing code in the package-aware virtual filesystem.

### DIFF
--- a/lute/require/include/lute/userlandvfs.h
+++ b/lute/require/include/lute/userlandvfs.h
@@ -3,7 +3,9 @@
 #include "lute/filevfs.h"
 #include "lute/modulepath.h"
 
-#include <map>
+#include "Luau/DenseHash.h"
+#include "Luau/StringUtils.h"
+
 #include <optional>
 #include <string>
 #include <utility>
@@ -17,9 +19,17 @@ struct Identifier
     std::string name;
     std::string version;
 
-    bool operator<(const Identifier& other) const
+    bool operator==(const Identifier& other) const
     {
-        return std::tie(name, version) < std::tie(other.name, other.version);
+        return std::tie(name, version) == std::tie(other.name, other.version);
+    }
+};
+
+struct IdentifierHashDefault
+{
+    size_t operator()(const Identifier& id) const
+    {
+        return Luau::detail::DenseHashDefault<std::string>()(Luau::format("%s:%s", id.name.c_str(), id.version.c_str()));
     }
 };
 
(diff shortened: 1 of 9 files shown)


lute/require/src/clivfs.cpp
```diff
@@ -23,6 +23,9 @@ static std::optional<std::string> readCliModule(const std::string& path)
 
 static bool isCliDirectory(const std::string& path)
 {
+    if (path == "@cli")
```
vrn-sn, line 26: We don't embed the root directory of the CLI commands into `CliVfs`, but the old "fake root" hack was previously hiding this broken behavior. This fixes it by effectively embedding the root directory under the (internal) modulepath `@cli`. This is probably the simplest way to fix this without needing to touch `luthier.luau`.

lute/require/include/lute/userlandvfs.h
```diff
@@ -73,8 +73,9 @@ class UserlandVfs
     std::string getCurrentPath() const;
 
 private:
-    UserlandVfs(std::map<Identifier, Info> allDependencies, std::string generatedRootLuaurc);
-    NavigationStatus jumpToDependencySubtreeImpl(Identifier identifier);
+    UserlandVfs(std::vector<Identifier> directDependencies, std::map<Identifier, Info> allDependencies);
```
andyfriesen: If you wanted, you could use `DenseHashMap` from `Luau/Common`

andyfriesen (commented):
I can't speak to the design, but the C++ all seems fine to me.

lute/require/include/lute/userlandvfs.h
vrn-sn: Done!

Vighnesh-V (approved):

PR #651, opened 2025-12-05
@std/test Tech Debt: Refactor how we provide debug information via assertions + make runtime error detection less brittle

Vighnesh-V:
This PR attempts to refactor the test library to be slightly more maintainable. While trying to bring https://github.com/luau-lang/lute/pull/519 up to date, I experienced a lot of brittleness with our test libraries ability to detect when a test failed due to runtime errors. The test runner also produces debug information in an adhoc manner multiple times in different places, which really sucks.

This PR ports over the [frktest](https://github.com/itsfrank/frktest/blob/main/src/assert_impl.luau) implementation of assertions. Rather than having `assertions` immediately yield control flow via `error`, all assertions internally return a typed `failure?`. When this failure is emitted, we extract all of the debug information needed for tests, exactly once at the point of the assertion 'failure'. This means that the test runner no longer has to know how to produce debug information (woo separation of concerns!), and the actual assertion implementation is responsible for this.  

Another freebie is that we only have to produce stacktraces in exactly one place - I've gone ahead and added that, along with updating the runtime error tests.

## Future Work
- The stacktraces currently produces are not very good - this is because `lute run` compiles the script with all optimizations on, and the inling negatively impacts the quality of stacktraces. I discussed an option with @vrn-sn to lower the optimization level for .test.luau files.
- The current implementation of `asserts` behaves like `REQUIRE` - control flow is immediately stopped at this point.
Since all assertion implementations are wrapped by a function that determines how to handle the control flow, we can now also provide an implementation of `CHECK`, where you can fail a test and resume control flow.

### DIFF
--- a/lute/std/libs/test/runner.luau
+++ b/lute/std/libs/test/runner.luau
@@ -2,9 +2,9 @@
 -- @std/test/runner
 -- Standard test runner library for Luau
 
-local types = require("./types")
 local asserts = require("./assert")
-local path = require("@std/path")
+local failure = require("./failure")
+local types = require("./types")
 
 type Test = types.test
 type TestCase = types.testcase
@@ -31,14 +31,11 @@ function runner.run(env: TestEnvironment): TestRunResult
 			for _, tc in suite.cases do
 				local failedtest: FailedTest = {
 					test = `{suite.name}.{tc.name}`,
-					assertion = "beforeall",
-					error = `beforeall hook failed: {err}`,
-					filename = "",
-					linenumber = 0,
+					failure = failure.lifecycle("beforeall", err),
 				}
 				table.insert(failures, failedtest)
 			end
-			return tostring(err)
+			return err
 		end
 	end
 
@@ -48,13 +45,10 @@ function runner.run(env: TestEnvironment): TestRunResult
 			failed += 1
 			local failedtest: FailedTest = {
 				test = testFullName,
-				assertion = "beforeeach",
-				error = `beforeeach hook failed: {err}`,
-				filename = "",
-				linenumber = 0,
+				failure = failure.lifecycle("beforeeach", err),
 			}
 			table.insert(failures, failedtest)
-			return tostring(err)
+			return err
 		end
 	end
 
@@ -63,13 +57,10 @@ function runner.run(env: TestEnvironment): TestRunResult
 		return function(err)
 			local failedtest: FailedTest = {
 				test = testFullName,
-				assertion = "aftereach",
-				error = `aftereach hook failed: {err}`,
-				filename = "",
-				linenumber = 0,
+				failure = failure.lifecycle("aftereach", err),
 			}
 			table.insert(failures, failedtest)
-			return tostring(err)
+			return err
 		end
 	end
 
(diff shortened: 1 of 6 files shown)


lute/std/libs/test/types.luau
```diff
@@ -3,6 +3,7 @@
 -- Standard test types library for Luau
 
 -- Test Reporting
+local failure = require("./failure")
 local assert = require("./assert")
```
skberkeley: unless there's a good reason not to, i think we should try to define all types in `types.luau`, and have `failure` and `assert` import those types from `types.luau`

lute/std/libs/test/runner.luau
```diff
@@ -3,7 +3,9 @@
 -- Standard test runner library for Luau
 
 local types = require("./types")
+local failure = require("./failure")
 local asserts = require("./assert")
```
skberkeley: nit: alpha sort

lute/std/libs/test/assert_impl.luau
skberkeley, line 1: does this need to be a separate file now?

lute/std/libs/test/runner.luau
```diff
@@ -25,87 +27,70 @@ function runner.run(env: TestEnvironment): TestRunResult
 	-- Error handler for beforeall hook
 	local function beforeallErrorHandler(suite: TestSuite)
 		return function(err)
+			local strerr = tostring(err)
```
skberkeley: this local's not doing a lot here

lute/std/libs/test/runner.luau
```diff
@@ -25,87 +27,70 @@ function runner.run(env: TestEnvironment): TestRunResult
...
-			return tostring(err)
+			return strerr
 		end
 	end
 
 	-- Error handler for beforeeach hook
 	local function beforeeachErrorHandler(testFullName: string)
 		return function(err)
 			failed += 1
+			local strerr = tostring(err)
```
skberkeley: do we need a new local here?

lute/std/libs/test/runner.luau
```diff
@@ -25,87 +27,70 @@ function runner.run(env: TestEnvironment): TestRunResult
...
 			table.insert(failures, failedtest)
-			return tostring(err)
+			return strerr
 		end
 	end
 
 	-- Error handler for aftereach hook
 	local function afterachErrorHandler(testFullName: string)
 		return function(err)
+			local strerr = tostring(err)
```
skberkeley: do we need a new local here?

lute/std/libs/test/runner.luau
```diff
@@ -25,87 +27,70 @@ function runner.run(env: TestEnvironment): TestRunResult
...
-			return tostring(err)
+			return strerr
 		end
 	end
 
 	-- Error handler for afterall hook
 	local function afterallErrorHandler(suiteName: string)
 		return function(err)
 			failed += 1
+			local strerr = tostring(err)
```
skberkeley: do we need a new local here?

lute/std/libs/test/types.luau
Vighnesh-V: I explicitly wrote out the type of `asserts` to break the cyclic dependency. This is fixed now.

lute/std/libs/test/assert_impl.luau
Vighnesh-V: nope - deleted.

lute/std/libs/test/runner.luau
Vighnesh-V: nope - also removed it for the other lifecycle handlers, since `failure.lifecycle` handles stringifying the error now.

skberkeley (approved):

PR #654, opened 2025-12-05
Fixes issues with gitignore parsing in commands lib

skberkeley:
I encountered some `.gitignore` parsing bugs when I was trying to run `lute lint` in the lint directory. I've:
- Ported over a fix that went into `luau-lang/codemod` but for some reason didn't make it into `lute transform` when I initially ported it
- Added some additional fixups needed in `.gitignore` parsing
- Removed some redundant rules from our actual `.gitignore`
- Added tests for `cli/lib/files` which is built on the ignore parsing. the ignore stuff doesn't have a very nice api surface, so i opted to test it using the `files` stuff instead
- Use system-specific path separators so it works on Windows (I would love to eventually port this stuff to just run on path objects so the OS stuff is abstracted away)

### DIFF
--- a/lute/cli/commands/lib/ignore.luau
+++ b/lute/cli/commands/lib/ignore.luau
@@ -166,7 +166,7 @@ local function matchesGlob(glob: Glob, filepath: string, isDirectory: boolean): 
 	-- both: at start of path OR after any directory separator
 	if glob.matchAnywhere then
 		local matchAtStart = filepath:match("^" .. glob.pattern) ~= nil
-		local matchAfterDir = filepath:match("/" .. glob.pattern) ~= nil
+		local matchAfterDir = filepath:match(separator .. glob.pattern) ~= nil
 
 		return matchAtStart or matchAfterDir
 	else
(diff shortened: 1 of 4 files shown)


tests/cli/lib/files.test.luau
```diff
@@ -0,0 +1,269 @@
...
+	suite:case("parse negation pattern", function(assert)
+		-- Setup
+		fs.createdirectory(testPath)
+		local gitignorePath = path.join(testPath, ".gitignore")
+		fs.writestringtofile(gitignorePath, "*.lua\n!important.lua\n")
+
+		local shouldIgnorePath = path.join(testPath, "debug.lua")
+		fs.writestringtofile(shouldIgnorePath, "")
+
+		local shouldNotIgnorePath = path.join(testPath, "important.lua")
```
annieetang: `.luau`
*insert :meow_glare:

lute/cli/commands/lib/ignore.luau
```diff
@@ -86,17 +89,20 @@ local function parseGlob(glob: string): Glob
...
 		elseif char == "?" then
-			-- Match any single character, except for /
-			table.insert(lua_parts, "[^/]")
+			-- Match any single character, except for directory separator
+			table.insert(lua_parts, `[^{separator}]`)
 			idx = idx + 1
 		elseif char == "-" then
 			table.insert(lua_parts, "%-")
 			idx = idx + 1
+		elseif char == "/" or char == "\\" then
```
annieetang, line 103: can we use `separator` here as well?

skberkeley: no, i want to avoid assuming which separator people are writing in their .gitignore files

annieetang (approved):

PR #657, opened 2025-12-06
Implements `lute pkgrun <file>`: a subcommand that runs Lute with package awareness

vrn-sn:
Implement a simple layer that bridges the gap between Loom's generated `loom.lock` file and Lute's `Package::RequireVfs` (introduced in #339).

Running a file with `lute pkgrun <file>` searches for a lockfile called `loom.lock` before executing `<file>`. We use regex matching to parse out the set of dependencies from the lockfile, which are expected to be installed in a `Packages` folder in the same directory as `loom.lock`. Using this, we construct a `Package::RequireVfs`, which automatically injects aliases to dependencies when executing user code.

Because of limitations with the current `loom.lock` format, there are still several improvements we need to make. I've left a few TODO comments related to this. Long term, regex matching is also not a great way to parse `loom.lock`, but I didn't want to embed a C++ TOML parser or deal with the complexities of embedding `@batteries/toml` and calling into it from C++. Instead, we can probably use Luau syntax for `loom.lock` since Lute can easily parse Luau tables. We could potentially base the format on [`.config.luau` files](https://rfcs.luau.org/config-luauconfig.html).


aatxe:
Worth noting batteries has a luau-implemented toml parser as well, which you could probably finesse a use of here, albeit maybe weirdly.

vrn-sn:
> Worth noting batteries has a luau-implemented toml parser as well, which you could probably finesse a use of here, albeit maybe weirdly.

Yeah, I had considered that. I ended up just going with the regex since it was a relatively simple temporary solution (we've been talking about replacing this with Luau syntax anyway). If the TOML parser were embedded into C++ like the `@lute` or `@std` libraries, it would be a bit easier to imagine—but even then, crossing the C++/Luau layer would be more complex than the regex.

aatxe (approved):

PR #658, opened 2025-12-08
Cleans up the std path library and adds `path.relative`

skberkeley:
I cleaned up the types in paths so that `path = setmetatable<pathdata, pathMT>`, which I think is a little more digestible. I also added a `path.relative` function, which I need to get my gitignore fixes working on Windows

### DIFF
--- a/lute/std/libs/path/posix/types.luau
+++ b/lute/std/libs/path/posix/types.luau
@@ -1,13 +1,11 @@
-export type pathMT = {
-	__tostring: () -> string,
-}
+local pathinterface = require("../pathinterface")
 
 export type pathdata = {
 	parts: { string },
 	absolute: boolean,
 }
 
-export type path = setmetatable<pathdata, pathMT>
+export type path = setmetatable<pathdata, pathinterface.pathinterface>
 
 export type pathlike = string | path | pathdata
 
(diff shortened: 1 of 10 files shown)


lute/std/libs/path/posix/types.luau
```diff
@@ -1,12 +1,14 @@
-export type pathinterface = {
+export type pathMT = {
```
aatxe:
```suggestion
export type pathinterface = {
```

lute/std/libs/path/types.luau
```diff
@@ -1,7 +1,7 @@
 local posixtypes = require("@std/path/posix/types")
 local win32types = require("@std/path/win32/types")
 
-export type path = posixtypes.path | win32types.path
+export type path = posixtypes.pathdata | win32types.pathdata
```
aatxe: Do we not wish for paths to still have their interface here...?

aatxe (commented):
My expectation here would be that there's a single interface that is used as the metatable type for _both_ `win32` and `posix` paths and that `posix.path`, for example, would be `setmetatable<posix.pathdata, pathinterface>` and the overarching path type would be either `posix.path | win32.path` or `setmetatable<posix.pathdata | win32.pathdata, pathinterface>`.

lute/std/libs/path/types.luau
skberkeley: oops, i missed this when refactoring, thanks for the catch

aatxe (approved):

PR #660, opened 2025-12-08
Surfaces some require errors in `lute run`

skberkeley:
I was running into a scenario where I was trying to `lute run` a file in a directory containing a folder with the same name as the file (ie `scratch.luau` and a folder called `scratch`). This was causing an ambiguous require, but lute would just report that `scratch.luau does not exist`, which was quite frustrating. I've added some piping to bubble the error up the stack, plus a test.

### DIFF
--- a/lute/cli/src/climain.cpp
+++ b/lute/cli/src/climain.cpp
@@ -43,7 +43,6 @@
 #include <Windows.h>
 #endif
 
-#include <iostream>
 #include <memory>
 #include <string>
 #include <vector>
@@ -310,7 +309,8 @@ static int assertionHandler(const char* expr, const char* file, int line, const 
     return 1;
 }
 
-static std::optional<std::string> getWithRequireByStringSemantics(std::string filePath)
+// Returns whether the filePath could be resolved to a valid file path, and an optional string containing either the valid path or an error message
+static std::pair<bool, std::optional<std::string>> getWithRequireByStringSemantics(std::string filePath)
 {
     std::string normalized = normalizePath(std::move(filePath));
 
@@ -324,26 +324,28 @@ static std::optional<std::string> getWithRequireByStringSemantics(std::string fi
 
     std::optional<ModulePath> mp = ModulePath::create(std::move(rootOfPath), std::move(restOfPath), isFile, isDirectory);
     if (!mp)
-        return std::nullopt;
+        return {false, "Could not initialize ModulePath instance."};
 
     ResolvedRealPath resolved = mp->getRealPath();
     if (resolved.status != NavigationStatus::Success)
-        return std::nullopt;
+        return {false, resolved.errMsg};
 
     if (resolved.type == ResolvedRealPath::PathType::File)
-        return resolved.realPath;
+        return {true, resolved.realPath};
 
-    return std::nullopt;
+    return {false, resolved.errMsg};
 };
 
-static std::optional<std::string> getValidPath(std::string filePath)
+// Returns whether the filePath could be resolved to a valid file path, and an optional string containing either the valid path or an error message
+static std::pair<bool, std::optional<std::string>> getValidPath(std::string filePath)
 {
-    if (std::optional<std::string> path = getWithRequireByStringSemantics(filePath))
-        return *path;
+    auto [ok, path] = getWithRequireByStringSemantics(filePath);
+    if (ok)
+        return {true, path};
 
     // Only fallback to checking .lute/* if the original path has no extension.
     if (filePath.find('.') != std::string::npos)
-        return std::nullopt;
+        return {ok, path};
 
     std::string fallbackPath = joinPaths(".lute", filePath);
     size_t fallbackSize = fallbackPath.size();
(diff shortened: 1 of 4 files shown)


lute/require/include/lute/modulepath.h
```diff
@@ -31,6 +31,7 @@ struct ResolvedRealPath
     std::string realPath;
     std::optional<std::string> relativePath;
     PathType type;
+    std::optional<std::string> errMsg;
```
vrn-sn:
```suggestion
```
Could we avoid surfacing the error message here? ModulePath is used in various places in our codebase with different purposes in each context, so I think we should leave the error message reporting to the consumer.

lute/require/src/modulepath.cpp
```diff
@@ -119,7 +119,15 @@ ResolvedRealPath ModulePath::getRealPath() const
     if (isADirectory(partialRealPath))
     {
         if (resolvedType)
-            return {NavigationStatus::Ambiguous};
+        {
```
vrn-sn: Same here, I think this should still return `{NavigationStatus::Ambiguous}` and leave the error message reporting to the consumer.

lute/cli/src/climain.cpp
```diff
@@ -310,7 +309,8 @@ static int assertionHandler(const char* expr, const char* file, int line, const
     return 1;
 }
 
-static std::optional<std::string> getWithRequireByStringSemantics(std::string filePath)
+// Returns whether the filePath could be resolved to a valid file path, and an optional string containing either the valid path or an error message
+static std::pair<bool, std::optional<std::string>> getWithRequireByStringSemantics(std::string filePath)
```
vrn-sn: Can we instead return a string and return an explicit error in all cases?

lute/cli/src/climain.cpp
```diff
@@ -324,26 +324,30 @@ static std::optional<std::string> getWithRequireByStringSemantics(std::string fi
 
     std::optional<ModulePath> mp = ModulePath::create(std::move(rootOfPath), std::move(restOfPath), isFile, isDirectory);
     if (!mp)
-        return std::nullopt;
+        return {false, "Could not initialize ModulePath instance."};
 
     ResolvedRealPath resolved = mp->getRealPath();
-    if (resolved.status != NavigationStatus::Success)
-        return std::nullopt;
+    if (resolved.status == NavigationStatus::Ambiguous)
```
vrn-sn: Small detail, but using a switch-case block here is slightly preferred for enums since it'll give us exhaustivity checking for free. Something like this:

```cpp
    std::pair<bool, std::string> result;
    switch (resolved.status)
    {
    case NavigationStatus::Success:
        if (resolved.type == ResolvedRealPath::PathType::File)
            result = {true, resolved.realPath};
        else
            result = {false, "Path is a directory, not a file."};
        break;
    case NavigationStatus::Ambiguous:
        result = {false, "Unable to tell whether path is a file or directory. Is there a same-named file or directory?"};
        break;
    case NavigationStatus::NotFound:
        result = {false, "File or directory not found."};
        break;
    }
    return result;
```

That example code also handles creating an explicit error message for what's currently the `std::nullopt` case (allowing us to remove the optional in the return type of this function).

lute/cli/src/climain.cpp
```diff
@@ -354,10 +358,11 @@ static std::optional<std::string> getValidPath(std::string filePath)
         fallbackPath += ext;
 
         if (isFile(fallbackPath))
-            return fallbackPath;
+            return {true, fallbackPath};
     }
 
-    return std::nullopt;
+
+    return {false, std::nullopt};
```
vrn-sn: Same in this function. We can just return `std::pair<bool, std::string>` to avoid the caller  needing to repeat the "file not found" message. (Also, maybe naming this `res` or something like that makes more sense than `path`, to be more similar to `pcall`'s interface and make a bit more sense in the erroring case.)

```suggestion
    return {false, path};
```

lute/cli/src/climain.cpp
```diff
@@ -400,10 +405,13 @@ int handleRunCommand(int argc, char** argv, int argOffset, LuteReporter& reporte
     Runtime runtime;
     lua_State* L = setupCliState(runtime);
 
-    std::optional<std::string> validPath = getValidPath(filePath);
-    if (!validPath)
+    auto [ok, validPath] = getValidPath(filePath);
+    if (!ok)
     {
-        reporter.formatError("Error: File '%s' does not exist.", filePath.c_str());
+        if (validPath)
```
vrn-sn: If we can remove the optional strings, then we just need a single error message path here (and no `if (validPath)` check, which is a bit easier to reason about.

lute/require/src/modulepath.cpp
```diff
@@ -119,7 +119,9 @@ ResolvedRealPath ModulePath::getRealPath() const
     if (isADirectory(partialRealPath))
     {
         if (resolvedType)
+        {
             return {NavigationStatus::Ambiguous};
+        }
```
vrn-sn:
```suggestion
            return {NavigationStatus::Ambiguous};
```
nit

vrn-sn (approved):

PR #661, opened 2025-12-09
Ensures that Lute will not block other runtime threads during basic file system operations (open, close, read, write)

### LINKED ISSUE #642: fs.read blocks the main lute runtime thread
```
local t = require("@lute/task")
local fs = require("@std/fs")
task.create(function()
      while true do
	 print("Woke up")
	 t.wait(0.5)
      end
      
end)
local f = fs.open("/dev/random", "r")
local x = fs.read(f)
print("finished")
fs.close(f)
```

If you read a bunch of bytes for a long time, that shouldn't block the printing of "Woke up". 

The fix is to pass a callback to uv_read in the fs.read function and yield the thread, so that the runtime can go to schedule other stuff and libuv can handle other i/o.

Vighnesh-V:
This PR ensures that `fs.(open, close, read, write)` are non-blocking - that is, other `Luau` threads that could continue to run, will not blocked by long i/o. This fixes https://github.com/luau-lang/lute/issues/642, whereby a very long read, e.g. to /dev/random could cause the main Lute Runtime thread to block synchronously (which would prevent stepping the uv event loop as well). I've tested this locally and it works as expected, but I'm not sure how to write a test for this in a cross platform way.

This PR also:
- Adds a base `UVRequest` type - responsible for interfacing with the lute runtime to resume the original luau thread. During destruction, will invoke the uv_*_cleanup function.
- Makes File Handles a lightuserdata.
- Rewrites `stdfs.writefiletostring` and `stdfs.readstringfromfile` in terms of the low level `lute` primitives that don't block anymore.
- Removes `readasync`, `writefiletostring` and `readstringfromfile` from the `lute` api surface.

### DIFF
--- a/definitions/fs.luau
+++ b/definitions/fs.luau
@@ -91,16 +91,4 @@ function fs.rmdir(path: string): ()
 	error("not implemented")
 end
 
-function fs.readfiletostring(filepath: string): string
-	error("not implemented")
-end
-
-function fs.writestringtofile(filepath: string, contents: string): ()
-	error("not implemented")
-end
-
-function fs.readasync(filepath: string): string
-	error("not implemented")
-end
-
 return fs
--- a/lute/fs/src/fs.cpp
+++ b/lute/fs/src/fs.cpp
@@ -11,6 +11,8 @@
 
 #include <cstdio>
 #include <cstring>
+
+#include "fs_impl.h"
 #ifdef _WIN32
 #include <direct.h>
 #else
@@ -69,6 +71,24 @@ const char* UV_DIRENT_TYPES[] = {
     UV_TYPENAME_CHAR,
     UV_TYPENAME_BLOCK,
 };
+
+static UVFile* getFileHandle(lua_State* L, int index)
+{
+    if (!lua_islightuserdata(L, index))
+    {
+        luaL_errorL(L, "Error: expected file handle");
+        return nullptr;
+    }
+
+    auto* handle = static_cast<UVFile*>(lua_tolightuserdata(L, index));
+    if (!handle)
+    {
+        luaL_errorL(L, "Error: invalid file handle");
+        return nullptr;
+    }
+
+    return handle;
+}
 
 std::optional<int> setFlags(const char* c, int* openFlags)
 {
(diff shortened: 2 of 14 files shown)


lute/fs/src/fs.cpp
```diff
@@ -70,6 +72,24 @@ const char* UV_DIRENT_TYPES[] = {
     UV_TYPENAME_BLOCK,
 };
 
+static UVFile* getFileHandle(lua_State* L, int index)
+{
+    if (!lua_islightuserdata(L, index))
+    {
+        luaL_errorL(L, "Error: expected file handle");
+        return nullptr;
+    }
```
aatxe, line 80:
```suggestion
    if (!lua_islightuserdata(L, index))
        luaL_errorL(L, "Error: expected file handle");
```

`luaL_errorL` is noret, no?

lute/fs/src/fs.cpp
```diff
@@ -70,6 +72,24 @@ const char* UV_DIRENT_TYPES[] = {
...
+
+    auto* handle = static_cast<UVFile*>(lua_tolightuserdata(L, index));
+    if (!handle)
+    {
+        luaL_errorL(L, "Error: invalid file handle");
+        return nullptr;
+    }
```
aatxe, line 86:
```suggestion
    if (!handle)
        luaL_errorL(L, "Error: invalid file handle");
```

lute/fs/src/fs.cpp
```diff
@@ -785,108 +659,6 @@ int listdir(lua_State* L)
...
-        offset += bytesWritten;
-        numBytesLeftToWrite -= bytesWritten;
-    } while (numBytesLeftToWrite > 0);
-
-    cleanup(writeBuffer, sizeof(writeBuffer), *handle);
-    return 0;
-}
-
 struct ResumeCaptureInformation
 {
```
aatxe: Is this struct deadcode now?

lute/fs/src/fs_impl.cpp
```diff
@@ -0,0 +1,261 @@
...
+
+    static FSClose* from(uv_fs_t* req)
+    {
+        return static_cast<FSClose*>(req->data);
+    }
+
+    UVFile* file = nullptr;
+};
+
+int open_impl(lua_State* L, const char* path, int flags, int mode)
```
annieetang, line 72: for the `_impl` functions, do we want to keep those in `fs.cpp` or is the goal to move all of it into `fs_impl.cpp`? I feel like we have a lot of files that are just wrappers calling each other and this might make it a little more confusing to track, since right now it's already std/fs -> lute/fs -> fs.cpp and now we're adding -> fs_impl.cpp?

annieetang: one thought is could we have the structs be defined in a separate file (maybe called something else instead of `fs_impl`) but keep implementation still in fs.cpp?

lute/runtime/include/lute/uvutils.h
```diff
@@ -21,4 +25,70 @@ typedef int (*BufferWriter)(char* buffer, size_t* size);
 // Abstracts away buffer management when getting strings from libuv functions.
 Luau::Variant<std::string, UvError> getStringFromUv(BufferWriter bufferWriter, size_t initialBufferSize = 256);
 
+// Cleanup functions for UV request types
+template<typename ReqT>
+void cleanup_uv_req(ReqT* req)
+{
+}
```
annieetang: (this is me still not being familiar with c++) do we need this when `uvutils.cpp` defines it as well?

lute/fs/src/fs_impl.cpp
```diff
@@ -0,0 +1,261 @@
...
+
+struct FSRead : FSRequest
+{
+    FSRead(lua_State* L, UVFile* file)
+        : FSRequest(L)
+        , file(file)
+    {
+        chunk.resize(4096);
+        iov = uv_buf_init(chunk.data(), chunk.size());
+        buffer.reserve(4096);
```
aatxe: can we pull `4096` here out into a constant for like `kChuckSize`? It's getting repeated everywhere.

definitions/fs.luau
```diff
@@ -91,16 +91,4 @@ function fs.rmdir(path: string): ()
...
-	error("not implemented")
-end
-
-function fs.writestringtofile(filepath: string, contents: string): ()
-	error("not implemented")
-end
-
-function fs.readasync(filepath: string): string
-	error("not implemented")
-end
```
annieetang, line 104: also this is so exciting tysm

lute/fs/src/fs_impl.cpp
```diff
@@ -0,0 +1,261 @@
...
+        flags,
+        mode,
+        [](uv_fs_t* req)
+        {
+            auto r = FSRequest::from(req);
+            auto result = req->result;
+            if (result < 0)
+            {
+                r->fail("Error opening file %s: %s", req->path, uv_strerror(result));
+                delete r;
```
aatxe: Why do we need a manual delete here? I would expect RAII to do exactly this.

lute/fs/src/fs_impl.cpp
```diff
@@ -0,0 +1,261 @@
...
+                [result](lua_State* L)
+                {
+                    auto* file = new UVFile();
+                    file->fd = result;
+                    lua_pushlightuserdata(L, file);
+                    return 1;
+                }
+            );
+
+            delete r;
```
aatxe: Why do we need a manual delete here? I would expect RAII to do exactly this.

lute/fs/src/fs_impl.cpp
```diff
@@ -0,0 +1,261 @@
...
+        );
+        delete r;
+        return;
+    }
+
+    // Append the read data to our buffer
+    r->buffer.insert(r->buffer.end(), r->chunk.begin(), r->chunk.begin() + bytesRead);
+
+    // Zero out chunk buffer before next read
+    std::fill(r->chunk.begin(), r->chunk.end(), 0);
```
aatxe, line 134: We're zeroing out the chunk because the read might not write the full size, right? i.e. it might write _fewer_ bytes into the chunk than a previous read did? If that's the right reasoning, you should describe that here so that other people can read and understand it, rather than commenting descriptions of the code.

aatxe: And if it's the wrong reasoning, you should still describe the right reasoning here. :)

lute/fs/src/fs_impl.cpp
```diff
@@ -0,0 +1,261 @@
...
+        handle->fd.value(),
+        [](uv_fs_t* req)
+        {
+            auto r = FSClose::from(req);
+            auto result = req->result;
+
+            if (result < 0)
+            {
+                r->fail("Error closing file: %s", uv_strerror(result));
+                delete r;
```
aatxe: Why do we need a manual delete here? I would expect RAII to do exactly this.

lute/fs/src/fs_impl.cpp
```diff
@@ -0,0 +1,261 @@
...
+            }
+
+            r->succeed(
+                [](lua_State* L)
+                {
+                    return 0;
+                }
+            );
+
+            delete r;
```
aatxe: Why do we need a manual delete here? I would expect RAII to do exactly this.

aatxe (commented):
Already looking a lot better, but some comments on this implementation.

lute/runtime/include/lute/uvutils.h
aatxe: It's templated, so yeah.

aatxe: Essentially, this is defining a default "for any type, it's empty." and the one in the `cpp` file is a specific instantiation of the template for `uv_fs_t*`.

Vighnesh-V: Yup. The feature is called explicit specialization in C++ (I think). Each UVRequest is parameterized by a different request type or handle, and each uv_request_t may or may not have its own cleanup method.

For example: `uv_fs_t` has `uv_fs_req_cleanup(req)`, and `uv_handle_t` has `uv_unref/uv_close()`. This lets us provide one implementation of these for each `UVRequest<T>` instantiation, but implement a generic `cleanup_uv_req` function that can be called in the destructor implementation.

I think the one annoying part of this design is that you won't get compiler warnings if you forgot to define a specialization.

lute/fs/src/fs_impl.cpp
Vighnesh-V: So, the motivation for fs_impl is that it shouldn't have any Luau -> C++ marshalling goop. That's why all the impls have all the data they need - they're not doing any parsing. I can see it going both ways, but I like this separation since when you're debugging it makes it pretty easy to stick a breakpoint in this file and see if we did all the parsing correctly.

lute/fs/src/fs_impl.cpp
Vighnesh-V: Yup - I've gone ahead and made this work via RAII + the other places.

lute/fs/src/fs.cpp
Vighnesh-V: Got rid of the return nullptr.

lute/fs/src/fs_impl.cpp
Vighnesh-V: yup!

lute/fs/src/fs.cpp
Vighnesh-V: Yup - I missed it.

Vighnesh-V:
I've made some small changes to how we create these requests. There is a little bit of nuance because the underlying C request struct will store a raw pointer in the data field, however, `std::unique_ptr` doesn't know the req_t has taken ownership of this. That's why at the end of issuing a request, we have to invoke `release()`.

To solve this problem, I've introduced a ScopedUVRequest (open to other names too). This is a struct that will call release() on the pointer it owns at the end of the block. When you're in the uv callback, you explicitly call `retake` to gain ownership of the unique_pointer, where it will be automatically cleaned up at the end of the callback. 

Read and write are unique cases because they require multiple libuv calls to complete. In this case, you move the unique_ptr back into one of these ScopedUVRequests which will automatically call `release()` signalling that the callback invoked by the subsequent libuv call will handle deletion (or continue to chain it).

lute/fs/src/fs.cpp
```diff
@@ -343,21 +222,14 @@ int open(lua_State* L)
...
     }
 
-    if (std::optional<FileHandle> result = openHelper(L, path, mode, &openFlags))
+    std::optional<int> modeFlags = setFlags(mode, &openFlags);
+    if (!modeFlags)
     {
-        createFileHandle(L, *result);
-        return 1;
+        luaL_errorL(L, "Error decoding mode: %s\n", mode);
+        return 0;
```
annieetang: no need to return right?

lute/fs/src/fs_impl.cpp
annieetang: sounds good! I was just wondering if we should name it something different but no pref

lute/fs/src/fs.cpp
Vighnesh-V: Yup, this is noret. I've fixed this.

annieetang (approved):
lgtm ty for doing this!!
will defer to @aatxe for final 👍  tho haha

aatxe (approved):

PR #668, opened 2025-12-09
CLI refactor: move out require setup code into separate source files

vrn-sn:
Move all setup for the various CLI-facing virtual filesystems into their own files: `requiresetup.{h,cpp}`. Our `climain.cpp` file was getting a bit confusing to parse (and would have gotten worse after merging #657).

Ran clang-format, and also made some trivial changes to clean up some other files in CLI.


aatxe (approved):

aatxe:
Non-user-visible code refactors probably don't need to be in the release notes, so they should be untagged.

vrn-sn:
Sounds good!

PR #672, opened 2025-12-10
Creates a new `@std/net` library with the `request` API exposed

vrn-sn:
This is needed for the package manager. Very barebones for now, and we will still need to expose the other parts of `@lute/net` under `@std/net`.

### DIFF
--- a/lute/std/libs/net.luau
+++ b/lute/std/libs/net.luau
@@ -10,7 +10,7 @@ export type metadata = luteNet.Metadata
 export type response = luteNet.Response
 
 function net.request(url: string, metadata: metadata?): response
-    return luteNet.request(url, metadata)
+	return luteNet.request(url, metadata)
 end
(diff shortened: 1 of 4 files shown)


lute/std/libs/net.luau
```diff
@@ -0,0 +1,16 @@
...
+local net = {}
+
+export type metadata = luteNet.Metadata
+export type response = luteNet.Response
+
+function net.request(url: string, metadata: metadata?): response
+	return luteNet.request(url, metadata)
+end
+
+return net
```
annieetang: could you do `table.freeze(net)`

annieetang (approved):

lute/std/libs/net.luau
vrn-sn: Done, also updated `luau.luau`.

PR #674, opened 2025-12-10
Fixes JSON string serialization and deserialization and adds tests

annieetang:
Added all of the tests for JSON strings from https://github.com/nst/JSONTestSuite, and fixes all of the bugs related to strings


lute/std/libs/json.luau
```diff
@@ -277,75 +276,105 @@ local function deserializeNumber(state: DeserializerState)
...
-
-local function deserializeString(state: DeserializerState): string
+	-- Skip opening quote "
 	state.cursor += 1
 
 	local startPos = state.cursor
 
-	if currentByte(state) == string.byte('"') then
-		state.cursor += 1
+	local result = {}
```
aatxe: Is there a reason we're using a table here and not a buffer? This code all looks _very_ prime for using a buffer.

annieetang: not really- I was just following some other function implementations but I'll use a buffer

aatxe (approved):

PR #678, opened 2025-12-11
Supports parsing Luau-syntax lock files in `lute pkgrun <file>`

vrn-sn:
We make use of the existing Luau-syntax configuration file extraction API to parse our Luau-syntax lock file (`loom.lock.luau`). The temporary regex hack has been deleted.

### DIFF
--- a/lute/cli/src/packagerun.cpp
+++ b/lute/cli/src/packagerun.cpp
@@ -1,9 +1,11 @@
 #include "lute/packagerun.h"
 
+#include "lute/userlandvfs.h"
+
 #include "Luau/FileUtils.h"
+#include "Luau/LuauConfig.h"
 
 #include <optional>
-#include <regex>
 #include <string>
 #include <string_view>
 
@@ -22,7 +24,7 @@ std::optional<std::string> getAbsolutePathToNearestLockfile(std::string entryFil
     std::optional<std::string> currentPath = getParentPath(entryFile);
     while (currentPath)
     {
-        std::string lockfilePath = joinPaths(*currentPath, "loom.lock");
+        std::string lockfilePath = joinPaths(*currentPath, "loom.lock.luau");
         if (isFile(lockfilePath))
             return lockfilePath;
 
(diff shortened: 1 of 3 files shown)


lute/cli/src/packagerun.cpp
```diff
@@ -32,19 +34,63 @@ std::optional<std::string> getAbsolutePathToNearestLockfile(std::string entryFil
...
+    // TODO: support timing out Luau-syntax configurations
+    std::optional<Luau::ConfigTable> configOpt = Luau::extractConfig(lockfileContents, {});
+    if (!configOpt)
+        return {};
+
+    Luau::ConfigTable config = std::move(*configOpt);
+    if (!config.contains("package"))
+        return {};
+
+    Luau::ConfigTable* packageTable = config["package"].get_if<Luau::ConfigTable>();
```
Vighnesh-V, line 65: is `config["package"]` safe to do?

vrn-sn: Yeah, since we check on the line above:
```cpp
if (!config.contains("package"))
        return {};
```

lute/cli/src/packagerun.cpp
```diff
@@ -32,19 +36,69 @@ std::optional<std::string> getAbsolutePathToNearestLockfile(std::string entryFil
...
+            return {};
+
+        if (!package->contains("name") || !package->contains("rev"))
+            return {};
+
+        const Luau::ConfigValue* name = (*package).find("name");
+        const Luau::ConfigValue* rev = (*package).find("rev");
+
+        const std::string* nameStr = name->get_if<std::string>();
+        const std::string* revStr = rev->get_if<std::string>();
```
Vighnesh-V: Can we just inine the ->get_if here since that is safe to do?

Vighnesh-V (changes requested):
Left one nit, after that it's good to go.

Vighnesh-V (approved):

lute/cli/src/packagerun.cpp
vrn-sn: Thanks!

PR #680, opened 2025-12-11
Adds CI jobs which call luthier with a bootstrapped lute

skberkeley:


aatxe (approved):

vrn-sn:
@skberkeley @aatxe when we tried this in the past, I think it slowed down all our CI quite a bit, just fyi. (Some CI jobs taking 10+ minutes). It might make more sense as a cron job.

aatxe:
@vrn-sn we just looked at the times and no-bootstrap macOS is almost as slow as bootstrap linux, so I think we're just gonna do linux bootstrap in CI as the compromise

PR #686, opened 2025-12-12
Fixes JSON string deserialization

annieetang:
There were some bug reports from @Vighnesh-V and @wmccrthy about deserialized objects returning `nil` when correctly accessed as a table, and that led to us discovering there were issues with the `deserializationString` logic I wrote in https://github.com/luau-lang/lute/pull/674 even though the test cases were passing. I essentially rewrote the logic to be more similar to the original implementation of it and just added invalid case checks to make it RFC compliant, so this new version now works with actual use cases (see the test case)

### DIFF
--- a/lute/std/libs/json.luau
+++ b/lute/std/libs/json.luau
@@ -278,6 +278,17 @@ local function deserializeNumber(state: DeserializerState)
 	return num
 end
 
+local function decodeSurrogatePair(high, low): string?
+	local highVal = tonumber(high, 16)
+	local lowVal = tonumber(low, 16)
+	if not highVal or not lowVal then
+		return nil -- Invalid
+	end
+	-- Calculate the actual Unicode codepoint
+	local codepoint = 0x10000 + ((highVal - 0xD800) * 0x400) + (lowVal - 0xDC00)
+	return utf8.char(codepoint)
+end
+
 local function deserializeString(state: DeserializerState): string
 	-- Must start with "
 	if currentByte(state) ~= string.byte('"') then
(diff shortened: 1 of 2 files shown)


lute/std/libs/json.luau
```diff
@@ -504,7 +498,7 @@ deserialize = function(state: DeserializerState): value
 	if string.match(fourChars, "^null") then
 		state.cursor += 4
 		return json.null
-	elseif string.match(fourChars, "^true") then
```
annieetang, line 507: oops one sec will fix this

lute/std/libs/json.luau
```diff
@@ -278,6 +278,17 @@ local function deserializeNumber(state: DeserializerState)
...
 end
 
+local function decodeSurrogatePair(high, low): string?
+	local highVal = tonumber(high, 16)
+	local lowVal = tonumber(low, 16)
+	if not highVal or not lowVal then
+		return nil -- Invalid
+	end
+	-- Calculate the actual Unicode codepoint
+	local codepoint = 0x10000 + ((highVal - 0xD800) * 0x400) + (lowVal - 0xDC00)
```
Vighnesh-V, line 288: These are very magic numbers to me - is there something I can read to understand these better? Would you be able to add a more detailed comment here explaining why this works (or is this just a standard unicode transform).

lute/std/libs/json.luau
```diff
@@ -289,89 +300,71 @@ local function deserializeString(state: DeserializerState): string
...
+			state.cursor += 1
+			local source = string.sub(state.src, startPos, state.cursor - 2)
 
-		-- End of string
-		if c == '"' then
-			state.cursor = startPos + 1
-			return buffer.tostring(buf, bufferPos)
-		end
+			-- Validate escape sequences
+			local temp = string.gsub(source, "\\\\", "") -- Remove valid \\ first
```
Vighnesh-V: style nit: move this comment above the line pls.

lute/std/libs/json.luau
```diff
@@ -440,6 +433,7 @@ local function deserializeObject(state: DeserializerState): object
 	state.cursor += 1
 
 	local current = {}
+	current[object_key] = true -- mark to identify as object
```
Vighnesh-V: Should we only be doing this when we've successfully parsed the object?

Vighnesh-V (changes requested):

lute/std/libs/json.luau
Vighnesh-V: also, - comment isn't super useful, can probably get rid of it.

lute/std/libs/json.luau
annieetang: I unfortunately am not 100% sure either this is copied and pasted from the original code haha

tests/std/json.test.luau
```diff
@@ -55,11 +55,29 @@ test.suite("JSONParsingTestSuite", function(suite)
...
+	end)
+
+	suite:case("json_asobject_test", function(assert)
+		local myjsonstr = [[
+		{
+			"a": true,
+			"b": {
+				"c": true
+			}
+		}
```
Vighnesh-V, line 68: one last comment - can you extend this test case so that there is an array keyed by "d" inside of the table "b"? Can you show that iterating over the array produces the correct keys.

Vighnesh-V (approved):
Looks good, just one more addition to the test case.

PR #692, opened 2025-12-14
Fix `vm.create` child VMs missing native libs and make spawned runtimes libuv-safe

### LINKED ISSUE #691: Lute libraries are not available from loaded vms
While upgrading luau in the FrameworkBenchmarks, I noticed [the simple parallel code](https://github.com/TechEmpower/FrameworkBenchmarks/tree/master/frameworks/Luau/lute) no longer works. It seems like the built-in libraries are unavailable when used from a module loaded via VM.

 This seems to be as of 0.1.0-nightly.20251020.

```
./src/parallel.luau:7: Failed to spawn, target module error: ./src/serve.luau:1: 'net' is not a lute library[C] function require
./src/serve.luau:1

stacktrace:

stacktrace:
[C] function create
./src/parallel.luau:7
```

Nicell:
## Summary
This PR fixes a regression introduced in `0.1.0-nightly.20251020` where `vm.create()` child VMs did not have native `@lute/*` modules registered (e.g. `@lute/net`). As a result, child VMs would fall back to `definitions/*.luau` and error with `"not implemented"` (repro: `examples/parallel_serve.luau`). Closes #691.

Additionally, once child VMs can load native libraries again, multiple runtimes running on different OS threads must not share and concurrently drive `uv_default_loop()`. This PR gives each spawned runtime its own libuv loop and wires the native libraries to use the runtime’s loop.

## User-visible fixes
- `examples/parallel_serve.luau` works again (no longer hits `definitions/net.luau: not implemented` when running `@lute/net` inside a `vm.create()` VM).
- Spawned VMs that use libuv-backed native libraries (net/task/fs/process/io) no longer rely on the process-global default loop across multiple threads.

## Key changes
- `lute/vm/src/spawn.cpp`
  - Register native `@lute/*` modules in child VMs (matching the main CLI runtime behavior).
  - Spawned runtimes allocate a dedicated libuv loop and are registered for process-lifetime tracking.

- `lute/runtime/include/lute/runtime.h`, `lute/runtime/src/runtime.cpp`
  - `Runtime` now owns an optional dedicated `uv_loop_t` and exposes `getUvLoop()`.
  - Added a small spawned-runtime registry (`registerSpawnedRuntime`, `waitForSpawnedRuntimes`) so the CLI can stay alive while spawned runtimes still have work.

- `lute/cli/src/climain.cpp`
  - After the main script finishes, wait for spawned runtimes to stop having work (since child VMs no longer contribute to `uv_default_loop()` activity).

- libuv-backed libraries now use the calling runtime’s loop instead of `uv_default_loop()`:
  - `lute/net/src/net.cpp` (also minimal locking around global server maps for multi-VM access)
  - `lute/task/src/task.cpp`
  - `lute/fs/src/fs.cpp`, `lute/fs/src/fs_impl.cpp`
  - `lute/process/src/process.cpp`
  - `lute/io/src/io.cpp`

## Tests
- Added regression coverage for `vm.create()` module loading:
  - `tests/src/vmcreate.test.cpp`
  - `tests/src/vm/vm_requirer.luau`
  - `tests/src/vm/vm_helper.luau`

## Repro / Verification
- Repro (previously failed on `20251020`+):  
  `lute tools/luthier.luau run lute -- ./examples/parallel_serve.luau`

- Run tests:  
  `lute tools/luthier.luau run Lute.Test`

## Notes
This avoids undefined behavior from multiple runtime threads concurrently creating handles on and calling `uv_run()` on the single process-global `uv_default_loop()` (and the associated uWebSockets loop).

_This was worked on by Codex for an hour, neat. I need to spend some time reviewing everything before opening this from draft_

### DIFF
--- a/lute/fs/src/fs.cpp
+++ b/lute/fs/src/fs.cpp
@@ -233,7 +233,8 @@ int open(lua_State* L)
 int fs_remove(lua_State* L)
 {
     uv_fs_t unlink_req;
-    int err = uv_fs_unlink(uv_default_loop(), &unlink_req, luaL_checkstring(L, 1), nullptr);
+    uv_loop_t* loop = reinterpret_cast<uv_loop_t*>(getRuntime(L)->getUvLoop());
+    int err = uv_fs_unlink(loop, &unlink_req, luaL_checkstring(L, 1), nullptr);
     uv_fs_req_cleanup(&unlink_req);
 
     if (err)
@@ -248,7 +249,8 @@ int fs_mkdir(lua_State* L)
     int mode = luaL_optinteger(L, 2, 0777);
 
     uv_fs_t req;
-    int err = uv_fs_mkdir(uv_default_loop(), &req, path, mode, nullptr);
+    uv_loop_t* loop = reinterpret_cast<uv_loop_t*>(getRuntime(L)->getUvLoop());
+    int err = uv_fs_mkdir(loop, &req, path, mode, nullptr);
     uv_fs_req_cleanup(&req);
 
     if (err)
@@ -262,7 +264,8 @@ int fs_rmdir(lua_State* L)
     const char* path = luaL_checkstring(L, 1);
 
     uv_fs_t rmdir_req;
-    int err = uv_fs_rmdir(uv_default_loop(), &rmdir_req, path, nullptr);
+    uv_loop_t* loop = reinterpret_cast<uv_loop_t*>(getRuntime(L)->getUvLoop());
+    int err = uv_fs_rmdir(loop, &rmdir_req, path, nullptr);
     uv_fs_req_cleanup(&rmdir_req);
 
     if (err)
@@ -276,7 +279,8 @@ int fs_stat(lua_State* L)
     const char* path = luaL_checkstring(L, 1);
 
     uv_fs_t stat_req;
-    int err = uv_fs_stat(uv_default_loop(), &stat_req, path, nullptr);
+    uv_loop_t* loop = reinterpret_cast<uv_loop_t*>(getRuntime(L)->getUvLoop());
+    int err = uv_fs_stat(loop, &stat_req, path, nullptr);
 
     if (err)
     {
@@ -353,7 +357,8 @@ int fs_copy(lua_State* L)
     auto* req = new uv_fs_t();
     req->data = new ResumeToken(getResumeToken(L));
 
-    int err = uv_fs_copyfile(uv_default_loop(), req, path, dest, 0, defaultCallback);
+    uv_loop_t* loop = reinterpret_cast<uv_loop_t*>(getRuntime(L)->getUvLoop());
+    int err = uv_fs_copyfile(loop, req, path, dest, 0, defaultCallback);
 
     if (err)
     {
@@ -373,7 +378,8 @@ int fs_link(lua_State* L)
     auto* req = new uv_fs_t();
     req->data = new ResumeToken(getResumeToken(L));
 
-    int err = uv_fs_link(uv_default_loop(), req, path, dest, defaultCallback);
+    uv_loop_t* loop = reinterpret_cast<uv_loop_t*>(getRuntime(L)->getUvLoop());
+    int err = uv_fs_link(loop, req, path, dest, defaultCallback);
 
     if (err)
     {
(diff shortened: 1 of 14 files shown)


aatxe:
cc @vrn-sn this isn't obsoleted by the changes you made, right?

vrn-sn:
> cc @vrn-sn this isn't obsoleted by the changes you made, right?

Partially, this part is solved by my previous PR:
- Register native @lute/* modules in child VMs (matching the main CLI runtime behavior).

The rest of the logic in this PR that relates to libuv is unrelated to what I merged in.

lute/runtime/include/lute/runtime.h
```diff
@@ -14,6 +14,7 @@
 #include <vector>
 
 struct lua_State;
+struct uv_loop_s;
```
Vighnesh-V, line 17: This is an internal libuv type - you'll want to store `uv_loop_t`

lute/runtime/include/lute/runtime.h
```diff
@@ -44,6 +45,9 @@ struct Runtime
     Runtime();
     ~Runtime();
 
+    bool useDedicatedUvLoop();
+    uv_loop_s* getUvLoop() const;
```
Vighnesh-V, line 49:
```suggestion
    uv_loop_t* getUVLoop() const;
```

lute/runtime/include/lute/runtime.h
```diff
@@ -92,6 +99,8 @@ struct Runtime
     std::thread runLoopThread;
 
     std::atomic<int> activeTokens;
+
+    bool ownsUvLoop = false;
```
Vighnesh-V, line 103: We could do one runtime, one loop - see my comment below, since we might not actually need this?

lute/runtime/include/lute/runtime.h
```diff
@@ -77,6 +81,9 @@ struct Runtime
     // Shorthand for global state
     lua_State* GL = nullptr;
 
+    // Event loop for this runtime; defaults to `uv_default_loop()`, but can be dedicated via `useDedicatedUvLoop`.
+    uv_loop_s* uvLoop = nullptr;
```
Vighnesh-V, line 85: This adds a lot of complexity. Instead of defaulting, just initialize a `uv_loop` manually for this runtime.

lute/runtime/src/runtime.cpp
```diff
@@ -36,6 +84,54 @@ Runtime::~Runtime()
...
+        uvLoop = nullptr;
+    }
+}
+
+bool Runtime::useDedicatedUvLoop()
+{
+    if (ownsUvLoop)
+        return true;
+
+    uv_loop_t* loop = new uv_loop_t();
```
Vighnesh-V, line 119: can move into the RuntimeConstructor, along with the uv_loop_init call.

lute/runtime/src/runtime.cpp
```diff
@@ -7,8 +7,55 @@
...
 
+#include <algorithm>
 #include <assert.h>
+#include <chrono>
+#include <thread>
 #include <string>
+#include <vector>
+
+static std::mutex spawnedRuntimeMutex;
+static std::vector<std::weak_ptr<Runtime>> spawnedRuntimes;
```
Vighnesh-V, line 18: These shouldn't be static data( I don't think we need a mutex here either?) Each runtime is going to be on it's own event loop.

lute/fs/src/fs.cpp
```diff
@@ -233,7 +233,8 @@ int open(lua_State* L)
 int fs_remove(lua_State* L)
 {
     uv_fs_t unlink_req;
-    int err = uv_fs_unlink(uv_default_loop(), &unlink_req, luaL_checkstring(L, 1), nullptr);
+    uv_loop_t* loop = reinterpret_cast<uv_loop_t*>(getRuntime(L)->getUvLoop());
```
Vighnesh-V, line 236: Probably don't want to re-interpret cast here - this should just be a uv_loop_t.

Vighnesh-V: + ditto for all the places below.

Vighnesh-V:
Left some preliminary feedback - I think with Varuns PR we can make this one considerably smaller.

> This avoids undefined behavior from multiple runtime threads concurrently creating handles on and calling uv_run() on the single process-global uv_default_loop() (and the associated uWebSockets loop).

I'm not sure there is undefined behaviour - the default loop is single threaded right? So each coroutine only runs on the main thread, and continuations are only processed in the main thread when `uv_run` is called.

Vighnesh-V (changes requested):

PR #694, opened 2025-12-15
Make `@std` and `@lute` aliases nonoverridable

vrn-sn:
We use the new `to_alias_override` Luau.Require callback to ensure that the `@std` and `@lute` aliases always point at the embedded libraries, not ones on disk that the user may have set up using `.luaurc` or `.config.luau` files.

@Vighnesh-V, this means that `@std/test` will always point at the embedded module now.

### DIFF
--- a/lute/require/src/packagerequirevfs.cpp
+++ b/lute/require/src/packagerequirevfs.cpp
@@ -1,5 +1,7 @@
 #include "lute/packagerequirevfs.h"
 
+#include "lute/modulepath.h"
+
 #include "Luau/FileUtils.h"
 
 #include "lua.h"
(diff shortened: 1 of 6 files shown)


lute/require/src/packagerequirevfs.cpp
```diff
@@ -76,8 +78,15 @@ NavigationStatus RequireVfs::toAliasFallback(lua_State* L, std::string_view alia
         return NavigationStatus::Success;
     }
 
-    vfsType = VFSType::Userland;
-    return userlandVfs.toAliasFallback(aliasUnprefixed);
```
Vighnesh-V, line 80: hype!

Vighnesh-V (approved):

PR #695, opened 2025-12-16
Hierarchical run command discovery

jkelaty-rbx:
When invoking lute, it will attempt to discover a script named with the provided argument, or within a `.lute` directory. This change supports the same discovery mechanism, but additionally supports discovering scripts in hierarchical `.lute` directories.

As a practical example, I provided a luthier shorthand to demonstrate how this mechanism can be used. Instead of `lute tools/luthier.luau ...args` we can now invoke `lute luthier ...args`, and this can now be invoked from within any nested directory of lute.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,7 @@
 __pycache__
 .cache
 .clangd
+**/.DS_Store
 compile_commands.json
 *~
 extern/*/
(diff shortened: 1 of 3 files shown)


aatxe (approved):

aatxe:
Great addition, thanks!

PR #696, opened 2025-12-16
Enables the `@lute` libraries in spawned VMs by implementing a new `LuteVfs`

### LINKED ISSUE #691: Lute libraries are not available from loaded vms

vrn-sn:
Resolves #691. As @Nicell points out in #692, we have some libuv memory safety issues that have been revealed with this change, so they'll need to be fixed outside of this PR.


Nicell (commented):
could you add a test case where we require a lute module through a VM spawn?

Nicell (approved):

PR #698, opened 2025-12-18
Add `tableext.toset`; reorder methods for better grouping

Alexander-L-Robert:
Felt like tableext should have a `toset` util.

I reordered tableext methods to group dictionary and array methods more distinctly 
(`keys` and `toset` are the boundary as they convert between the two table formats)

This repo seems to avoid comments where possible so I left out labeling those sections.

EDIT: I'm now realizing there's no ideal way to do this. The current implementation of toset of an array of numbers can create sparse arrays. 

We could possibly:
- expect users to treat this like a dictionary, even if it creates a sparse array.
- use `tostring()` to convert to string keys and not support tables. Update the signature to only allow primitive types that can be converted via tostring().
- create an object that would store the original type of the list then when trying to access the key, it would convert the key back to that type for its metaproperties like `__pair`.
  - to support tables, it would have to deep compare so identical tables with different pointers could still index the same key. Maybe add an optional arg that would allow for a custom comparison function for tables this if that's the original list's value type.

### DIFF
--- a/lute/std/libs/tableext.luau
+++ b/lute/std/libs/tableext.luau
@@ -67,8 +67,8 @@ function tableext.reverse<T>(tbl: { T }, inplace: boolean?)
 	return new
 end
 
-function tableext.toset<T>(tbl: { T }): { [T]: boolean }
-	local set = {}
+function tableext.toset<T>(tbl: { T }): { [T]: true }
+	local set: { [T]: true } = {}
 	for _, v in tbl do
 		set[v] = true
 	end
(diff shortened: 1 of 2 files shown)


lute/std/libs/tableext.luau
```diff
@@ -56,6 +46,24 @@ function tableext.keys<K, V>(tbl: { [K]: V }): { K }
 	return keys
 end
 
+function tableext.toset<T>(tbl: { T }): { [T]: true }
+	local set: { [T]: true } = {}
```
Alexander-L-Robert: I was getting an inference error without this type assignment
```
TypeError: Type
	'{| [T]: boolean |}'
could not be converted into
	'{ [T]: true }'
caused by:
  Property '[indexer value]' is not compatible.
Type 'boolean' could not be converted into 'true' in an invariant contextLuau[1000](https://luau-lang.org/typecheck)
```

PR #699, opened 2025-12-18
Add `tableext.toset`

Alexander-L-Robert:
Felt like tableext should have a `toset` util.

### DIFF
--- a/tests/std/tableext.test.luau
+++ b/tests/std/tableext.test.luau
@@ -62,6 +62,27 @@ test.suite("TableExt", function(suite)
 		assert.tableeq(b, { 4, 5, 6 })
 		assert.tableeq(c, { 6, 5, 4 })
 	end)
+
+	suite:case("toset", function(assert)
+		local a = { 1, 2, 3, 2, 1 }
+		local setA = tableext.toset(a)
+		assert.eq(setA[1], true)
+		assert.eq(setA[2], true)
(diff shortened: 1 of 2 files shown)


tests/std/tableext.test.luau
```diff
@@ -62,6 +62,39 @@ test.suite("TableExt", function(suite)
...
+
+		local foundKeys: { number } = {}
+		for key, value in setA do
+			assert.eq(value, true)
+			table.insert(foundKeys, key)
+		end
+		assert.eq(#foundKeys, setSize)
+		assert.eq(setA[foundKeys[1]], true)
+		assert.eq(setA[foundKeys[2]], true)
+		assert.eq(setA[foundKeys[3]], true)
```
Alexander-L-Robert, line 89: Sanity check to verify that sets created using an array of numbers doesn't break from having gaps between the number keys.

This repo seems to avoid comments where possible so I left this comment out of code but can add if you think it's appropriate.

aatxe: You can author comments that explain why things are the way they are. There's no desire or commitment to avoiding comments in general, though I would probably tell people to omit comments that are just a direct explanation of the code that immediately follows it since those kinds of comments get stale all the time and add very little value. Explanations of why are great, high-level descriptions of overarching decisions that affect everything that follows are great, etc.

aatxe (approved):

PR #700, opened 2025-12-18
Group tableext methods pertaining to arrays and dicts

Alexander-L-Robert:
I reordered tableext methods to group dictionary and array methods more distinctly
(I ensured `keys` and `toset` are at the boundary as they convert between the two table formats)

This PR is intended to be merged after:
- https://github.com/luau-lang/lute/pull/699

### DIFF
--- a/lute/std/libs/tableext.luau
+++ b/lute/std/libs/tableext.luau
@@ -5,6 +5,8 @@
 
 local tableext = {}
 
+-- Functions on general tables
+
 function tableext.map<K, A, B>(table: { [K]: A }, f: (A) -> B): { [K]: B }
 	local new = {}
 
@@ -27,34 +29,6 @@ function tableext.filter<K, V>(table: { [K]: V }, predicate: (V) -> boolean): { 
 	return new
 end
 
-function tableext.any<T>(arr: { T }, predicate: (T) -> boolean): boolean
-	for _, v in arr do
-		if predicate(v) then
-			return true
-		end
-	end
-	return false
-end
-
-function tableext.all<T>(arr: { T }, predicate: (T) -> boolean): boolean
-	for _, v in arr do
-		if not predicate(v) then
-			return false
-		end
-	end
-	return true
-end
-
-function tableext.extend<T>(tbl: { T }, ...: { T })
-	local extensions = table.pack(...)
-	extensions.n = nil :: any
-	for _, extension in extensions do
-		for _, entry in extension do
-			table.insert(tbl, entry)
-		end
-	end
-end
-
 function tableext.combine<K, V>(tbl: { [K]: V }, ...: { [K]: V }): { [K]: V }
 	local tables = table.pack(...)
 	tables.n = nil :: any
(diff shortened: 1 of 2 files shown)


aatxe:
We have no allergy to comments, you can leave comments, especially if you actually want to retain the organization you're proposing here because it is very easy for someone to make a change later and not see it.

aatxe:
This probably needs a rebase.

lute/std/libs/tableext.luau
```diff
@@ -5,6 +5,8 @@
 
 local tableext = {}
 
+-- Methods using dictionary tables
```
skberkeley:
```suggestion
-- Functions on general tables
```

These functions technically work for array-like tables too

lute/std/libs/tableext.luau
```diff
@@ -56,6 +48,26 @@ function tableext.keys<K, V>(tbl: { [K]: V }): { K }
 	return keys
 end
 
+-- Methods using array tables
```
skberkeley:
```suggestion
-- Functions on array-like tables
```

tests/std/tableext.test.luau
```diff
@@ -62,6 +62,41 @@ test.suite("TableExt", function(suite)
...
+
+		local foundKeys: { number } = {}
+		for key, value in setA do
+			assert.eq(value, true)
+			table.insert(foundKeys, key)
+		end
+		assert.eq(#foundKeys, setSize)
+		assert.eq(setA[foundKeys[1]], true)
+		assert.eq(setA[foundKeys[2]], true)
+		assert.eq(setA[foundKeys[3]], true)
```
skberkeley, line 89: i don't really understand what this part is checking that's different from above?

tests/std/tableext.test.luau
```diff
@@ -62,6 +62,41 @@ test.suite("TableExt", function(suite)
...
+
+		local b = { "a", "b", "c", "a" }
+		local setB = tableext.toset(b)
+		assert.eq(setB["a"], true)
+		assert.eq(setB["b"], true)
+		assert.eq(setB["c"], true)
+		assert.eq(setB["d"], nil)
+		assert.eq(#tableext.keys(setB), 3)
+		assert.neq(#tableext.keys(setB), #b)
+	end)
```
skberkeley, line 103: can you also test the case of `toset({})`?

skberkeley (changes requested):

tests/std/tableext.test.luau
Alexander-L-Robert: This was shown as part of a bad diff and these changes are already in primary. I've fixed the diff and branch to be up to date now.

This change was:
> `-- Sanity check: sets created using an array of numbers don't break from having "gaps" between the number keys.`

It was confirming that dictionaries-like table made with number keys don't have any issues that a sparse array-like table would have from iteration.

I think we discussed it async briefly and said it wouldn't hurt to add, but feel free to remove it if you don't think it's adding any value.

Alexander-L-Robert:
@skberkeley I've gotten this branch up to date again and fixed the areas you highlighted.

skberkeley (approved):

PR #702, opened 2026-01-02
Fix require interception

bjcscat:
I'm not sure if there's a better way to do this but the nature of the require being defined in @cli/test means requires won't be from the test module itself which breaks pretty much everything.

### DIFF
--- a/lute/cli/commands/test/init.luau
+++ b/lute/cli/commands/test/init.luau
@@ -42,17 +42,26 @@ end
  the lute repo, which sucks. Intercepting @std/test requires works correctly and ensures that the correct @std/test instance gets populated
 ]]
 --
-local function lutetestrequire(req: string)
+
+local testRequireBytecode = luau.compile([[
+	local req = ...
+
 	if req == "@std/test" then
+		local test = require("@std/test")
 		return table.freeze({ case = test.case, suite = test.suite, run = function() end })
 	end
+	
 	return require(req) :: any
-end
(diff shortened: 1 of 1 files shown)


lute/cli/commands/test/init.luau
```diff
@@ -42,17 +42,26 @@ end
...
 	return require(req) :: any
-end
+]])
 
 local function loadtests(testfiles: { path.path })
-	local fenv: { [unknown]: unknown } = setmetatable({ ["require"] = lutetestrequire }, { __index = _G })
 	-- Load all test files first
 	for _, p in testfiles do
+		local fenv: { [unknown]: unknown } = setmetatable(
+			{ ["require"] = luau.load(testRequireBytecode, path.format(p)) },
```
Vighnesh-V, line 61: Clever! I didn't realize you could `luau.load` like this and the loaded function would be considered as part of the passed in chunkname!

Vighnesh-V:
@bjcscat thanks for this fix - we've been dealing with this issue for a while and we didn't know how to make it so that the overloaded require was considered part of the test module itself so that transitive requires succeeded! I am currently experimenting with another fix that might mean we won't have to even override require here - I will test that and report back. If it doesn't work, I'll give this an upvote.

Vighnesh-V:
Removed the require interception as part of https://github.com/luau-lang/lute/pull/705 - which should also fix transitive requires.

Issue #634, opened 2025-12-03
`lute test` needs to work in CI


Vighnesh-V:
Currently, the `lute test` command isn't being used to run all the tests in CI because the base64 tests are unable to find the batteries alias. The reason for this is that `lute test` intercepts the `require` calls made by `luau.loadbypath`. Because there is a chunkname discrepancy between the intercepting require, and the module being executed, the `loadbypath` fails. We need to fix this before `lute test` can be used as the test command in CI.

PR #705, opened 2026-01-05
Runs `lute test` as part of CI

Vighnesh-V:
Thanks to recent changes to how require works, namely that `@std` requires will always resolve to the single builtin instance in the runtime (ignoring aliases), `lute test` can now handle transitive and batteries requires correctly without needing to intercept requires at all. This was the last blocker remaining before `lute test` could be used in CI. 
This PR:
a) updates CI to call `lute test`
b) removes the require interception that `lute test` did
b) removes the explicit call to test.run() within the test files in our repo, since test discovery handles this

I've left `test.run()` in for now, in case there is some desire to use the testing library without `lute test` (and the standard library tests in `std.test.luau` exercise this behaviour).  I'm open to opinions about what to do with it, and can address those in a followup.

Resolves issue #634

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,18 +53,7 @@ jobs:
         run: rm .luaurc && mv .luaurc.ci .luaurc
 
       - name: Run Luau Tests (POSIX)
-        if: runner.os != 'Windows'
-        shell: bash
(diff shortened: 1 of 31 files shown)


annieetang (approved):
let's gooo, super excited it's working 🥳

PR #706, opened 2026-01-06
Improves lute runtime performance when scheduling many threads

Vighnesh-V:
The lute runtime currently schedules coroutines in a std::vector, and we always pop off the first thread in the vector to run. For small numbers of coroutines, this is okay, but when you get to 10k+ this becomes very inefficient as we have to move all the other elements one over `O(n)`. This PR just replaces the `std::vector` with a `Luau::VecDeque`, making the pop operation `O(1)`.

I've also included a profiling script that I used to profile lute since I keep forgetting the `xctrace` incantation - it can be invoked by:
```
./tools/profile.sh /path/to/.luau(defaults to profile.luau)
```

### DIFF
--- a/lute/runtime/include/lute/runtime.h
+++ b/lute/runtime/include/lute/runtime.h
@@ -3,6 +3,7 @@
 #include "lute/ref.h"
 
 #include "Luau/Variant.h"
+#include "Luau/VecDeque.h"
 
 #include <atomic>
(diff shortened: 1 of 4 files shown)


vrn-sn (approved):
Awesome! Curious, do you have any numbers on the improvement?

Issue #710, opened 2026-01-08
`lute check` needs to support the new Luau type solver


Vighnesh-V:
Currently, `lute check` defaults to the old solver under the hood. We should force this on by default.

To test that the new solver is being used:
```
function add(a,b)
    return a + b
end

local vec2 = {}

function vec2.new(x, y)
    return setmetatable({x = x or 0, y = y or 0}, {__add = function(v1, v2) return {x = v1.x + v2.x, y = v1.y + v2.y} end})
end

add(1,1)
add(vec2.new(0, 0), vec2.new(1,1))
```

This should typecheck with no errors - the old luau solver will raise an error in the add function.

PR #711, opened 2026-01-08
Lute check uses the new solver by default

raymondnumbergenerator:
Addresses #710 

Updates `lute check` to use the new solver by default.

Fails the test as expected without the change
```
Failed Tests (1):

❌ lute check.uses new solver
	/Users/rng/dev/lute/tests/cli/check.test.luau:32
		eq: 1 ~= 0
```


lute/cli/src/tc.cpp
```diff
@@ -22,8 +22,6 @@ declare fs: {
     close: (file) -> (),
     read: (file) -> string,
     write: (file, string) -> (),
-    readfiletostring : (string) -> string,
-    writestringtofile : (string, string) -> (),
```
raymondnumbergenerator, line 26: i think these were removed?
https://github.com/luau-lang/lute/commit/5e354a471832805f2a31d3163ba6b56406e1328c#diff-eb93529c45e1e3acc6cfcfc6803aad8ec9f06acc3b665e394ae684a45f39085dL94-L98

Vighnesh-V: We can actually delete all of these - we shouldn't be getting type information from here anymore!

raymondnumbergenerator: oops missed one more

tests/cli/check.test.luau
```diff
@@ -0,0 +1,35 @@
+local fs = require("@std/fs")
+local path = require("@std/path")
+local process = require("@std/process")
+local system = require("@std/system")
+local test = require("@std/test")
+
+local lutePath = path.format(process.execpath())
+local tmpDir = system.tmpdir()
+
+test.suite("lute check", function(suite)
```
Vighnesh-V:
```suggestion
test.suite("LuteTypeCheck", function(suite)
```

Just so this looks a bit nicer in the terminal! Same for the test case if possible.

tests/cli/check.test.luau
```diff
@@ -0,0 +1,35 @@
...
+	suite:case("uses new solver", function(assert)
```
Vighnesh-V: Nice! Can we also add a C++ test that invokes `typecheck` - the function exported by tc.h?  
If you need to create temporary directories / work with files we have platform agnostic helper functions in Luau/FileUtils.h and lute/fileutils.h (we're trying not to use the <filesystem> header because it's got some problems).

If you need an example of how to write a test, you can check out the tests in `require.test.cpp` or `compile.test.cpp`!

lute/cli/src/tc.cpp
```diff
@@ -186,6 +184,7 @@ int typecheck(const std::vector<std::string>& sourceFilesInput, LuteReporter& re
     LuteFileResolver fileResolver;
     Luau::LuteConfigResolver configResolver(mode);
     Luau::Frontend frontend(&fileResolver, &configResolver, frontendOptions);
+    frontend.setLuauSolverMode(Luau::SolverMode::New);
 
     Luau::registerBuiltinGlobals(frontend, frontend.globals);
     Luau::LoadDefinitionFileResult loadResult =
```
Vighnesh-V: We can delete this line along with `kLuteDefinitions`!

Vighnesh-V (changes requested):
Looks good! Just left a comment asking for a C++ test + deleting the out of date definition files.

tests/cli/check.test.luau
```diff
@@ -0,0 +1,39 @@
...
+	return setmetatable({ x = x or 0, y = y or 0 }, {
+		__add = function(v1, v2)
+			return { x = v1.x + v2.x, y = v1.y + v2.y }
+		end,
+	})
+end
+add(1, 1)
+add(vec2.new(0, 0), vec2.new(1, 1))
+
+add(vec2.new(0, 0), vec2.new(1,1))
```
Vighnesh-V: Instead of having this code be created in a temp directory for this test, now that this code exists on disk in the repo, can we make this test use the `newsolver.luau` file?

Vighnesh-V (approved):
Left one comment - can you make the test in `check.test.luau` use the `newsolver.luau` file instead?

PR #715, opened 2026-01-10
Embed and expose the package manager prototype under `lute pkg install` and `lute pkg run`

vrn-sn:
Very early prototype that will likely change substantially in the next few weeks as we start getting feedback and iterate on the design.

If the current directory has a `loom.config.luau` in it like the following example:
```luau
-- loom.config.luau
return {
	package = {
		name = "MyProject",
		version = "0.1.0",
		dependencies = {
			["pretty-print"] = { rev = "main", source_kind = "github", source = "https://github.com/afujiwara-roblox/pretty-print" },
			base64 = { rev = "main", source_kind = "github", source = "https://github.com/afujiwara-roblox/base64" },
			toml = { rev = "main", source_kind = "github", source = "https://github.com/afujiwara-roblox/toml" },
		},
	},
}
```

Running `lute pkg install` will install the declared dependencies into a `Packages` folder in the current directory. (No global package cache support yet, but there are no technical reasons we can't do that—this is just simpler as a starting point.)

When the dependencies are installed, a `loom.lock.luau` file will be generated. When this file is present in a project, we can opt-in to package awareness by using `lute pkg run <file.luau>`. With package awareness enabled, case-insensitive aliases to the installed packages are automatically made available to the Lute runtime, allowing us to run a script like this:

```luau
-- These three dependencies will be required successfully if:
-- 1. They were declared in a manifest file (`loom.config.luau`).
-- 2. The `lute pkg install` subcommand successfully installed them into the `Packages` folder.
-- 3. This script was run with package awareness enabled (`lute pkg run <this_file.luau>`).

local toml = require("@toml")
local prettyprint = require("@pretty-print")
local base64 = require("@base64")

local toml_content = [[
[config]
key = "TG9vbSBpcyBzbyBjb29sISB3b3dvd293"
key2 = "VGhlIGJlc3QgdGhpbmcgc2luY2Ugc2xpY2VkIGJyZWFkIQ=="
]]

-- Deserialize the TOML content using the @toml library
local config = toml.deserialize(toml_content) :: any

print("------------------------------")
print("This is the original config")
print(config)
print("------------------------------")
print("Pretty printed original config:")
print(prettyprint(config))
print("------------------------------")
-- Decode the base64 encoded keys using the @base64 library
config.config.key = buffer.tostring(base64.decode(buffer.fromstring(config.config.key)))
config.config.key2 = buffer.tostring(base64.decode(buffer.fromstring(config.config.key2)))

print("Pretty printed decoded config:")
print(prettyprint(config))
print("------------------------------")
```

This prototype still has rough edges that we (primarily @afujiwara-roblox and @vrn-sn) hope to iron out as we receive feedback.


Vighnesh-V:
Is there a way we should review this? Like a suggested reading order?

vrn-sn:
@Vighnesh-V I think the big thing is to just ignore the `extern` folder. The `pp.luau` file should be an exact copy of our batteries version, and unzip is from https://github.com/0x5eal/luau-unzip.

Ignoring that, it's not too much to review:
```bash
> git diff --stat origin/primary... -- lute/cli/commands/loom/src/loom  
 lute/cli/commands/loom/src/loom/commands/install.luau          | 35 +++++++++++++++++++++++++++++++++
 lute/cli/commands/loom/src/loom/lockfile/lockfile.luau         | 22 +++++++++++++++++++++
 lute/cli/commands/loom/src/loom/lockfile/lockfile.test.luau    | 80 ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 lute/cli/commands/loom/src/loom/loom.luau                      | 14 ++++++++++++++
 lute/cli/commands/loom/src/loom/loom_types.luau                | 10 ++++++++++
 lute/cli/commands/loom/src/loom/manifest/project_manifest.luau | 51 ++++++++++++++++++++++++++++++++++++++++++++++++
 lute/cli/commands/loom/src/loom/package_source/git.luau        | 79 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 lute/cli/commands/loom/src/loom/resolver/package.luau          | 36 ++++++++++++++++++++++++++++++++++
 lute/cli/commands/loom/src/loom/resolver/resolution.luau       | 46 ++++++++++++++++++++++++++++++++++++++++++++
 lute/cli/commands/loom/src/loom/utils/fs_utils.luau            | 33 +++++++++++++++++++++++++++++++
 10 files changed, 406 insertions(+)
```

I'd probably start at the `lute/cli/commands/loom/init.luau` entry point and follow the logic from there to `lute/cli/commands/loom/src/loom/commands/install.luau` and so on (the flow is mostly linear—it just jumps to different files).

aatxe:
Didn't we implement support already for CLI commands referencing `@batteries` correctly without exposing it? If so, can we remove the duplication of `pp.luau`?

aatxe:
I think it also probably makes sense for this to just be `lute pkg` if its entire identity is going to be existing as a subcommand of lute (which I think is fine to do, to be clear).

vrn-sn:
> Didn't we implement support already for CLI commands referencing @batteries correctly without exposing it? If so, can we remove the duplication of pp.luau?

I don't think this is implemented yet.

> I think it also probably makes sense for this to just be lute pkg if its entire identity is going to be existing as a subcommand of lute (which I think is fine to do, to be clear).

Sounds good, do we want something like `lute pkg install`? Also, another point we briefly discussed was running with package awareness: do we still want that to be under `lute pkgrun` or to just merge it into the existing default `lute run` behavior?

aatxe:
I'd expect `lute pkg install` and `lute pkg run` for right now, if we also want to back run into the normal run, which we might, that feels like a different issue in my mind?

vrn-sn:
> I'd expect `lute pkg install` and `lute pkg run` for right now

Just pushed these changes now.

lute/cli/commands/pkg/src/loom/commands/install.luau
```diff
@@ -0,0 +1,35 @@
+local resolution = require("../resolver/resolution")
+local loom_types = require("../loom_types")
+local project_manifest = require("../manifest/project_manifest")
```
aatxe: It feels very werid that all of these things are snake_case in their name. We've been otherwise doing luacase modules, I think, even for cli stuff that isn't part of the API, and camelCase for non-API identifiers, and this code departs from that.

lute/cli/commands/pkg/src/loom/loom_types.luau
```diff
@@ -0,0 +1,10 @@
+local Git = require("./package_source/git")
+export type package_source = {
```
aatxe:
```suggestion
export type PackageSource = {
```

lute/cli/commands/pkg/src/loom/loom_types.luau
```diff
@@ -0,0 +1,10 @@
...
+	git: Git.git,
+}
+
+export type loom = {
```
aatxe:
```suggestion
export type Loom = {
```

lute/cli/commands/pkg/src/loom/manifest/projectmanifest.luau
```diff
@@ -0,0 +1,51 @@
...
+		"OtherPackage": "1.0.0",
+	}
+}
+]]
+
+export type ProjectManifest = {
+	package: PackageManifest,
+}
+
+function fromTable(table: any): ProjectManifest
```
aatxe, line 29: `any` seems like a bad type here, it's at least assuming that the thing being passed in is a table, so I'd expect `{[unknown]: unknown}` or `table` or `{}`.

lute/cli/commands/pkg/src/loom/package_source/git.luau
```diff
@@ -0,0 +1,79 @@
+local fs_utils = require("../utils/fs_utils")
+local net = require("@std/net")
+local unzip = require("../../extern/unzip")
+
+export type git = {
```
aatxe:
```suggestion
export type Git = {
```

aatxe (changes requested):

lute/cli/commands/pkg/src/loom/commands/install.luau
vrn-sn: Yeah, I'll go through it and look for this type of thing. This is mainly because it was developed in a separate repository, and we weren't paying attention to style as much.

lute/cli/commands/pkg/src/loom/manifest/projectmanifest.luau
vrn-sn: I can't seem to get this to type check if I put anything more specific. I believe that was the motivation for writing this function — a bridge where we use runtime checks to go from `any` to `ProjectManifest`, but we do seem to be assuming it's a `table` right now. Maybe we can take a look together tomorrow? I might be hitting some bugs in the type system.

aatxe: at minimum, we can put a `// FIXME(Luau)` comment here  to look at the type, but yeah, I would believe that bugs and/or limitations prevent us from getting something useful here.

lute/cli/commands/pkg/src/loom/commands/install.luau
vrn-sn: Updated every instance of this I could find.

lute/cli/commands/pkg/src/loom/manifest/projectmanifest.luau
vrn-sn: Added a comment with the intention to address it outside of this PR. (Can make a ticket in the Lute repo for this as a follow-up action item.)

lute/cli/commands/pkg/src/loom/lockfile/lockfile.luau
```diff
@@ -0,0 +1,22 @@
...
+-- 	version: string?,
+-- 	source_kind: string?,
+-- 	source: string?,
+-- }
+
+-- TODO: Support path/registry dependencies
+export type LockFilePackage = {
+	name: string,
+	rev: string,
+	source_kind: string,
```
aatxe:
```suggestion
	sourceKind: string,
```

lute/cli/commands/pkg/src/loom/manifest/projectmanifest.luau
```diff
@@ -0,0 +1,53 @@
+-- TODO: Support path/registry dependencies
+export type DependencySpec = {
+	rev: string,
+	source_kind: string,
```
aatxe:
```suggestion
	sourceKind: string,
```

lute/cli/commands/pkg/src/loom/resolver/package.luau
```diff
@@ -0,0 +1,36 @@
+local lockfile = require("../lockfile/lockfile")
+local manifest = require("../manifest/projectmanifest")
+export type Package = {
+	name: string,
+	rev: string,
+	source_kind: string,
```
aatxe:
```suggestion
	sourceKind: string,
```

aatxe (approved):

PR #716, opened 2026-01-10
Fix `parseGlob` handling of `.` character

wmccrthy:
### Problem

The `parseGlob` function in `ignore` does not currently escape the `.` character; `.` is a magic character in lua, so if we don't escape it, it will match ANY character instead of the literal `.` we intend to match for.

This leads to issues where directories are incorrectly matched and ignored due to file extension globs from the parsed `gitignore`.

For instance, if we have `*.something` in our gitignore, any directory path ending in `something` will be ignored. I've added a small case to the `files` test covering this sort of scenario.

### Solution

Ensure `.` is escaped properly in `parseGlob`

### DIFF
--- a/tests/cli/lib/files.test.luau
+++ b/tests/cli/lib/files.test.luau
@@ -271,4 +271,3 @@ test.suite("cli/lib/files", function(suite)
 		fs.removedirectory(testPath, { recursive = true })
 	end)
 end)
-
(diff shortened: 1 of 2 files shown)


Nicell (approved):

tests/cli/lib/files.test.luau
```diff
@@ -265,3 +271,5 @@ test.suite("cli/lib/files", function(suite)
 		fs.removedirectory(testPath, { recursive = true })
 	end)
 end)
+
+test.run()
```
wmccrthy:
```suggestion
```
remove, old convention

tests/cli/lib/files.test.luau
```diff
@@ -265,3 +271,4 @@ test.suite("cli/lib/files", function(suite)
 		fs.removedirectory(testPath, { recursive = true })
 	end)
 end)
+
```
wmccrthy:
```suggestion
```

PR #717, opened 2026-01-11
Split net library into client and server

Nicell:
I'd like to get websockets in upstream, as I've been using my branch of it for a bit with success. First, though, splitting net into server and client should make it cleaner and easier to review the upcoming websocket PRs... Also the file was getting really long.

### DIFF
--- a/lute/net/src/client.cpp
+++ b/lute/net/src/client.cpp
@@ -1,0 +1,264 @@
+#include "lute/net.h"
+#include "lute/runtime.h"
+
+#include "Luau/DenseHash.h"
+
+#include "lua.h"
+#include "lualib.h"
+
+#include "curl/curl.h"
+
+#include <memory>
+#include <string>
+#include <utility>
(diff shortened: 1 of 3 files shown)


lute/net/src/client.cpp
```diff
@@ -0,0 +1,264 @@
...
+            break;
+
+        lua_pushcfunction(L, func, name);
+        lua_setfield(L, -2, name);
+    }
+
+    lua_setreadonly(L, -1, 1);
+
+    return 1;
+}
```
aatxe, line 264: Environment setup seems to be broken now with how they're split. I want to make all of the modules/submodules for the runtime libraries more uniform here, but in general, we should have open for `@lute/net/client` and open for `@lute/net/server` and an open for `@lute/net` that, at the very least, exposes those two tables as `client` and `server` within it.

aatxe (changes requested):

PR #718, opened 2026-01-12
Ensure `ExprTableItem` nodes are serialized with `location`

wmccrthy:
`AstExprTableItem` nodes are not serialized with location right now.

I encountered a bug due to this behavior while using `lute lint` CLI:
```
Error linting file '/Users/wmccarthy/git/roblox/luau-ui-ecosystem-lints/repro_nil_location.luau': @std/syntax/init.luau:14: attempt to index nil with 'beginline'
```

If you follow the trace from std/syntax/init.luau:14, you'll see the error stems from calls to syntax.span.subsumes in lint's init file ([see here](https://github.com/luau-lang/lute/blob/primary/lute/cli/commands/lint/init.luau?rgh-link-date=2026-01-07T21%3A07%3A33Z#L150-L160)).
Here, `nodeIsSuppressed` is used as the default visitor function, in other words, it is called at every visited node as the visitor walks the tree. It relies on the assumption that every node has location field, which is not true. `ExprTableItem` nodes, specifically, do not have a `location` field.

Adding `location` to table item nodes will fix this issue and help avoid any similar future issues.


definitions/luau.luau
skberkeley, line 1: tiny nit: can you put `location` first to match the other type definitions?

tests/std/syntax/parser.test.luau
```diff
@@ -499,6 +499,27 @@ test.suite("parseExpr", function(suite)
 		assert.eq(expr.operator.text, "::")
 		assert.eq(expr.annotation.tag, "reference")
 	end)
+
+	-- table items serialized with location field
+	suite:case("ExprTableItem serialized with location field", function(assert)
+		local tableExpr = parser.parseexpr("{ 1, foo = bar, [2] = baz }")
```
skberkeley: nit: can you make the expression multiline so we have some different line values?

lute/luau/src/luau.cpp
```diff
@@ -462,27 +462,36 @@ struct AstSerialize : public Luau::AstVisitor
 
         lua_rawcheckstack(L, 2);
 
+        // Compute end position: after separator if present, otherwise at value end
+        Luau::Position endPosition = cstNode->separator
+            ? Luau::Position{cstNode->separatorPosition->line, cstNode->separatorPosition->column + 1} // +1 accounts for separator length
+            : item.value->location.end;
```
skberkeley: This should always be `item.value-> location.end`, because we want to think of the node as having the (optional) key and value as its subnodes. For example, in `{ x = 3 }` the table entry node is `x = 3`, rather than just `x`. Could you add a test to `lint.test.luau` for the following to confirm?
```luau
local y = {
    -- lute-lint-ignore(divide_by_zero)
    x = 3 / 0
}

local z = 5 / 0
```

skberkeley (changes requested):

skberkeley (approved):

PR #719, opened 2026-01-12
Extends `lute lint` to support linting an input string

skberkeley:
The bulk of the changes are refactoring some of the linting logic to extract the common logic between linting a file and a string. Also, paths in reported lint violations are now optional since they're not relevant to linting string input.

The broader context for this PR is that it allows linting to be consumed by VSCode (and maybe other) extensions, since the source of truth for file contents is whatever the editor/LSP server has in memory, rather than what's on disk in the file.

### DIFF
--- a/lute/cli/commands/lint/lsp/init.luau
+++ b/lute/cli/commands/lint/lsp/init.luau
@@ -25,7 +25,7 @@ local function tag(t: lintTypes.tag): number
 	end
 end
 
-local function diagnostic(lint: lintTypes.LintViolation): lspTypes.Diagnostic
+function lsp.diagnostic(lint: lintTypes.LintViolation): lspTypes.Diagnostic
 	return table.freeze({
 		range = table.freeze({
 			start = table.freeze({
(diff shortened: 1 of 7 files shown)


lute/cli/commands/lint/lsp/init.luau
```diff
@@ -39,7 +39,7 @@ local function diagnostic(lint: lintTypes.LintViolation): lspTypes.Diagnostic
 		}),
 		severity = severity(lint.severity),
 		code = lint.lintname,
-		source = "lute lint", -- LUAUFIX: Errors because this isn't inferred as a singleton
+		source = "lute lint" :: "lute lint", -- LUAUFIX: Errors because this isn't inferred as a singleton
```
aatxe: This doesn't error now, right? Obviously still a LUAUFIX since we expect inference to trigger here, but just want to confirm

aatxe (approved):

lute/cli/commands/lint/lsp/init.luau
skberkeley: Sorry yeah, I'll clarify that this is covering up a type error

PR #720, opened 2026-01-13
Adds any and all (for array-like tables) to tableext

skberkeley:
I had a need for `any` when working on something else, and added `all` along the way

### DIFF
--- a/tests/std/tableext.test.luau
+++ b/tests/std/tableext.test.luau
@@ -109,6 +109,12 @@ test.suite("TableExt", function(suite)
 			return v > 5
 		end)
 		assert.eq(hasGreaterThanFive, false)
+
+		a = {}
+		local anyInEmpty = tableext.any(a, function(v: number)
+			return true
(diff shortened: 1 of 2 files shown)


tests/std/tableext.test.luau
```diff
@@ -97,4 +97,30 @@ test.suite("TableExt", function(suite)
 		assert.eq(#tableext.keys(setB), 3)
 		assert.neq(#tableext.keys(setB), #b)
 	end)
+
+	suite:case("any", function(assert)
```
Vighnesh-V, line 101: test requests - can you also add two tests for the vacuous case?

Vighnesh-V (approved):
Left one comment, but this is a good addition.

tests/std/tableext.test.luau
skberkeley: i prefer the term degenerate, but yeah, good idea

PR #723, opened 2026-01-14
Adds autofix support to `lute lint`

skberkeley:
Adds support for autofixing reported lint violations (if the lint rules provide suggested fixes). The core logic for applying the fixes is pretty simple: split the content by line and do the replacements to the relevant lines. For now, we don't handle the cases where suggested fixes overlap (although my approach would probably be to just ignore fixes that overlap with previously applied fixes). We rerun the lints after applying fixes and report any violations that are still left. It's possible that those reported violations have suggested fixes, but we ignore them to prevent the chance of looping between suggested fixes infinitely.

### DIFF
--- a/lute/cli/commands/lint/init.luau
+++ b/lute/cli/commands/lint/init.luau
@@ -31,6 +31,8 @@ OPTIONS:
 						  as individual lint rules. If unspecified, the default lint rules are used.
   -j, --json              Output lint violations in JSON format matching the LSP diagnostic spec.
   -s, --string-input	  Lint the provided string input instead of reading from files.
+  --auto-fix			  Automatically apply fixes for lint violations that provide a suggested fix.
+						  Assumes that suggested fixes do not overlap.
 
 PATHS:
   Path(s) to the Luau file(s) or folders containing Luau files to be linted. Only files with .luau or .lua extensions will be linted.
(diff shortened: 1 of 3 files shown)


lute/cli/commands/lint/init.luau
```diff
@@ -244,18 +246,98 @@ local function lintString(
...
+	location: syntax.span,
+	replacement: string,
+}
+
+local function applySuggestedFixes(source: string, fixes: { fix }): string
+	local lines = source:split("\n")
+
+	-- Sort fixes by starting location descending to avoid messing up locations of earlier fixes
+	table.sort(fixes, function(a: fix, b: fix)
+		return not (a.location < b.location)
```
Vighnesh-V, line 259: is there a reason we can't write a.location >= b.location? I'm guessing the runtime doesn't derive that for us automatically?

lute/cli/commands/lint/init.luau
```diff
@@ -244,18 +246,98 @@ local function lintString(
...
+
+	fs.writestringtofile(inputFilePath, newContent)
+
+	return lintString(newContent, lintRules, inputFilePath)
 end
 
 local function lintPaths(
 	inputFilePaths: { string },
-	lintRules: { types.LintRule }
+	lintRules: { types.LintRule },
```
Vighnesh-V, line 339: If you anticipate needing to add more arguments here, you could also extract these into a `type lintoptions` and then configure this function by passing a table.

lute/cli/commands/lint/init.luau
```diff
@@ -330,6 +417,10 @@ local function main(...: string)
 	end
 
 	if stringInput ~= nil then
+		if args:has("auto-fix") then
+			print("Auto-fix has no effect on string input.")
```
Vighnesh-V, line 421: Should this be an exit(1)?

Vighnesh-V (approved):
Looks good, just left a few comment.s

lute/cli/commands/lint/init.luau
skberkeley: yeah only `__lt` is defined

lute/cli/commands/lint/init.luau
skberkeley: this is the last feature i had planned for linting but i'll keep that in mind!

lute/cli/commands/lint/init.luau
skberkeley: i'd rather that linting still proceed in this case, and just let the user know that --auto-fix isn't going to do anything

PR #725, opened 2026-01-14
Add CLI docs

Nicell:
Adds basic CLI docs to start for now


annieetang (commented):
spoke about this on slack but we can get our team to manually write the CLI docs to include more info on it than what's just in the help dialogue for each (especially examples), so maybe we can update this PR to just set up the index file and not need the autogen for each command? ty for the work tho and pushing us to have better docs!

annieetang (approved):
ty!!

PR #726, opened 2026-01-14
Improve reference docs

Nicell:
If the file is init and has children exported, create an index.md file so structure is right (for std)

image

Creates index files for sidebar.

image
image

Should be merged after #725


annieetang (approved):
thanks sm!

annieetang:
~~wait actually can you add the table to `@std` too~~
(nvm it does! thanks!)

PR #730, opened 2026-01-15
Brings in profiler and codecoverage from Luau

Vighnesh-V:
This PR just pulls in the Luau CLI's Profiler and Code Coverage modules into `lute`. We're doing this instead of exposing them from Luau because these features are not really products, so much as best effort attempts, so we don't necessarily want to expose them from Luau yet.

This is just copied from Luau/CLI - no changes have been made.

### DIFF
--- a/lute/cli/CMakeLists.txt
+++ b/lute/cli/CMakeLists.txt
@@ -25,9 +25,11 @@ add_library(Lute.CLI.lib STATIC)
 target_sources(Lute.CLI.lib PRIVATE
     include/lute/climain.h
     include/lute/compile.h
+    include/lute/coverage.h
     include/lute/fileutils.h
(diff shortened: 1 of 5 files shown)


skberkeley:
These aren't exposed as subcommands yet right?

vrn-sn (approved):
Let's hook these in and expose them soon so that we don't have dead files sitting around. When we do that, we need to get rid of all these `printf`s. (Approving this for now, if you want to do that in multiple PRs.)

Vighnesh-V:
@skberkeley Yeah, these are explicitly not being exposed as an option in `lute run` just yet, I'll wire in profiling in separate PR.

RE: getting rid of the printf's - will do in a separate pass.

Issue #709, opened 2026-01-08
The remaining file system apis need to be made asynchronous.


Vighnesh-V:
A number of fs apis don't pass callbacks to the underlying libuv call, which makes them occur synchronously. We should make it so that all of these work asynchronously.

annieetang:
checklist:

- [x] fs.open
- [x] fs.read
- [x] fs.write
- [x] fs.close
- [x] fs.remove
- [x] fs.stat
- [x] fs.exists
- [x] fs.type
- [x] fs.watch
- [x] fs.link
- [x] fs.symlink
- [x] fs.copy
- [x] fs.mkdir
- [x] fs.listdir
- [x] fs.rmdir

PR #731, opened 2026-01-15
Ensures `fs.mkdir` is asynchronous

annieetang:
Addresses #709's journey of making our file system APIs asynchronous

### DIFF
--- a/lute/fs/src/fs.cpp
+++ b/lute/fs/src/fs.cpp
@@ -247,19 +247,22 @@ int remove(lua_State* L)
     return remove_impl(L, path);
 }
 
-int fs_mkdir(lua_State* L)
+int mkdir(lua_State* L)
 {
+    int nArgs = lua_gettop(L);
+    if (nArgs < 1)
+    {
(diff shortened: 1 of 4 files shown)


lute/fs/src/fs.cpp
```diff
@@ -247,19 +247,22 @@ int remove(lua_State* L)
...
+    {
+        luaL_errorL(L, "Error: no path supplied\n");
+    }
+
+    if (nArgs > 2)
+    {
+        luaL_errorL(L, "Error: too many arguments supplied\n");
+    }
     const char* path = luaL_checkstring(L, 1);
     int mode = luaL_optinteger(L, 2, 0777);
```
annieetang, line 263: I just realized our API doesn't actually allow users to provide a mode

```lua
function fs.mkdir(path: string): ()
```

Do we want to add `mode` as an option or keep it hidden?

Vighnesh-V: Lets add it as an optional param, but in a separate PR.

Vighnesh-V (approved):
nice!

PR #736, opened 2026-01-16
Adds `mode` option to `fs.createdirectory`

annieetang:
Previously, `fs.createdirectory` and `fs.mkdir` did not provide an option to change the permissions when creating a directory - we always defaulted to octal 0777 which is read/write/execute for everyone (owner, group and others). 

This PR adds `mode: number?` to the `fs.mkdir` signature and the `fs.createdirectory` `createdirectoryoptions` type so users can pass in a hexadecimal literal that corresponds to the octal permission.

I do realize that having the user pass in a hex isn't super great user experience and it would be better if we had some equivalent of how the `fs.open`'s `mode` works with `"r" | "w" | ...`, but discussed with @Vighnesh-V and having an optional param that exposes `mode` in the first place is better than nothing for now!

now if we want to specify read-only permissions for a directory, we can write:

```lua
local dir = path.join(tmpdir, "read_only_dir")
local mode = 0x124 -- octal 0444 (read-only for everyone)

fs.createdirectory(dir, { makeparents = true, mode = mode })
```

### DIFF
--- a/definitions/fs.luau
+++ b/definitions/fs.luau
@@ -59,7 +59,9 @@ function fs.type(path: string): FileType
 	error("not implemented")
 end
 
-function fs.mkdir(path: string): ()
+--- Creates a directory at the specified path.
+--- @param mode, if specified, should be a hexadecimal literal that corresponds to a valid octal permission mode (e.g., 0x1FF for octal 0777 permissions). Default read, write, execute for everyone permissions (0777) will be used if mode not specified.
+function fs.mkdir(path: string, mode: number?): ()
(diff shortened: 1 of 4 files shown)


annieetang:
just kidding! after further discussion we're gonna not allow users to specify the mode through `fs.createdirectory` or `fs.mkdir` to try to be platform agnostic (given `mode` is currently not implemented on Windows). Will close this and keep as reference and put a comment in a separate PR to use our `std/process` library if a user wants to set the mode for a directory on unix!

PR #737, opened 2026-01-16
Updates `fs.mkdir` to no longer support setting permissions

annieetang:
See #736

### DIFF
--- a/definitions/fs.luau
+++ b/definitions/fs.luau
@@ -59,6 +59,8 @@ function fs.type(path: string): FileType
 	error("not implemented")
 end
 
+--- Creates a directory at the specified path.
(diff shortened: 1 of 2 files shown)


Vighnesh-V (approved):

annieetang:
Gonna leave this open until we have a consensus as a team on how to handle platform specific filesystem operations to avoid back and forth changes if we change our minds!

annieetang:
following up! after talking about it during deep dive we'll just not allow users to modify the permissions it at all for now so I'll merge this PR. but it brings up a separate conversation of how we present APIs in Lute that aren't necessarily available on all OSes (OS's?), since we have `@std/path` separated by windows and posix, but other APIs (including this one now) aren't necessarily separated

PR #741, opened 2026-01-20
Ensures `fs.stat` and `fs.type` are asynchronous

annieetang:
part of #709 !

### DIFF
--- a/lute/fs/src/fs.cpp
+++ b/lute/fs/src/fs.cpp
@@ -25,29 +25,6 @@
 #include <stdlib.h>
 #include <string>
 
-#include <sys/stat.h>
-
-
-#if !defined(S_ISREG) && defined(S_IFMT) && defined(S_IFREG)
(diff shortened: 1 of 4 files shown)


lute/fs/src/fs.cpp
```diff
@@ -634,7 +486,7 @@ int listdir(lua_State* L)
                         lua_pushstring(L, dir.name);
                         lua_setfield(L, -2, "name");
 
-                        lua_pushstring(L, UV_DIRENT_TYPES[dir.type]);
+                        lua_pushstring(L, fs::UV_DIRENT_TYPES[dir.type]);
```
annieetang, line 489: this won't be needed once `listdir` is in `fs_impl.cpp` (when re-implementing as async)

Vighnesh-V: sure, that works!

Vighnesh-V (approved):

PR #743, opened 2026-01-20
Ensures `fs.rmdir` is asynchronous

annieetang:
part of https://github.com/luau-lang/lute/issues/709 😃👍

### DIFF
--- a/lute/fs/src/fs.cpp
+++ b/lute/fs/src/fs.cpp
@@ -265,18 +265,22 @@ int mkdir(lua_State* L)
     return mkdir_impl(L, path, mode);
 }
 
-int fs_rmdir(lua_State* L)
+int rmdir(lua_State* L)
 {
+    int nArgs = lua_gettop(L);
+    if (nArgs < 1)
+    {
+        luaL_errorL(L, "Error: no path supplied\n");
+    }
+
+    if (nArgs > 1)
+    {
+        luaL_errorL(L, "Error: too many arguments supplied\n");
+    }
+
     const char* path = luaL_checkstring(L, 1);
 
-    uv_fs_t rmdir_req;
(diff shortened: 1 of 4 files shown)


lute/fs/src/fs.cpp
```diff
@@ -265,18 +265,22 @@ int mkdir(lua_State* L)
...
 }
 
-int fs_rmdir(lua_State* L)
+int rmdir(lua_State* L)
 {
-    const char* path = luaL_checkstring(L, 1);
+    int nArgs = lua_gettop(L);
+    if (nArgs < 1)
+    {
+        luaL_errorL(L, "Error: no path supplied\n");
```
Vighnesh-V: Small suggestion here - maybe `rmdir: no path supplied`?

lute/fs/src/fs.cpp
```diff
@@ -265,18 +265,22 @@ int mkdir(lua_State* L)
...
+    }
 
-    uv_fs_t rmdir_req;
-    int err = uv_fs_rmdir(uv_default_loop(), &rmdir_req, path, nullptr);
-    uv_fs_req_cleanup(&rmdir_req);
+    if (nArgs > 1)
+    {
+        luaL_errorL(L, "Error: too many arguments supplied\n");
```
Vighnesh-V: ditto here.

Vighnesh-V (approved):
left a comment about error messages, but otherwise looks good!

lute/fs/src/fs.cpp
annieetang: sounds good and def more useful, thanks!

annieetang: i'll make the errors for the other functions better in a follow up!

PR #744, opened 2026-01-20
Fixes a crash in listdir

Vighnesh-V:
This PR just pulls in @checkraisefold 's fix here: https://github.com/luau-lang/lute/pull/740. The main change is to check the return error code from `uv_fs_scandir`.

Using the UVRequest abstractions a) helps make this function suitable for use cooperatively (async), and handles the additional memory management improvments that @checkraisefold added. 

Helps address one of the functions listed in https://github.com/luau-lang/lute/issues/709

### DIFF
--- a/lute/fs/src/fs.cpp
+++ b/lute/fs/src/fs.cpp
@@ -459,65 +459,15 @@ int type(lua_State* L)
 
 int listdir(lua_State* L)
 {
-    const char* path = luaL_checkstring(L, 1);
-
-    auto* req = new uv_fs_t();
-    req->data = new ResumeToken(getResumeToken(L));
-
-    int err = uv_fs_scandir(
-        uv_default_loop(),
-        req,
-        path,
-        0,
-        [](uv_fs_t* req)
-        {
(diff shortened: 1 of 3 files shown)


lute/fs/src/fs.cpp
```diff
@@ -461,63 +461,7 @@ int listdir(lua_State* L)
 {
     const char* path = luaL_checkstring(L, 1);
```
annieetang, line 468: do you think it's worthwhile having an arg check here as well? something like 

```lua
    int nArgs = lua_gettop(L);
    if (nArgs < 1)
    {
        luaL_errorL(L, "listdir: no path supplied\n");
    }

    if (nArgs > 2)
    {
        luaL_errorL(L, "listdir: too many arguments supplied\n");
    }
```

annieetang (approved):
lgtm other than an error handling suggestion!

lute/fs/src/fs.cpp
annieetang: (maybe we can just move this into a helper function at some point bc almost all of these functions have a check like this)

PR #745, opened 2026-01-20
Ensures `fs.exists`, `fs.link`, `fs.symlink`, `fs.copy` are all asynchronous

annieetang:
Another four `fs` functions for https://github.com/luau-lang/lute/issues/709!


lute/fs/src/fs_impl.cpp
```diff
@@ -98,6 +102,19 @@ struct FSClose : FSRequest
...
+    FSTwoPaths(lua_State* L, const char* src, const char* dest)
+        : FSRequest(L)
+        , src(src)
+        , dest(dest)
+    {
+    }
+
+    const std::string src;
+    const std::string dest;
+};
```
annieetang: made this for better error messaging for `link`, `symlink`, and `copy` since they all take a `src` and `dest`. lmk if this name / approach is fine!

Vighnesh-V: Maybe, FSPathPairRequest?

lute/fs/src/fs_impl.cpp
```diff
@@ -13,7 +13,11 @@
 #else
 #include <unistd.h>
 #endif
-
+#include <fcntl.h>
+#include <filesystem>
```
Vighnesh-V: We shouldn't have any code that depends on this header in the repo.

lute/fs/src/fs_impl.cpp
```diff
@@ -434,6 +483,103 @@ int type_impl(lua_State* L, const char* path)
...
+        }
+    );
+
+    return lua_yield(L, 0);
+}
+
+int symlink_impl(lua_State* L, const char* path, const char* dest)
+{
+    uvutils::ScopedUVRequest<FSTwoPaths> req{L, path, dest};
+    int flags = std::filesystem::is_directory(path) ? UV_FS_SYMLINK_DIR : 0;
```
Vighnesh-V: we shouldn't have this code here as it depends on the filesystem header.

lute/fs/src/fs_impl.cpp
```diff
@@ -13,7 +13,11 @@
 #else
 #include <unistd.h>
 #endif
-
+#include <fcntl.h>
```
Vighnesh-V: what does this header get us?

lute/fs/src/fs_impl.cpp
Vighnesh-V: oh I see. This was in the original implementation. While we're here, let's fix this and get rid of the filesystem reference.

Vighnesh-V: The UV_FS_SYMLINK_DIR flag is only used on windows btw. Can you add a isDirectory function to `lute/fileutils.h` and have it call the `stat` syscall - I think this is portable for windows and unix platforms.

lute/fs/src/fs_impl.cpp
```diff
@@ -13,7 +13,11 @@
...
+#include <filesystem>
+#include <memory>
+#include <stdlib.h>
+#include <string>
```
Vighnesh-V: Can you check which of these headers are necessary?

Vighnesh-V (changes requested):
Left a couple comments - the big one is that we shouldn't depend on the filesystem header.

lute/fs/src/fs_impl.cpp
annieetang: yup! sounds good thanks!

Vighnesh-V (approved):

PR #746, opened 2026-01-21
Adds a profiler to `lute run`

Vighnesh-V:
You can invoke the profiler with:
```
lute --profile <--frequency 1000> <--profile-output filename> myfile.luau
```

to get a trace file that can be inspected by ui.perfetto.dev.

This profiler works by sampling callstacks at a given frequency, assigning the delta time between samples to _each_ frame in the currently observed callstack(which is a simplifying assumption, but if your sample frequency is high enough, end up being okay). By tracking callstacks between calls from the VM to `interrupt`, we can emit begin and end events by diffing the stacks, nesting as appropriate. For example, a callstack like:
``` 
main -> foo -> bar -> baz
```
should produce:
```
BEGIN_main, 
   BEGIN_foo, 
     BEGIN_bar, 
        BEGIN_baz, 
        END_baz, 
      END_bar, 
   END_foo, 
END_main
```


annieetang:
Exciting!! Could you also add this to CLI docs?

docs/cli/run.md
```diff
@@ -14,8 +15,30 @@ is equivalent to:
 lute <script.luau> [args...]
 ```
 
+```bash
+lute --profile [--profile-output somefile] [--frequency <number> ] <script.luau> [args...]
+```
+will profile the execution of `script.luau` at the provided frequency, outputting the trace to a file called
```
skberkeley, line 22: can you add the units of frequency?

skberkeley: oh i see it below, although it might still be useful here too

docs/cli/run.md
```diff
@@ -14,8 +15,30 @@ is equivalent to:
...
 ## Options
 
 ### `-h, --help`
 
 Display this usage message.
+
+### `--profile`
+
+Enable profiling for the script being passed to `lute run`. Default sampling frequency is
+10000Hz, and default profile name is <datetime>_<filename>.json.
```
skberkeley:
```suggestion
10000Hz, and default output filename is <datetime>_<filename>.json.
```

examples/profile_test.luau
skberkeley, line 1: i wonder if it's useful to include the profile we get with this example too

docs/cli/run.md
skberkeley: i think we should also document the default frequency here and below

lute/cli/src/climain.cpp
```diff
@@ -125,12 +132,17 @@ bool runBytecode(
...
+    std::optional<ProfileOptions> profileOptions
 )
 {
     // module needs to run in a new thread, isolated from the rest
     lua_State* L = lua_newthread(GL);
 
+    if (profileOptions)
+    {
+        profilerStart(L, profileOptions->frequency);
+    }
```
skberkeley:
```suggestion
        profilerStart(L, profileOptions->frequency);
```
nit

lute/cli/src/climain.cpp
```diff
@@ -183,9 +210,14 @@ static bool runFile(Runtime& runtime, const char* name, lua_State* GL, int progr
 
     std::string chunkname = "@" + normalizePath(name);
 
-    std::string bytecode = Luau::compile(*source, copts());
+    Luau::CompileOptions opts = copts();
+    if (profileOptions)
+    {
+        opts.optimizationLevel = 2;
+    }
```
skberkeley: this feels worth documenting
should we also expose it as a cli option?

lute/cli/src/climain.cpp
```diff
@@ -280,6 +314,30 @@ int handleRunCommand(int argc, char** argv, int argOffset, bool packageAwareness
...
+        }
+        else if (strcmp(currentArg, "--frequency") == 0)
+        {
+            if (i + 1 >= argc)
+            {
+                reporter.reportError("Error: --frequency requires a value argument.");
+                reporter.reportOutput(packageAwareness ? PKGRUN_HELP_STRING : RUN_HELP_STRING);
+                return 1;
+            }
+            profileOpts.frequency = std::stoi(argv[++i]);
```
skberkeley, line 337: should we error if these options are present and `--profile` isn't?

lute/cli/src/climain.cpp
```diff
@@ -309,6 +367,32 @@ int handleRunCommand(int argc, char** argv, int argOffset, bool packageAwareness
...
+            char dateTimeStr[20];
+            std::strftime(dateTimeStr, sizeof(dateTimeStr), "%Y-%m-%d_%H-%M-%S", localTime);
+
+            std::string baseName = filePath;
+            size_t lastSlash = baseName.find_last_of("/\\");
+            if (lastSlash != std::string::npos)
+                baseName = baseName.substr(lastSlash + 1);
+            size_t lastDot = baseName.find_last_of('.');
+            if (lastDot != std::string::npos)
+                baseName = baseName.substr(0, lastDot);
```
skberkeley: naive question: is there a filesystem path library in c++ we could use?

skberkeley (commented):
i'm not 100% sure how the profiler code works so i'm going to leave it to someone else to review (unless you have the time to explain it to me lol)

lute/cli/src/climain.cpp
vrn-sn: There is `<filesystem>` but we don't use it for portability reasons. Although we do have our own `Luau/FileUtils.h`—might be some overlap that we could use.

docs/cli/run.md
Vighnesh-V: Yup, I'll add this.

examples/profile_test.luau
Vighnesh-V: That's a good idea, I can add that.

lute/cli/src/climain.cpp
Vighnesh-V: You know, I think this is a hallucination by me. I read: https://luau.org/guides/profile/ and thought by 'optimized version of the interpreter' that the code itself had to be compiled by optimization level 2. I think what that was saying is in a optimized build of _lute_, you should run the profiler.

I can delete this code.

Vighnesh-V: Separately, we should introduce a command line flag to help set the optimization level.

lute/cli/src/climain.cpp
Vighnesh-V: yeah, we are trying to avoid the <filesystem> header where possible. We also have lute/fileutils.h - I could check that and Luau/FileUtils.h, and if this function isn't present, add it there.

lute/cli/src/climain.cpp
Vighnesh-V: I really wish we had a better C++ argparsing solution. I'll add the error handling here.

Vighnesh-V: I added this!

lute/cli/src/profiler.cpp
```diff
@@ -1,75 +1,109 @@
...
-            stack += ',';
-            if (ar.linedefined > 0)
-                stack += std::to_string(ar.linedefined);
-        }
+    while (iterPrev > 0 && iterCurr > 0 && gProfiler.previousStack[iterPrev - 1] == currentStack[iterCurr - 1])
+    {
+        commonDepth++;
+        iterPrev--;
+        iterCurr--;
+    }
```
Vighnesh-V, line 94: @skberkeley Instead of doing the std::reverse followed by iteration forward, I did the iteration in reverse like we talked about. The reason that we're checking > 0 and then subtracting 1 is that size_t types are unsigned, so if we subtracted one on lines 83 (where the iteration starts), we would underflow to SIZE_MAX.

tests/std/syntax/parser.test.luau
```diff
@@ -127,6 +127,9 @@ test.suite("Parser", function(suite)
 			if entry.type ~= "file" then
 				continue
 			end
+			if entry.name.sub(entry.name, -5) == ".json" then
```
skberkeley: dym using `stringext.hassuffix` instead for readability?

skberkeley (approved):
lgtm, although i would appreciate another set of eyes on the profiling logic

lute/cli/src/profiler.cpp
```diff
@@ -1,75 +1,109 @@
...
 
-        stack.clear();
+    std::vector<FrameInfo> currentStack;
+    lua_Debug dbg;
+    for (int level = 0; lua_getinfo(L, level, "sln", &dbg); ++level)
+    {
+        FrameInfo frame;
+        frame.name = dbg.name ? dbg.name : "(anonymous)";
+        frame.file = dbg.short_src ? dbg.short_src : "(unknown)";
```
vrn-sn:
```suggestion
        frame.file = dbg.source ? dbg.source : "(unknown)";
```

I think `source` is a better fit since the dump will be consumed by tooling. (Allows users to trace back to the exact filepath without truncation.)

lute/cli/src/profiler.cpp
```diff
@@ -1,75 +1,109 @@
...
+        frame.line = dbg.linedefined;
+        currentStack.push_back(frame);
```
vrn-sn:
```suggestion
        currentStack.push_back(std::move(frame));
```

lute/cli/src/profiler.cpp
```diff
@@ -1,75 +1,109 @@
...
+        commonDepth++;
+        iterPrev--;
+        iterCurr--;
+    }
 
-        if (!stack.empty())
-        {
-            gProfiler.data[stack] += elapsedTicks;
-        }
+    for (size_t i = 0; i < gProfiler.previousStack.size() - commonDepth; i++)
```
vrn-sn, line 96: Sanity check: `gProfiler.previousStack.size() - commonDepth` means this is all the frames above the common depth that we've detected as having ended, and we emit "ended" events in reverse call order.

lute/cli/src/profiler.cpp
```diff
@@ -1,75 +1,109 @@
...
+    {
+        const FrameInfo& frame = gProfiler.previousStack.at(i);
+        gProfiler.events.emplace_back('E', frame, gProfiler.currentTimestamp);
+    }
 
-        if (gc > 0)
-        {
-            gProfiler.gc[gc] += elapsedTicks;
-        }
+    for (size_t i = currentStack.size() - commonDepth; i > 0; i--)
```
vrn-sn, line 102: Sanity check: `currentStack.size() - commonDepth` means this is all new frames in this timestep, and we emit "begin" events in reverse order of the reversed frame list, so in normal call order.

lute/common/src/fileutils.cpp
```diff
@@ -164,4 +164,18 @@ bool isDirectory(const std::string& path)
...
 
+std::string getFilenameWithoutExtension(const std::string& path)
+{
+    std::string base = path;
+    size_t lastSlash = base.find_last_of("/\\");
+    if (lastSlash != std::string::npos)
+        base = base.substr(lastSlash + 1);
+    // Once we've stripped off the last '/' in the path, we need to trim everything after .
+    // in order to handle .foo.bar suffixes
+    size_t firstDot = base.find('.');
```
vrn-sn: Don't we want to strip away only what comes after the final dot?

i.e. `some/file.test.ext` -> `some/file.test`

lute/cli/src/profiler.cpp
Vighnesh-V: yup exactly.

lute/cli/src/profiler.cpp
Vighnesh-V: yup, exactly. Note that the frame list isn't reversed:

E.g.

```
vector<FrameInfo>{quux, baz, bar, foo, main}
```
Means that main called foo called bar ....

We iterate from the point of first difference in reverse to emit the begin events.

lute/common/src/fileutils.cpp
Vighnesh-V: I think when I wrote this I was of the opinion that all the extensions e.g. {foo.tar.xz |foo.test.luau} should just map to foo. For the dump name, all I wanted was <timestamp>_file.json.

lute/cli/src/profiler.cpp
Vighnesh-V: If you emit source, you get this:
```
@./examples/profile_test.luau
```

in the dump. I think short src works better here.

Vighnesh-V: image

lute/common/src/fileutils.cpp
skberkeley: I'm with Varun here, this is also what Node and Lute do (if this is the inverse of path.extname)

lute/cli/src/profiler.cpp
vrn-sn: I guess I mean to use `source` but just remove the `@` prefix since the rest of the chunkname will be the file path. `short_src` automatically truncates and is more meant for displaying to CLI frontends.

Vighnesh-V: This is what short_src produces:
image

lute/cli/src/profiler.cpp
Vighnesh-V: adding this!

lute/cli/src/profiler.cpp
vrn-sn: That's only because your example file path isn't long enough to need truncation. If it's long enough, you'll end up with something with `...` in the string.

lute/common/src/fileutils.cpp
Vighnesh-V: Oh this func also strips the path off as well.

Vighnesh-V: Cool - I'll tweak it to index to the last dot.

Vighnesh-V: done!

vrn-sn (approved):

PR #747, opened 2026-01-21
Add executable bit to release binaries

lukadev-0:
GitHub does not preserve permissions when uploading files as an artifact, causing the executable bit of the binaries within the release artifacts to be lost. This makes Lute not executable when installing it using [mise-en-place's GitHub backend](https://mise.jdx.dev/dev-tools/backends/github.html).

This PR re-adds the executable bit to the release artifacts.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -131,7 +131,7 @@ jobs:
           mkdir release
           for dir in ./artifacts/*; do
             base_name=$(basename "$dir")
-            chmod +x "$dir"/* # Artifacts don't preserve permissions, so re-add them.
+            chmod 755 "$dir"/* # Artifacts don't preserve permissions, so re-add them.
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -131,6 +131,7 @@ jobs:
           mkdir release
           for dir in ./artifacts/*; do
             base_name=$(basename "$dir")
+            chmod +x "$dir"/* # Artifacts don't preserve permissions, so re-add them.
```
vrn-sn:
```suggestion
            chmod 755 "$dir"/* # Artifacts don't preserve permissions, so re-add them.
```

I think we should just hardcode everything while we're at it so that we're not having to trust GitHub to preserve the other permissions.

vrn-sn (approved):

PR #748, opened 2026-01-21
Adds initial `lute doc` CLI

annieetang:
Initial version of a new `lute doc` Command Line Interface! 

tldr; this is basically a refactor of the `scripts/reference.luau` file via CLI to make reviewing easier

Details:

This currently follows the exact logic of `scripts/reference.luau` that string matches generating the `@lute` and `@std` parts of lute.luau.org/reference as well as the default pages, so it's very much hard coded and not intended on being used by any other repo yet, but I wanted to get the CLI set up and iterate incrementally on this afterwards to avoid massive PRs.

Current workflow:

```bash
$ lute doc
```

is the equivalent of 

```bash
$ lute run scripts/reference.luau
```

There's an option to specify the output directory, which github actions needs because it's run in a different cwd I think? So you can also run with 

```bash
$ lute doc -o ./specified-docs-folder
```

That generates this lute repo's `definitions/` and `std/` docs the exact same way the website shows it as.

#### Immediate next steps (PRs):

- [x] Refactor and add an `--module` option that specifies which module to generate, so the current `generateAllDocs` won't be hard coding the `definitions` and `std` paths.

     * Note: for this I'd then want `--module` to be a required option and not have the CLI generate docs for everything since the script isn't reliable yet. 

- [ ]  Update the `processDirectory` to use the `typeofmodule` API in `@lute/lute` that gets the actual type info of the module instead of this string matching pattern. 

    * `luau.typeofmodule` is currently returning a string representation of the module type so this code will probably look a bit ugly too until we're unblocked on having that API be represented in some data structure form.

- [ ] Have `lute doc` generate docs for everything in the current working directory by default, so `generateAllDocs` actually generates all docs.

then there's more after that but just listing these for now!


annieetang:
open to thoughts abt splitting up the PRs and combining or doing some steps with the initial PR if that's more user friendly?

Also, I avoided adding the docs for this for now because it's not ready to be used by anyone else other than the GitHub actions bot, so I didn't want it to appear on our site yet in https://lute.luau.org/cli/

lute/cli/commands/doc/init.luau
```diff
@@ -30,6 +33,22 @@ type PropertySignature = {
 	blockEq: string?,
 }
 
+local function printHelp()
+	print([[
+Usage: lute doc [OPTIONS]
+
+Generate type-directed Markdown documentation for Luau module(s).
```
skberkeley: is it truly type directed rn?

lute/cli/commands/doc/init.luau
```diff
@@ -329,26 +331,72 @@ local function generateLibraryIndex(title: string, alias: string, modulePath: pa
...
+
+	-- Generate std subfolder index with table of modules
+	local stdIndex = generateLibraryIndex("std", "std", stdlibPath)
+	fs.writestringtofile(path.join(stdlibReferencePath, "index.md"), stdIndex)
+
+	-- Mock module in test/ for testing purposes
+
+	-- local testPath = path.join(process.cwd(), "test")
+	-- local testReferencePath = path.join(process.cwd(), "test")
+	-- processDirectory(testPath, testReferencePath, testPath, "test")
```
skberkeley: can you remove this commented code and add a test?

skberkeley (changes requested):

lute/cli/commands/doc/init.luau
annieetang: no i forgot to remove that when I decided to just use the old version for this initial pr, thx for catch!

lute/cli/commands/doc/init.luau
annieetang: added test file but I can't run `lute doc` yet through the tests bc the script hard-codes the output directories and I get a write permission denied with `process.run`, but the github actions `build-docs-site` runs `lute doc` if that's good with you until I do the next step

annieetang: nvm got it working, thx @skberkeley

tests/cli/doc.test.luau
```diff
@@ -0,0 +1,45 @@
...
+		assert.eq(result.exitcode, 0)
+		assert.strcontains(result.stdout, "Documentation generation complete!")
+	end)
+
+	suite:case("generate test docs specified directory", function(assert)
+		local result = process.system(`cd docs && {lutePath} doc -o {path.format(tmpOutputDir)}`)
+
+		assert.eq(result.exitcode, 0)
+		assert.strcontains(result.stdout, "Documentation generation complete!")
+		assert.strcontains(result.stdout, tostring(tmpOutputDir)) -- "Files written to: <output dir>"
```
skberkeley, line 50: Sorry, one last ask: can you do some validation that files were created as expected too?

tests/cli/doc.test.luau
```diff
@@ -36,10 +37,18 @@ test.suite("lute-doc", function(suite)
 	end)
 
 	suite:case("generate test docs specified directory", function(assert)
+		if fs.exists(tmpOutputDir) then
+			fs.removedirectory(tmpOutputDir, { recursive = true })
+		end
+
+		fs.createdirectory(tmpOutputDir)
```
skberkeley, line 44: we should probably clean this up at the end too

skberkeley (approved):

tests/cli/doc.test.luau
annieetang: tyty

PR #751, opened 2026-01-22
Adds a CI job that prevents us from adding new type errors to the lute repo.

Vighnesh-V:
This PR adds a script called `check.luau` that gathers up every .luau file in `std/libs/` and `tests/` and passes them to the `lute check` command. `check.luau` takes an optional `--update` argument that will update a list of all the type errors. I'm using the diffext battery to show the diff, but I think we need to add some more pretty printing options for that to be useful atm.

I've also added a CI job that will prevent PR's from merging if they introduce a new type error.


tools/check.luau
```diff
@@ -0,0 +1,179 @@
+--!strict
+
```
skberkeley: idt we need this, `.luaurc` has strict mode enabled already

tools/check.luau
```diff
@@ -0,0 +1,179 @@
...
+local process = require("@std/process")
+local cli = require("@batteries/cli")
+local difftext = require("@batteries/difftext")
+
+local FAILLIST_PATH = "tools/check-faillist.txt"
+
+local TYPECHECK_PATHS = {
+	"tests",
+	"lute/std/libs",
+}
```
skberkeley, line 16: you should add `lute/cli`, `tools`, and `examples` too

tools/check.luau
```diff
@@ -0,0 +1,179 @@
...
+		table.insert(lines, `{err.file}:{err.line}:{err.col}-{err.colEnd}`)
+	end
+
+	return table.concat(lines, "\n") .. "\n"
+end
+
+local function main(args: { string })
+	local parser = cli.parser()
+	parser:add("update", "flag", { help = "Update the faillist file", aliases = { "u" } })
+	parser:add("check", "positional", { help = "This script" })
```
skberkeley, line 139: wut?

tools/check.luau
```diff
@@ -0,0 +1,179 @@
...
+	return files
+end
+
+local function getAllFiles(): { string }
+	local allFiles: { string } = {}
+	for _, dir in TYPECHECK_PATHS do
+		local files = collectFiles(dir)
+		for _, file in files do
+			table.insert(allFiles, file)
+		end
```
skberkeley: `tableext.extend`?

tools/check.luau
```diff
@@ -0,0 +1,179 @@
...
+	end
+	table.sort(allFiles)
```
skberkeley: If we statically sort TYPECHECK_PATHS, idt we need this given you're sorting above too?

tools/check.luau
```diff
@@ -0,0 +1,179 @@
...
+		if a.file ~= b.file then
+			return a.file < b.file
+		end
+		if a.line ~= b.line then
+			return a.line < b.line
+		end
+		return a.col < b.col
+	end)
+	return errors
+end
```
skberkeley, line 124: is it worth doing a sorted insert instead?

tools/check.luau
Vighnesh-V: I did this to sort the paths by filename too E.g.

`commonPrefix/...../a.luau` 
should be before
``commonPrefix/...../b.luau`

Vighnesh-V: I don't think walk() makes any guarantees about iteration order.

Vighnesh-V: Oh I see what you mean - collectFiles also sorts.

tools/check.luau
Vighnesh-V: talked offline - we're leaving it as is since it's simpler to reason about.

tools/check.luau
Vighnesh-V: the name of the script itself gets passed e.g. "check.luau" or "luthier.luau" as an argument to scripts run with the run subcommand. Luthier handles this by just adding an argument for the "self script". E.g.:
```
args:add("help", "flag", { help = "show the help message" })
args:add("luthier", "positional", { help = "the luthier script", required = true }) -- should this be handled automatically?
args:add("subcommand", "positional", { help = "command to execute", required = true })
```
This just handles that. It's really weird, I know.

skberkeley: oh i see, maybe worth documenting that it's a dummy arg to consume a positional arg?

skberkeley (approved):

PR #752, opened 2026-01-22
Support `target` in lint diagnostics

wmccrthy:
We can use additional diagnostic metadata like a `target` URI to surface additional information about a lint violation in Mandolin [(vscode API)](https://code.visualstudio.com/api/references/vscode-api#Diagnostic).

This PR supports this by:
- extending `LintViolation` type to include `target`
- extending `Diagnostic` type to align with the vscode API
- serializing `diagnostic.code` as an object with `value` and `target` fields, if the passed violation includes a `target`

This provides a baseline that can be easily consumed by Mandolin to surface such metadata. One can imagine the convenience of being able to follow a lint violation to relevant docs, or something similar:
Screenshot 2026-01-22 at 9 58 52 AM


wmccrthy:
I wonder if we should include `target` in `lute lint` CLI output too

lute/cli/commands/lint/types.luau
```diff
@@ -14,6 +14,7 @@ export type LintViolation = {
 		read location: syntax.span?, -- if nil, applies to whole violation location
 		read fix: string,
 	}?,
+	read target: string?,
```
skberkeley: can you add a comment explaining what this is for so that future lint rule authors know how to use it?

lute/cli/commands/lint/lsp/types.luau
```diff
@@ -1,7 +1,7 @@
 export type Diagnostic = {
 	read range: Range, -- zero-based
 	read severity: number, -- 1: Error, 2: Warning, 3: Information, 4: Hint
-	read code: string, -- lint name
+	read code: Code, -- lint name | value: lint name, target: URI for additional information
```
skberkeley: nit: can we add an extra `codeDescription` field instead to match the [`LSP` spec for a Diagnostic](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#diagnostic)? i agree that what you have right now aligns nicer with how Mandolin currently works, but I'd like to keep in mind the eventual goal of having Mandolin implement/consume a full-fledged LSP server

wmccrthy: oh sure! do you think it's worth renaming the `target` field on the LintViolation type for better alignment w/ this?

skberkeley (approved):

Issue #708, opened 2026-01-08
`lute compile` fails to detect a luaurc when the entry point is not in the same directory as the luaurc.


Vighnesh-V:
```
repo/
    .luaurc -- defines @ext -> ./src/batteries
    src/
        entry.luau -> requires @ext/file
        batteries/
            file.luau
```

Running `lute compile src/entry.luau` from the repo/ directory will fail.

Vighnesh-V:
Fixed by @raymondnumbergenerator in https://github.com/luau-lang/lute/pull/753

PR #753, opened 2026-01-22
Fix `lute compile` fails to detect a luaurc when the entry point is not in the same directory as the luaurc

raymondnumbergenerator:
Addresses https://github.com/luau-lang/lute/issues/708

In the linked issue, removing the entry point limit to the `.luaurc` discovery doesn't fix the issue. When doing so, `src` doesn't exist as a path in the bundle VFS and it crashes at runtime when trying to `require("@ext/bat")`

The LCR calculation calculates `LCR("repo/src/entry.luau", "repo/src/batteries/bat.luau") = src`, and the final bundle VFS gets laid out as:
```
@bundle
  entry.luau
  batteries
    bat.luau
```

The `.luaurc` has this, which doesn't actually match with the bundle VFS.
```
    "aliases": {
        "ext": "./src/batteries/"
    }
```

So when trying to `require("@ext/bat")`, lute tries to navigate to `@bundle/src/batteries`. `"@bundle/src"` doesn't exist and crashes [here](https://github.com/luau-lang/lute/blob/e11f8b0c0d933e7bbd0463c9213a364fb966f920/lute/require/src/modulepath.cpp#L119).

This fix adds any discovered `.luaurc` to the LCR calculation, which makes the paths between the bundle VFS and the `.luaurc` at the repo root symmetric.

Alternatively, could have used the LCR to strip prefixes in the bundled `.luaurc`.

### DIFF
--- a/tests/src/staticrequires.test.cpp
+++ b/tests/src/staticrequires.test.cpp
@@ -126,20 +126,23 @@ TEST_CASE_FIXTURE(LuteFixture, "staticrequiretracer_require_alias")
 {
     std::string luteProjectRoot = getLuteProjectRootAbsolute();
     std::string testDir = joinPaths(luteProjectRoot, "tests/src/staticrequires");
-    std::string entryPoint = joinPaths(testDir, "lib/requirealias.luau");
+    std::string entryPoint = joinPaths(testDir, "dep/requirealias.luau");
 
     StaticRequireTracer tracer{getReporter()};
     tracer.trace(entryPoint);
 
     auto pairs = tracer.getStaticRequirePairs();
 
-    // requirealias.luau requires @example/option.luau, should resolve correctly
+    // requirealias.luau requires @otherlib/example.luau, should resolve correctly
     REQUIRE(pairs.size() == 2);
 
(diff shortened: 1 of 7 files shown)


tests/src/staticrequires.test.cpp
```diff
@@ -122,6 +122,34 @@ TEST_CASE_FIXTURE(LuteFixture, "staticrequiretracer_relative_paths")
...
+    std::string luteProjectRoot = getLuteProjectRootAbsolute();
+    std::string testDir = joinPaths(luteProjectRoot, "tests/src/staticrequires");
+    std::string entryPoint = joinPaths(testDir, "dep/requirealias.luau");
+
+    StaticRequireTracer tracer{getReporter()};
+    tracer.trace(entryPoint);
+
+    auto pairs = tracer.getStaticRequirePairs();
+
+    // requirealias.luau requires @otherlib/example.luau, should resolve correctly
```
Vighnesh-V: I think the comment might be out of date - this should require @nested/example right?

Vighnesh-V (approved):
One nit: comment might be out of date. Otherwise, looks good!

PR #759, opened 2026-01-27
Adds dev readme to docs

skberkeley:
Adds README describing how to run docsite locally for testing/editing documentation

### DIFF
--- a/docs/.gitignore
+++ b/docs/.gitignore
@@ -2,3 +2,4 @@ node_modules
 .vitepress/cache
 dist
 reference
+docs/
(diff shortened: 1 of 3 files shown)


docs/.gitignore
```diff
@@ -2,3 +2,4 @@ node_modules
 .vitepress/cache
 dist
 reference
+docs/
```
annieetang: I think we need this, no? some of the files in `docs/` need to be in our repo bc they're manually written

annieetang: if you don't want the `README.md` to display with vitepress, there's an exclusion option in `.vitepress/config.mts`, maybe `excludeByGlobPattern`?

skberkeley: this ignores `docs/docs`, which is autogenerated by `npx vitepress`

annieetang (approved):
thx!

PR #764, opened 2026-01-28
Fix bug preventing `lute pkg run` from detecting files with `require-by-string` semantics

vrn-sn:
In this code, we created `validPath` from `filePath` using require-by-string semantics, but we forgot to pass `validPath` along when converting it into an absolute path.

This manifests when doing `lute pkg run ./run/this/script`, where we're trying to run `./run/this/script.luau`.


skberkeley (approved):
dym adding a test suite for `lute pkg run`, even if it's just for this case?

vrn-sn:
Oops, I had auto-merge enabled

vrn-sn:
Going to just file it as a pending issue. Aiden and I need to add unit tests for `loom-core` anyway.

PR #768, opened 2026-01-29
CI improvements: no unnecessary bootstrapping and Slack bot integration

vrn-sn:
This PR aims to balance fast PR checks with ensuring that we never ship broken binaries.

PRs:
- Bootstrapped builds (i.e. 3 sequential fresh builds) no longer run as part of PR gated commit checks. This improves velocity, especially for stacked PRs, where each upstream PR being merged in requires a new run of the full CI suite for downstream PRs.
- Actual speedup: 
  - ~9.5 minutes down to ~6.5 minutes (~1.5x) when we get unlucky with a slow MacOS runner.
  - ~9.5 minutes down to ~5 minutes (~1.9x) in the typical case.

When a PR is merged in to `primary`:
- We re-run all the same checks, _including all bootstrapped builds_. If a build-breaking issue made it through PR checks without being detected, it will be caught here. This leads us to:

The new "Lute Build Stability" Slack bot: when CI status at the tip of the `primary` branch _changes_ (to prevent noise), we send a Slack message to an internal channel with one of these messages.
- 🚨 Lute CI has started failing
- ✅ Lute CI is passing again

To avoid automatically shipping a broken release, the nightly release pipeline early-exits (skipping the release for that day) until CI is passing again.


vrn-sn:
Looks like we've shaved the time for build-docs-site, lute-check, and lute-lint down from 7–8 min. to around 3–4 min.

.github/workflows/ci.yml
```diff
@@ -37,10 +37,22 @@ jobs:
...
+            echo "proceed=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "proceed=false" >> "$GITHUB_OUTPUT"
+          fi
+
       - uses: actions/checkout@v4
 
       - name: Setup and Build Lute
         id: build_lute
+        if: ${{ steps.should_run.outputs.proceed == 'true' }}
```
skberkeley, line 60: do you need to add `needs: [ should_run ]`?

.github/workflows/ci.yml
```diff
@@ -186,10 +197,48 @@ jobs:
...
             echo "All jobs succeeded"
+            echo "result=success" >> "$GITHUB_OUTPUT"
           else
             echo "One or more jobs failed"
+            echo "result=failure" >> "$GITHUB_OUTPUT"
             exit 1
           fi
+      - name: Post-aggregation notification
+        if: >
+          always() &&
```
skberkeley, line 217: isn't this redundant?

.github/workflows/ci.yml
```diff
@@ -186,10 +197,48 @@ jobs:
...
+          if [[ "$LAST_STATUS" == "success" ]] && [[ "${{ steps.aggregator.outputs.result }}" == "failure" ]]; then
+            TEXT=":rotating_light: *Lute CI has started failing*"
+          elif [[ "$LAST_STATUS" == "failure" ]] && [[ "${{ steps.aggregator.outputs.result }}" == "success" ]]; then
+            TEXT=":white_check_mark: *Lute CI is passing again*"
+          fi
+
+          if [[ -z "$TEXT" ]]; then
+            echo "No status change detected. Skipping notification."
+            exit 0
+          fi
```
skberkeley:
```suggestion
          if [[ "$LAST_STATUS" == "${{ steps.aggregator.outputs.result }}"  ]]; then
            echo "No status change detected. Skipping notification."
            exit 0
          elif [[ "$LAST_STATUS" == "failure" ]]; then
            TEXT=":white_check_mark: *Lute CI is passing again*"
          elif [[ "$LAST_STATUS" == "success" ]]; then
            TEXT=":rotating_light: *Lute CI has started failing*"
          fi
```

.github/workflows/ci.yml
```diff
@@ -37,10 +37,22 @@ jobs:
...
     steps:
+      # Non-bootstrap jobs always run; all jobs run on pushes to primary branch.
+      - name: Determine if this matrix job should run
+        id: should_run
+        shell: bash
+        run: |
+          if [[ "${{ matrix.use-bootstrap }}" != "true" || ( "${{ github.event_name }}" == "push" && "${{ github.ref }}" == "refs/heads/primary" ) ]]; then
+            echo "proceed=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "proceed=false" >> "$GITHUB_OUTPUT"
```
skberkeley, line 53: can we not just `exit 0` here instead of threading this state forward?

vrn-sn: There needs to be a way to prevent the later steps from executing. If this just exits, they'll still run.

.github/workflows/ci.yml
vrn-sn: From what I understand, it has to be there, otherwise a failure in the previous step will prevent this step from running.

.github/workflows/ci.yml
vrn-sn: Don't think so, steps are executed sequentially.

.github/workflows/ci.yml
skberkeley: oh i see, and i suppose `exit 1` might report a failure?

.github/workflows/release.yml
```diff
@@ -21,6 +21,17 @@ jobs:
       contents: read
     
     steps:
+      - name: Validate CI status
+        run: |
+          LAST_STATUS=$(curl -s -H "Authorization: Bearer ${{ secrets.GITHUB_TOKEN }}" \
```
Vighnesh-V, line 26: slick :)

.github/workflows/ci.yml
vrn-sn: Yeah, exactly.

.github/workflows/ci.yml
vrn-sn: pushed this, thx!

.github/workflows/ci.yml
```diff
@@ -22,25 +22,50 @@ jobs:
...
           - os: macos-latest
             options: --enable-sanitizers
+            run-on-event: all
+          - os: ubuntu-latest
+            options: --c-compiler gcc --cxx-compiler g++
+            use-bootstrap: true
+            run-on-event: pull_request
+          - os: ubuntu-latest
+            options: --c-compiler clang --cxx-compiler clang++
+            use-bootstrap: true
```
Vighnesh-V: Why do these need to all use bootstrap if we're running this in a separate pass?

skberkeley (approved):

PR #769, opened 2026-01-29
Adds `TypeSerializer` and updates `luau.typeofmodule` to return a `TypePack` of `Type`s

annieetang:
### Problem

`luau.typeofmodule` currently returns a `string` representing the module's return type, which is hard for users to actually use

### Solution

To move towards a useable Luau documentation generator (and the `lute doc` CLI), this PR exposes a similar version of Luau's `Type` and `TypePack` data structures (but it's just a table of the `Luau::Type` properties), which can then be used to represent a given module's return type as a data structure instead of a giant string.

We can then serialize with the exposed functions `serializeType(TypeId ty, lua_State* L)` or `serializeTypePack(TypePackId tp, lua_State* L)`. It's very similar to how `TypeFunctionRuntime` and `TypeFunctionRuntimeBuilder` work, but we weren't able to use it because it's interlaced with memory allocation and runtime stuff that clashes with Lute runtime things

So most of this is wrapper code of the `Luau::TypeVisitor`, with the `serialize()` functions following a similar pattern as the `AstSerializer`, but these ones are based on the below `type` signature (and see https://luau.org/types-library/). 

```lua
export type Type = {
	tag: "nil" | "unknown" | "never" | "any" | "boolean" | "number" | "string" | "buffer" | "thread" | "singleton" | "negation" | "union" | "intersection" | "table" | "metatable" | "function" | "extern" | "generic",

	-- for singleton type
	value: string | boolean | nil,

	-- for negation type
	inner: Type,

	-- for union and intersection types
	components: { Type },

	-- for table type
	properties: { [string]: { read: Type?, write: Type? } },
	indexer: { index: Type }?, -- TODO: add readresult and writeresult
	-- TODO: readindexer: { index: type, result: type } }?,
	-- TODO: writeindexer: { index: type, result: type } }?,
	metatable: Type?,

	-- for function type
	parameters: { head: { Type }?, tail: Type },
	returns: { head: { Type }?, tail: Type },
	generics: { Type },
	genericpacks: { TypePack },

	-- for extern type
	-- 'properties', 'metatable' are shared with table type
	parent: Type?,

	-- for generic type
	name: string?,
	ispack: boolean,
}

export type TypePack = {
	head: { Type }?,
	tail: Type?,
}
```

so `luau.typeofmodule: TypePack` and users can parse this table to process the type of a module!

Additionally, `TypeSerializer` doesn't serialize some types like
* `BoundType`, `FreeType`, `ErrorType`, `NoRefineType`, `BlockedType`, `PendingExpansionType`, `TypeFunctionInstanceType`
* `BoundTypePack`, `FreeTypePack`, `ErrorTypePack`, `BlockedTypePack`, `TypeFunctionInstanceTypePack`

since these aren't user facing

### Future Work

1. Most likely a `typevisitor` library (like `visitor.luau`) to help traverse and parse the `TypePack` and `Type` types, because the updated example looks ugly
2. This implementation does just represent everything in a table of values, so it's not the full userdata where you can call `tabletype:properties()` for ex, but I think building on that in a subsequent PR is reasonable? (i.e., more how `Span` works I think?) 😄


lute/luau/src/type.cpp
```diff
@@ -0,0 +1,554 @@
...
+#include "Luau/TypeFwd.h"
+#include "Luau/VisitType.h"
+
+#include "lua.h"
+
+namespace Luau
+{
+
+// Serializer for runtime types, modeled after AstSerialize but for TypeId/TypePackId
+struct TypeSerialize : public Luau::TypeVisitor
```
Vighnesh-V: Can you add a `final` specifier here?

lute/luau/src/type.cpp
```diff
@@ -0,0 +1,554 @@
...
+{
+    lua_State* L;
+    const char* tag;
```
Vighnesh-V: Does this tag need to be initialized?

lute/luau/src/type.cpp
```diff
@@ -0,0 +1,554 @@
...
+
+    // Helper to push the "tag" field
+    void pushTag(const char* tag)
+    {
+        lua_pushstring(L, tag);
+        lua_setfield(L, -2, "tag");
+    }
+
+    // Helper to set the "properties" field for table and extern types
+    void setPropertiesFields(std::map<Name, Property> props)
```
Vighnesh-V:
```suggestion
    void setPropertiesFields(const std::map<Name, Property>& props)
```

lute/luau/src/type.cpp
```diff
@@ -0,0 +1,554 @@
...
+                traverse(*prop.writeTy);
+                lua_setfield(L, -2, "write");
+            }
+            else
+            {
+                lua_pushnil(L);
+                lua_setfield(L, -2, "write");
+            }
+
+            lua_settable(L, -3);
```
Vighnesh-V: Might be nice to make the body of the for loop it's own helper function

lute/luau/src/type.cpp
```diff
@@ -0,0 +1,554 @@
...
+        lua_setfield(L, -2, "returns");
+
+        // Generics
+        lua_createtable(L, ftv.generics.size(), 0);
+        for (size_t i = 0; i < ftv.generics.size(); i++)
+        {
+            traverse(ftv.generics[i]);
+            lua_rawseti(L, -2, i + 1);
+        }
+        lua_setfield(L, -2, "generics"); // do we need to serialize genericpacks?
```
Vighnesh-V: functions can take generic packs I'm p sure.

lute/luau/src/type.cpp
```diff
@@ -0,0 +1,554 @@
...
+    }
+
+    // Luau table type:
+    // properties: { [type]: { read: type?, write: type? } },
```
Vighnesh-V: isn't properties a map from name -> { read: type?, write type? }

lute/luau/src/type.cpp
```diff
@@ -0,0 +1,554 @@
...
+        }
+
+        // Indexer
+        if (ttv.indexer)
+        {
+            lua_createtable(L, 0, 3);
+            traverse(ttv.indexer->indexType);
+            lua_setfield(L, -2, "index");
+
+            traverse(ttv.indexer->indexResultType);
```
Vighnesh-V: Is traverse guaranteed to push a type onto the stack?

tests/std/luau.test.luau
```diff
@@ -7,7 +7,7 @@ local test = require("@std/test")
 local tmpDir = system.tmpdir()
 
 test.suite("LuauTypeOfModuleSuite", function(suite)
```
Vighnesh-V, line 9: Can you create a new test suite, either in this file or another that explicitly tests type serialization. I would love to have some explicit cases for each kind of type. I also think some c++ tests _could_ be valuable too, but I'm not actually sure how to inspect a pushed luau type to verify it's structure.

lute/luau/src/type.cpp
```diff
@@ -0,0 +1,554 @@
...
+            tag = "string";
+            break;
+        case PrimitiveType::Thread:
+            tag = "thread";
+            break;
+        case PrimitiveType::Buffer:
+            tag = "buffer";
+            break;
+        default:
+            LUAU_UNREACHABLE();
```
Vighnesh-V: In this case, note that the tag could be `nil` if we hit a primitive type that doesn't exist - e.g. if luau adds a new primitive and we havent added a check here in this case.

For this, we should probably emit a `lua_error` and the user can pcall the typeof function.

Vighnesh-V (changes requested):

lute/luau/src/type.cpp
annieetang: ah nope that was left from an older iteration, thx!

lute/luau/src/type.cpp
annieetang: talked in person and decided that after `traverse` we're gonna add a `lua_rawcheckstack` to check if something got added to avoid setting a field when nothing was actually pushed

tests/std/luau.test.luau
annieetang: yup will add these in a separate PR since this one's getting pretty large

definitions/luau.luau
```diff
@@ -686,8 +686,61 @@ function luau.resolverequire(path: string, fromchunkname: string): string
 	error("not implemented")
 end
 
--- For now, we return a string representation of the module's type, but we will expand it to some Luau data structure representation of Type (similar to the AST types) in a subsequent PR.
-function luau.typeofmodule(modulepath: string): string
+export type Type = {
+	tag: "nil"
+		| "unknown"
```
Vighnesh-V, line 691: I think we're missing metatable type here?

lute/luau/src/type.cpp
```diff
@@ -0,0 +1,584 @@
...
+            tag = "string";
+            break;
+        case PrimitiveType::Thread:
+            tag = "thread";
+            break;
+        case PrimitiveType::Buffer:
+            tag = "buffer";
+            break;
+        default:
+            tag = "nil";
```
Vighnesh-V: You can skip setting this tag, since lua_error will throw.

lute/luau/src/type.cpp
```diff
@@ -0,0 +1,584 @@
...
+            lua_createtable(L, 0, 3);
+            traverse(ttv.indexer->indexType);
+            lua_rawcheckstack(L, 2);
+            lua_setfield(L, -2, "index");
+
+            traverse(ttv.indexer->indexResultType);
+            lua_rawcheckstack(L, 2);
+            lua_setfield(L, -2, "readresult");
+
+            traverse(ttv.indexer->indexResultType);
```
Vighnesh-V: We're doing this twice here it looks like. However, ttv.indexer->indexResultType is the same type? Wouldn't this get caught by the type visitor seenSet?

Also, don't indexers have the same read and write type for their result? Does it make sense to make these different?

definitions/luau.luau
```diff
@@ -686,8 +686,61 @@ function luau.resolverequire(path: string, fromchunkname: string): string
...
+	properties: { [string]: { read: Type?, write: Type? } },
+	indexer: { index: Type, readresult: Type, writeresult: Type }?,
+	readindexer: { index: Type, result: Type }?,
+	writeindexer: { index: Type, result: Type }?,
+	metatable: Type?,
+
+	-- for function type
+	parameters: { head: { Type }?, tail: Type },
+	returns: { head: { Type }?, tail: Type },
+	generics: { Type },
```
Vighnesh-V, line 728: This should have generic packs right?

Vighnesh-V (changes requested):

definitions/luau.luau
annieetang: yeah haha, i wonder if these should also be updated in the `EmbeddedBuiltinDefinitions.cpp` in Luau... cuz these were copied over from there and i'm realizing that some of it is a bit outdated

annieetang:
Following up from discussion w/ vighnesh: going to leave out parts of `indexer`, including the `readindexer` and `writeindexer` properties for a separate PR since this one is getting large and I want to spend some more time thinking about how to diverge the read / write index results and I'm not sure the current `traverse()` implementation let's us do that when visiting the same type (bc the indexer has the same read and write type currently)

Vighnesh-V (approved):
Approved, but in the follow up we should add the C++ tests and ensure that table indexers can have read and write result types.

PR #777, opened 2026-01-31
Add `AstExprTableItem` to `AstExpr` union

wmccrthy:
I found that `AstExprTableItem` is missing from the `AstExpr` type definition. This was leading to annoying type errors in a side project.


skberkeley:
This is deliberate, `AstExprTableItem` is _not_ an `AstExpr`. For example, it doesn't make sense to have an `AstExprTableItem` instance be a subnode of an `AstExprBinary`.

PR #778, opened 2026-02-02
Adds tests to TypeSerializer pt 1

annieetang:
(following up from https://github.com/luau-lang/lute/pull/769)
tests tests tests for `TypeSerializer`

and adds error handling for types that we shouldn't serialize, so all overrides are implemented now

there will be a pt 2 for: FunctionType, TableType, MetatableType, ExternType, TypePack, VariadicTypePack, GenericTypePack tests

### DIFF
--- a/lute/luau/src/luau.cpp
+++ b/lute/luau/src/luau.cpp
@@ -2913,11 +2913,6 @@ int typeofmodule_luau(lua_State* L)
 
     // Serialize and push the return type
     serializeTypePack(modulePtr->returnType, L);
-
-    // check the table was created and set it read-only
-    lua_rawcheckstack(L, 1);
-    lua_setreadonly(L, -1, 1);
-
     return 1;
 }
 
--- a/lute/luau/src/luau.cpp
+++ b/lute/luau/src/luau.cpp
@@ -2913,6 +2913,7 @@ int typeofmodule_luau(lua_State* L)
 
     // Serialize and push the return type
     serializeTypePack(modulePtr->returnType, L);
+
     return 1;
 }
 
--- a/lute/luau/src/luau.cpp
+++ b/lute/luau/src/luau.cpp
@@ -2914,6 +2914,10 @@ int typeofmodule_luau(lua_State* L)
     // Serialize and push the return type
     serializeTypePack(modulePtr->returnType, L);
 
+    // check the table was created and set it read-only
+    lua_rawcheckstack(L, 1);
+    lua_setreadonly(L, -1, 1);
+
     return 1;
 }
 
--- a/lute/luau/src/type.cpp
+++ b/lute/luau/src/type.cpp
@@ -5,6 +5,7 @@
 #include "Luau/VisitType.h"
 
 #include "lua.h"
+#include "lualib.h"
 
 namespace Luau
 {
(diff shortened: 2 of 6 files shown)


lute/luau/src/luau.cpp
```diff
@@ -2914,6 +2914,10 @@ int typeofmodule_luau(lua_State* L)
     // Serialize and push the return type
     serializeTypePack(modulePtr->returnType, L);
 
+    // check the table was created and set it read-only
+    lua_rawcheckstack(L, 1);
```
skberkeley: doesn't this check that there's a slot available? if you're just editing a value already on the stack idt you need this

vrn-sn: Was going to comment the same thing. If we are worried about it, we could always `LUAU_ASSERT` that `lua_gettop(L)` did increase by 1 after calling `serializeTypePack`.

Although, is there a reason we don't just do the `lua_setreadonly` in the `serializeTypePack` body itself?

vrn-sn: Another idea: a common pattern is to have APIs that touch the Luau stack like `serializeTypePack` return the number of values pushed to the stack (in this case, 1). If we did that, we could just leave that function responsible for communicating this information back up to this caller.

annieetang: fixed in https://github.com/luau-lang/lute/pull/780

lute/luau/src/type.cpp
```diff
@@ -558,6 +559,84 @@ struct TypeSerialize final : public Luau::TypeVisitor
...
+    // Non-Serialized Types
+
+    bool visit(TypeId ty) override
+    {
+        // NOTE: `TypeSerialize` should explicitly visit _all_ types and type packs,
+        // otherwise it's prone to serializing types that should not be serialized.
+        LUAU_ASSERT(false);
+        LUAU_UNREACHABLE();
+    }
+    bool visit(TypeId ty, const BoundType& btv) override
```
skberkeley, line 523: nit: can you put whitespace between function defs?

tests/src/typeserializer.test.cpp
```diff
@@ -0,0 +1,308 @@
...
+#include "lute/type.h"
+
+#include "Luau/TypeArena.h"
+#include "Luau/TypeFunction.h"
+
+using namespace Luau;
+
+static void requireStringField(lua_State* L, const char* field, const char* expected)
+{
+    lua_getfield(L, -1, field);
```
skberkeley, line 14: do we need a checkstack here?

tests/src/typeserializer.test.cpp
```diff
@@ -0,0 +1,308 @@
...
+    lua_pop(L, 1);
+}
+
+// Primitive Types
+
+TEST_CASE_FIXTURE(TypeSerializeFixture, "serialize_primitive_nil_type")
+{
+    TypeId ty = arena.addType(PrimitiveType{PrimitiveType::NilType});
+
+    Luau::serializeType(L, ty);
```
skberkeley: should we check the value returned by `serializeType` and `serializeTypePack`?

tests/src/typeserializer.test.cpp
```diff
@@ -0,0 +1,308 @@
...
+TEST_CASE_FIXTURE(TypeSerializeFixture, "serialize_string_singleton")
+{
+    TypeId ty = arena.addType(SingletonType{StringSingleton{"hello"}});
+
+    Luau::serializeType(L, ty);
+
+    REQUIRE(lua_istable(L, -1));
+    requireStringField(L, "tag", "singleton");
+
+    lua_getfield(L, -1, "value");
```
skberkeley: why can't you use `requireStringField` here?

tests/src/typeserializer.test.cpp
```diff
@@ -0,0 +1,308 @@
...
+
+static void requireBoolField(lua_State* L, const char* field, bool expected)
+{
+    lua_getfield(L, -1, field);
+    REQUIRE(lua_isboolean(L, -1));
+    CHECK(lua_toboolean(L, -1) == expected);
+    lua_pop(L, 1);
+}
+
+// Primitive Types
```
skberkeley, line 28: aren't these tests all missing a `lua_pop`?

tests/src/typeserializer.test.cpp
```diff
@@ -0,0 +1,308 @@
...
+
+    lua_rawgeti(L, -1, 1);
+    requireStringField(L, "tag", "number");
+    lua_pop(L, 1);
+
+    lua_rawgeti(L, -1, 2);
+    requireStringField(L, "tag", "string");
+    lua_pop(L, 1);
+
+    lua_pop(L, 1); // components
```
skberkeley: if only there were some way to pop multiple elements off the stack

tests/src/typeserializer.test.cpp
```diff
@@ -0,0 +1,308 @@
...
+    lua_getfield(L, -1, "inner");
+    REQUIRE(lua_istable(L, -1));
+    requireStringField(L, "tag", "number");
+    lua_pop(L, 1);
+}
+
+TEST_CASE_FIXTURE(TypeSerializeFixture, "serialize_union_type")
+{
+    TypeId numberTy = arena.addType(PrimitiveType{PrimitiveType::Number});
+    TypeId stringTy = arena.addType(PrimitiveType{PrimitiveType::String});
```
skberkeley, line 175: do the fixtures share the arena? it might be worth throwing these into a setup function to save memory if so

skberkeley (changes requested):
It seems like these tests need `checkstacks`, since you're growing the stack in each case

tests/src/typeserializer.test.cpp
annieetang: forgot to update it after i added the helper!

tests/src/typeserializer.test.cpp
annieetang: re slack: will use `REQUIRE_EQ`!

tests/src/typeserializer.test.cpp
annieetang: 🥲👍

tests/src/typeserializer.test.cpp
annieetang: updated!

Vighnesh-V (approved):

vrn-sn (approved):

skberkeley (approved):

PR #780, opened 2026-02-03
Cleans up `TypeSerialize`

annieetang:
Some cleaning up of `TypeSerialize` after talking with @Vighnesh-V and @vrn-sn 
Changes: 
* update API of public `serializeType` and `serializeTypePack` funcs to match luau definitions & sets the table to be read only
* removing check stacks that come after internal `serialize` traverse functions bc upon further discussion and implementation of all of the traverse functions in https://github.com/luau-lang/lute/pull/778, we don't need to add the extra checks, we will just error if we reach a type or typepack that we aren't serializing
* renaming raw checkstacks to be regular `lua_checkstack` and making sure the size is just enough (each `lua_pushxxx` followed by a `lua_setfield` removes space on the top of the `lua_state` so we only need to allocate a spot for non-temporary pushes (i.e., `lua_createtable`) and a slot to do the temporary pushes)
* updating setPropertiesFields to also check its stack before pushing

### DIFF
--- a/lute/luau/src/type.cpp
+++ b/lute/luau/src/type.cpp
@@ -5,9 +5,16 @@
 #include "Luau/VisitType.h"
 
 #include "lua.h"
+#include "lualib.h"
 
 namespace Luau
 {
+
+static void checkStack(lua_State* L, int n)
+{
+    if (!lua_checkstack(L, n))
+        luaL_error(L, "stack overflow while serializing type");
+}
 
 // Serializer for runtime types, modeled after AstSerialize but for TypeId/TypePackId
 struct TypeSerialize final : public Luau::TypeVisitor
@@ -30,7 +37,7 @@ struct TypeSerialize final : public Luau::TypeVisitor
     // Helper to push a Name, Property pair into the properties table
     void pushProperty(const std::string& name, const Property& prop)
     {
-        lua_checkstack(L, 3); // 1 for key + 1 for value table + 1 for traverse/nil
+        checkStack(L, 3); // 1 for key + 1 for value table + 1 for traverse/nil
 
         // push the Name string as the key for the property
         lua_pushstring(L, name.c_str());
@@ -55,7 +62,7 @@ struct TypeSerialize final : public Luau::TypeVisitor
     // Helper to set the "properties" field for table and extern types
     void setPropertiesFields(const std::map<Name, Property>& props)
     {
-        lua_checkstack(L, 1); // 1 for properties table
+        checkStack(L, 1); // 1 for properties table
         lua_createtable(L, 0, props.size());
 
         for (const auto& [name, prop] : props)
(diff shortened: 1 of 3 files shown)


lute/luau/src/type.cpp
```diff
@@ -433,32 +422,31 @@ struct TypeSerialize final : public Luau::TypeVisitor
...
-    // Luau GenericTypePack is { head: nil, tail: generic type with ispack=true }
+    // Luau GenericTypePack is
+    // name: string?
+    // ispack: boolean
     void serialize(TypePackId tp, const GenericTypePack& gtp)
     {
-        lua_rawcheckstack(L, 2);
+        lua_checkstack(L, 2);
         lua_createtable(L, 0, 3);
+        pushTag("generic");
```
annieetang, line 403: i realize i implemented this wrong before lol

lute/luau/src/type.cpp
```diff
@@ -284,8 +279,8 @@ struct TypeSerialize final : public Luau::TypeVisitor
     // [TODO] writeindexer: { index: type, result: type } }?,
     void serialize(TypeId ty, const TableType& ttv)
     {
-        lua_rawcheckstack(L, 2);
-        lua_createtable(L, 0, 4);
+        lua_checkstack(L, 2);
+        lua_createtable(L, 0, 3);
```
skberkeley, line 267: why is different?

lute/luau/src/type.cpp
```diff
@@ -318,20 +312,19 @@ struct TypeSerialize final : public Luau::TypeVisitor
     }
 
     // Luau metatable type:
+    // table: type,
     // metatable: type?
     void serialize(TypeId ty, const MetatableType& mtv)
     {
-        lua_rawcheckstack(L, 2);
-        lua_createtable(L, 0, 2);
```
skberkeley, line 325: how did this work before lol

lute/luau/src/type.cpp
```diff
@@ -433,32 +422,31 @@ struct TypeSerialize final : public Luau::TypeVisitor
...
         if (!gtp.name.empty())
         {
             lua_pushstring(L, gtp.name.c_str());
             lua_setfield(L, -2, "name");
         }
-        lua_pushboolean(L, true); // ispack = true for generic type packs
+        else
+        {
+            lua_pushnil(L);
+            lua_setfield(L, -2, "name");
```
skberkeley: can't this go outside the `if/else`

lute/luau/src/type.cpp
annieetang: I'm only adding the tag, properties and indexer rn so it just needs those

lute/luau/src/type.cpp
annieetang: I didn't add a test for checking the specific field yet so I guess it just didn't error

lute/luau/src/type.cpp
annieetang: sg i updated the other ones too

lute/luau/src/type.cpp
Vighnesh-V: rawcheckstack also throws an os error, but checkstack will throw a runtime error that can be caught.

Vighnesh-V (approved):

lute/luau/src/type.cpp
vrn-sn: (Meaning `lua_rawcheckstack` just blindly grows the stack, so it _might_ result in an OOM that the OS then has to handle. We should try to avoid `lua_rawcheckstack` in user-facing APIs since it could lead to user-caused OOM crashes.)

PR #781, opened 2026-02-03
Ensures that the lute runtime initializes it's own event loop

Vighnesh-V:
We need to support parallelism in Lute, and to do that each spawned runtime must have it's own event loop. This is because the uv_loop_t is not safe to share between threads.


vrn-sn (approved):
Seems reasonable to me!

PR #782, opened 2026-02-03
Removes unused step from release CI

skberkeley:
Noticed there was an unused step. Also did some reformatting to be more consistent with the rest of the file

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -129,6 +129,11 @@ jobs:
         with:
           sparse-checkout: CMakeBuild/get_version.cmake
 
+      # Used for lute --version
+      - name: Set nightly version suffix
+        if: needs.check.outputs.nightly_tag
+        run: echo "LUTE_VERSION_SUFFIX=${{ needs.check.outputs.nightly_tag }}" >> "$GITHUB_ENV"
+
       - name: Download Artifact
         uses: actions/download-artifact@v4
         with:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -101,6 +101,11 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v4
 
+      # Used for lute --version
+      - name: Set nightly version suffix
+        if: needs.check.outputs.nightly_tag
+        run: echo "LUTE_VERSION_SUFFIX=${{ needs.check.outputs.nightly_tag }}" >> "$GITHUB_ENV"
+
       - name: Setup and Build Lute
         id: build_lute
         uses: ./.github/actions/setup-and-build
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -98,27 +98,23 @@ jobs:
     runs-on: ${{ matrix.os }}
 
     steps:
-    - name: Checkout code
-      uses: actions/checkout@v4
-
-    - name: Set nightly version suffix
-      if: needs.check.outputs.nightly_tag
-      run: echo "LUTE_VERSION_SUFFIX=${{ needs.check.outputs.nightly_tag }}" >> "$GITHUB_ENV"
```
vrn-sn, line 106: Haven't looked at this code in a while: are we sure this won't break `lute --version`?

.github/workflows/release.yml
```diff
@@ -98,27 +98,23 @@ jobs:
...
+          target: Lute.CLI
+          config: release
+          options: ${{ matrix.options }}
+          token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Upload Artifact
+        uses: actions/upload-artifact@v4
+        with:
+          path: ${{ steps.build_lute.outputs.exe_path }}
+          name: ${{ matrix.artifact_name }}
```
Vighnesh-V, line 122: How come we don't need this? Isn't this what uploads the release artifact for the subsequent step to publish? e.g. https://github.com/luau-lang/lute/actions/runs/21460431298/job/61811832195

skberkeley: Sorry the white space changes are obscuring things but the step I'm removing is the one Varun commented on

.github/workflows/release.yml
skberkeley: oh you're right, didn't notice get_version.cmake

vrn-sn: If removing this would break `--version`, we should add a comment above this saying that it looks like a no-op but is actually setting up the environment for the build system to set up the version.

Relevant PR if you could investigate: https://github.com/luau-lang/lute/pull/390.

vrn-sn: Oh okay, cool. If you want to still make the whitespace changes and add a comment here, I'd be happy to approve that.

skberkeley: added a clarifying comment

.github/workflows/release.yml
```diff
@@ -133,6 +129,11 @@ jobs:
         with:
           sparse-checkout: CMakeBuild/get_version.cmake
 
+      # Used for lute --version
```
vrn-sn: I think it needs to be passed at build-time?

vrn-sn: i.e. we should preserve the original order of steps

skberkeley: oh mb, i didn't mean to move it

vrn-sn (approved):

PR #784, opened 2026-02-05
Boilerplate for Lute Lint Configuration System + Support for Ignores

wmccrthy:
This PR adds boilerplate for a `lute lint` configuration system, based on the rough spec [here](https://roblox.atlassian.net/wiki/spaces/~7120203c8951ed160f49bdb87cbf260629a70f/pages/4328751232/Spec).
This initial PR involves:
- adding a `config` arg to the CLI
- loading `config` at the start of lute lint execution flow, falling back to an empty table if no config file is given or present in the current working dir
- support for the `ignores` field, which configures global file ignores using lua patterns and/or direct paths.


lute/cli/commands/lint/init.luau
```diff
@@ -353,6 +358,18 @@ local function lintPaths(
 		while curr ~= nil do
 			local inputFilePath = pathLib.parse(curr)
 
+			local isIgnored = false
+			for _, ignore in ignores do -- ignore matches
+				if string.match(pathLib.format(curr), pathLib.format(ignore)) then
```
wmccrthy: this will only support lua patterns right now; I wonder if it's worth extracting some of the logic from `lib/ignore` so we support gitignore-style syntax instead

lute/cli/commands/lint/init.luau
```diff
@@ -12,6 +12,7 @@ local tableext = require("@std/tableext")
 local triviaUtils = require("@std/syntax/utils/trivia")
 local types = require("@self/types")
 local visitor = require("@std/syntax/visitor")
+local confgiUtils = require("@self/configutils")
```
```suggestion
local configUtils = require("@self/configutils")
```

lute/cli/commands/lint/types.luau
```diff
@@ -23,4 +23,27 @@ export type LintRule = {
...
+
+export type GlobalsConfig = { [string]: unknown }
+
+export type LintConfig = {
+	ignores: { path.pathlike }?, -- globs/paths to ignore as we walk lintee files
+	globals: GlobalsConfig?, -- globals to pass down to all rules via context
+	rules: {
+		paths: { path.pathlike }?, -- paths to local rules to use
+		[string]: RuleConfig,
+	}?,
```
skberkeley: nit: can we leave these additional fields/types for the followup PRs?

lute/cli/commands/lint/init.luau
```diff
@@ -402,6 +421,22 @@ local function main(...: string)
...
 	end
 
+	local lintConfig: types.LintConfig = {}
+	local configPath = args:get("config")
+	if configPath == nil then
+		-- search in calling dir
+		local cwd = process.cwd()
+		configPath = pathLib.join(cwd, ".config.luau")
+	end
+	if fs.exists(configPath :: pathLib.pathlike) then
```
skberkeley: is this cast needed? if there's type error do you mind sending it?

lute/cli/commands/lint/init.luau
```diff
@@ -402,6 +421,22 @@ local function main(...: string)
...
+		local success, loadedConfig = pcall(luau.loadbypath, configPath :: pathLib.pathlike, nil)
```
```suggestion
		local success, loadedConfig = pcall(luau.loadbypath, configPath :: pathLib.pathlike)
```

lute/cli/commands/lint/configutils.luau
```diff
@@ -0,0 +1,31 @@
...
+	if candidate.rules then
+		assert(typeof(candidate.rules) == "table", "config.rules must be a table")
+		if candidate.rules.paths then
+			assert(typeof(candidate.rules.paths) == "table", "config.rules.paths must be a table of paths")
+		end
+	end
+end
+
+local function extractConfig(candidate: any): types.LintConfig
+	local luteLintConfig = candidate.lute and candidate.lute.lint
```
skberkeley:
```suggestion
	local luteLintConfig = if candidate.lute == nil then nil else candidate.lute.lint
```

lute/cli/commands/lint/configutils.luau
```diff
@@ -0,0 +1,31 @@
...
+	end
+	if candidate.globals then
+		assert(typeof(candidate.globals) == "table", "config.globals must be a table of [string] : unknown")
+	end
+	if candidate.rules then
+		assert(typeof(candidate.rules) == "table", "config.rules must be a table")
+		if candidate.rules.paths then
+			assert(typeof(candidate.rules.paths) == "table", "config.rules.paths must be a table of paths")
+		end
+	end
```
skberkeley: ditto, can we bump these to the followups?

lute/cli/commands/lint/configutils.luau
```diff
@@ -0,0 +1,31 @@
+local types = require("./types")
+
+-- we expect .config.luau file structured s.t top level table has: table.lute.lint field
+local function assertValidConfig(candidate: any)
+	if candidate.ignores then
+		assert(typeof(candidate.ignores) == "table", "config.ignores must be a table of paths")
```
skberkeley:
```suggestion
		assert(typeof(candidate.ignores) == "table", "config.lute.lint.ignores must be an array of filepath globs")
```

lute/cli/commands/lint/init.luau
```diff
@@ -337,12 +340,14 @@ end
...
 	lintRules: { types.LintRule },
-	autofixEnabled: boolean
+	autofixEnabled: boolean,
+	lintConfig: types.LintConfig
 ): { [pathLib.path]: { types.LintViolation } }
 	if VERBOSE then
 		print("Filtering input files by gitignore")
 	end
 	local inputFiles = files.getSourceFiles(inputFilePaths, VERBOSE)
+	local ignores = lintConfig.ignores or {}
```
```suggestion
	local ignores = if lintConfig.ignores ~= nil then lintConfig.ignores else {}
```

lute/cli/commands/lint/init.luau
```diff
@@ -352,6 +357,19 @@ local function lintPaths(
...
+			for _, ignore in ignores do -- ignore matches
+				if inputFilePathString:match(pathLib.format(ignore)) then
+					isIgnored = true
+					break
+				end
+			end
+			if isIgnored then
+				curr = walker()
+				continue
+			end
```
skberkeley: i don't think this is the best solution; it means that people are using Luau patterns instead of globs and somewhat complicates the logic here. i think we should instead try to add logic to either extend `files.getSourceFiles` to take an optional array of ignores, or pull some logic out of `@cli/ignore` to have a simpler version of `IgnoreResolver` which is initialized with a set of globs rather than doing .gitignore lookups for each filepath it checks

skberkeley: this might mean doing a separate PR which we merge before this one

skberkeley (changes requested):

lute/cli/commands/lint/init.luau
wmccrthy: agreed, i think a simpler `IgnoreResolver` will fit in best here.

lute/cli/commands/lint/init.luau
wmccrthy: Included nil as per the following type err when only passing 2 args: `TypeError: Argument count mismatch. Function expects 3 arguments, but only 2 are specified`

lute/cli/commands/lint/init.luau
wmccrthy: Yea, seeing:
```
TypeError: Expected this to be
	'path | path | pathdata | pathdata | string | string'
but got
	'(path | path | string)?'; 
the 2nd component of the union is `nil`, which is not a subtype of `path | path | pathdata | pathdata | string | string`
```

lute/cli/commands/lint/types.luau
wmccrthy: Sure, are you cool with having them commented out for now? Or would your prefer fully omitting

skberkeley: yeah that's fine given that we have the followup work planned soon

lute/cli/commands/lint/init.luau
skberkeley: hmm ok, can you leave the cast out and leave the error in? that seems like an erroneous type error

skberkeley: you'll have to run `lute tools/check.luau <path to lute> --update` and push the updated faillist to get through ci

lute/cli/commands/lint/init.luau
skberkeley: gotcha, i would similarly leave out the cast/nil and leave the type error and update the faillist, since this should work

lute/cli/commands/lint/init.luau
wmccrthy: opened https://github.com/luau-lang/lute/pull/796 so we can re-use logic in this PR

lute/cli/commands/lint/init.luau
```diff
@@ -401,6 +418,22 @@ local function main(...: string)
...
 
+	local lintConfig: types.LintConfig = {}
+	local configPath = args:get("config")
+	if configPath == nil then
+		-- search in calling dir
+		local cwd = process.cwd()
+		configPath = pathLib.join(cwd, ".config.luau")
+	end
+	if fs.exists(configPath) then
+		local success, loadedConfig = pcall(luau.loadbypath, configPath :: pathLib.pathlike)
```
```suggestion
		local success, loadedConfig = pcall(luau.loadbypath, configPath)
```
It's generally more helpful for us to report type errors rather than suppress them, since they demonstrate bugs in the new solver

lute/cli/commands/lint/configutils.luau
```diff
@@ -0,0 +1,22 @@
...
+
+-- we expect .config.luau file structured s.t top level table has: table.lute.lint field
+local function assertValidConfig(candidate: any)
+	if candidate.ignores then
+		assert(typeof(candidate.ignores) == "table", "config.lute.lint.ignores must be an array of filepath globs")
+	end
+end
+
+local function extractConfig(candidate: any): types.LintConfig
+	local luteLintConfig = if candidate.lute ~= nil then (candidate.lute :: any).lint else nil
```
skberkeley:
```suggestion
	local luteLintConfig = if candidate.lute ~= nil then candidate.lute.lint else nil
```
I think `candidate: any` should already be suppressing an error, if it doesn't it's probably a bug

lute/cli/commands/lint/configutils.luau
```diff
@@ -0,0 +1,22 @@
+local types = require("./types")
+
+-- we expect .config.luau file structured s.t top level table has: table.lute.lint field
+local function assertValidConfig(candidate: any)
+	if candidate.ignores then
+		assert(typeof(candidate.ignores) == "table", "config.lute.lint.ignores must be an array of filepath globs")
```
skberkeley, line 13: just a thought: do we want to do some stronger validation here, eg check that all keys are numbers and values are strings?

tests/cli/lint.test.luau
```diff
@@ -2614,4 +2614,107 @@ end
...
+		-- fs.writestringtofile(path.format(configFile), invalidRulesConfig)
+		-- result = process.run({ lutePath, "lint", "-c", path.format(configFile), path.format(configFile) })
+		-- assert.eq(result.exitcode, 1)
+		-- assert.strcontains(result.stderr, "config.rules must be a table")
+	end)
+
+	suite:case("lute lint accounts for configured ignores (config passed as arg)", function(assert)
+		local ignoredFile = path.join(tmpDir, "ignored.luau")
+		local violatingContents = [[
+			a = b
```
skberkeley, line 2674: can you extend to check a file that doesn't get ignored?

tests/cli/lint.test.luau
```diff
@@ -2614,4 +2614,107 @@ end
...
+		assert.strnotcontains(result.stdout, "divide_by_zero")
+	end)
+
+	suite:case("lute lint accounts for configured ignores (config auto-detected)", function(assert)
+		local ignoredFile = path.join(tmpDir, "ignored.luau")
+		local violatingContents = [[
+			a = b
+			b = a
+			c = 1 / 0
+		]]
```
skberkeley, line 2713: it's probably worth creating a subdir inside `tmpDir` since this test could become flaky if the runner has other `.config.luau` files in its tmpDir already

lute/cli/commands/lint/configutils.luau
wmccrthy: Yea seems like a bug, I haven't seen this type err before: `TypeError: Type '~nil' does not have key 'lint'Luau[1002](https://luau-lang.org/typecheck)`

lute/cli/commands/lint/configutils.luau
wmccrthy: that seems appropriate

tests/cli/lint.test.luau
```diff
@@ -2688,6 +2691,11 @@ end
 		assert.eq(result.exitcode, 0) -- no errors
 		assert.strnotcontains(result.stdout, "almost_swapped")
 		assert.strnotcontains(result.stdout, "divide_by_zero")
+
+		result = process.run({ lutePath, "lint", "-c", path.format(configFile), path.format(notIgnoredFile) })
+		assert.eq(result.exitcode, 1) -- violations caught
+		assert.strcontains(result.stdout, "[almost_swapped]")
+		assert.strcontains(result.stdout, "[divide_by_zero]")
```
skberkeley: i'd like to test that the config ignoring works when `lute lint` is called in a directory rather on individual files, so can you do the same thing as below where you put all the files in a directory? would also be great if you could test config ignore's interaction with `.gitignore` files (i assume the desired behavior is to do the union of the ignores)

tests/cli/lint.test.luau
skberkeley: can you put this in the subdir too?

skberkeley (changes requested):

tests/cli/lint.test.luau
wmccrthy: Good call, I'll add. There's no new logic to specifically account for `.gitignore`, but since lute lint gets all the paths from `files.getSourceFiles`, they should already be accounted for.

tests/cli/lint.test.luau
```diff
@@ -2614,4 +2614,168 @@ end
...
+		assert.strnotcontains(result.stdout, "almost_swapped")
+		assert.strnotcontains(result.stdout, "divide_by_zero")
+
+		fs.removedirectory(testSubDir, {
+			recursive = true,
+		})
+	end)
 end)
+
+test.run()
```
skberkeley:
```suggestion

```

skberkeley (approved):

tests/cli/lint.test.luau
wmccrthy: whoops

PR #785, opened 2026-02-05
Fixes std/path metatable type errors

annieetang:
Fixes a type error in path lib's `posix` and `win32` files (came up when I was trying to serialize the type of these modules), thx for helping fix @skanosue !


skberkeley (approved):
i assume you'll need to update lute check but looks good!

PR #786, opened 2026-02-06
Adds a lint rule for unused locals

skberkeley:
Adds a new lint rule to flag unused locals! Notable differences from the builtin Luau linter is that loop variables and function arguments will now be warned on if unused (you can still prefix with `_` to silence).

Some followup work I'd like to do:
- Classify builtin APIs by whether they actually read their arguments, like Selene does. For example, `table.insert` shouldn't actually count as having "used" its first argument.
- Use @wmccrthy's work on lint configs to add configuration options to the rule
- Add a lint rule or extend this one to flag unused globals


lute/cli/commands/lint/rules/unused_variable.luau
```diff
@@ -0,0 +1,202 @@
...
+local visitorLib = require("@std/syntax/visitor")
+
+local name = "unused_variable"
+local message = "Unused variable detected. Prefix this variable with '_' to indicate that it's intentionally unused."
+
+local function unusedLocals(block: syntax.AstStatBlock): { syntax.AstLocal }
+	local visitor: visitorLib.Visitor & {
+		unusedLocals: { syntax.AstLocal },
+		scopes: { { [syntax.AstLocal]: boolean } },
+		ignoreReferences: { [syntax.AstLocal]: boolean },
```
aatxe:
```suggestion
		ignoreReferences: { [syntax.AstLocal]: true },
```

Arguably could just be this, right? You're using it as a set.

aatxe (approved):

lute/cli/commands/lint/rules/unused_variable.luau
skberkeley: Huh yeah, that's what I intended but I must've forgotten to change

lute/std/libs/syntax/visitor.luau
```diff
@@ -939,7 +948,9 @@ local function create(visit: ((types.AstNode) -> boolean)?): Visitor
 	if visit then
 		return {
 			visitStatBlock = visit,
-			visitStatBlockEnd = visit,
+			visitStatBlockEnd = function(s)
+				visit(s)
+			end,
```
Vighnesh-V, line 953: Why do these have to be wrapped in function?

Vighnesh-V (approved):
Left one comment. More generally, we should offer a way to configure the linter so we can just specify directories to exclude (e.g. examples).

skberkeley:
> Left one comment. More generally, we should offer a way to configure the linter so we can just specify directories to exclude (e.g. examples).

Yep, @wmccrthy is working on linter configurations, including ignore options

PR #787, opened 2026-02-06
Adds handling cyclic type references in `TypeSerializer` + tests

annieetang:
More follow up on `TypeSerializer` improvements. This PR creates a reference table in `TypeSerializer` to handle cyclic & shared type references, since Luau types can contain cycles (e.g., `type Node = { value: number, next: Node }` )

When we're serializing the types to luau tables, we need to detect and handle these cycles to avoid infinite recursion or missing serializations. To do this, I think we can maintain a mapping of each `TypeId`/`TypePackId` to its serialized table. So when we encounter a type we've already serialized, detected by the `cycle` functions provided by the `Luau::TypeVisitor`, we retrieve and reuse the existing table instead of creating a new one, preserving the cyclic structure in the serialization.

Also adds test case for this cycle, as well as `FunctionType`, and `TableType` tests

### DIFF
--- a/lute/luau/src/type.cpp
+++ b/lute/luau/src/type.cpp
@@ -1,5 +1,6 @@
 #include "lute/type.h"
 
+#include "Luau/ToString.h"
 #include "Luau/Type.h"
 #include "Luau/TypeFwd.h"
 #include "Luau/VisitType.h"
@@ -20,11 +21,63 @@ static void checkStack(lua_State* L, int n)
 struct TypeSerialize final : public Luau::TypeVisitor
 {
     lua_State* L;
+    int refsTableIndex; // Reference table for handling cyclic & shared type references. Maps TypeId/TypePackId (as lightuserdata) to the corresponding serialized table on the stack.
 
     explicit TypeSerialize(lua_State* L)
         : Luau::TypeVisitor{"TypeSerialize", /*skipBoundTypes=*/false}
         , L(L)
     {
+        // Create the refs table and store a reference to it. This table will be used to track already serialized types to handle cycles and shared references.
+        lua_createtable(L, 0, 0);
+        refsTableIndex = lua_gettop(L);
+    }
+
+    ~TypeSerialize()
+    {
+        // Pop the refs table from the stack when the serializer is destroyed
+        if (refsTableIndex > 0 && lua_gettop(L) >= refsTableIndex)
+            lua_remove(L, refsTableIndex);
+    }
+
+    void cycle(TypeId ty) override
+    {
+        lua_pushlightuserdata(L, (void*)ty);
+        lua_gettable(L, refsTableIndex);
+
+        if (lua_isnil(L, -1))
+        {
+            luaL_error(L, "TypeSerialize: cycle detected while serializing type, but TypeId %s was not found in refs table", Luau::toString(ty).data());
+        }
+    }
+
+    void cycle(TypePackId tp) override
+    {
+        lua_pushlightuserdata(L, (void*)tp);
+        lua_gettable(L, refsTableIndex);
+
+        if (lua_isnil(L, -1))
+        {
+            luaL_error(L, "TypeSerialize: cycle detected while serializing type pack, but TypePackId %s was not found in refs table", Luau::toString(tp).data());
+        }
+    }
+
+    // Register this type's serialized table in the refs table, using the TypeId as the key.
+    // Called after creating the serialized table for a type, but before traversing any child types.
+    void registerType(TypeId ty)
+    {
+        checkStack(L, 3);
+        lua_pushlightuserdata(L, (void*)ty); // TypeId pointer as key
+        lua_pushvalue(L, -2); // Copy of the serialized type table as value
+        lua_rawset(L, refsTableIndex); // refs[ty] = serializedTable
+    }
+
+    // Similar to registerType but for TypePackId
+    void registerTypePack(TypePackId tp)
+    {
+        checkStack(L, 3);
+        lua_pushlightuserdata(L, (void*)tp);
+        lua_pushvalue(L, -2);
+        lua_rawset(L, refsTableIndex);
     }
 
     // Helper to push the "tag" field that every Luau type has
@@ -77,6 +130,7 @@ struct TypeSerialize final : public Luau::TypeVisitor
     {
         checkStack(L, 2); // 1 for table + 1 for space to push/set remaining fields
         lua_createtable(L, 0, 1);
+        registerType(ty);
 
         const char* tag = nullptr;
         switch (ptv.type)
@@ -110,6 +164,7 @@ struct TypeSerialize final : public Luau::TypeVisitor
     {
         checkStack(L, 2);
         lua_createtable(L, 0, 1);
+        registerType(ty);
 
         pushTag("any");
     }
@@ -118,6 +173,7 @@ struct TypeSerialize final : public Luau::TypeVisitor
     {
         checkStack(L, 2);
         lua_createtable(L, 0, 1);
+        registerType(ty);
 
         pushTag("unknown");
     }
@@ -126,6 +182,7 @@ struct TypeSerialize final : public Luau::TypeVisitor
     {
         checkStack(L, 2);
         lua_createtable(L, 0, 1);
+        registerType(ty);
 
         pushTag("never");
     }
@@ -136,6 +193,7 @@ struct TypeSerialize final : public Luau::TypeVisitor
     {
         checkStack(L, 2);
         lua_createtable(L, 0, 2);
+        registerType(ty);
 
         pushTag("singleton");
 
@@ -154,6 +212,7 @@ struct TypeSerialize final : public Luau::TypeVisitor
     {
         checkStack(L, 2);
         lua_createtable(L, 0, 2);
+        registerType(ty);
 
         pushTag("negation");
 
@@ -167,6 +226,7 @@ struct TypeSerialize final : public Luau::TypeVisitor
     {
         checkStack(L, 3); // 1 for table + 1 for components table + 1 for traverse
         lua_createtable(L, 0, 2);
+        registerType(ty);
 
         pushTag("union");
 
@@ -185,6 +245,7 @@ struct TypeSerialize final : public Luau::TypeVisitor
     {
         checkStack(L, 3); // 1 for table + 1 for components table + 1 for traverse
         lua_createtable(L, 0, 2);
+        registerType(ty);
 
         pushTag("intersection");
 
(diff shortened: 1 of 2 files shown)


lute/luau/src/type.cpp
```diff
@@ -284,6 +348,8 @@ struct TypeSerialize final : public Luau::TypeVisitor
         {
             lua_createtable(L, 0, 3);
             traverse(ttv.indexer->indexType);
+            lua_setfield(L, -2, "index");
```
annieetang, line 355: this fixes a bug I found testing where I forgot to set the field after pushing the index type, leading to a stackoverflow in massive types like `definitions/luau.luau`

tests/src/typeserializer.test.cpp
```diff
@@ -233,7 +233,172 @@ TEST_CASE_FIXTURE(TypeSerializeFixture, "serialize_generic_type")
...
+    FunctionType ftv{{genericTy}, {}, argTypes, retTypes};
+    TypeId ty = arena.addType(ftv);
+
+    lua_checkstack(L, 3);
+    REQUIRE_EQ(Luau::serializeType(L, ty), 1);
+
+    REQUIRE(lua_istable(L, -1));
+    requireStringField(L, "tag", "function");
+
+    lua_getfield(L, -1, "parameters");
```
annieetang, line 288: we can def do some improvements to writing this test suite later since this is getting pretty gnarly but I feel like that can be a follow up if necesary

lute/luau/src/type.cpp
```diff
@@ -20,11 +21,63 @@ static void checkStack(lua_State* L, int n)
...
+        }
+    }
+
+    // Register this type's serialized table in the refs table, using the TypeId as the key.
+    // Called after creating the serialized table for a type, but before traversing any child types.
+    void registerType(TypeId ty)
+    {
+        checkStack(L, 3);
+        lua_pushlightuserdata(L, (void*)ty); // TypeId pointer as key
+        lua_pushvalue(L, -2); // Copy of the serialized type table as value
```
skberkeley: does this push a value copy of the serialized table or a reference?

lute/luau/src/type.cpp
```diff
@@ -20,11 +21,63 @@ static void checkStack(lua_State* L, int n)
...
+    }
+
+    ~TypeSerialize()
+    {
+        // Pop the refs table from the stack when the serializer is destroyed
+        if (refsTableIndex > 0 && lua_gettop(L) >= refsTableIndex)
+            lua_remove(L, refsTableIndex);
+    }
+
+    void cycle(TypeId ty) override
```
skberkeley, line 44: i might be missing something, but when does `cycle` get called?

lute/luau/src/type.cpp
```diff
@@ -20,11 +21,63 @@ static void checkStack(lua_State* L, int n)
...
+        lua_createtable(L, 0, 0);
+        refsTableIndex = lua_gettop(L);
+    }
+
+    ~TypeSerialize()
+    {
+        // Pop the refs table from the stack when the serializer is destroyed
+        if (refsTableIndex > 0 && lua_gettop(L) >= refsTableIndex)
+            lua_remove(L, refsTableIndex);
+    }
```
skberkeley, line 42: should we error if this condition doesn't hold? i don't see how `refsTableIndex` gets mutated

lute/luau/src/type.cpp
annieetang: called in the `traverse()` functions (the implementation is in `Luau::TypeVisitor`

lute/luau/src/type.cpp
annieetang: ~~oh `lua_pushvalue` pushes a copy~~

sorry correction we talked abt it on slack - lua_pushvalue(L, -2) makes a copy of the table reference, not a copy of the table so it's doing the equivalent of `local b = a` and `b` and `a` point to the same thing.

also added another test case to check this (checks `Node.next.next....`)

tests/src/typeserializer.test.cpp
```diff
@@ -233,7 +233,172 @@ TEST_CASE_FIXTURE(TypeSerializeFixture, "serialize_generic_type")
...
+{
+    GenericType gtp;
+    gtp.name = "T";
+    TypeId genericTy = arena.addType(gtp);
+    TypeId numberTy = arena.addType(PrimitiveType{PrimitiveType::Number});
+
+    TypePackId argTypes = arena.addTypePack(TypePack{{numberTy}, std::nullopt});
+    TypePackId retTypes = arena.addTypePack(TypePack{{}, std::nullopt});
+
+    FunctionType ftv{{genericTy}, {}, argTypes, retTypes};
```
skberkeley: nit: can you comment the stringification of the type you're testing?

lute/luau/src/type.cpp
```diff
@@ -20,11 +21,65 @@ static void checkStack(lua_State* L, int n)
...
+        // Pop the refs table from the stack when the serializer is destroyed
+        if (refsTableIndex > 0 && lua_gettop(L) >= refsTableIndex)
+            lua_remove(L, refsTableIndex);
+        else
+            luaL_error(L, "TypeSerialize: reference table index corrupted during serialization");
+    }
+
+    void cycle(TypeId ty) override
+    {
+        lua_pushlightuserdata(L, (void*)ty);
```
skberkeley, line 47: we should probably checkstack here and in the other cycle right?

lute/luau/src/type.cpp
```diff
@@ -20,11 +21,65 @@ static void checkStack(lua_State* L, int n)
...
+        }
+    }
+
+    // Register this type's serialized table in the refs table, using the TypeId as the key.
+    // Called after creating the serialized table for a type, but before traversing any child types.
+    void registerType(TypeId ty)
+    {
+        checkStack(L, 3);
+        lua_pushlightuserdata(L, (void*)ty); // TypeId pointer as key
+        lua_pushvalue(L, -2); // Copy of the serialized type table as value
```
skberkeley:
```suggestion
        lua_pushvalue(L, -2); // Reference to the serialized type table as value
```

lute/luau/src/type.cpp
```diff
@@ -383,6 +454,7 @@ struct TypeSerialize final : public Luau::TypeVisitor
     {
         checkStack(L, 2);
         lua_createtable(L, 0, 2);
+        registerTypePack(tp);
```
skberkeley, line 459: do variadics and normal type packs not need tags?

tests/src/typeserializer.test.cpp
```diff
@@ -233,7 +263,239 @@ TEST_CASE_FIXTURE(TypeSerializeFixture, "serialize_generic_type")
...
+    TypeId ty = arena.addType(ftv);
+
+    lua_checkstack(L, 3);
+
+    // { tag: "function", parameters: { head: { { tag: "number" } }, tail: nil }, returns: { head: nil, tail: nil }, generics: { { tag: "generic", name: "T", ispack: false } }, genericpacks: {} }
```
skberkeley, line 282:
```suggestion
    // <T>(number) -> ()
    // { tag: "function", parameters: { head: { { tag: "number" } }, tail: nil }, returns: { head: nil, tail: nil }, generics: { { tag: "generic", name: "T", ispack: false } }, genericpacks: {} }
```

tests/src/typeserializer.test.cpp
```diff
@@ -233,7 +263,239 @@ TEST_CASE_FIXTURE(TypeSerializeFixture, "serialize_generic_type")
...
+    REQUIRE(lua_istable(L, -1));
+    lua_rawgeti(L, -1, 1);
+    requireStringField(L, "tag", "number");
+    lua_pop(L, 3); // head, parameters, function
+
+    lua_getfield(L, -1, "returns");
+    REQUIRE(lua_istable(L, -1));
+    lua_getfield(L, -1, "head");
+    REQUIRE(lua_isnil(L, -1)); // no return types
+    lua_pop(L, 2); // returns, function
```
skberkeley: no tail checks?

tests/src/typeserializer.test.cpp
```diff
@@ -233,7 +263,239 @@ TEST_CASE_FIXTURE(TypeSerializeFixture, "serialize_generic_type")
...
+
+    lua_getfield(L, -1, "generics");
+    REQUIRE(lua_istable(L, -1));
+    lua_rawgeti(L, -1, 1);
+    requireStringField(L, "tag", "generic");
+    requireStringField(L, "name", "T");
+    lua_pop(L, 2); // generics, function
```
skberkeley, line 316: no `ispack` check?

tests/src/typeserializer.test.cpp
```diff
@@ -233,7 +263,239 @@ TEST_CASE_FIXTURE(TypeSerializeFixture, "serialize_generic_type")
...
+    Property prop;
+    prop.readTy = numberTy;
+    prop.writeTy = stringTy;
+    ttv.props["x"] = prop;
+
+    TypeId ty = arena.addType(ttv);
+
+    lua_checkstack(L, 3);
+
+    // { tag: "table", properties: { x: { read: { tag: "number" }, write: { tag: "string" } } } }
```
skberkeley, line 339:
```suggestion
    // { read x: number, write x: string }
    // { tag: "table", properties: { x: { read: { tag: "number" }, write: { tag: "string" } } } }
```

tests/src/typeserializer.test.cpp
```diff
@@ -233,7 +263,239 @@ TEST_CASE_FIXTURE(TypeSerializeFixture, "serialize_generic_type")
...
+    // The 'read' type of 'next' should be the same table as the root
+    // We verify by checking that it has the same properties
+    requireStringField(L, "tag", "table");
+    lua_getfield(L, -1, "properties");
+    REQUIRE(lua_istable(L, -1));
+    
+    // Verify it has both 'value' and 'next' properties (proving it's the same cycle)
+    lua_getfield(L, -1, "value");
+    REQUIRE(lua_istable(L, -1));
+    lua_pop(L, 1);
```
skberkeley: if the values on the stack are references, can we instead directly compare the two values to prove it's a cycle? right now it doesn't "prove" that it's a cycle, just that it's a table with value and next props

skberkeley: we can also get rid of the next test too

annieetang: sg! I updated to use `lua_equal`

lute/luau/src/type.cpp
annieetang: talked on slack & will have this addition in a separate PR

skberkeley (approved):

PR #792, opened 2026-02-09
Bundles the batteries utilities for use by lute cli commands

Vighnesh-V:
This PR: 
- Adds a pass to the build script that bundles the `batteries` library into C++
- Enables the CLI VFS to return required modules from the `@batteries` alias, for use only by lute cli commands
- Updates the `doc` and `lint` subcommands to use the @batteries code.
- Also fixed a couple of type checking errors in luthier


tools/luthier.luau
```diff
@@ -426,7 +524,7 @@ local function generateTypeDefinitionFiles()
 
 	writeStringToFile(
 		projectRelative("lute", "cli", "commands", "setup", "generated", "definitions.luau"),
-		string.format(TYPEDEF_PATTERN, stringDictionary)
```
Vighnesh-V, line 427: These constants are used in exactly one place, and it's here.

tools/luthier.luau
```diff
@@ -4,13 +4,10 @@ local crypto = require("@lute/crypto")
 local fs = require("@lute/fs")
 local process = require("@lute/process")
 local system = require("@lute/system")
-
+local stdfs = require("@std/fs")
```
vrn-sn: this will break bootstrapped builds

Vighnesh-V: you right. The reason I originally added this was because I didn't create the intermediate directories (which will need to exist anyway). I've changed this back to use fs.mkdir instead, since it only needs to create the generated directory.

vrn-sn (approved):
Approving, but two requests:

1. Would it be possible to clean up Luthier? We now have 3 different places where we do almost identical generation/out-of-date checking logic. It seems like we could split out a submodule of Luthier to do all this and call into it for std, cli, and now batteries.
2. We could store a piece of state in an `enum class` member of `CliVfs` representing whether we're currently managing the CLI commands VFS or the batteries VFS. Then, you don't need to do the `getIdentifier()` and check prefix each time.

PR #793, opened 2026-02-09
Infra: fix bootstrap builds not using bootstrapped Lute

vrn-sn:
Found another bug in our bootstrapping logic: we bootstrap two builds and then use the Foreman-installed Lute for the third build. Aliases only live for a single shell instance, so they don't persist between steps.

We no longer invoke Foreman if bootstrapping, and we use a `$LUTE` environment variable to point to the correct executable.

Context: broken release build successfully built the first two but failed to build the third since it was calling into Foreman's binary (https://github.com/luau-lang/lute/actions/runs/21840297510/job/63022500151).

### DIFF
--- a/.github/actions/setup-and-build/action.yml
+++ b/.github/actions/setup-and-build/action.yml
@@ -30,6 +30,7 @@ runs:
   using: "composite"
   steps:
     - name: Install tools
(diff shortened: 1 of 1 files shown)


Vighnesh-V (approved):
Good catch!

PR #796, opened 2026-02-09
Separate ignore and glob parsing into isolated module

wmccrthy:
To support the implementation of `lute lint` configuration system, where we'd like to allow users to specify file ignores with an array of globs, this PR separates `ignore.luau` logic used to parse ignores and created ignoreData objects into their own module. There should be no logical change to `IgnoreResolver`, the changes simply ensure we can use `parseIgnoreContents` and `isIgnored` in isolation.


lute/cli/commands/lib/parseIgnores.luau
```diff
@@ -0,0 +1,193 @@
...
+	ignores: { Glob },
+	whitelists: { Glob },
+	next: GitignoreData?,
+}
+
+local separator = if system.win32 then "\\" else "/"
+
+--- Converts a glob pattern into a Luau string pattern for efficient matching
+--- The output pattern should be matched against a filepath that is relative to the .gitignore location
+local function parseGlob(glob: string): Glob
```
skberkeley, line 21: can you update to use the pattern used in `@std` files? ie initialize an empty table and all the function declarations add methods to the table. it makes the export surface of the file easier to reason about as you're reading through

skberkeley (approved):
just needs the one refactor, otherwise looks good!

skberkeley (approved):

PR #798, opened 2026-02-10
Fixes bootstrap builds failing due to not creating an empty batteries file

Vighnesh-V:
Added the batteries templates and updated the bootstrap script.


tools/templates/batteries_header.h
```diff
@@ -0,0 +1,7 @@
+// This file is auto-generated by luthier.luau. Do not edit.
+#pragma once
+
+#include <string_view>
+#include <utility>
+
+extern const std::pair<std::string_view, std::string_view> lutebatteries[13];
```
vrn-sn: Should be size 1, I believe.

vrn-sn (approved):
Just one small thing, feel free to merge after testing locally!

tools/templates/batteries_header.h
Vighnesh-V: good catch, I've updated this!

PR #799, opened 2026-02-10
Updates serialization of TypePack, VariadicTypePack, GenericTypePack

annieetang:
And adds tests for these!
Adding `tag` for a pack, variadic pack, and generic pack, as well as updating the luau definition to include specific fields for each, like we have for `Type`. And various fixes to the serialization after adding tests and realizing that some of the initial implementation wasn't exactly correct

### DIFF
--- a/definitions/luau.luau
+++ b/definitions/luau.luau
@@ -724,8 +724,8 @@ export type Type = {
 	metatable: Type?,
 
 	-- for function type
-	parameters: { head: { Type }?, tail: Type },
-	returns: { head: { Type }?, tail: Type },
+	parameters: TypePack,
+	returns: TypePack,
 	generics: { Type },
 	genericpacks: { TypePack },
 
@@ -739,8 +739,17 @@ export type Type = {
 }
 
 export type TypePack = {
+	tag: "typepack" | "variadic" | "generic",
 	head: { Type }?,
-	tail: Type?,
+	tail: TypePack?,
+
+	-- for variadic type packs
+	type: Type?,
+	hidden: boolean?,
+
+	-- for generic type packs
+	name: string?,
+	ispack: boolean,
 }
 
 function luau.typeofmodule(modulepath: string): TypePack?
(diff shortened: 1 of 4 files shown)


definitions/luau.luau
```diff
@@ -739,8 +739,17 @@ export type Type = {
...
 	head: { Type }?,
-	tail: Type?,
+	tail: TypePack?,
+
+	-- for variadic type packs
+	type: Type?,
+	hidden: boolean?,
+
+	-- for generic type packs
+	name: string?,
```
skberkeley: why are these optional? the various fields aren't optional on the regular Type definition

lute/luau/src/type.cpp
```diff
@@ -420,13 +420,15 @@ struct TypeSerialize final : public Luau::TypeVisitor
         lua_setfield(L, -2, "metatable"); 
     }
 
-    // Luau TypePack is { head: {type}?, tail: type? }
+    // Luau TypePack is { head: {type}?, tail: typepack? }
```
skberkeley: nit: why is this formatted inconsistently from the other doc comments?

tests/src/typeserializer.test.cpp
```diff
@@ -437,7 +437,96 @@ TEST_CASE_FIXTURE(TypeSerializeFixture, "serialize_cyclic_table_type")
...
+}
+
+TEST_CASE_FIXTURE(TypeSerializeFixture, "serialize_generic_type_pack")
+{
+    GenericTypePack gtp;
+    gtp.name = "T";
+
+    TypePackId tp = arena.addTypePack(gtp);
+
+    lua_checkstack(L, 2);
```
skberkeley:
```suggestion
    lua_checkstack(L, 1);
```

definitions/luau.luau
annieetang: updated ty, as well as genericType's name

skberkeley (approved):

Vighnesh-V (approved):

vrn-sn (approved):

Issue #621, opened 2025-11-27
AstExprInstantiate requires serialization logic in lute/luau


checkraisefold:
Given the merging of https://github.com/luau-lang/luau/pull/1980 and inclusion in Luau 0.701, AstExprInstantiate (the turbofish) will actually start to see use. There's currently no serialization logic in lute's luau parser for it.

checkraisefold:
blocked by https://github.com/luau-lang/luau/pull/2115

checkraisefold:
> blocked by [luau-lang/luau#2115](https://github.com/luau-lang/luau/pull/2115)

704 fixes the visitor, no longer an upstream block

skberkeley:
Blocked due to a visitor bug in `AstExprInstantiate`

aatxe:
@skberkeley this is fixed in the latest release, right? Or is this not ready until next week?

skberkeley:
Yeah fixed, I'll resume on it

PR #801, opened 2026-02-11
Extends AST serialization to support AstExprInstantiate

skberkeley:
Addresses #621


lute/luau/src/luau.cpp
```diff
@@ -920,6 +920,34 @@ struct AstSerialize : public Luau::AstVisitor
...
+        serializeNodePreamble(node, "instantiate", "expr");
+
+        node->expr->visit(this);
+        lua_setfield(L, -2, "expr");
+
+        serializeToken(cstNode.leftArrow1Position, "<");
+        lua_setfield(L, -2, "leftarrow1");
+
+        serializeToken(cstNode.leftArrow2Position, "<");
+        lua_setfield(L, -2, "leftarrow2");
```
aatxe, line 939: Should we serialize these as two separate tokens? Or one combined `<<` token?

lute/luau/src/luau.cpp
```diff
@@ -920,6 +920,34 @@ struct AstSerialize : public Luau::AstVisitor
...
+
+        serializePunctuated(node->typeArguments, cstNode.commaPositions, ",");
+        lua_setfield(L, -2, "typearguments");
+
+        serializeToken(cstNode.rightArrow1Position, ">");
+        lua_setfield(L, -2, "rightarrow1");
+
+        serializeToken(cstNode.rightArrow2Position, ">");
+        lua_setfield(L, -2, "rightarrow2");
```
aatxe, line 948: Ditto, should serialize this as two separate tokens? Or one combined `>>` token?

tests/parserExamples/explicit-generic-instantiation-1.luau
```diff
@@ -0,0 +1,5 @@
+function foo<A, B...>(_x: A, ...: B...)
+	return nil :: any
+end
+
+local _x = foo<<string, number...>>("hello", 1, 2, 3)
```
aatxe: trailing newline

```suggestion
local _x = foo<<string, number...>>("hello", 1, 2, 3)

```

tests/parserExamples/.styluaignore
```diff
@@ -0,0 +1 @@
+explicit-generic-instantiation-1.luau
```
aatxe: trailing newline
```suggestion
explicit-generic-instantiation-1.luau

```

aatxe (approved):
Approving because I'll accept this, but I want to at least ask you to consider if we should instead do combined tokens for the applicator syntax.

lute/luau/src/luau.cpp
skberkeley: Talked offline, the C++ parser implementation has these as separate tokens, which means we need to serialize them as separate tokens on the Lute side too to retain whitespace/positioning info

PR #806, opened 2026-02-11
Fixes a few typechecking errors in lute

Vighnesh-V:
There is a type error in the std test library caused by the type of the assertions table not being in the `asserts.luau` file anymore. This caused a number of downstream type errors in test scripts, because they were adding annotation for an assertions table that didn't exist in that file.
Edit: also fixed an fs.luau tc error.


skberkeley (approved):
sick

PR #808, opened 2026-02-11
Lint Config: Support Passing `Context` thru to Rules

wmccrthy:
This PR implements two next steps for lute lint's configuration system: rule context and rule-specific ignores.

`Context` is a table of `globals` and rule specific `options`, as configured by the user. For example, given a configuration of:
```
{
...,
	globals = {
		myGlobal = 1
	},
	rules = {
		myRule = {
			options = {
				myOption = 1
			}
		}
	}
}
```
we will the following context to each call of `myRule`:
```
{
	globals = {
		myGlobal = 1
	},
	options = {
		myOption = 1
	}
}
```

This enables rule authors to write logic w/ configurability in mind, and extend their analyses to use-cases beyond the default.

Rule-specific ignores extend the global-lint ignores (implemented in this PR), but are evaluated per-rule. This enables more granular ignore behavior, where we can effectively turn off rules for certain files / paths.


lute/cli/commands/lint/init.luau
```diff
@@ -336,14 +346,14 @@ local function lintFile(
...
-	return lintString(newContent, lintRules, inputFilePath)
+	return lintString(newContent, lintRules, inputFilePath, lintConfig)
 end
 
 local function lintPaths(
 	inputFilePaths: { string },
 	lintRules: { types.LintRule },
 	autofixEnabled: boolean,
-	_lintConfig: types.LintConfig, -- will be used in follow-up PRs
+	lintConfig: types.LintConfig, -- will be used in follow-up PRs
```
```suggestion
	lintConfig: types.LintConfig,
```

lute/cli/commands/lint/init.luau
```diff
@@ -227,8 +228,16 @@ local function lintString(
...
 	local directiveCache = {}
+	local rulesConfig = if config.rules ~= nil then config.rules else {}
 	for _, rule in lintRules do
-		local success, err = pcall(rule.lint, ast, inputFilePath)
+		local ruleConfig: types.RuleConfig = if rulesConfig[rule.name] ~= nil then rulesConfig[rule.name] else {}
+
+		local context: types.RuleContext = {
+			globals = if config.globals ~= nil then config.globals else {},
+			options = if ruleConfig.options ~= nil then ruleConfig.options else {},
+		}
```
skberkeley: we should do this at the `lintPaths` level, instead of redoing it for every file

skberkeley (changes requested):

lute/cli/commands/lint/init.luau
wmccrthy: We can move some of the derivation out for sure, but we'll need to pass thru globals and rule configs to this to `lintString` regardless, since I assume this is the level at which we'll do things like:
- override violation severity w/ the configured severity
- not run rules according to rule-specific ignores

skberkeley: yeah sorry, i just meant we shouldn't do the computation of globals and options at this level

lute/cli/commands/lint/init.luau
```diff
@@ -65,10 +65,30 @@ local function loadLintRule(path: string): types.LintRule
...
+	rules: { types.LintRule },
+	ruleConfigs: types.RuleConfigurations,
+	ruleIgnores: types.RuleIgnores
+)
+	table.insert(rules, rule)
+	local ruleConfig: types.RuleConfig = if config.rules ~= nil and config.rules[rule.name] ~= nil
+		then config.rules[rule.name]
+		else {}
+	ruleConfigs[rule.name] = ruleConfig
+	ruleIgnores[rule.name] = parseIgnores.parseIgnoreContents(rootLocation, ruleConfig.ignores or {})
```
```suggestion
    if ruleConfig.ignores ~= nil then
        ruleIgnores[rule.name] = parseIgnores.parseIgnoreContents(rootLocation, ruleConfig.ignores)
    end
```

lute/cli/commands/lint/types.luau
```diff
@@ -19,31 +20,41 @@ export type LintViolation = {
 }
 
 export type LintRule = {
-	read lint: (ast: syntax.AstStatBlock, sourcepath: path.path?) -> { LintViolation },
+	read lint: (ast: syntax.AstStatBlock, sourcepath: path.path?, context: RuleContext) -> { LintViolation },
```
skberkeley, line 23:
```suggestion
	read lint: (ast: syntax.AstStatBlock, sourcepath: path.path?, context: RuleContext?) -> { LintViolation },
```

lute/cli/commands/lint/init.luau
```diff
@@ -228,7 +252,28 @@ local function lintString(
 	local violations: { types.LintViolation } = {}
 	local directiveCache = {}
 	for _, rule in lintRules do
-		local success, err = pcall(rule.lint, ast, inputFilePath)
+		if inputFilePath then
+			-- check if rule-specific ignores omit the current path
+			local ruleIgnores = configData.ruleIgnores[rule.name]
+			if
+				#ruleIgnores.ignores > 0
```
skberkeley: what if we set `ruleIgnores[rule.name] = nil` instead of `{}`?

lute/cli/commands/lint/init.luau
```diff
@@ -228,7 +252,28 @@ local function lintString(
...
+				and parseIgnores.isIgnored(
+					configData.ruleIgnores[rule.name],
```
```suggestion
					ruleIgnores,
```

lute/cli/commands/lint/init.luau
```diff
@@ -228,7 +252,28 @@ local function lintString(
...
+					pathLib.format(inputFilePath),
+					fs.type(inputFilePath) == "dir"
```
skberkeley: can you put this in a variable outside the for loop?

lute/cli/commands/lint/init.luau
```diff
@@ -228,7 +252,28 @@ local function lintString(
...
+				)
+			then
+				continue
+			end
+		end
+
+		local ruleConfig = configData.ruleConfigs[rule.name]
+		local context = {
+			globals = configData.globals,
+			options = if ruleConfig.options then ruleConfig.options else {},
```
skberkeley, line 274:
```suggestion
			options = ruleConfig.options,
```
this should be fine, since `RuleContext.options` is `unknown`

tests/cli/lint.test.luau
```diff
@@ -2669,7 +2669,9 @@ end
 
 	suite:case("lute lint accounts for configured ignores (config passed as arg)", function(assert)
 		local testSubDir = path.join(tmpDir, "testDir")
-		fs.createdirectory(testSubDir)
+		if not fs.exists(testSubDir) then
+			fs.createdirectory(testSubDir)
+		end
```
skberkeley: maybe just use the `modulePath` directory? see the top of the file

skberkeley: sorry, should've pointed you to it earlier

tests/cli/lint.test.luau
```diff
@@ -2779,4 +2785,225 @@ end
...
+		return {
+			lute = {
+				lint = {
+					globals = {
+						["testGlobal"] = 1,
+					},
+					rules = {
+						["testRule"] = {
+							options = {
+								["testOption"] = 1,
```
skberkeley: can you make the values different just in case

tests/cli/lint.test.luau
```diff
@@ -2779,4 +2785,225 @@ end
...
+			violatorFile,
+			[[
+		local x = 1 / 0
+		local a, b
+		a = b
+		b = a
+]]
+		)
+
+		local configFile = path.join(testSubDir, ".config.luau")
```
skberkeley, line 2573: this test makes an implicit assumption that the test runner's working directory isn't `testSubDir`. we should either specify the working directory in the process.run calls or put these files in different directories

skberkeley (changes requested):
just some small comments!

lute/cli/commands/lint/init.luau
wmccrthy: good shout, that should simplify

lute/cli/commands/lint/init.luau
wmccrthy: No prob. Also, I just added this check to be safe but will `inputFilePath` ever be a dir here? It seems like all paths come downstream from `files.getSourceFiles`, which I think should only output file paths?

lute/cli/commands/lint/init.luau
wmccrthy: Yea, I suppose it's a question of if we want users to have to check if `options` are nil in lint rule implementations or not. My thinking was:
- they will have to check if individual fields on options are nil regardless (tho i suppose this is only if defined options is a table)
- having to check if `options` itself is nil, then the fields on top of that feels defensive and might be annoying on the user side, so at the very least assuring options is an empty table avoids the former.

What do you think? Maybe it's worth enforcing a little stricter that options must be a table?

lute/cli/commands/lint/types.luau
wmccrthy: From the perspective of having written a bunch of rules, I think it'd be nicer devX if `context` itself was never nil, for similar reasons as I mentioned keeping options non-nil in [this comment.](https://github.com/luau-lang/lute/pull/808/changes#r2800943256)

lute/cli/commands/lint/init.luau
skberkeley: we do already have the globals as `{ [ string ] : unknown }`, so maybe it's worth just doing the same for options?

lute/cli/commands/lint/init.luau
skberkeley: yeah, `getSourceFiles` only returns files, so this could just be `false` instead

lute/cli/commands/lint/types.luau
skberkeley: sure, that's reasonable

lute/cli/commands/lint/init.luau
wmccrthy: Yea, that interface makes sense to me. Just curious, might it cause downstream type issues if rule authors try to index into `globals` / `options` with `.index` syntax as opposed to `['index']`? I know i've ran into similar nits before

skberkeley: I think both should work:
image
Could've been an older bug with the new solver

wmccrthy: 🔥

skberkeley (approved):
looks good!

annieetang:
67 👀

PR #811, opened 2026-02-11
Improves fs.watch implementation by using deque

annieetang:
Updates implementation of `fs.watch` with the `deque` from batteries instead of `table.remove(_, 1)`. This still doesn't help with resolvig the perf issues we've been having with `fs.watch` but using the deque is still better than removing from the front of a table


skberkeley:
> Updates implementation of `fs.watch` with the `deque` from batteries to improve the performance and timeout issues we've been having that's most likely related to `table.remove(_, 1)`

I would update this to say that it probably doesn't fix the perf issues we've been having with `fs.watch`

skberkeley (approved):

PR #812, opened 2026-02-12
Adds unused variable docs and links violations to it

skberkeley:
Also removed a cast from the unused variable list to report it in the faillist instead


annieetang (approved):
ty for adding the LUAUFIX comment!

PR #813, opened 2026-02-12
Introduces an explicit Batteries VFS to handle requires from @std libraries and cli commands.

Vighnesh-V:
This is a refactor PR that splits out the Batteries specific behavior from the CliVFS

### DIFF
--- a/lute/require/include/lute/requirevfs.h
+++ b/lute/require/include/lute/requirevfs.h
@@ -1,5 +1,6 @@
 #pragma once
 
+#include "lute/batteriesvfs.h"
 #include "lute/bundlevfs.h"
 #include "lute/clivfs.h"
 #include "lute/filevfs.h"
@@ -53,6 +54,7 @@ private:
         Cli,
         Bundle,
         Lute,
+        Batteries, // Only for internal use
     };
 
     VFSType vfsType = VFSType::Disk;
@@ -60,6 +62,7 @@ private:
     FileVfs fileVfs;
     StdLibVfs stdLibVfs;
     LuteVfs luteVfs;
+    BatteriesVfs batteriesVfs;
     std::optional<CliVfs> cliVfs = std::nullopt;
     std::optional<BundleVfs> bundleVfs = std::nullopt;
 };
--- a/lute/require/src/requirevfs.cpp
+++ b/lute/require/src/requirevfs.cpp
@@ -10,12 +10,14 @@
 #include "lualib.h"
 
 RequireVfs::RequireVfs(CliVfs cliVfs)
-    : cliVfs(std::move(cliVfs))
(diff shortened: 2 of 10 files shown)


Vighnesh-V:
Going to put this in draft mode again because it looks like there is an issue with the lute-check script!

lute/require/include/lute/requirevfs.h
```diff
@@ -53,13 +54,15 @@ class RequireVfs : public IRequireVfs
...
         Lute,
+        Batteries, // Only for internal use
     };
 
     VFSType vfsType = VFSType::Disk;
 
     FileVfs fileVfs;
     StdLibVfs stdLibVfs;
     LuteVfs luteVfs;
+    BatteriesVfs batteriesVfs;
```
annieetang, line 65: curiosity q! is there a reason we aren't making this `std::optional` like cli or bundle?

annieetang (dismissed):
lgtm - ty for implementing this!

lute/require/include/lute/requirevfs.h
Vighnesh-V: Yeah! The Standard Library is guaranteed to exist (and will probably call batteries, as of this new change). So if the StdlibVFS is guaranteed to exist, the BatteriesVFS needs to as well.

vrn-sn: Long-term, feel like this should just be a `Luau::Variant` of all the possible VFSs. It's such a huge waste of memory to store all of them at all times. (my fault)

lute/require/src/requirevfs.cpp
vrn-sn: While we're at it, worth setting vfsType here too? Or is that not appropriate for this case? Can't remember the details of BundleVfs that well.

vrn-sn (approved):

lute/require/src/requirevfs.cpp
Vighnesh-V: probably worth it :), I'll go ahead and do that

vrn-sn: Does bundled code just execute entirely within the bundle? (i.e. there's no FileVfs active ever, even for the entry point)?

lute/require/include/lute/requirevfs.h
Vighnesh-V: Are you also suggesting switching the active VFS in the variant when you switch VFSTypes? Each VFS is stateful, so that might actually be a problem.

Vighnesh-V: Otherwise, there's only like one instance of this per runtime, so honestly not too much memory overhead.

annieetang:
ty!!! this helps with https://github.com/luau-lang/lute/pull/811

PR #817, opened 2026-02-12
fix lint severity mapping

wmccrthy:
I found it odd that my lint "warning" violations kept showing up with blue squiggles instead of the red I'm more used to; did a little digging and it looks like we our severity mapping was off by one, see [VS Code API def here:](https://code.visualstudio.com/api/references/vscode-api#DiagnosticSeverity)

Screenshot 2026-02-12 at 2 51 21 PM

This fixes mapping.


wmccrthy:
Making change on the Mandolin side instead

PR #820, opened 2026-02-13
Adds a no-default-lints option to lute lint

skberkeley:
Also updates lute lint's behavior to always run default rules unless the option is specified


tests/cli/lint.test.luau
```diff
@@ -35,6 +35,7 @@ type lintTestParams = {
 	expectedExitCode: number,
 	rulePath: string?,
 	expectedAutofixContents: string?,
+	disableDefaultLints: true?,
```
annieetang, line 38: `boolean?`?

lute/cli/commands/lint/init.luau
```diff
@@ -28,6 +28,7 @@ local VERBOSE = false
 
 local function printHelp()
 	print(USAGE)
+	-- Ordering of options: help + verbose fist, then options with aliases in alphabetical order, then options without aliases in alphabetical order
```
annieetang, line 31: we could probably pull `printHelp()` into a generic thing in CLI to keep these consistent across all commands later?

tests/cli/lint.test.luau
skberkeley: No, this is deliberate so that you can only pass `true` or not provide it at all

lute/cli/commands/lint/init.luau
skberkeley: yeah, that would be a nice thing to have

annieetang (approved):
lgtm ty!

lute/cli/commands/lint/init.luau
Vighnesh-V: printHelp() is already a utility in the batteries cli parser! I think it's args:help() or something to that effect.

PR #822, opened 2026-02-16
Support Loading Local Rules from Lint Config

wmccrthy:
This PR adds supports for loading rules from paths specified in the lute lint config file under `config.lute.lint.rules.paths`.


lute/cli/commands/lint/types.luau
```diff
@@ -38,7 +38,7 @@ export type LintConfig = {
 	ignores: { path.pathlike }?, -- globs/paths to ignore as we walk lintee files
 	globals: GlobalsConfig?, -- globals to pass down to all rules via context
 	rules: {
-		-- paths: { path.pathlike }?, -- paths to local rules to use
+		paths: { path.pathlike }?, -- paths to local rules to use
```
skberkeley: small bikeshed: is it worth having this as `LintConfig.paths` instead of `LintConfig.rules.paths`? we could then rename `rules` to `ruleConfigs` as well

tests/cli/lint.test.luau
```diff
@@ -2781,4 +2781,147 @@ return {
...
+					}
+				}
+			end
+		}
+		]]
+			fs.writestringtofile(cliProvidedRulePath, cliProvidedRuleContents)
+
+			local configFile = path.join(testSubDir, ".config.luau")
+			-- Escape backslashes so the path survives being embedded in a Luau string literal (Windows)
+			local escapedRulePath = path.format(configuredRulePath):gsub("\\", "\\\\")
```
skberkeley: are you intending for this to be an absolute path?

tests/cli/lint.test.luau
```diff
@@ -2781,4 +2781,147 @@ return {
...
+			}
+		}
+		]]):format(escapedRulePath)
+			fs.writestringtofile(configFile, configContents)
+
+			local ignoredDir = path.join(testSubDir, "ignored")
+			local notIgnoredDir = path.join(testSubDir, "not_ignored")
+			fs.createdirectory(ignoredDir)
+			fs.createdirectory(notIgnoredDir)
+			local ignoredFile = path.join(ignoredDir, "ignored.luau")
```
skberkeley, line 2952: is this test passing rn? if so it smells like a bug because the ignore glob doesn't match the directory name you're trying to ignore

skberkeley (dismissed):

lute/cli/commands/lint/types.luau
wmccrthy: I think that's fair. Maybe `LintConfig.rulePaths`? I'd imagine ppl may get confused abt whether `paths` refers to paths to the rules to run or files to lint

skberkeley: yep, that sounds good!

tests/cli/lint.test.luau
wmccrthy: No, just ran into a weird windows only bug w/ how the path string was inserted by `string.format`. Turns out I can also just use `%q` instead of `%s` and it takes care of the formatting overhead here ([src](https://www.lua.org/pil/12.1.html#:~:text=To%20quote%20an%20arbitrary%20string,All%20rights%20reserved.)).

tests/cli/lint.test.luau
wmccrthy: hm was definitely a bug; been doing some digging, when `assert.strcontains` is passed 1 for the `instances` arg, it seems to evaluate as true if there are 1 or more instances of the given pattern in the given string. Is that intended? Seems kinda misleading IMO, i'd been using this pattern to assert that there was ONLY 1 instance in a couple cases

wmccrthy: To be clear, there were other issues w/ the logic I was testing, just wanted to note the above since I found it surprising

skberkeley: huh that does sound misleading, i'll take a look at it

tests/cli/lint.test.luau
```diff
@@ -2177,7 +2177,7 @@ end
 		-- 	return {
 		-- 		lute = {
 		-- 			lint = {
-		-- 				rules = "should be table"
+		-- 				ruleConfigs "should be table"
 		-- 			}
 		-- 		}
 		-- 	}
```
skberkeley, line 2184: can we uncomment this stuff now?

skberkeley (commented):
can you update the tests you added to use the new lint test helper?

tests/cli/lint.test.luau
wmccrthy: was thinking about addressing all of these in a follow up PR, [ticket here](https://roblox.atlassian.net/browse/UIBLOX-4126), to keep this PR as slim as possible. Happy to add here if you're cool reviewing that now

wmccrthy:
> can you update the tests you added to use the new lint test helper?

Moved the first over but I'm not sure the second case is a good candidate for the helper (without overcomplicating the helper itself). We're creating two separate lintee files and putting them in different nested dirs, so it'd b quite a lift for the helper. Wdyt?

tests/cli/lint.test.luau
skberkeley: yeah sure, we can just do it in a followup

skberkeley:
> > can you update the tests you added to use the new lint test helper?
> 
> Moved the first over but I'm not sure the second case is a good candidate for the helper (without overcomplicating the helper itself). We're creating two separate lintee files and putting them in different nested dirs, so it'd b quite a lift for the helper. Wdyt?

yeah that's fair, it's hard to test the ignore logic with the helper rn

skberkeley (commented):
sorry, just one more thing: can you test specifying the rule in the config via a relative path? i believe it's only testing absolute right now

wmccrthy:
> sorry, just one more thing: can you test specifying the rule in the config via a relative path? i believe it's only testing absolute right now

good call

tests/cli/lint.test.luau
```diff
@@ -2820,4 +2821,153 @@ return {
...
+			})
+			assert.eq(result.exitcode, 1) -- errors
+			assert.strnotcontains(result.stdout, "ignored.luau")
+			assert.strcontains(result.stdout, "valid.luau")
+			assert.strcontains(result.stdout, "Configured Rule Ran with options.a = 1") -- options are correctly passed through
+		end
+	)
 end)
+
+test.run()
```
skberkeley: can you configure whatever coding tool you're using to stop adding these?

tests/cli/lint.test.luau
```diff
@@ -2820,4 +2821,153 @@ return {
...
+			}
+		}
+	end
+}
+		]]
+		-- Create the config-provided rules (loaded via config.rulepaths, NOT via -r)
+		local configuredAbsoluteRulePath = path.format(path.join(tmpDir, "configuredAbsRule.luau"))
+		fs.writestringtofile(
+			configuredAbsoluteRulePath,
+			(ruleTemplate :: any):format(
```
skberkeley, line 2845: why do you need an any cast here?

tests/cli/lint.test.luau
wmccrthy: my bad 🤦 i'm manually adding cause some weird local bugs I've been having w/ `lute test` and always forget to remove before pushing

tests/cli/lint.test.luau
wmccrthy:
```
TypeError: We cannot statically check the type of `string.format` when called with a format string that is not statically known.
If you'd like to use an unchecked `string.format` call, you can cast the format string to `any` using `:: any`.
```
Type issue when formatting a string that is declared as a local and formatted thereafter, as opposed to formatted in the declaration.

wmccrthy: Can leave the errr and add a `LUAUFIX` comment if we'd prefer?

skberkeley: Can you use `:: string` instead? Casting to any throws away a lot of type information which might suppress actual type errors

wmccrthy: I think the solver is already interprating `ruleTemplate` as a string, it's the formatting that it can't statically resolve. Casting to string yields the same error:
```
TypeError: We cannot statically check the type of `string.format` when called with a format string that is not statically known.  
If you'd like to use an unchecked `string.format` call, you can cast the format string to `any` using `:: any`.
```

lute/cli/commands/lint/init.luau
```diff
@@ -102,8 +102,9 @@ local function registerRule(
 	ruleConfigs: internalTypes.RuleConfigurations,
 	ruleIgnores: internalTypes.RuleIgnores
 )
-	local ruleConfig: internalTypes.RuleConfig = if config.rules ~= nil and config.rules[rule.name] ~= nil
-		then config.rules[rule.name] :: internalTypes.RuleConfig -- Refinement doesn't support dynamic indexing, so we need to assert the type here
+	local ruleConfig: internalTypes.RuleConfig = if config.ruleconfigs ~= nil
+			and config.ruleconfigs[rule.name] ~= nil
+		then config.ruleconfigs[rule.name] :: internalTypes.RuleConfig -- Refinement doesn't support dynamic indexing, so we need to assert the type here
 		else {}
```
hgoldstein, line 107: There's several ways to write this one w/o an assertion
```suggestion
	local ruleConfig: internalTypes.RuleConfig = if config.ruleconfigs ~= nil
			then config.ruleconfigs[rule.name] or {}
			else {}
```

tests/cli/lint.test.luau
hgoldstein: We're nixing this error but for now `:: any` please.

tests/cli/lint.test.luau
```diff
@@ -2820,4 +2821,151 @@ return {
...
+		)
+		local configuredRelativePath = path.format(path.join(tmpDir, "configuredRelativeRule.luau"))
+		fs.writestringtofile(
+			configuredRelativePath,
+			(ruleTemplate :: any):format(
+				"configuredRelativeRule",
+				"configuredRelativeRule",
+				"configuredRelativeRule ran."
+			)
+		)
```
hgoldstein, line 2859: I know I am a bit late to this, but does it make more sense to write out all of these structures to disk rather than re-make them every time?

hgoldstein (approved):
The tests seem fine but I'd recommend you change these by just reifying some of the directory structures in `lute` itself rather than setting up and tearing them down each time, similar to what you did for the codemod tests.

lute/cli/commands/lint/init.luau
wmccrthy: Nice, thanks! Curious what the difference is here at the solver level? I guess why `config.ruleconfigs[rule.name] ~= nil` isn't refined as you'd expect

tests/cli/lint.test.luau
wmccrthy: The general structure of all the lint tests seems to follow: create tmp files for violation / custom rules and teardown after.

I assume this was originally done due to the lack of a snapshot testing framework, which is now solved 🎉 , but I feel like it makes more sense to respect existing structures here rather than using a different standard just for our additions.

lute/cli/commands/lint/init.luau
hgoldstein: > ... why `config.ruleconfigs[rule.name] ~= nil` isn't refined as you'd expect

We didn't write support for it: it's _probably_ doable but also I think it might be in poor taste.

tests/cli/lint.test.luau
hgoldstein: Fair enough!

lute/cli/commands/lint/init.luau
wmccrthy: Gotcha

PR #826, opened 2026-02-17
Cleans up some Luau AST types

skberkeley:
Refactors some Luau AST types to be more consistent. The general naming convention is `Ast<node category><node type>`, ie `AstExprBinary`. _Only_ node types which are part of the category in their name should follow this convention, to avoid confusion around whether they are full nodes or not. For example, we previously had `AstTypeFunctionParameter` which I renamed to `AstFunctionTypeParameter` and `AstExprTableItemList` which I renamed to `AstTableExprListItem`. 

I also simplified some of the top level AST type definitions, since our tagged union inference is stronger now.


tools/check-faillist.txt
```diff
@@ -264,108 +266,70 @@ tests/std/process.test.luau:123:41-42
...
+tests/std/syntax/printer.test.luau:733:11-31
+tests/std/syntax/printer.test.luau:746:11-31
+tests/std/syntax/printer.test.luau:759:11-31
+tests/std/syntax/printer.test.luau:772:11-31
+tests/std/syntax/printer.test.luau:785:11-31
+tests/std/syntax/printer.test.luau:798:11-31
+tests/std/syntax/printer.test.luau:811:11-31
+tests/std/syntax/printer.test.luau:825:11-31
+tests/std/syntax/printer.test.luau:841:11-100
+tests/std/syntax/printer.test.luau:872:11-100
```
hgoldstein: What's going on with these?

skberkeley: talked offline, these are code too complex errors

aatxe: Can we check which limits are getting tripped here?

skberkeley: I'm seeing `NormalizationTooComplex` when I stick it in a Luau unit test

hgoldstein (approved):
You'll probably need to merge in my changes to the fail-list but otherwise LGTM

PR #827, opened 2026-02-17
Don't Error when Loading Invalid Lint Rule

wmccrthy:
`lute lint` will fail if any of the paths derived from the passed `-r` arg are _not_ valid lint rules. I think it'd be a nicer UX for consumers to be able to call `lute lint -r {path_to_suite}/src ...` rather than maintaining some dir that is exclusively the rule implementations themselves (which will prob b a bunch of re-exports in most cases anyway).

This PR refactors the loading process to silently ignore invalid lint rules, such that `lute lint` will exhaustively load all rules underneath the path passed to `-r`, without requiring each discovered .luau file to be a rule implementation


lute/cli/commands/lint/init.luau
```diff
@@ -66,11 +66,11 @@ local function assertIsLintRule(rule: unknown, path: string)
 	)
 end
 
-local function loadLintRule(path: string): types.LintRule
+local function loadLintRule(path: string): types.LintRule?
 	local loaded = luau.loadbypath(path)
-	assertIsLintRule(loaded, path)
+	local success, _ = pcall(assertIsLintRule, loaded, path)
```
wmccrthy: getting a type error in `lute check` / check CI here, but not seeing it locally...

skberkeley: `lute check` should run with all flags, so you might need to change your luau lsp settings to see all the errors

skberkeley: why don't we just refactor `assertIsLintRule` to be `isLintRule` and return a boolean?

skberkeley: can we also move the verbose printing into that function so we can also include the reason why a path isn't a lint rule?

wmccrthy: sure thing

skberkeley (approved):

PR #829, opened 2026-02-18
Fix various Luau type errors and update Lute's workspace to enable all flags

hgoldstein:
I went through `tools/check-faillist.txt` and tried to fix as many errors as I could reasonably find. There are definitely some errors that I've fixed by paving over bugs in the new solver (for example all of the `table.sort` invocations). Additionally I've noted some interesting new solver bugs:
* Comparing anything with `nil` should not be an error
* Errant warning on `#` on a metatable with `__len`
* There are some spots where we need better table types: read-only indexers, for example.
* There's a couple refinement bugs in `traverseTable` in `lute/batteries/pp.luau`, one of which seems very new.
* We should add an overload to pcall that allows taking a function that returns an empty pack.

I've also updated the VSCode workspace to enable all flags in `luau-lsp`: this should help with some reported mismatches between the faillist and folks editor experience.


batteries/difftext/printdiff.luau
```diff
@@ -67,11 +67,11 @@ local function formatLineSideHeader(
 	local newStr = tostring(if lineNumbers then lineNumbers.newLine else "")
 	maxNumWidth = maxNumWidth or 1
 	if operation == "EQUAL" then
-		return string.format(`  %{maxNumWidth}s %{maxNumWidth}s| `, oldStr, newStr)
+		return string.format(`  %{maxNumWidth}s %{maxNumWidth}s| ` :: any, oldStr, newStr)
```
Vighnesh-V, line 70: What was the original typechecker error here?

batteries/pp.luau
```diff
@@ -90,14 +90,17 @@ local function traverseTable(
...
+	local keys: { unknown } = {}
 
 	-- Collect all keys, not just primitives
 	for key in dataTable do
 		table.insert(keys, key)
 	end
 
-	table.sort(keys, function(a: string, b: string): boolean
+	-- NOTE: We shouldn't need the annotations on `a` and `b`, but
+	-- the new solver is not happy.
```
Vighnesh-V: Can you cut a ticket for this if it doesn't already exist?

Vighnesh-V: This pattern happens a bunch, and it would be nice to have this fixed.

definitions/luau.luau
```diff
@@ -172,6 +172,9 @@ export type AstExprTableItemList = {
 	value: AstExpr,
 	separator: Token<"," | ";">?,
 	istableitem: true,
+	-- This is here so that the type of
+	-- AstExprTableItem is coherent.
+	key: nil,
```
skberkeley: why do we need this? `AstExprTableItem` should ideally be a tagged union based on `kind`

tests/std/fs.test.luau
```diff
@@ -209,6 +209,9 @@ test.suite("FsSuite", function(suite)
 
 		local success, err = pcall(function()
 			fs.removedirectory(dir, { recursive = false })
+			-- NOTE: We should add an overload to `pcall` that allows
+			-- passing a function that returns no arguments.
```
Vighnesh-V: Ticket pls if it doesn't exist already. You can assign it to me.

Vighnesh-V (commented):
Small comments, but otherwise looks fine.

skberkeley (approved):

batteries/difftext/printdiff.luau
aatxe, line 70: can we add a `-- FIXME(Luau): ` comment here explaining the bug if a bug is to blame for this not working?

batteries/pp.luau
```diff
@@ -90,14 +90,17 @@ local function traverseTable(
 	local output = ""
 	local indentStr = string.rep("  ", indent)
 
-	local keys = {}
+	-- LUAUFIX: We shouldn't need this annotation.
```
```suggestion
	-- FIXME(Luau): We shouldn't need this annotation.
```

batteries/pp.luau
```diff
@@ -90,14 +90,17 @@ local function traverseTable(
...
+	local keys: { unknown } = {}
 
 	-- Collect all keys, not just primitives
 	for key in dataTable do
 		table.insert(keys, key)
 	end
 
-	table.sort(keys, function(a: string, b: string): boolean
+	-- NOTE: We shouldn't need the annotations on `a` and `b`, but
```
```suggestion
	-- FIXME(Luau): We shouldn't need the annotations on `a` and `b`, but
```

tests/std/fs.test.luau
```diff
@@ -209,6 +209,9 @@ test.suite("FsSuite", function(suite)
 
 		local success, err = pcall(function()
 			fs.removedirectory(dir, { recursive = false })
+			-- NOTE: We should add an overload to `pcall` that allows
```
aatxe:
```suggestion
			-- FIXME(Luau): We should add an overload to `pcall` that allows
```

tests/std/json.test.luau
```diff
@@ -69,7 +69,7 @@ test.suite("JSONParsingTestSuite", function(suite)
 		]]
 
 		local maybeJson = json.deserialize(myjsonstr)
-		local obj = json.asobject(maybeJson)
+		local obj: any = json.asobject(maybeJson)
```
aatxe: Is this covering a bug? I'm pretty sure there is an "object" type from the json library, so shouldn't this just be that?

tests/std/process.test.luau
```diff
@@ -76,23 +76,23 @@ test.suite("ProcessSuite", function(suite)
 		end, "process.run requires a non-empty table of arguments")
 
 		assert.erroreq(function()
-			process.run({ "echo", "hello" }, { cwd = {} })
+			process.run({ "echo", "hello" }, { cwd = {} :: any })
```
aatxe, line 84: a FIXME(Luau) comment somewhere in this file to say that all the any casts on tables are bidirectional typing problems?

tools/luthier.luau
```diff
@@ -194,7 +194,7 @@ local function traverseFileTree(path: string, callback: (string, string) -> ())
 		callback(currentDir.path, currentDir.relativePath)
 
 		local contents = fs.listdir(currentDir.path)
-		table.sort(contents, function(a, b)
+		table.sort(contents, function(a: fs.DirectoryEntry, b: fs.DirectoryEntry)
```
aatxe, line 198:
```suggestion
        -- FIXME(Luau): bidirectional typing
		table.sort(contents, function(a: fs.DirectoryEntry, b: fs.DirectoryEntry)
```

tools/luthier.luau
```diff
@@ -691,12 +691,16 @@ local function getConfigureArguments()
 	local configArgs =
 		{ "-G=Ninja", "-B " .. projectPath, "-DCMAKE_BUILD_TYPE=" .. config, "-DCMAKE_EXPORT_COMPILE_COMMANDS=1" }
 
-	if args:get("cxx-compiler") then
-		table.insert(configArgs, "-DCMAKE_CXX_COMPILER=" .. args:get("cxx-compiler"))
+	local cxx_compiler = args:get("cxx-compiler")
+
+	if cxx_compiler then
```
aatxe, line 696:
```suggestion
	local cxx_compiler = args:get("cxx-compiler")
	if cxx_compiler then
```

tools/luthier.luau
```diff
@@ -691,12 +691,16 @@ local function getConfigureArguments()
...
+		table.insert(configArgs, "-DCMAKE_CXX_COMPILER=" .. cxx_compiler)
 	end
 
-	if args:get("c-compiler") then
-		table.insert(configArgs, "-DCMAKE_C_COMPILER=" .. args:get("c-compiler"))
+	local c_compiler = args:get("c-compiler")
+
+	if c_compiler then
```
aatxe, line 701:
```suggestion
	local c_compiler = args:get("c-compiler")
	if c_compiler then
```

tools/luthier.luau
```diff
@@ -707,7 +711,9 @@ local function getConfigureArguments()
 end
 
 local function readTuneFile(path: string): Tune
-	return toml.deserialize(readFileToString(path))
+	-- We're promising that the thing we read _actually_ looks like a `Tune`,
+	-- this could be ill advised.
```
aatxe, line 714: Yeah, we ought to implement validation that it complies with this type. Pretty trivial to do, I can do a follow-up PR with it after this one.

batteries/difftext/printdiff.luau
hgoldstein: Passing a non-literal string to `string.format`.

definitions/luau.luau
hgoldstein: Rewrote the code that was tripping over this.

batteries/pp.luau
hgoldstein: It's bidirectional inference: there's several tickets for us.

batteries/difftext/printdiff.luau
hgoldstein: I'm gonna leave these in for now: this is a check for not using dynamic strings with `string.format`, and this in fact a dynamic string (being interpolated). We may remove this in the future, sounds like.

tests/std/process.test.luau
hgoldstein: They're not, it's all intentionally passing arguments with incorrect types. AFACT `cwd` is meant to be a `string`, not a table, it's more obvious with the other tests, I think.

tests/std/json.test.luau
hgoldstein: I'll rewrite this: it's not covering a bug, more that we treat `json.asobject` as a type predicate: if it returns true we assume that the given value is an `object`.

tests/std/process.test.luau
aatxe: Probably worth a comment that these tests are intentionally ill-typed then?

PR #830, opened 2026-02-18
Fix Lute Lint `types` Dependency

wmccrthy:
The Lint CLI's `types` file is currently dependent on `parseIgnores`. This leads to errors when locally using the lint files included in `lute setup`, since `parseIgnores` is not (and should not be) bundled into `lute setup`'s definition files.

This PR duplicates the `Glob` and `GitignoreData` type definitions, so we can remove the `parseIgnores` dependency


skberkeley:
Closing in favor of #832, thanks for bringing this up!

PR #831, opened 2026-02-18
Add support custom rules and config in lint test helper

skberkeley:
felt the need for this refactor given the addition of configs to lute lint

### DIFF
--- a/tests/cli/lint.test.luau
+++ b/tests/cli/lint.test.luau
@@ -51,20 +51,12 @@ local function lintTestHelper(assert: testTypes.asserts, params: lintTestParams)
 	-- Do
 	local lintArgs = { lutePath, "lint" }
 
-	if params.rulePath then
-		if params.customRuleContents then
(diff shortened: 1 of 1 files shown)


tests/cli/lint.test.luau
```diff
@@ -47,14 +49,28 @@ local function lintTestHelper(assert: testTypes.asserts, params: lintTestParams)
 	fs.writestringtofile(violatorPath, params.content)
 
 	-- Do
-	-- Run the transformer on the transformee
 	local lintArgs = { lutePath, "lint" }
 
 	if params.rulePath then
```
vrn-sn: Maybe a little cleaner to do:
```
if both then
    error
elseif one then
    logicForOne
elseif two then
    logicForTwo
end
```

vrn-sn (approved):

tests/cli/lint.test.luau
skberkeley: true, i'll do that

PR #832, opened 2026-02-19
Separates internal lint types into another file

skberkeley:
This ensures that they don't leak into the type definitions generated by `lute setup`. This was previously causing issues because they relied on types related to `.gitignore` parsing, which they were requiring using a relative path which was breaking in the typedef files.


annieetang (approved):
lgtm but thoughts on having `types/init.luau` and `types/internal.luau`? doesn't really matter to me but just a style thought

skberkeley:
> lgtm but thoughts on having `types/init.luau` and `types/internal.luau`? doesn't really matter to me but just a style thought

We wouldn't reexport `internal` from `init` so I don't think it changes things by that much

PR #835, opened 2026-02-19
Adds empty if block lint rule

skberkeley:
Matches https://kampfkarren.github.io/selene/lints/empty_if.html


lute/cli/commands/lint/rules/empty_if_block.luau
```diff
@@ -0,0 +1,130 @@
...
+		return false
+	end
+
+	-- If comments_count is false, we consider blocks with only comments as empty
+	if not commentsCount then
+		return true
+	end
+
+	-- If comments_count is true, check if there are any comments
+	return not (containsComment(leadingTrivia) or containsComment(trailingTrivia))
```
aatxe:
```suggestion
    -- if comments_count is true, the block is not empty when it contains comments.
	if containsComment(leadingTrivia) or containsComment(trailingTrivia) then
		return false
	end

	return true
```

aatxe (approved):

PR #836, opened 2026-02-19
Lute Lint: ensure `rootLocation` is always absolute

wmccrthy:
When running `lute lint ./ -c ./config.lint.luau` locally, Vighnesh ran into an issue with the configs ignore logic:
```
@std/path/posix/init.luau:209: Cannot compute relative path between absolute and relative paths
stacktrace:
[C] function error
@std/path/posix/init.luau:209 function relative
@std/path/init.luau:83 function relative
@cli/lib/parseIgnores.luau:165 function isIgnored
@cli/lint/init.luau:448 function lintPaths
@cli/lint/init.luau:588 function main
@cli/lint/init.luau:611
```
Since Vighnesh passed a relative path for the `-c` arg, the `configPath` and `rootLocation` derived from it in the main linting flow were both using relative paths, thus our error.
As the linter walks the fs, it checks each against our `ignoreData` to determine if the current path should be ignored / linted. All of these walked paths are absolute, which is ensured by [this logic](https://github.com/luau-lang/lute/blob/primary/lute/cli/commands/lib/files.luau#L55-L57) in `getSourceFiles` (where we derive lintee filepaths from). Thus, if `rootLocation` is relative, an error will be thrown. To avoid this, we can normalize the `rootLocation` to be absolute **always**


lute/cli/commands/lint/init.luau
```diff
@@ -523,7 +523,7 @@ local function main(...: string)
 
 	local ignores = if lintConfig.ignores ~= nil then lintConfig.ignores else {}
 	local rootLocation = if fs.exists(configPath)
```
Vighnesh-V, line 529: fs.exists, and fs.type is file here.

tools/check-faillist.txt
```diff
@@ -55,7 +55,7 @@ lute/cli/commands/lint/init.luau:512:56-65
 lute/cli/commands/lint/init.luau:512:56-65
 lute/cli/commands/lint/init.luau:512:56-65
 lute/cli/commands/lint/init.luau:525:36-45
-lute/cli/commands/lint/init.luau:526:39-48
+lute/cli/commands/lint/init.luau:526:55-64
```
Vighnesh-V: Whats the type error here?  Can we fix it?

Vighnesh-V (changes requested):
Can you add a test for this?

tools/check-faillist.txt
wmccrthy: Error is:
```
TypeError: Expected this to be
	'path | path | pathdata | pathdata | string | string'
but got
	'(path | path | string)?'; 
```

we can fix it by casting `configPath` to `pathlike` in the `pathLib.dirname` call:
```
local rootLocation = if fs.exists(configPath)
		then pathLib.format(pathLib.resolve(pathLib.dirname(configPath)))
		else pathLib.format(process.cwd())
```

wmccrthy: I think when the error first came up, Sora wanted to keep it around for insights into the solver's shortcomings. Lemme confirm

wmccrthy: yea, same error but [slightly different case](https://github.com/luau-lang/lute/pull/784#discussion_r2785098690)

Vighnesh-V: Cool, no worries on fixing it then, since this just moved around.

Vighnesh-V (approved):

PR #839, opened 2026-02-21
Freezes parsed AST nodes to be immutable

skberkeley:
This is a prerequisite for making AST APIs available in Engine.
I wrote a benchmark where we try to parse every luau file in the `lute` repo and benchmarked three approaches (thanks Hunter for the suggestions):
- Manually freezing in the same pass that the AST is serialized
- Freezing in an extra pass in the C++
- Freezing using a userland function
and the second approach was a nice middleground which didn't sacrifice too much performance while having a stronger correctness guarantee from the simplified code.

Numbers for those curious:
```
$hyperfine  "./userland_freeze_lute ./scratch1.luau" "./vm_freeze_lute ./scratch1.luau" "./vm_extra_pass_freeze_lute ./scratch1.luau" --warmup 5 --show-output
Benchmark 1: ./userland_freeze_lute ./scratch1.luau
  Time (mean ± σ):      3.230 s ±  0.023 s    [User: 2.832 s, System: 0.392 s]
  Range (min … max):    3.186 s …  3.265 s    10 runs
 
Benchmark 2: ./vm_freeze_lute ./scratch1.luau
  Time (mean ± σ):      2.562 s ±  0.026 s    [User: 2.170 s, System: 0.386 s]
  Range (min … max):    2.515 s …  2.598 s    10 runs
 
Benchmark 3: ./vm_extra_pass_freeze_lute ./scratch1.luau
  Time (mean ± σ):      2.875 s ±  0.038 s    [User: 2.482 s, System: 0.389 s]
  Range (min … max):    2.845 s …  2.963 s    10 runs
 
Summary
  ./vm_freeze_lute ./scratch1.luau ran
    1.12 ± 0.02 times faster than ./vm_extra_pass_freeze_lute ./scratch1.luau
    1.26 ± 0.02 times faster than ./userland_freeze_lute ./scratch1.luau
```

### DIFF
--- a/lute/luau/src/luau.cpp
+++ b/lute/luau/src/luau.cpp
@@ -39,7 +39,7 @@ static void deepFreeze(lua_State* L)
     lua_rawcheckstack(L, 2);
 
     lua_pushnil(L);
-    while (lua_next(L, -1) != 0)
+    while (lua_next(L, -2) != 0)
     {
         // Freeze the value if it's a table
         if (lua_istable(L, -1))
--- a/lute/luau/src/luau.cpp
+++ b/lute/luau/src/luau.cpp
@@ -29,6 +29,32 @@
 namespace luau
 {
 
+// Recursively freezes the table at the top of the stack and any descendant tables.
+// The table must be at the top of the stack when called.
+static void deepFreeze(lua_State* L)
+{
+    if (!lua_istable(L, -1))
+        luaL_error(L, "expected table to freeze");
+
+    lua_rawcheckstack(L, 2);
+
+    lua_pushnil(L);
+    while (lua_next(L, -1) != 0)
+    {
+        // Freeze the value if it's a table
+        if (lua_istable(L, -1))
+            deepFreeze(L);
+
+        // Pop value, keep key for next iteration
+        lua_pop(L, 1);
+    }
+
+    if (lua_getreadonly(L, -1))
+        return;
+
+    lua_setreadonly(L, -1, 1);
+}
+
 struct StatResult
 {
     std::shared_ptr<Luau::Allocator> allocator;
@@ -131,7 +157,7 @@ struct Span
 
 static Span checkSpan(lua_State* L, int index)
 {
-	lua_checkstack(L, 1);
+    lua_checkstack(L, 1);
 
     if (!lua_istable(L, index))
         luaL_typeerror(L, index, "span");
@@ -2823,6 +2849,8 @@ int luau_parse(lua_State* L)
     }
     lua_setfield(L, -2, "lineoffsets");
 
+    deepFreeze(L);
+
     return 1;
 }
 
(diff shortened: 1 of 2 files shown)


lute/luau/src/luau.cpp
```diff
@@ -2845,6 +3026,8 @@ int luau_parseexpr(lua_State* L)
         luaL_error(L, "parsing failed:\n%s", fullError.c_str());
     }
 
+    lua_rawcheckstack(L, 1);
```
hgoldstein: Why do we need this?

lute/luau/src/luau.cpp
```diff
@@ -596,12 +600,16 @@ struct AstSerialize : public Luau::AstVisitor
...
+
             lua_rawseti(L, -2, i + 1);
         }
+
+        lua_setreadonly(L, -1, 1);
     }
 
     // For correct trivia computation, everything must end up going through serializeToken
-    void serializeToken(Luau::Position position, const char* text, int nrec = 0)
+    void serializeToken(Luau::Position position, const char* text, int nrec = 0, bool freezeToken = true)
```
hgoldstein: Why do we have an extra argument here?

lute/luau/src/luau.cpp
```diff
@@ -1208,18 +1286,15 @@ struct AstSerialize : public Luau::AstVisitor
         node->condition->visit(this);
         lua_setfield(L, -2, "condition");
 
-        if (node->hasThen)
-        {
-            serializeToken(cstNode->thenPosition, "then");
-            lua_setfield(L, -2, "thenkeyword");
+        // Lute isn't in the business of serializing incomplete ASTs
+        LUAU_ASSERT(node->hasThen);
```
hgoldstein: Is that true? I feel like if you were implementing, say, an LSP on top of Lute, you might expect that the AST you're given is incomplete (because the user is writing it).

hgoldstein (changes requested):
Do we need to freeze the table made in the constructor for `AstSerialize`? What about the location structs? Mainly I think you need to handle incomplete ASTs.

One question I have is: how bad would it be if we "deep froze" the table at the very end? Just recurse through the table itself rather than freeze at every junction when serializing.

lute/luau/src/luau.cpp
aatxe: it's at best "true at time of writing," i suppose? but yeah, we'd like to be able to deal with partial ASTs IMO.

lute/luau/src/luau.cpp
```diff
@@ -29,6 +29,32 @@
...
+
+        // Pop value, keep key for next iteration
+        lua_pop(L, 1);
+    }
+
+    if (lua_getreadonly(L, -1))
+        return;
+
+    lua_setreadonly(L, -1, 1);
+}
```
hgoldstein, line 56: Does it make more sense to just have the table check at the top and early exit if otherwise? I get wanting to error, but we're also not exposing this to userland (for now).

hgoldstein (approved):
See comment: this seems fine

lute/luau/src/luau.cpp
skberkeley: I think erroring might provide a better debugging experience so i'd prefer it as it is rn

PR #848, opened 2026-02-26
Move crypto `runtime` tests into `lute`

aatxe:
We seem to have ended up with two separate folders for runtime tests authored in luau, which seems weird and we can obviously clean that up.


hgoldstein (approved):

Vighnesh-V:
probably have to update the faillist:
- `lute tools/check.luau /path/to/rebuilt/luau/binary/ --update`

PR #849, opened 2026-02-26
Fix some @std/fs typechecking errors

Vighnesh-V:
Not too interesting here. One of these is a case where we do some platform specific stuff, so Luau can't actually refine paths based on this information.

### DIFF
--- a/lute/std/libs/fs.luau
+++ b/lute/std/libs/fs.luau
@@ -7,6 +7,7 @@ local deque = require("@batteries/collections/deque")
 local fs = require("@lute/fs")
 local pathlib = require("@std/path")
 local sys = require("@std/system")
+local win32paths = require("@std/path/win32")
 
 local fslib = {}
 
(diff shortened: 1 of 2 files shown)


aatxe (approved):

lute/std/libs/fs.luau
```diff
@@ -178,7 +180,7 @@ end
 
 --- Note: for loops do not support yielding generalized iterators, so we cannot use fs.walk as `for path in fs.walk(...) do` directly. A while loop can be used instead. See example/walk_directory.luau for usage.
 function fslib.walk(path: pathlike, options: walkoptions?): () -> path?
-	local queue = { path }
+	local queue: { path } = { pathlib.parse(path) }
```
hgoldstein, line 183: Is the `{ path }` annotation required?

Vighnesh-V: yeah - it makes the return type of the inner function line up correctly.

hgoldstein: I'd appreciate a minimal repro there 🙂

Vighnesh-V: otherwise, the type system only sees later on that elements from this table are truthy.

Vighnesh-V: lemme get you one

Vighnesh-V: Have one, will add it to a separate set of repros we discussed.

PR #854, opened 2026-03-05
lute/luau: make `span`s into an ordinary frozen table

aatxe:
In #595, we implemented a `span` userdata to replace the table with two locations in order to speed up the overall performance of the parsing library. I rabbitholed a little bit when I did this because, upon further reflection, the performance benefit is only from the fact that we were previously doing three heap allocations per span versus one with the change. The userdata added some extra complexity and "weirdness" to the API, and I'd like to clean things up before a proper stable release. Toward that end, this PR just makes `span` into an ordinary frozen table. It still supports comparison via `<` and it still has the same fields, but it's just a table!

### DIFF
--- a/lute/luau/src/luau.cpp
+++ b/lute/luau/src/luau.cpp
@@ -121,8 +121,6 @@ struct Trivia
     std::string_view text;
 };
 
-// the userdata version of `Luau::Location` because exposing this as a table was, unfortunately, very impractical
-// it happens too much all over the entire AST to do reasonably.
 struct Span
 {
     uint32_t beginLine;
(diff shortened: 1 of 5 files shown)


lute/luau/src/luau.cpp
```diff
@@ -131,38 +129,67 @@ struct Span
...
+        luaL_typeerror(L, index, "span");
+    int beginColumn = lua_tonumber(L, -1);
+
+    if (lua_getfield(L, index, "endline") == LUA_TNIL)
+        luaL_typeerror(L, index, "span");
+    int endLine = lua_tonumber(L, -1);
+
+    if (lua_getfield(L, index, "endcolumn") == LUA_TNIL)
+        luaL_typeerror(L, index, "span");
+    int endColumn = lua_tonumber(L, -1);
```
vrn-sn, line 156: The `lua_tonumber` API doesn't consume values from the stack, so we're adding 4 values to the stack in this function. I think it doesn't match up with the `lua_checkstack(L, 2)` assumption I saw near the call sites of this function.

I think we can just `lua_pop(L, 1)` after each `lua_tonumber(L, -1)`.

vrn-sn (approved):
LGTM, just a small comment!

PR #857, opened 2026-03-06
chore: clean up a bunch of the type errors and analysis warnings in the codebase

aatxe:
I'm trying to burn down a lot of the type errors, clangd warnings, etc. throughout the codebase. This PR does a bunch of that, and leaves comments describing instances of Luau bugs that we had to work around as necessary. The workarounds are typically either a cast or explicit type instantiation that shouldn't be necessary. Some of the changes are not workarounds, and are just cases where a cast was correct to do.


lute/cli/commands/pkg/loom-core/extern/unzip/utils/path.luau
```diff
@@ -44,7 +44,7 @@ local function replaceBackslashes(input: string, replacement: "/"): string
 		return "\\\\" .. input:sub(3):gsub("\\", replacement)
 	else
 		-- Replace all single backslashes
-		return input:gsub("\\", replacement)
+		return (input:gsub("\\", replacement))
```
Vighnesh-V, line 47: this one is really annoying :(

lute/cli/commands/transform/init.luau
```diff
@@ -13,7 +13,7 @@ local function exhaustiveMatch(value: never): never
 end
 
 local function loadMigration(path: string): types.Migration
-	local success, loaded = pcall(luau.loadbypath, path)
+	local success, loaded = pcall(luau.loadbypath, path, nil)
```
Vighnesh-V, line 16: was the error here that pcall expects an optional here, and we were undersaturating?

lute/std/libs/tableext.luau
```diff
@@ -50,15 +50,15 @@ end
 
 -- Functions on array-like tables
 
-function tableext.toset<T>(tbl: { T }): { [T]: true }
+function tableext.toset<T>(tbl: {T}): { [T]: true }
```
Vighnesh-V: stylua should have caught this one.

tests/std/snapshots/runtime_error_doesnt_report_xpcall_case.snap.luau
```diff
@@ -1,8 +1,9 @@
 local test = require("@std/test")
 
 test.case("access_field_of_nil", function(assert)
-	local t = nil
+	local t = nil :: any
 	local value = t.field -- This will cause an error
+	print(value)
 end)
```
Vighnesh-V, line 7: Can we just add things in tests/std/snapshots to the exclude directory in the check script?

Vighnesh-V (approved):
Left a few comments. TLDR:
- stylua
- probably want to add stuff in pkg/loom ... and tests/std/snapshot to the exclude directory in the check script.

lute/cli/commands/transform/init.luau
aatxe: It's about `luau.loadbypath` itself, it takes two arguments, the second of which is optional, but matching typepacks (as must be done by the signature of pcall) requires the arities to match exactly.

lute/std/libs/tableext.luau
aatxe: Stylua doesn't like this actually, and it wants it to be the other way. I find that surprising, but it will be fixed back when we can rerun stylua on this PR again (which is after it cuts a release, which @JohnnyMorganz indicated would hopefully be soon).

tests/std/snapshots/runtime_error_doesnt_report_xpcall_case.snap.luau
aatxe: Why shouldn't our tests be well-typed?

PR #862, opened 2026-03-08
Use embedded sources for `@std`, `@lute`, and `@batteries` in `lute check`

### LINKED ISSUE #773: `lute check` should use the embedded versions of `@std` and `@lute` for typechecking in the same way that `lute run`does

ActualMasterOogway:
This PR ensures that `lute check` correctly resolves and uses embedded source code for the `@std`, `@lute`, and `@batteries` namespaces during static analysis.

Previously, `lute check` used a basic file resolver that only searched the physical disk. Since Lute's standard library and runtime definitions are embedded directly into the executable, `require` calls to these namespaces would fail to resolve. This caused Luau to treat standard modules as `any`, effectively disabling typechecking for code that relied on Lute's built-in APIs.

- **Refactored `resolveRequire`**: Updated the core resolution logic in `lute/luau/src/resolverequire.cpp` to support both physical disk and virtual embedded paths.
- **Implemented `readSourceFromVfs`**: Added a utility to retrieve source code directly from Lute's internal memory buffers when an `@` alias is used.
- **Updated Resolvers**: Integrated the new resolution and reading logic into `LuteModuleResolver` and `LuteFileResolver`.
- **Build System Integration**: Updated `lute/luau/CMakeLists.txt` to link the analysis library against `Lute.Std`, `Lute.Lute`, and `Lute.Batteries`, ensuring the embedded sources are available at compile time.
- **Improved Global Registry**: Ensured the `require` function is correctly recognized as a built-in global during the analysis phase.

Fixes #773


ActualMasterOogway:
Please add the bug and cli labels to this PR, as I am unable to

lute/luau/src/resolverequire.cpp
```diff
@@ -156,8 +264,48 @@ std::optional<std::string> resolveRequire(std::string requirePath, std::string r
...
+            return std::string(result.contents);
+    }
+    else if (name.rfind("@batteries", 0) == 0)
+    {
+        BatteryModuleResult result = getBatteryModule(name);
+        if (result.type == BatteryModuleType::Module)
+            return std::string(result.contents);
+    }
+
+    return readFile(name);
```
Vighnesh-V: Each VFS already implements this functionality, so this is duplicated here. Can you implement the methods `getIdentifier()` and `getContents()` and have each of them call out to the appropriate vfs and then delete this method.

Also, the bit associated with the `LuteModuleResult` needs to go into the LuteVFS.

lute/luau/CMakeLists.txt
```diff
@@ -16,5 +16,5 @@ target_sources(Lute.Luau PRIVATE
 
 target_compile_features(Lute.Luau PUBLIC cxx_std_17)
 target_include_directories(Lute.Luau PUBLIC "include" ${LIBUV_INCLUDE_DIR})
-target_link_libraries(Lute.Luau PRIVATE Lute.Common Lute.Require Lute.Runtime uv_a Luau.Analysis Luau.Ast Luau.Compiler Luau.CLI.lib Luau.Require Luau.VM)
+target_link_libraries(Lute.Luau PRIVATE Lute.Common Lute.Require Lute.Runtime Lute.Lute Lute.Std Lute.Batteries uv_a Luau.Analysis Luau.Ast Luau.Compiler Luau.CLI.lib Luau.Require Luau.VM)
```
Vighnesh-V:
```suggestion
target_link_libraries(Lute.Luau PRIVATE Lute.Common Lute.Require Lute.Runtime uv_a Luau.Analysis Luau.Ast Luau.Compiler Luau.CLI.lib Luau.Require Luau.VM)
```

These should be link dependencies to the require module, as only the vfs's there need to handle accessing embedded modules.

lute/luau/include/lute/moduleresolver.h
```diff
@@ -1,6 +1,7 @@
 #pragma once
 
 #include "Luau/FileResolver.h"
+#include "lute/resolverequire.h"
```
Vighnesh-V: unused import?

lute/luau/include/lute/resolverequire.h
```diff
@@ -6,4 +6,6 @@
 struct lua_State;
 
 std::optional<std::string> resolveRequire(std::string requirePath, std::string requirerChunkname, std::string* error);
```
Vighnesh-V: The resolve require api currently resolves a require path from a different chunkname for explicit use by tools in the file system - this return is intended to be a filesystem path. With this change, the resolveRequire api will now evaluate non-file system paths, which means tools that use `loadbypath` will fail.

Instead of doing this, can you expose a similar method for typechecking?

lute/luau/src/moduleresolver.cpp
```diff
@@ -24,7 +24,7 @@ std::optional<Luau::ModuleInfo> LuteModuleResolver::resolveModule(const Luau::Mo
 
         std::string error;
         std::string requirerChunkname = "@" + context->name;
-        std::optional<std::string> absolutePath = resolveRequire(requirePath, std::move(requirerChunkname), &error);
+        std::optional<std::string> absolutePath = ::resolveRequire(requirePath, std::move(requirerChunkname), &error);
```
Vighnesh-V: With the change I suggested above, can you make it so that resolving a builtin module returns both the path, and the source and cache both of them on the module resolver?

lute/luau/src/resolverequire.cpp
```diff
@@ -12,11 +19,11 @@
 #include <optional>
 #include <string>
 
-// FileVfsContext
-class FileVfsContext : public Luau::Require::NavigationContext
+// LuteVfsContext
+class LuteVfsContext : public Luau::Require::NavigationContext
```
Vighnesh-V: Can you keep the existing FileVFSContext and introduce this as a separate struct?

lute/luau/src/resolverequire.cpp
```diff
@@ -30,7 +37,20 @@ class FileVfsContext : public Luau::Require::NavigationContext
     std::optional<std::string> getAlias(const std::string& alias) const override;
     std::optional<std::string> getConfig() const override;
 
-    FileVfs vfs;
+    FileVfs fileVfs;
+    StdLibVfs stdLibVfs;
+    LuteVfs luteVfs;
```
Vighnesh-V: This LuteVfs currently does nothing. The issue here is that most instances of LuteVfs don't do much except to validate that you can navigate to particular paths, and so you're just resetting to these paths and then completely separately getting the resources they contain. 

Can you introduce a LuteTypeCheckVfs that's a subtype of LuteVfs. This should just correct implement `getIdentifier()` and `getContents()`. To be clear, the reponsibility of these vfs's is to help in navigating requires, and to get the contents at those paths.

lute/luau/src/resolverequire.cpp
```diff
@@ -75,49 +95,137 @@ static NC::ConfigStatus convert(ConfigStatus status)
...
+    {
+        vfsType = VFSType::Batteries;
+        return convert(batteriesVfs.resetToPath(path));
+    }
+    else
+    {
+        vfsType = VFSType::Disk;
+        return convert(fileVfs.resetToPath(path));
+    }
 }
```
Vighnesh-V: Can you also implement toAliasOverride? We want to match the behaviour here to what the requirevfs does.

lute/luau/src/resolverequire.cpp
```diff
@@ -75,49 +95,137 @@ static NC::ConfigStatus convert(ConfigStatus status)
...
+std::optional<std::string> LuteVfsContext::getAlias(const std::string& alias) const
 {
+    if (alias == "std")
+        return "@std";
+    if (alias == "lute")
+        return "@lute";
+    if (alias == "batteries")
+        return "@batteries";
     return std::nullopt;
 }
```
Vighnesh-V: I don't think you need to implement this method.

lute/luau/src/resolverequire.cpp
```diff
@@ -75,49 +95,137 @@ static NC::ConfigStatus convert(ConfigStatus status)
...
+    switch (vfsType)
+    {
+    case VFSType::Disk:
+        return fileVfs.getConfig();
+    case VFSType::Std:
+        return stdLibVfs.getConfig();
+    case VFSType::Lute:
+        return luteVfs.getConfig();
+    case VFSType::Batteries:
+        return batteriesVfs.getConfig();
```
Vighnesh-V: none of these have any config since they're all embedded and their respective VFS's handle getting modules from fs-esque paths.  You can just return the fileVfs config here and replaces these with a default:block.

lute/cli/src/tc.cpp
Vighnesh-V: I think* this entire struct is no longer necessary. You'd be able to use the LuteModuleResolver in typechecking.

lute/luau/src/resolverequire.cpp
```diff
@@ -75,49 +95,137 @@ static NC::ConfigStatus convert(ConfigStatus status)
...
+    switch (vfsType)
+    {
+    case VFSType::Disk:
+        return convert(fileVfs.getConfigStatus());
+    case VFSType::Std:
+        return convert(stdLibVfs.getConfigStatus());
+    case VFSType::Lute:
+        return convert(luteVfs.getConfigStatus());
+    case VFSType::Batteries:
+        return convert(batteriesVfs.getConfigStatus());
```
Vighnesh-V: these can be collapsed to Absent.

Vighnesh-V (changes requested):
First off, thank you for contributing this, this is something we've been wanting to do. At a high level, here are the things that need to change:
* Can you introduce a separate LuteTypeCheckContext instead of modifying the FileVFSContext, since the resolveRequire api relies on getting filesystem paths.
*Can you introduce a separate 'resolveForTypeCheck` method that returns a pair of <path, source> that gets cached on the ModuleResolver?
* LuteVfs - can you introduce a subclass of LuteVfs for typechecking that returns the actual contents and paths for lute definitions.
* Testing: There are a couple tests you can write in C++ to help us know whether or not this is implemented correctly:
    *   Full end to end typecheck of a luau file that requires every file in the standard library and every file in the lute library.
    * individual calls out to the resolveRequire api from a variety of different paths. For example:
        * Requiring something in std from some file called main.luau.
        * Requiring @self/platform from @std/system, to check that the vfs knows how to go from system/init.luau -> system/platform.luau
        * Requires from std -> lute.
        * Requires from std -> batteries.
        * Requires from batteries -> std.

You'll probably also have to update the typecheck faillist while you're at it.

lute/luau/src/resolvemodule.cpp
```diff
@@ -142,7 +305,7 @@ std::optional<std::string> resolveModule(std::string requirePath, std::string re
         return std::nullopt;
     }
 
-    FileVfsContext context{requirerChunkname.substr(1)};
+    LuteTypeCheckContext context{requirerChunkname};
```
Vighnesh-V: I think this is leftover - can you revert just these changes to `resolveModule`, since the expectation for clients is that they get a file system path here.

lute/luau/src/resolvemodule.cpp
```diff
@@ -74,50 +180,107 @@ static NC::ConfigStatus convert(ConfigStatus status)
     return result;
 }
 
-FileVfsContext::FileVfsContext(std::string requirerChunkname)
-    : requirerChunkname(std::move(requirerChunkname))
+// LuteTypeCheckContext
```
Vighnesh-V: Not a useful comment, can you clip this?

lute/luau/src/resolvemodule.cpp
```diff
@@ -11,26 +18,125 @@
...
+            LuteModuleResult result = getLuteModule(identifier);
+            if (result.type == LuteModuleType::Module)
+                return std::string(result.contents);
+        }
+        else if (identifier.rfind("@batteries", 0) == 0)
+        {
+            BatteryModuleResult result = getBatteryModule(identifier);
+            if (result.type == BatteryModuleType::Module)
+                return std::string(result.contents);
+        }
```
Vighnesh-V, line 242: You should replace this section with checking the vfs type and then calling vfs.getSource(). For example:
```
case VFSType::Batteries:
    return batteriesVfs.readSource();
```

If you do this, you won't have to link `Lute.Lute Lute.Std Lute.Batteries` in this CMake. 
Remember, the `LuteTypeCheckVFSContext` is stateful - navigating it updates state , and the VFS tracks which module you're requiring as well as if it's a valid path.

lute/luau/src/resolvemodule.cpp
Vighnesh-V: That way you can also revert the changes needed to the `staticrequires` module. You should know that batteries requires are only used for the cli commands - we should be not exposing them to anyone.

lute/cli/src/staticrequires.cpp
```diff
@@ -110,10 +110,10 @@ void StaticRequireTracer::trace(const std::string& entryPoint)
 
         for (const auto& req : requiresInFile)
         {
-            // Skip warning for built-in libraries (@std and @lute)
+            // Skip warning for built-in libraries (@std, @lute, @batteries)
```
Vighnesh-V: See my comment below - resolveModule should only be handling file system paths.

lute/cli/src/tc.cpp
```diff
@@ -12,25 +12,16 @@ struct LuteFileResolver : Luau::LuteTypeCheckModuleResolver
 {
```
Vighnesh-V: Can you delete the LuteFileResolver struct? It basically does the same thing as the `LuteTypeCheckModuleResolver`. You can just use the LuteTypeCheckModuleResolver directly as an argument to the frontend.

lute/luau/include/lute/resolvemodule.h
```diff
@@ -5,5 +5,13 @@
 
 struct lua_State;
 
+struct ResolvedModule
+{
+    std::string path;
+    std::optional<std::string> source;
```
Vighnesh-V: This should just be a string - a require for typechecking purposes cannot resolve to a directory, that should be treated as an error.

lute/luau/src/resolvemodule.cpp
```diff
@@ -11,26 +18,125 @@
 #include <optional>
 #include <string>
 
-// FileVfsContext
-class FileVfsContext : public Luau::Require::NavigationContext
+// LuteTypeCheckVfs
+struct LuteTypeCheckVfs
```
Vighnesh-V: Can you please keep the FileVFSContext around? That should be what serves the existing resolveRequire/resolveModule api. This can be a separate VFS.

lute/luau/src/resolvemodule.cpp
```diff
@@ -74,50 +180,107 @@ static NC::ConfigStatus convert(ConfigStatus status)
...
+    std::optional<std::string> getAlias(const std::string& alias) const override
+    {
+        if (alias == "std")
+            return "@std";
+        if (alias == "lute")
+            return "@lute";
+        if (alias == "batteries")
+            return "@batteries";
+        return std::nullopt;
+    }
```
Vighnesh-V, line 309: You don't have to override this method.

lute/luau/src/resolvemodule.cpp
```diff
@@ -74,50 +180,107 @@ static NC::ConfigStatus convert(ConfigStatus status)
...
+        case LuteTypeCheckVfs::VFSType::Disk:
+            return vfs.fileVfs.getConfig();
+        case LuteTypeCheckVfs::VFSType::Std:
+            return vfs.stdLibVfs.getConfig();
+        case LuteTypeCheckVfs::VFSType::Lute:
+            return vfs.luteVfs.getConfig();
+        case LuteTypeCheckVfs::VFSType::Batteries:
+            return vfs.batteriesVfs.getConfig();
+        }
+        return std::nullopt;
```
Vighnesh-V: Of these cases, only disk needs to defer to the getConfig method. None of the other VFS's support this.

lute/luau/src/resolvemodule.cpp
```diff
@@ -11,26 +18,125 @@
...
+            StdLibModuleResult result = getStdLibModule(identifier);
+            if (result.type == StdLibModuleType::Module)
+                return std::string(result.contents);
+        }
+        else if (identifier.rfind("@lute", 0) == 0)
+        {
+            LuteModuleResult result = getLuteModule(identifier);
+            if (result.type == LuteModuleType::Module)
+                return std::string(result.contents);
+        }
```
Vighnesh-V: You'll need to move this logic into the LuteVfs.

lute/luau/CMakeLists.txt
Vighnesh-V: Once you change the implementation of `getSource` to call the appropriate vfs's getSource, you can remove linking these three in.

Vighnesh-V (changes requested):

lute/luau/src/resolvemodule.cpp
ActualMasterOogway: > You don't have to override this method.

It's a declared as virtual in [RequireNavigator.h](https://github.com/luau-lang/luau/blob/9e2984fd/Require/include/Luau/RequireNavigator.h#L97), so it has to be overridden

Vighnesh-V (approved):
approved, but I made a few tweaks myself to prevent user code from requiring batteries.

PR #863, opened 2026-03-09
Create `lute completion` command for shell tab-completion

### LINKED ISSUE #845: Shell Autocompletions
We should consider implementing shell autocompletions (at minimum for bash, ideally for fish as well) for both the embedded lute subcommands _and_ for discovery of and autocompletion of scripts located in `.lute`.

tyulpan:
Adds a `lute completion <shell> [options]` command to resolve #845. 
This command was inspired by a similar command in [gh](https://cli.github.com/manual/gh_completion) and other popular CLI tools.

Supported shells:
- `bash`
- `fish`
- `zsh`

Completions cover:
- all embedded `lute` subcommands
- per-command flags
- scripts located in `.lute` (.luau/.lua)

All supported shells have been tested manually.


tyulpan:
Updated the test file to match the new naming.

Refreshed the scripts to align with changes from the past 1.5 months:
- PR #888
- PR #936
- PR #950
- PR #1033

Hopefully, nothing was overlooked.

Nicell:
One thing that comes to mind when reviewing this is now there's more things to keep up to date when we update the CLI. Our help command is already out of date IIRC. I wonder if we should have more of a generator pattern for creating things like help messages and completions to avoid incorrect/stale information that doesn't match what the commands actually do.

I don't think that needs to block this PR but just thinking about it and this PR is related.

lute/cli/commands/completion/init.luau
Nicell, line 1: could this file benefit from @batteries/cli? might be good for consistency and help with flag parsing to be more forgiving to the end user

docs/cli/completion.md
```diff
@@ -0,0 +1,66 @@
...
+
+```
+lute completion <shell> [options]
+```
+
+## Supported shells
+
+### - `bash`
+### - `fish`
+### - `zsh`
```
Nicell: this renders kinda weird 🤔 should we drop the `-` or the `###`? leaning towards dropping `###`

lute/cli/src/climain.cpp
```diff
@@ -77,6 +78,10 @@ Lint Options:
 			--rules <path>          Path to a single lint rule or a directory containing multiple lint rules.
 			                        If not specified, default lint rules are used.
 
+Completion Options:
+	lute completion <shell> [options]
+		Generates a completion script for the specified shell.
```
Nicell, line 83: should we include the supported shells here?

tyulpan:
>One thing that comes to mind when reviewing this is now there's more things to keep up to date when we update the CLI. Our help command is already out of date IIRC. I wonder if we should have more of a generator pattern for creating things like help messages and completions to avoid incorrect/stale information that doesn't match what the commands actually do.

Agreed. It would be useful to create a `spec.luau` file for each command. 
It could return a simple `CommandSpec` table defined in `@batteries/cli` and serve as the single source of truth for the command.

Something like:
```luau
export type CommandSpec = {
      name: string,
      summary: string,
      description: string?,
      flags: { FlagSpec }?,
      options: { OptionSpec }?,
      subcommands: { CommandSpec }?, 
      examples: { string }?,
  }
  ```
  
  I think this would simplify CLI updates and maintenance.

PR #864, opened 2026-03-09
Modifies `assert.strcontains` so that specifying 1 instance asserts for exactly 1 instance

skberkeley:
Previous behavior was to check for at least one instance of the specified substring, which was a bit misleading.


lute/std/libs/test/assert.luau
```diff
@@ -139,21 +139,24 @@ local function erroreq<A...>(func: (A...) -> ...unknown, expectedErrorMessage: s
 end
 
 local function strcontains(haystack: string, needle: string, msg: string?, instances: number?): failure?
```
Vighnesh-V: Can you swap msg? and instances? while you're here?

Vighnesh-V (changes requested):

lute/std/libs/test/assert.luau
skberkeley: I thought about it but It'd be a lot bigger of a refactor so I opted not to

Vighnesh-V: Actually, there aren't that many assert.strcontains instances that pass both msg and number - I count like maybe 5 ish?

skberkeley: Done!

Vighnesh-V (approved):

PR #865, opened 2026-03-09
Fixes/silences type errors in the syntax library

skberkeley:
A lot of these were code too complex errors (most likely in Normalization based on a previous example I minimized) that I had to silence with any casts.


lute/std/libs/syntax/utils/init.luau
```diff
@@ -230,12 +230,12 @@ function utils.isTypePack(n: types.AstNode): types.AstTypePack?
...
-	return if n.istoken then n else nil
+function utils.isToken(n: types.AstNode | types.Token): types.Token?
+	return if (n :: types.Token).istoken then n :: types.Token else nil
 end
 
-function utils.isBaseToken(n: types.AstNode): types.Token?
-	return if n.istoken and not n.kind then n else nil
+function utils.isBaseToken(n: types.AstNode | types.Token): types.Token?
+	return if (n :: types.Token).istoken and not (n :: any).kind then n :: types.Token else nil -- LUAUFIX: use (n :: types.AstNode) instead of (n :: any) once it no longer triggers code too complex
 end
```
hgoldstein, line 235: Is there something we could be doing better here? It feels weird to have all of these casts to token. If _we_ need to cast to `types.Token` all over the place then I can imagine a consumer having to do similar casting, fighting against the type system, etc.

skberkeley: I think the cleanest thing might be to refactor AST nodes which are also tokens (e.g. literals) so that they have a token field instead of being intersected with `Token` like they currently are. I think this what you mentioned Hack does? That's a bit more of a nontrivial refactor though, and should probably be a separate PR

lute/std/libs/syntax/utils/trivia.luau
```diff
@@ -4,9 +4,12 @@ local types = require("../types")
 local retrieverLib = {}
 
 function retrieverLib.leftmosttrivia(n: types.AstNode): { types.Trivia }
-	local q: query.query<types.Token> = query.findallfromroot(n, function(n)
-		return if n.istoken ~= nil and n.istoken then n else nil
-	end)
+	local q: query.query<types.Token> = query.findallfromroot(
+		n :: any,
```
Vighnesh-V: do you know why this particular issue happens? If possible, could you get a small repro using `https://play.luau.org/`.

lute/std/libs/syntax/utils/trivia.luau
```diff
@@ -27,9 +30,12 @@ function retrieverLib.leftmosttrivia(n: types.AstNode): { types.Trivia }
...
 function retrieverLib.rightmosttrivia(n: types.AstNode): { types.Trivia }
-	local q: query.query<types.Token> = query.findallfromroot(n, function(n)
-		return if n.istoken ~= nil and n.istoken then n else nil
-	end)
+	local q: query.query<types.Token> = query.findallfromroot(
+		n :: any,
+		function(n: types.Token): types.Token?
+			return if n.istoken ~= nil and n.istoken then n else nil
+		end :: any
+	) -- LUAUFIX: Remove any casts once query.findallfromroot no longer emits code too complex
```
Vighnesh-V: ditto

lute/std/libs/syntax/printer.luau
```diff
@@ -38,8 +38,8 @@ local function printTriviaList(self: PrintVisitor, trivia: { types.Trivia })
 end
 
 local function printToken(self: PrintVisitor, token: types.Token)
-	if self.replacements[token] ~= nil then
-		self:printReplacement(token, self.replacements[token])
+	if self.replacements[token :: types.AstNode] ~= nil then
+		self:printReplacement(token :: types.AstNode, self.replacements[token :: types.AstNode])
```
Vighnesh-V: I recently fixed an issue involving comparing indexer keys to nil - could you see if this repros still? It might be that we don't have to do these casts.

lute/std/libs/syntax/printer.luau
```diff
@@ -106,9 +106,8 @@ local function printReplacement(self: PrintVisitor, node: types.AstNode, replace
 
 	if typeof(replacement) == "string" then
 		self:write(replacement)
-	elseif replacement.kind ~= nil then -- LUAUFIX: type solver upset comparing string singleton union to nil
-		-- TODO: preserve trivia
-		visitor.visit(replacement, self)
+	elseif replacement.kind ~= nil then
```
Vighnesh-V, line 109: ditto here.

lute/std/libs/syntax/utils/trivia.luau
skberkeley: spoke offline, it's most likely an issue in normalization (and possibly other subsystems)

lute/std/libs/syntax/printer.luau
skberkeley: I don't think it would make a difference; `replacements` has type `{ [ AstNode ] : replacement }`, and `Token` isn't a subtype of `AstNode`. I think this should actually be fixed by pushing tokens as always being fields on AstNodes, so I might move this PR into draft and rebase

Vighnesh-V (approved):

PR #872, opened 2026-03-11
Updates AstNode types so that they are never also Tokens

skberkeley:
I also added `Token` to the union which defines `AstNode`, and changed its `istoken` field to `kind = "token"`, as this plays nicer with how the type system interacts with tagged unions. Overall, I feel that this aligns our AST types to play nicer with the type system.


lute/cli/commands/lint/rules/unused_variable.luau
```diff
@@ -29,18 +29,25 @@ local function getWriteOnlyArgs(func: syntax.AstExpr, writeOnlyAPIs: smallTrie):
 		end
 
 		return entry.vals
-	elseif
-		(func.tag == "indexname" and func.expression.tag == "global")
-		or (func.tag == "index" and func.expression.tag == "global" and func.index.tag == "string") -- LUAUFIX: Refinement bug here: func should be narrowed to AstExprIndexExpr (CLI-190889)
```
hgoldstein, line 34: We should probably keep this FIXME comment around.

lute/luau/src/luau.cpp
```diff
@@ -837,34 +840,51 @@ struct AstSerialize : public Luau::AstVisitor
 
     void serialize(Luau::AstExprConstantNil* node)
     {
-        serializeToken(node->location.begin, "nil", preambleSize);
```
hgoldstein, line 840: I think I am a bit confused: would it make more sense to change `serializeToken` to do all of this work instead? e.g. assume there's some table on the stack, and do the work needed to set the member named `token` to the correct kind of token?

hgoldstein (dismissed):
Mainly my comment in `luau.cpp`

lute/luau/src/luau.cpp
skberkeley: I don't think so, the vast majority of `serializeToken` calls happen for keywords like `for` or `function` which aren't assigned to a `token` field. I'll go back and remove the table size argument from `serializeToken` too, since we're never making tokens with extra fields anymore.

lute/cli/commands/lint/rules/unused_variable.luau
skberkeley: Oh, I got rid of it because the original code was no longer erroring. The hovertypes are still pretty unhelpful but I think that's a separate problem

Vighnesh-V (approved):

PR #874, opened 2026-03-12
Fixes the task.spawn, task.defer, and task.resume methods

### LINKED ISSUE #522: Result of `task.spawn` cannot be used with `coroutine.status`
The return value of `task.spawn` doesn't actually seem to return a `thread`. At least not in a way that `coroutine.status` recognizes. 

When attempting to pass the thread in, `coroutine.status` errors with `invalid argument #1 to 'status' (thread expected, got function)`

Minimal repro:

```luau
-- scripts/repro.luau

local thread = task.spawn(function()
	print("do something for a while")
	task.wait(2)
	print("done")
end)

local status = coroutine.status(thread)

print(status)
```

And here's the output:

```
➜  flipbook git:(main) ✗ lute --version         
0.1.0-nightly.20251104
➜  flipbook git:(main) ✗ lute scripts/repro.luau
do something for a while
./scripts/repro.luau:9: invalid argument #1 to 'status' (thread expected, got function)
stacktrace:
[C] function status
./scripts/repro.luau:9
done
```

Vighnesh-V:
Resolves #522 

This PR refactors `task.spawn, defer, resume` to share most of their logic for creating the stack and handing control off to newly created threads. The cause of the bug in #522 was that the stack didn't get set up correctly in the parent stack due to `lua_xmove` popping from the top, so the value that got returned was not callable with `coroutine.status` (because it wasn't a thread). The code I've added documents the invariant that I wanted to maintain along with helpers for correctly setting up the stack.

I've added tests for the following cases:
{task.defer, task.spawn, task.resume } called with { thread | function (except for task.resume, which only takes a thread passed { 0 arguments to resume with, > 0 arguments to resume with }.

I also introduced 'deferSelf' for the argumentless version of task.defer that says to immediately yield the current thread.

### DIFF
--- a/lute/task/src/task.cpp
+++ b/lute/task/src/task.cpp
@@ -108,7 +108,7 @@ struct LuaThread
         int toPush = lua_gettop(parent);
         // We only want to resume with the actual number of arguments here, regardless of if the first argument is a thread or a function
         int numResumeArgs = toPush > 1 ? toPush - 1 : 0;
-        if (!lua_checkstack(L, 1))
+        if (!lua_checkstack(parent, 1))
             luaL_error(L, "Not enough stack space to create a new thread");
         if (lua_isfunction(parent, 1))
         {
@@ -163,7 +163,7 @@ struct LuaThread
         if (!ok)
         {
             runtime->reportError(child);
-            return 1;
+	    return 1;
         }
 
         return 1;
--- a/lute/task/src/task.cpp
+++ b/lute/task/src/task.cpp
@@ -94,7 +94,7 @@ namespace task
 
 struct LuaThread
 {
-    LuaThread(lua_State* L)
+    LuaThread(lua_State* L, std::string_view caller = "")
         : parent(L)
         , runtime(getRuntime(parent))
     {
@@ -123,7 +123,7 @@ struct LuaThread
         }
         else
         {
-            luaL_error(parent, "Expected a thread or function as the first argument.");
+            luaL_error(parent, "%s: Expected a thread or function as the first argument.", caller.data());
         }
         this->resumeArgs = numResumeArgs;
 
@@ -163,7 +163,7 @@ struct LuaThread
         if (!ok)
         {
             runtime->reportError(child);
-	    return 1;
+            return 1;
         }
 
         return 1;
@@ -184,7 +184,7 @@ private:
 
 int lua_spawn(lua_State* L)
 {
-    LuaThread newThread{L};
+    LuaThread newThread{L, "task.spawn"};
     return newThread.resume();
 }
 
@@ -196,7 +196,7 @@ int lua_deferSelf(lua_State* L)
         return 0;
     }
     lua_pushthread(L);
-    LuaThread newThread{L};
+    LuaThread newThread{L, "task.deferSelf"};
     newThread.defer();
     return lua_yield(L, 0);
 }
@@ -204,7 +204,7 @@ int lua_deferSelf(lua_State* L)
 int lua_defer(lua_State* L)
 {
 
-    LuaThread newThread{L};
+    LuaThread newThread{L, "task.defer"};
     return newThread.defer();
 }
 
@@ -215,7 +215,7 @@ int lute_resume(lua_State* L)
         luaL_error(L, "Expected a thread as the first argument.");
         return 0;
     }
-    LuaThread newThread{L};
+    LuaThread newThread{L, "task.resume"};
     return newThread.resume();
 }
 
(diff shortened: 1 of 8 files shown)


lute/task/src/task.cpp
```diff
@@ -88,14 +91,134 @@ static void yieldLuaStateFor(lua_State* L, uint64_t milliseconds, bool putDeltaT
...
+        // [function, args...]
+        // or
+        // [thread, args...]
+
+        // If passed a thread, we push the 0 or more arguments onto that threads stack and resume it.
+        // If passed a function, we push the function and the 0 or more arguments onto a newly created thread's stack
+        int toPush = lua_gettop(parent);
+        // We only want to resume with the actual number of arguments here, regardless of if the first argument is a thread or a function
+        int numResumeArgs = toPush > 1 ? toPush - 1 : 0;
+        if (!lua_checkstack(L, 1))
```
annieetang: should this checkstack be for `parent`

lute/task/src/task.cpp
```diff
@@ -88,14 +91,134 @@ static void yieldLuaStateFor(lua_State* L, uint64_t milliseconds, bool putDeltaT
...
+            return 1;
+        }
+
+        int resumeStatus = lua_resume(child, parent, resumeArgs);
+        bool ok = resumeStatus == LUA_OK || resumeStatus == LUA_YIELD || resumeStatus == LUA_BREAK;
+
+        if (!ok)
+        {
+            runtime->reportError(child);
+            return 1;
```
annieetang, line 166:
```suggestion

```
since we error, (and also `return 1` later anyways)

tests/lute/task.test.luau
```diff
@@ -22,4 +22,119 @@ test.suite("LuteTaskSuite", function(suite)
...
+		local thread = task.spawn(function()
+			task.wait(2)
+		end)
+
+		local status = coroutine.status(thread)
+		assert.eq(status, "suspended")
+		coroutine.close(thread)
+	end)
+
+	-- task.spawn with thread, 0 args
```
annieetang, line 48: is it possible to call `task.spawn()` / what happens if that happens?

annieetang (commented):
thx for fixing these!

tests/lute/task.test.luau
Vighnesh-V: The logic for constructing a thread expects the first thing on the stack to be either a thread or a function. If it's neither, an error is raised with the following message:
```
"Expected a thread or function as the first argument."
```
which models the type of the functions in the typedefs (you have to pass a thread / function)

lute/task/src/task.cpp
Vighnesh-V: L and parent are identical here, so this ends up working. I've switched this to parent!

lute/task/src/task.cpp
Vighnesh-V: runtime->reportError doesn't actually throw an error - it just prints to stderr. That being said, the fallthrough is alright here.

tests/lute/task.test.luau
annieetang: sounds good! do we want to add a test case showing this or nah

lute/task/src/task.cpp
```diff
@@ -88,14 +91,134 @@ static void yieldLuaStateFor(lua_State* L, uint64_t milliseconds, bool putDeltaT
...
+        }
+
+        int resumeStatus = lua_resume(child, parent, resumeArgs);
+        bool ok = resumeStatus == LUA_OK || resumeStatus == LUA_YIELD || resumeStatus == LUA_BREAK;
+
+        if (!ok)
+        {
+            runtime->reportError(child);
+	    return 1;
```
annieetang:
```suggestion

```

annieetang: or leave but add an indent for formatting?

annieetang (commented):
last thing- do we want to specify the error messages a bit more?

`luaL_error(L, "Expected a thread as the first argument.");` -> `luaL_error(L, "Lute resume: Expected a thread as the first argument.");` or something specific? 

since some of the error messages specify which function the error is coming from (e.g., `luaL_error(L, "task.deferSelf does not take any arguments");`

lute/task/src/task.cpp
Vighnesh-V: I'll fix the formatting :)

Vighnesh-V:
~~I think inside of `LuauThread` it's fine to just warn with `expected first argument to be thread`.~~

Edit: I've updated LuaThread to take a caller parameter, so we can provide slightly better error messages.

annieetang (approved):
thanks!!

PR #876, opened 2026-03-12
Renames `resolverequire` to `resolveModule` and return `path` instead of `string`

annieetang:
journey to shipping Lute V1


lute/luau/src/tcmoduleresolver.cpp
annieetang: hm ok so I couldn't rename this `resolveRequire` function to `resolveModule` because it conflicts with the `resolveModule` definition- lmk if this is fine or if we should figure out different namings

Vighnesh-V: Like it's calling itself recursively?

Vighnesh-V: You should be able to do Luau::resolveModule or ::resolveModule should work

annieetang: if i rename to `resolveModule` it thinks its calling itself recursively

Vighnesh-V (changes requested):

lute/luau/src/tcmoduleresolver.cpp
Vighnesh-V: If you write `::resolveModule` instead of just `resolveModule`, the compiler knows you're looking this file up in the global namespace.

annieetang: oh bet ty

Vighnesh-V (approved):

PR #878, opened 2026-03-13
Adds `getAllRequires` to `@std/luau`

### LINKED ISSUE #776: There should be a standard library method for 'get all the required modules for a given module'
Right now, folks have to write their own version of this analysis that looks like:
```
function dependson(src : string)
    local mod = parse(src)
    local requires = query(mod, function(req). return resolveRequire(req.path) end)
   return requires
end
```

We could just provide this for folks to use out of the box.

tyulpan:
Adds `getAllRequires(src: string): { string }` to `@std/luau`, which parses Luau source code and returns all of its required modules.
Resolves #776.

Now we can run this code:
```luau

local code = [=[
  local crypto = require("@lute/crypto")
  local base64 = require("@batteries/base64")

  local message = "Hello, world!"
  local digest = crypto.digest(crypto.hash.sha256, message)

  function hexify(digest)
    local hex = {}
    for i = 1, buffer.len(digest) do
      hex[i] = string.format("%02x", buffer.readu8(digest, i - 1))
    end
    return table.concat(hex)
  end

  function pretty(digest)
    return buffer.tostring(base64.encode(digest))
  end

  print(hexify(digest))
  print(pretty(digest))
]=]

local requires = luau.getAllRequires(code)

print(pp(requires))
```

and get:
```console
{
  "@lute/crypto",
  "@batteries/base64",
}
```

### DIFF
--- a/lute/std/libs/luau.luau
+++ b/lute/std/libs/luau.luau
@@ -7,6 +7,9 @@ local luteLuau = require("@lute/luau")
 
 local fs = require("@lute/fs")
 local path = require("@std/path")
+local syntax = require("@std/syntax")
+local utils = require("@std/syntax/utils")
+local query = require("@std/syntax/query")
 
 local luau = {}
 
@@ -43,6 +46,28 @@ end
 function luau.resolveModule(modulepath: string, frompath: path.pathlike): path.pathlike
 	local frompathString = if typeof(frompath) == "string" then frompath else path.format(frompath)
 	return luteLuau.resolveModule(modulepath, `@{frompathString}`)
+end
+
+function luau.getAllRequires(src: string): { string }
+	local ast = luteLuau.parse(src)
+
+	local requireCalls = query.findallfromroot(ast, utils.isExprCall):filter(function(call)
+		return call.func.tag == "global" and (call.func :: syntax.AstExprGlobal).name.text == "require"
+	end)
+
+	return requireCalls:maptoarray(function(call: syntax.AstExprCall): string?
+		local firstArg = call.arguments[1]
+		if firstArg == nil then
+			return nil
+		end
+
+		local argNode = firstArg.node
+		if argNode.tag ~= "string" then
+			return nil
+		end
+
+		return (argNode :: syntax.AstExprConstantString).text
+	end)
 end
 
 return table.freeze(luau)
--- a/tests/std/luau.test.luau
+++ b/tests/std/luau.test.luau
@@ -161,3 +161,54 @@ test.suite("LuauTypeOfModuleSuite", function(suite)
 		assert.eq(resolvedPath, expected)
 	end)
 end)
+
+test.suite("LuauGetAllRequiresSuite", function(suite)
+	suite:case("getAllRequiresBasic", function(assert)
+		local requires = luau.getAllRequires([[
+			local crypto = require("@lute/crypto")
+			local base64 = require("@batteries/base64")
+			print("hello")
+		]])
+		assert.eq(#requires, 2)
+		assert.eq(requires[1], "@lute/crypto")
+		assert.eq(requires[2], "@batteries/base64")
+	end)
+
+	suite:case("getAllRequiresEmpty", function(assert)
+		local requires = luau.getAllRequires([[
+			local hello = 1
+			print("world")
+		]])
+		assert.eq(#requires, 0)
+	end)
+
+	suite:case("getAllRequiresNonStringArg", function(assert)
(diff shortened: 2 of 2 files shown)


Vighnesh-V:
Hi. Thanks for your contribution! I want to ask what you thought about making this api return the fully resolved path to the require as well, in addition to the value passed to `require()`? The existing require tracer module does exactly this. Moreover, what should this do in the presence of dynamic requires?

tyulpan:
>I want to ask what you thought about making this api return the fully resolved path to the require as well, in addition to the value passed to require()?
 
Users can combine `getAllRequires` with the existing [`resolveModule`](https://lute.luau.org/reference/lute/luau.html#luau-resolvemodule) method to return resolved paths.

Something like:
```luau
for _, req in luau.getAllRequires(src) do
    local resolved = luau.resolveModule(req, filePath)
end
```

I think this behavior is more universal and reliable. 
Besides, there is no need to duplicate the logic from `StaticRequireTracer`.


>Moreover, what should this do in the presence of dynamic requires?

We just skip them([line](https://github.com/luau-lang/lute/pull/878/changes#diff-c9b899612b9d42fe9d8279d75b7472e1da60bc2596a1b1161dfa926af79b6ceaR69)), same as the `StaticRequireTracer`([line](https://github.com/luau-lang/lute/blob/6c347614b0d4755800811dd3203134706c9f2f4f/lute/cli/src/staticrequires.cpp#L28)) does. There is no way to resolve these without runtime evaluation.

lute/std/libs/luau.luau
```diff
@@ -45,4 +48,26 @@ function luau.resolveModule(modulepath: string, frompath: path.pathlike): path.p
 	return luteLuau.resolveModule(modulepath, `@{frompathString}`)
 end
 
+function luau.getAllRequires(src: string): { string }
+	local ast = luteLuau.parse(src)
+
+	local requireCalls = query.findallfromroot(ast, utils.isExprCall):filter(function(call)
+		return call.func.tag == "global" and (call.func :: syntax.AstExprGlobal).name.text == "require"
+	end)
```
Vighnesh-V, line 56: @skanosue do you have an opinion about making a `query.findrequirer` api? Other folks have asked for this as well for custom analyses passes.

skberkeley: I think it would be more appropriate in `utils`, something like `query.findallfromroot(ast, utils.isRequireCall)`. Unless I'm misunderstanding what you mean by a "requirer"?

Vighnesh-V:
I've been doing some thinking and I've decided to close this PR. This feature is useful, however, I don't think it's complete on it's own. As @skberkeley pointed out, this is a pretty short one-liner. If you wanted to write a code mod that operated on requires, it would be pretty easy to do. I suspect that most situations in which you wanted access to just the string arguments to `require`, would be quickly followed up with a request to resolveRequire, followed up with a module graph traversal. 

I think what we'd want here is to expose the module graph directly as a traversal. That means via an iterator, you get something like `{ requireContent: "@std/x/y/z", resolved: "/path/to/resolved/require", source: "/path/to/requirer", require: AstExpr }`. This is more general, and probably more useful for tooling writers.

I hope my reasoning makes sense! Thank you for this contribution.

tyulpan:
@Vighnesh-V The reasoning is perfectly clear. What is needed at the end is a structure for a more in-depth analysis. 

One thing remains unclear: how can I currently extract a table of arguments from all requires?
How can I do this with short one-liner?

I propose the following refactoring plan:
1. implement `utils.isRequireCall` in a separate `PR`.
2. replace `local requireCalls = query.findallfromroot(ast, utils.isExprCall):filter(...)` with `local requireCalls = query.findallfromroot(ast, utils.isRequireCall)`.
3. return all arguments, not just the string ones.
4. rename the function from `getAllRequires` to `getRequiresContent`.

It will be a pretty useful `helper` that spares users the trouble of parsing source code into an AST, running queries against it, and then doing additional mapping and filtering.

Vighnesh-V:
Hi @tyulpan , I think step 1 is perfectly reasonable, and I've approved the change already. What I am proposing instead of what you've suggested is a method that lets you traverse the module graph. So for example:
```file.luau
local x = require("@std/path")
local y = require("
```
```
for _, mod in luau.requires("path/to/file.luau") do
    --[[ mod : { 
           requireStr: -- the literal string passed to require
           resolved: -- the path to the resolved require
           require: -- the ast expression passed to require
    }
    ]]--
end

```

PR #882, opened 2026-03-14
Added path normalization in `check.luau`

### LINKED ISSUE #881: `tools/check.luau` does not support Windows
The issue is that paths aren't normalized, causing the tool to work improperly on Windows

#882

ActualMasterOogway:
Normalizing all paths in `tools/check.luau`, fixing issues on Windows

Fixes #881

Please add the `infra` tag as I am unable to

### DIFF
--- a/tools/check.luau
+++ b/tools/check.luau
@@ -23,9 +23,10 @@ local EXCLUDE_TYPECHECK = {
 	"tests/std/snapshots",
 }
 
-local function normalizePath(p: string): string
-	local normalized = string.gsub(p, "\\", "/")
-	return normalized
+local function posixify(p: path.pathlike): string
+	local normalized = path.normalize(p)
+	local posixy, _ = string.gsub(path.format(normalized), "\\", "/")
+	return posixy
 end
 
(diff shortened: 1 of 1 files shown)


Vighnesh-V (approved):

tools/check.luau
```diff
@@ -23,12 +23,17 @@ local EXCLUDE_TYPECHECK = {
 	"tests/std/snapshots",
 }
 
+local function normalizePath(p: string): string
+	local normalized = string.gsub(p, "\\", "/")
+	return normalized
+end
```
Vighnesh-V: There should be a path.normalize method, unless you meant standardizing on forward slash?

ActualMasterOogway: Yes, I did mean that. I've renamed the function to "posixify" in accordance with https://github.com/luau-lang/lute/blob/ed6e075979b85c4fd91f7648ac9a11a041a89e4c/lute/cli/commands/test/snap/runner.luau#L16 in the latest commit. Hope this clears everything up

Vighnesh-V: thank you, that makes sense.

PR #883, opened 2026-03-17
Use the new `math.nan` constant instead of `0 / 0`

vrn-sn:
Updates all existing uses of `0 / 0` in our codebase to `math.nan`, updates the suggested fix during linting, and adds a new lint warning for new uses of `0 / 0`.

### DIFF
--- a/examples/badisnan.luau
+++ b/examples/badisnan.luau
@@ -1,4 +1,0 @@
-return function(n)
-	--lute-lint-ignore(divide_by_zero)
-	return n == 0 / 0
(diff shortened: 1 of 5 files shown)


examples/badisnan.luau
```diff
@@ -1,4 +0,0 @@
-return function(n)
```
vrn-sn, line 1: This file doesn't seem to have much value, so I figured I'd remove it.

skberkeley (approved):

PR #885, opened 2026-03-18
Adds a hello world section to the guide/ for Lute

Vighnesh-V:
This is an introductory bit on how to write a simple hello world program in Lute. It also walks through the directory structure required to setup a project, which is useful in subsequent chapters.


annieetang (approved):
ty!

PR #886, opened 2026-03-18
Adds a simple guessing game tutorial to the guides/ section

Vighnesh-V:
This tutorial walks a user through some basics of writing programs with Luau:
1. How do you require standard library functionality
2. The difference between luau provided functions and lute provided functions
3. Writing a program that talks to the outside world
4. Input validation
5. Command line arguments in Luau

### DIFF
--- a/docs/guide/writing-a-guessing-game.md
+++ b/docs/guide/writing-a-guessing-game.md
@@ -37,9 +37,9 @@ hello
 ```
 
 ### Generating random numbers with Luau builtins
-Let's move on to implementing this guessing game. We need to generate a random number, whcih Luau already provides via the `random` function on the `math` table. 
+Let's move on to implementing this guessing game. We need to generate a random number, which Luau already provides via the `random` function on the `math` table. 
 
-This highlights an important distinction between Luau and Lute - **Luau** provides builtin functions that operate on the primitive types, e.g. `string`, `math`, `buffer`, etc, while **Lute** provides utilities for interacting with the outside world (for example, user input, the file system, http requests). In this case, the `math.random` function is provided by Luau and will suffice for our purposes. We'll set a limit of 100, so that math.random will not generate numbers that are too small.
+This highlights an important distinction between Luau and Lute - **Luau** provides builtin functions that operate on the primitive types, e.g. `string`, `math`, `buffer`, etc, while **Lute** provides utilities for interacting with the outside world e.g. user input, the file system, http requests. In this case, the `math.random` function is provided by Luau and will suffice for our purposes. We'll set a limit of 100, so that math.random will not generate numbers that are too small.
 
 ```luau
 local io = require("@std/io")
--- a/docs/guide/writing-a-guessing-game.md
+++ b/docs/guide/writing-a-guessing-game.md
@@ -8,28 +8,28 @@ order: 4
 We're going to walk through how you might build a small guessing game using Lute. Like we did in the previous chapter, make a new folder
 for this project, along with a script called `main.luau`.
 
-This is a program that will take a users input and compare it to a secret number that the program will generate. The user will then be prompted to input their guess,
-and the program will tell them "higher" or "lower" until they guess right.
+This is a program that will take a user's input and compare it to a secret number that the program will generate. The user will then be prompted to input their guess,
+and the program will tell them "Too high!" or "Too low!" until they guess right.
 
 
 ### Using the Lute standard library
-To start, we'll need some way of getting user input. Thankfully, Lute provides a utility for this in the `io` library. In `main.luau`, add:
+To start, we'll need some way of getting user input. Thankfully, Lute provides a utility for this in the [`io`](../reference/std/io) library. In `main.luau`, add:
 ```luau
-local iolib = require("@std/io")
+local io = require("@std/io")
 ```
 
 This is a require statement - the `@std` is an alias, and `io` is the name of the library. The `@std` and `@lute` aliases, have been reserved by Lute to refer to the standard library and Lute internal libraries respectively. You can `require` any set of scripts that return values, even ones you write locally.
 This module returns a table with one field - `input`, which waits for the user to type something into the terminal. We can assign this table a value so we can refer to it later.
 
 ```luau
-local iolib = require("@std/io")
+local io = require("@std/io")
 
 print("Say something! ")
-local input : string = iolib.input()
+local input : string = io.input()
 print(input)
 ```
 
-If you run this program with `lute main.luau` the program will pause until you input something, and then print that back out for you. For example:
+Now, run this with `lute main.luau`. The program will pause until you input something, then print it back to you:
 ```bash
 lute main.luau
 >>> Say something! hello
@@ -37,26 +37,33 @@ hello
 ```
 
 ### Generating random numbers with Luau builtins
-Let's move on to implementing this guessing game. We need to generate a random number, and Luau already provides this functionality for us, under the `math` table, which has a `random` function. This demonstrates a difference between Luau and Lute - Luau has builtins functions that operate on the primitive types, e.g. `string`, `math`, `buffer`, etc, but things like talking to your operating system or other programs must be handled by utilities that Lute provides. In this case, the `math.random` function is provided by Luau and will suffice for our purposes. We'll set a limit of 100, so that math.random will not generate numbers that are too small.
+Let's move on to implementing this guessing game. We need to generate a random number, whcih Luau already provides via the `random` function on the `math` table. 
+
+This highlights an important distinction between Luau and Lute - **Luau** provides builtin functions that operate on the primitive types, e.g. `string`, `math`, `buffer`, etc, while **Lute** provides utilities for interacting with the outside world (for example, user input, the file system, http requests). In this case, the `math.random` function is provided by Luau and will suffice for our purposes. We'll set a limit of 100, so that math.random will not generate numbers that are too small.
 
 ```luau
-local iolib = require("@std/io")
+local io = require("@std/io")
 local randomNumber : number = math.random(100)
 
 print("Guess a number: ")
-local input : string = iolib.input()
--- What do we do here?
+local input : string = io.input()
+local guess = tonumber(input)
 ```
 
-So now we have the guessed value - we should try to let the user know if they got it right or wrong. However, the output of `input()` is a string, and we want to compare it to a number. If you try to ask `x \< randomNumber`, this won't work correctly. We'll want to convert this using the `tonumber` function.
+So now we have the guessed value - we should try to let the user know if they got it right or wrong. 
+
+::: info
+The output of `input()` is a string, and we want to compare it to a number. If you try to ask `x \< randomNumber`, this won't work correctly. We'll want to convert this using the [`tonumber`](https://www.lua.org/manual/5.1/manual.html#pdf-tonumber) function.
+:::
+
 
 ```luau
-local iolib = require("@std/io")
+local io = require("@std/io")
 local randomNumber = math.random(100)
 
 print("Guess a number: ")
-local input : string = iolib.input()
-local guess = tonumber(input)
+local input : string = io.input()
+
 
 if guess > randomNumber then
     print("Too high!")
@@ -69,14 +76,16 @@ end
 
 So far so good! Ideally though, the user would be able to guess multiple times - in this example, the user has one chance. Let's even the odds a bit - we'll need to re-prompt the user if they got it wrong and exit out if they guess right.
 
+### Adding a Loop
+
 ```luau
-local iolib = require("@std/io")
+local io = require("@std/io")
 local randomNumber = math.random(100)
 
 local guessedRight = false
 while not guessedRight do
     print("Guess a number: ")
-    local input = iolib.input()
+    local input = io.input()
     local guess = tonumber(input)
     if guess > randomNumber then
         print("Too high!")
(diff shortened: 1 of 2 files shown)


docs/guide/writing-a-guessing-game.md
```diff
@@ -0,0 +1,203 @@
...
+for this project, along with a script called `main.luau`.
+
+This is a program that will take a users input and compare it to a secret number that the program will generate. The user will then be prompted to input their guess,
+and the program will tell them "higher" or "lower" until they guess right.
+
+
+### Using the Lute standard library
+To start, we'll need some way of getting user input. Thankfully, Lute provides a utility for this in the `io` library. In `main.luau`, add:
+```luau
+local iolib = require("@std/io")
```
annieetang: do we want our suggested naming convention to be `iolib = require("@std/io")` or just `io = require("@std/io")`? asking bc our [references page](https://lute.luau.org/reference/lute/io.html) has the 'suggested' require as just the module name (`io`) and I think it would be good to keep consistent

docs/guide/writing-a-guessing-game.md
```diff
@@ -0,0 +1,203 @@
...
+### Introduction
+We're going to walk through how you might build a small guessing game using Lute. Like we did in the previous chapter, make a new folder
+for this project, along with a script called `main.luau`.
+
+This is a program that will take a users input and compare it to a secret number that the program will generate. The user will then be prompted to input their guess,
+and the program will tell them "higher" or "lower" until they guess right.
+
+
+### Using the Lute standard library
+To start, we'll need some way of getting user input. Thankfully, Lute provides a utility for this in the `io` library. In `main.luau`, add:
```
annieetang: can you link the reference pages for the libraries/modules mentioned?

https://lute.luau.org/reference/lute/io.html

docs/guide/writing-a-guessing-game.md
```diff
@@ -0,0 +1,203 @@
...
+        print("Got it!")
+        guessedRight = true
+    end
+end
+```
+
+You can see that here, we've added a variable and a loop to check whether or not the user got the guess right. After checking at the beginning of the loop, we prompt them, they respond, and if they get it right, we update the value of guessedRight. Try running this program to see if you can guess the number!
+
+### Input Validation
+One thing we glanced over, was what if the input entered isn't a number? In this case, `tonumber` will return nil - and we forgot to check for that. Let's add that right now!
```
annieetang:
```suggestion
One thing we glanced over was what if the input entered isn't a number? In this case, `tonumber` will return nil - and we forgot to check for that. Let's add that right now!
```

docs/guide/writing-a-guessing-game.md
```diff
@@ -0,0 +1,203 @@
...
+local args = {...} -- ... are the arguments passed to this script, and they're being unpacked into an array.
+```
+
+The first argument is always the name of the script, with the remaining arguments following after. If the user passes a `--max <number>` argument, we should set the maximum value of the number being generated to that value, defaulting to a 100. We can write a small helper function to validate this information"
+
+```luau
+local iolib = require("@std/io")
+local args = {...}
+local function getArgs(args) : number
+    if #args < 3 ||  then
```
annieetang:
```suggestion
    if #args < 3 || then
```

docs/guide/writing-a-guessing-game.md
```diff
@@ -0,0 +1,203 @@
...
+        guessedRight = true
+    end
+end
+```
+
+If you try this program out now, you should be able to run it like: 
+```bash
+lute main.luau --max 50
+```
+to re-run the guessing game with maximum 50 guesses.
```
annieetang, line 214: could we have a small conclusion section that wraps up what we covered & the final commands / user flow of `main.luau`

something like:

### Conclusion
Congratulations! You've now built a guessing game using Lute. In this chapter, we covered:

* The Lute standard library (`@std` alias) (see [reference docs here](https://lute.luau.org/reference/std/))
* Luau builtins within Lute
* Command line argument handling
* ... [anything else nice to note]

docs/guide/writing-a-guessing-game.md
```diff
@@ -0,0 +1,203 @@
...
+order: 4
+---
+
+# Writing a Guessing Game
+
+### Introduction
+We're going to walk through how you might build a small guessing game using Lute. Like we did in the previous chapter, make a new folder
+for this project, along with a script called `main.luau`.
+
+This is a program that will take a users input and compare it to a secret number that the program will generate. The user will then be prompted to input their guess,
```
annieetang:
```suggestion
This is a program that will take a user's input and compare it to a secret number that the program will generate. The user will then be prompted to input their guess,
```

docs/guide/writing-a-guessing-game.md
```diff
@@ -0,0 +1,203 @@
...
+and the program will tell them "higher" or "lower" until they guess right.
```
annieetang: is it `"Too high!"` or `"Too low!"` now for the program output

docs/guide/writing-a-guessing-game.md
```diff
@@ -0,0 +1,203 @@
...
+
+```luau
+local iolib = require("@std/io")
+
+print("Say something! ")
+local input : string = iolib.input()
+print(input)
+```
+
+If you run this program with `lute main.luau` the program will pause until you input something, and then print that back out for you. For example:
```
annieetang:
```suggestion
Now, run this with `lute main.luau`. The program will pause until you input something, then print it back to you:
```

docs/guide/writing-a-guessing-game.md
```diff
@@ -0,0 +1,203 @@
...
+```bash
+lute main.luau
+>>> Say something! hello
+hello
+```
+
+### Generating random numbers with Luau builtins
+Let's move on to implementing this guessing game. We need to generate a random number, and Luau already provides this functionality for us, under the `math` table, which has a `random` function. This demonstrates a difference between Luau and Lute - Luau has builtins functions that operate on the primitive types, e.g. `string`, `math`, `buffer`, etc, but things like talking to your operating system or other programs must be handled by utilities that Lute provides. In this case, the `math.random` function is provided by Luau and will suffice for our purposes. We'll set a limit of 100, so that math.random will not generate numbers that are too small.
```
annieetang:
```suggestion
Let's move on to implementing this guessing game. We need to generate a random number, which Luau already provides via the `math.random` function, which can be found under the `math` table, which has a `random` function. 

This highlights an important distinction between Luau and Lute - **Luau** provides built-in functions that operate on the primitive types (e.g., `string`, `math`, `buffer`), while **Lute** provides utilities for interacting with the outside world (e.g., user input, file systems, shells)

In our guessing game case, the `math.random` function is provided by Luau and will suffice for our purposes. We'll set a limit of 100, so that `math.random` will not generate numbers that are too small.
```

docs/guide/writing-a-guessing-game.md
```diff
@@ -0,0 +1,203 @@
...
+```luau
+local iolib = require("@std/io")
+local randomNumber : number = math.random(100)
+
+print("Guess a number: ")
+local input : string = iolib.input()
+-- What do we do here?
+```
+
+So now we have the guessed value - we should try to let the user know if they got it right or wrong. However, the output of `input()` is a string, and we want to compare it to a number. If you try to ask `x \< randomNumber`, this won't work correctly. We'll want to convert this using the `tonumber` function.
```
annieetang:
```suggestion
So now we have the guessed value - we should try to let the user know if they got it right or wrong. However, the output of `input()` is a string, and we want to compare it to a number. If you try to ask `x \< randomNumber`, this won't work correctly. We'll want to convert this using the [`tonumber`](https://www.lua.org/manual/5.1/manual.html#pdf-tonumber) function.
```

docs/guide/writing-a-guessing-game.md
annieetang:
```suggestion
To start, we'll need some way of getting user input. Thankfully, Lute provides a utility for this in the [`io`](https://lute.luau.org/reference/lute/io.html) library. 

In `main.luau`, add:
```

docs/guide/writing-a-guessing-game.md
```diff
@@ -0,0 +1,203 @@
...
+### Generating random numbers with Luau builtins
+Let's move on to implementing this guessing game. We need to generate a random number, and Luau already provides this functionality for us, under the `math` table, which has a `random` function. This demonstrates a difference between Luau and Lute - Luau has builtins functions that operate on the primitive types, e.g. `string`, `math`, `buffer`, etc, but things like talking to your operating system or other programs must be handled by utilities that Lute provides. In this case, the `math.random` function is provided by Luau and will suffice for our purposes. We'll set a limit of 100, so that math.random will not generate numbers that are too small.
+
+```luau
+local iolib = require("@std/io")
+local randomNumber : number = math.random(100)
+
+print("Guess a number: ")
+local input : string = iolib.input()
+-- What do we do here?
```
annieetang: Do we need a full description for this section? If we want to consolidate how about something like this:

----

```luau
local io = require("@std/io")
local randomNumber: number = math.random(100)

print("Guess a number: ")
local input: string = io.input()

-- Convert the string input to a number for comparison
local guess = tonumber(input)
```

> **Note**: We use the [tonumber](https://www.google.com/search?q=https://www.lua.org/manual/5.1/manual.html%23pdf-tonumber) function because `io.input()` returns a string, and you cannot mathematically compare a string to a number (e.g., `"5" < 10` will throw an error).

docs/guide/writing-a-guessing-game.md
```diff
@@ -0,0 +1,203 @@
...
+if guess > randomNumber then
+    print("Too high!")
+elseif guess < randomNumber then
+    print("Too low!")
+else
+    print("Got it!")
+end
+```
+
+So far so good! Ideally though, the user would be able to guess multiple times - in this example, the user has one chance. Let's even the odds a bit - we'll need to re-prompt the user if they got it wrong and exit out if they guess right.
```
annieetang, line 77:
```suggestion
### Adding Loops and Logic

So far so good! To make the game playable, we'd ideally let the user guess multiple times. In our current version, the user has one chance. Let's even the odds a bit - we'll re-prompt the user with a `while` loop that continues until the user guesses correctly.
```

docs/guide/writing-a-guessing-game.md
```diff
@@ -0,0 +1,203 @@
...
+    if guess > randomNumber then
+        print("Too high!")
+    elseif guess < randomNumber then
+        print("Too low!")
+    else
+        print("Got it!")
+        guessedRight = true
+    end
+end
+```
```
annieetang, line 128: if we want to consolidate (but feel free to ignore!)

---

```luau
local iolib = require("@std/io")
local randomNumber = math.random(100)

local guessedRight = false
while not guessedRight do
    print("Guess a number: ")
    local input = iolib.input()
    local guess : number? = tonumber(input)
    if not guess then
        print(`Not a number: {input}`)
        continue
    end
    if guess > randomNumber then
        print("Too high!")
    elseif guess < randomNumber then
        print("Too low!")
    else
        print("Got it!")
        guessedRight = true
    end
end
```

> **Input Validation**: Notice the `if not guess` check. If a user types "hello" instead of "5", `tonumber` returns `nil`. Without this check, the program would crash. Using `continue` allows us to skip the rest of the loop and ask the user again.

docs/guide/writing-a-guessing-game.md
```diff
@@ -0,0 +1,203 @@
...
+
+### Handling command line arguments
+To finish up, let's see how we can pass arguments to programs in Lute to configure their behaviour. In this case, let's try to configure the maximum size of random number being generated. Lute programs have additional arguments passed to them as variadics from the top level. For example:
+
+```luau
+local args = {...} -- ... are the arguments passed to this script, and they're being unpacked into an array.
+```
+
+The first argument is always the name of the script, with the remaining arguments following after. If the user passes a `--max <number>` argument, we should set the maximum value of the number being generated to that value, defaulting to a 100. We can write a small helper function to validate this information"
```
annieetang:
```suggestion
The first argument (`args[1]`) is always the script name, with the remaining arguments following after. If the user passes a `--max <number>` argument, we should set the maximum value of the number being generated to that value, defaulting to a 100. We can write a small helper function to validate this information:
```

docs/guide/writing-a-guessing-game.md
```diff
@@ -0,0 +1,203 @@
...
+        print("Too low!")
+    else
+        print("Got it!")
+        guessedRight = true
+    end
+end
+```
+
+### Handling command line arguments
+To finish up, let's see how we can pass arguments to programs in Lute to configure their behaviour. In this case, let's try to configure the maximum size of random number being generated. Lute programs have additional arguments passed to them as variadics from the top level. For example:
```
annieetang:
```suggestion
To finish up, let's see how we can pass arguments to programs in Lute to configure their behaviour. 

For our game, let's allow users to configure the maximum size of random number being generated. In Lute, top-level arguments are passed as variadics (`...`). For example:
```

annieetang (commented):
just some rewording, suggestions to shorten some sections (feel free to ignore), and adding some hyperlinks to other docs!

docs/guide/writing-a-guessing-game.md
annieetang: oh or the summary you wrote in the PR description could be nice to have too

docs/guide/writing-a-guessing-game.md
Vighnesh-V: Fixed.

docs/guide/writing-a-guessing-game.md
Vighnesh-V: I kept this section but added the note about continue.

docs/guide/writing-a-guessing-game.md
```diff
@@ -0,0 +1,225 @@
...
+
+Now, run this with `lute main.luau`. The program will pause until you input something, then print it back to you:
+```bash
+lute main.luau
+>>> Say something! hello
+hello
+```
+
+### Generating random numbers with Luau builtins
+Let's move on to implementing this guessing game. We need to generate a random number, whcih Luau already provides via the `random` function on the `math` table.
```
annieetang:
```suggestion
Let's move on to implementing this guessing game. We need to generate a random number, which Luau already provides via the `random` function on the `math` table. 
```

docs/guide/writing-a-guessing-game.md
```diff
@@ -0,0 +1,225 @@
...
+```bash
+lute main.luau
+>>> Say something! hello
+hello
+```
+
+### Generating random numbers with Luau builtins
+Let's move on to implementing this guessing game. We need to generate a random number, whcih Luau already provides via the `random` function on the `math` table. 
+
+This highlights an important distinction between Luau and Lute - **Luau** provides builtin functions that operate on the primitive types, e.g. `string`, `math`, `buffer`, etc, while **Lute** provides utilities for interacting with the outside world (for example, user input, the file system, http requests). In this case, the `math.random` function is provided by Luau and will suffice for our purposes. We'll set a limit of 100, so that math.random will not generate numbers that are too small.
```
annieetang: oh can we do one style of listing examples in this sentence:
either the `, e.g. xxx, etc.` or the `(for example, ...)` for consistency

annieetang (approved):

PR #887, opened 2026-03-19
Adds `utils.isRequireCall`

tyulpan:
Right now, users have to write something like this to get `require` calls([example](https://github.com/luau-lang/lute/blob/e0faf05ee8d274be25c9e686e95c88af42d9589c/examples/query.luau#L7)):
```luau
local requireCalls = query
    .findallfromroot(ast, utils.isExprCall)
    :filter(function(call)
        local calledFunc = call.func
        return calledFunc.tag == "global" and calledFunc.name.text == "require"
    end)
```

writing their own `helpers`([example](https://github.com/luau-lang/lute/blob/e0faf05ee8d274be25c9e686e95c88af42d9589c/lute/cli/commands/lint/rules/unused_variable.luau#L19)):
```luau
local function isRequireCall(expr: syntax.AstExpr)
  return expr.tag == "call" and expr.func.tag == "global" and expr.func.name.text == "require"
end
```

It would be nice if `utils` had `utils.isRequireCall`. 
Then it would be simplified to:
```luau
local requireCalls = query.findallfromroot(ast, utils.isRequireCall)
```

Thanks to @Vighnesh-V for [highlighting](https://github.com/luau-lang/lute/pull/878/changes#r2934146238) the issue and @skberkeley for the `API` [suggestion](https://github.com/luau-lang/lute/pull/878/changes#r2934155678).


Vighnesh-V (approved):
This is really nice to have!

PR #888, opened 2026-03-19
Implements a `lute new <project>` subcommand.

Vighnesh-V:
I've added a subcommand to `lute` to make it easier to start new projects, similar to `zig init` or `cargo new`. This command is invoked with:
```
lute new <project-name>
``` 
and sets up a project with the following structure:
```
project-name/
    main.luau
    main.test.luau
    .config.luau
```
This command sets up the following default project configuration using the `.config.luau` format, with comments included for convenience.
```luau
return {
	lute = {
		lint = {
			-- Array of globbed strings, representing paths to exempt from linting,
			ignores = {},
			-- Global values passed to each linting rule,
			globals = {},
			-- Array of strings from which to load local lint rules,
			rulepaths = {},
			-- Per rule overrides,
			ruleconfigs = {
				-- Array of globbed strings, representing paths exempt from this rule,
				-- ignores: {string}?,
				-- Override a rule's default severity,
				-- severity: ("warn" | "error" | "info" | "hint")?,
				-- Pass custom options through to a rule,
				-- options: { [string]: unknown }?,
				-- Disable this rule,
				-- off: boolean?,
			},
		},
	},
	luau = {
		-- Default language mode to use for typechecking,
		languagemode = "strict",
		-- Aliases for builtin libraries and lint rules
		aliases = {
			lute = "~/.lute/typedefs/0.1.0/lute",
			std = "~/.lute/typedefs/0.1.0/std",
			lint = "~/.lute/typedefs/0.1.0/lint",
		},
	},
}
```

I think this is a good starting point, but I'm happy to take any suggestions for improvements to this command. I think this is a pretty sane set of defaults. I'd love to hear if folks have opinions on file names - `main.luau` isn't a common entry point, so we could do something like `projectName.luau` and `projectName.test.luau` instead of `main.luau` and `main.test.luau`.

Changes I've made:
1. Extract some of the common logic between `lute setup` and `lute new` into a common file.
2. Update `lute test` to search for tests from the current directory, instead of a dedicated tests/ directory, This is informed by a discussion I had with @Nicell about Luau engineers colocating tests with source files.

### DIFF
--- a/lute/cli/commands/lib/typedefs.luau
+++ b/lute/cli/commands/lib/typedefs.luau
@@ -1,0 +1,109 @@
+local path = require("@std/path")
+local process = require("@std/process")
+
+local TYPEDEFS_VERSION = "0.1.0"
+local TYPEDEFS_BASE_DIR = `.lute/typedefs/{TYPEDEFS_VERSION}`
+local TYPEDEFS_RELATIVE_DIR = `~/{TYPEDEFS_BASE_DIR}`
+local TYPEDEFS_PATH = path.join(process.homedir(), TYPEDEFS_BASE_DIR)
+
+local TEMPLATE_RC_FILE: { [any]: any } = {
+	aliases = {
+		lint = `{TYPEDEFS_RELATIVE_DIR}/lint`,
+		lute = `{TYPEDEFS_RELATIVE_DIR}/lute`,
+		std = `{TYPEDEFS_RELATIVE_DIR}/std`,
+	},
+}
+
+local function indent(level: number)
+	return string.rep("\t", level)
+end
+
+local function nest(src, parenIndent)
+	return `\{\n{src}\n{parenIndent}\}`
+end
+
+local function kv(k, v, desc: string?)
+	local kvpair = `{k} = {v}`
+	if desc then
+		kvpair = kvpair .. `, -- {desc}`
+	end
+	return kvpair
+end
+
+local function tbl(fields: { { value: string, description: string } } | { string }, level)
+	local buf: { string } = {}
+	local fieldIndent = indent(level + 1)
+	local parenIndent = indent(level)
+	for _, f in fields do
+		if typeof(f) == "string" then
+			table.insert(buf, `{fieldIndent}{f},`)
+		else
+			table.insert(buf, `{fieldIndent}-- {f.description}`)
+			table.insert(buf, `{fieldIndent}{f.value},`)
+		end
+	end
+
+	return nest(table.concat(buf, "\n"), parenIndent)
+end
+
+local function generateLuauConfig(languageMode: "strict" | "nonstrict"): string
+	local aliases = {
+		kv("lute", `"{TYPEDEFS_RELATIVE_DIR}/lute"`),
+		kv("std", `"{TYPEDEFS_RELATIVE_DIR}/std"`),
+		kv("lint", `"{TYPEDEFS_RELATIVE_DIR}/lint"`),
+	}
+	local aliasTable = tbl(aliases, 2)
+	local luau = tbl({
+		{
+			value = kv("languagemode", `"{languageMode}"`),
+			description = "Default language mode to use for typechecking",
+		},
+		{ value = kv("aliases", aliasTable), description = "Aliases for builtin libraries" },
+	}, 1)
+	return luau
+end
+
+local function generateLuteConfig(): string
+	local ruleconfigs = tbl({
+		`-- [RuleName] : \{`,
+		`-- Array of globbed strings, representing paths exempt from this rule`,
+		`-- ignores: \{string\}?`,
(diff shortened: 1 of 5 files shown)


lute/cli/commands/new/init.luau
```diff
@@ -0,0 +1,84 @@
...
+if not fs.exists(typedefs.TYPEDEFS_PATH) then
+	print("Running lute setup first...")
+	local result = process.run({ path.format(process.execpath()), "setup" }, { stdio = "inherit" })
+	if not result.ok then
+		print("Error: lute setup failed.")
+		process.exit(1)
+	end
+end
+
+fs.createdirectory(projectPath, { makeparents = true })
```
skberkeley: why `makeparents = true`? do we allow `lute new ./dir/proj`?

lute/cli/commands/new/init.luau
```diff
@@ -0,0 +1,84 @@
...
+
+local mainScript = path.join(projectPath, "main.luau")
+local testScript = path.join(projectPath, "main.test.luau")
+
+fs.writestringtofile(
+	mainScript,
+	[[print("Hello, world!")
```
skberkeley: nit:
```suggestion
	[[
print("Hello, world!")
```
idt leading newline gets written

lute/cli/commands/new/init.luau
```diff
@@ -0,0 +1,84 @@
...
+]]
+)
+
+fs.writestringtofile(
+	testScript,
+	[[local _test = require("@std/test")
```
skberkeley:
```suggestion
	[[
local _test = require("@std/test")
```

lute/cli/commands/new/init.luau
```diff
@@ -0,0 +1,84 @@
...
+)
+
+fs.writestringtofile(path.join(projectPath, ".config.luau"), typedefs.generateConfigDotLuau("strict"))
+
+print(`Created project '{projectName}'`)
+print("")
+print("To get started:")
+print(`\tcd {projectName}`)
+print("\tlute run main.luau")
+print("\tlute test")
```
skberkeley: nit: can we put this in a `[[]]` string and print it once?

lute/cli/commands/lib/typedefs.luau
```diff
@@ -0,0 +1,109 @@
...
+
+local TEMPLATE_RC_FILE: { [any]: any } = {
+	aliases = {
+		lint = `{TYPEDEFS_RELATIVE_DIR}/lint`,
+		lute = `{TYPEDEFS_RELATIVE_DIR}/lute`,
+		std = `{TYPEDEFS_RELATIVE_DIR}/std`,
+	},
+}
+
+local indent = function(level: number)
```
skberkeley: pls `local function`

lute/cli/commands/lib/typedefs.luau
```diff
@@ -0,0 +1,109 @@
...
+		kv("lint", `"{TYPEDEFS_RELATIVE_DIR}/lint"`),
+	}
+	local aliasTable = tbl(aliases, 2)
+	local luau = tbl({
+		{
+			value = kv("languagemode", `"{languageMode}"`),
+			description = "Default language mode to use for typechecking",
+		},
+		{ value = kv("aliases", aliasTable), description = "Aliases for builtin libraries" },
+	}, 1)
```
skberkeley, line 62: i won't die on this hill but i personally feel that a giant interpolated string would be more readable

tests/cli/new.test.luau
skberkeley, line 1: some of these seem like good candidates for snapshot testing

Vighnesh-V: Yeah that's what I was thinking, but snapshot testing isn't ready yet. I still have to work on my PR a little bit.

lute/cli/commands/new/init.luau
Vighnesh-V: Nah, I think this still looks ugly as a multiline or interpolated string.

lute/cli/commands/lib/typedefs.luau
Vighnesh-V: I feel like there will be more desire to set up sane defaults / create .config.luau with configurable defaults so I'm going to leave this as is.

lute/cli/commands/new/init.luau
Vighnesh-V: it doesn't! Updating now.

lute/cli/commands/new/init.luau
Vighnesh-V: autopiloted here, this shouldn't make parents.

Vighnesh-V:
Also converted `new` to use the `main(...)` format for defining entry points for cli commands.

skberkeley (approved):

PR #889, opened 2026-03-20
Adds `luau.requires` to `@std`

### LINKED ISSUE #776: There should be a standard library method for 'get all the required modules for a given module'

tyulpan:
Adds `luau.requires(filePath: path.pathlike): { RequireNode }` to `@std`, which parses Luau source code and returns all of its required modules in the following format:
```console
{
  resolvedPath: string?, -- the path to the resolved require
  requirePath: string?,  -- an argument passed to the require as a string literal
  requireExpr: luteLuau.AstExprCall, -- the ast expression passed to require
}
```

Resolves #776.
Related to PR #878.

Now we can run this code for [hash.luau](https://github.com/luau-lang/lute/blob/96d1afa75313ecb1639c12a2dc8a3c7149447a9b/examples/hash.luau):
```luau

local requires = luau.requires("./examples/hash.luau")

print(pp(requires))
```

and get:
```console
{
  {
    requirePath = "@lute/crypto",
    resolvedPath = "/path/to/lute/definitions/crypto.luau",
    requireExpr = {...}
  },
  {
    requirePath = "@batteries/base64",
    resolvedPath = "/path/to/lute/batteries/base64.luau",
    requireExpr = {...}
  },
}
```


lute/std/libs/luau.luau
```diff
@@ -7,11 +7,22 @@ local luteLuau = require("@lute/luau")
...
 local luau = {}
 
 export type bytecode = luteLuau.Bytecode
 
+export type TypePack = luteLuau.TypePack
+
+export type RequireNode = {
+	resolvedPath: string?,
+	requireContent: string?,
+	requireAst: luteLuau.AstExprCall,
```
aatxe:
```suggestion
	requireExpr: luteLuau.AstExprCall,
```

This seems like a better name for this field.

lute/std/libs/luau.luau
```diff
@@ -7,11 +7,22 @@ local luteLuau = require("@lute/luau")
...
 
 local luau = {}
 
 export type bytecode = luteLuau.Bytecode
 
+export type TypePack = luteLuau.TypePack
+
+export type RequireNode = {
+	resolvedPath: string?,
+	requireContent: string?,
```
aatxe:
```suggestion
	requirePath: string?,
```
This seems like a better name for this field.

aatxe (changes requested):

aatxe (approved):

PR #892, opened 2026-03-20
Rename all types to be PascalCase in lute APIs

annieetang:
in the journey to shipping a stable version of Lute, we've decided to rename all types in lute APIs to be `PascalCase` from now on. This PR updates all the usages and type definitions, so will probably be a breaking change for users when merged


lute/cli/commands/lint/rules/duplicate_keys.luau
```diff
@@ -13,7 +13,7 @@ local function lint(
 	sourcepath: path.path?,
```
annieetang, line 13: Also, I didn't update `path` and `pathlike` because there would be a ton of changes. Let me know if we still want to update that (might split into separate PR)

aatxe: Should definitely do it, but it's fine to do it in a separate PR!

definitions/crypto.luau
```diff
@@ -1,23 +1,23 @@
...
 
-function crypto.digest(hash: hash<any>, message: string | buffer): buffer
+function crypto.digest(hash: Hash<any>, message: string | buffer): buffer
 	error("not implemented")
 end
 
 crypto.secretbox = {}
 
-export type secretbox = {
+export type Secretbox = {
```
aatxe:
```suggestion
export type SecretBox = {
```

aatxe (approved):
Looks good, only note is that it should be `SecretBox`, not `Secretbox` for crypto.

definitions/crypto.luau
annieetang: ty!

PR #896, opened 2026-03-24
Adds `assert.that` to `@std/assert`

annieetang:
Adds and uses `assert.that(value: unknown, msg: string?): Failure?` in lieu of places we have `assert.eq(_, true)`

We considered naming this `assert.true` and `assert.truthy`, but agreed on `assert.that` since it's the [new (suggested) API for NUnit](https://docs.nunit.org/articles/nunit/writing-tests/assertions/assertion-models/constraint.html)


lute/std/libs/test/assert.luau
```diff
@@ -79,6 +79,14 @@ local function neq<T>(lhs: T, rhs: T, msg: string?): Failure?
 	return assertion(msg or `{lhs} == {rhs}`)
 end
 
+local function truthy(value: unknown, msg: string?): Failure?
+	if value then
+		return nil
+	end
+
+	return assertion(msg or `Expected {formatValue(value)} to be truthy.`)
```
Vighnesh-V:
```suggestion
	return assertion(msg or `Expected a truthy value, but got {formatValue(value)}`)
```

Vighnesh-V: Otherwise this error message says 'Expected nil to be truthy' when it fails, which is kind of a weird error message.

annieetang: ty! updated

Vighnesh-V (approved):
Hell yeah!

PR #897, opened 2026-03-25
Introduces an abstraction for libuv streams.

Vighnesh-V:
This PR adds a libuv stream abstraction and uses it in the process implementation.

I've done some refactoring on how we create processes and tried to document the structure a bit better. Some of this is just mechanically moving stuff around, but hopefully this is a touch more readable.


Nicell:
Which other constructs are we thinking would adopt this? io reading, something in fs, or maybe websockets? I think right now it's hard to tell if this is the right shape to me because it's only being used on process. I think maybe migrating a second interface or providing some high level overview of another future consumer like tcp streams could be useful.

Vighnesh-V:
@Nicell This is intended to wrap every stream interface provided by libuv- it is definitely intended to work for tcp, io etc. This is already a pretty dense PR so I don't want to migrate a second thing right now, but my intention is absolutely to do that in a follow up so it's a bit more manageable. Maybe I could introduce the tcp library next up?

I will probably end up needing to tweak the shape of UVStream/subclasses as I go, so there will definitely be more iteration. The overall goal here is to enable us to think less about the low level C constructs + interfacing with the runtime.

lute/runtime/src/UVStream.cpp
```diff
@@ -0,0 +1,8 @@
+#include "lute/UVStream.h"
```
vrn-sn: Do we need this file?

Edit: this is slightly misleading, sorry. I didn't mean the file being included, I mean the file I made the comment in (UVStream.cpp).

lute/runtime/src/uvutils.cpp
```diff
@@ -36,4 +36,27 @@ Luau::Variant<std::string, UvError> getStringFromUv(BufferWriter bufferWriter, s
...
+        auto var = currentEnvItems[i];
+        if (var.name && var.value && map.find(var.name) == map.end())
+            map[var.name] = var.value;
+    }
+    uv_os_free_environ(currentEnvItems, currentEnvCount);
+
+    return std::nullopt;
+}
+
+
```
vrn-sn:
```suggestion
```

lute/runtime/CMakeLists.txt
```diff
@@ -5,11 +5,13 @@ target_sources(Lute.Runtime PRIVATE
     include/lute/runtime.h
     include/lute/userdatas.h
     include/lute/UVRequest.h
+    include/lute/UVStream.h
     include/lute/uvutils.h
 
     src/ref.cpp
     src/runtime.cpp
     src/UVRequest.cpp
```
vrn-sn, line 13: Could we tweak the file naming to match the other lowercase filenames (for both UVRequest and UVStream)?

lute/process/src/process.cpp
```diff
@@ -24,28 +26,32 @@
...
-namespace process
+static int crossPlatformFileno(FILE* stream)
 {
+#ifdef _WIN32
+    return _fileno(stream);
+#else
+    return fileno(stream);
+#endif
+}
+
```
vrn-sn:
```suggestion
```

lute/process/src/process.cpp
```diff
@@ -174,209 +435,28 @@ struct ProcessHandle
         }
         else
         {
-            resumeToken->fail("Process error: " + error_msg);
+            std::string msg = "Process failed because: \n";
+            for (auto msg : state.getErrors())
+            {
+                msg += msg + "\n";
+            }
+            resumeToken->fail(msg);
```
vrn-sn:
```suggestion
            std::string msg = "Process failed because: \n";
            for (const auto& error : state.getErrors())
            {
                msg += error + "\n";
            }
            resumeToken->fail(msg);
```
Shadowing probably isn't what was intended here?

lute/process/src/process.cpp
```diff
@@ -174,209 +435,28 @@ struct ProcessHandle
...
         }
 
-        resumeToken.reset();
+        closeHandles();
```
vrn-sn: I think we lost the call to `resumeToken.reset()` in this final path - eventually it will get freed once we `self.reset()`, but that won't happen until the close callbacks finish.

lute/process/src/process.cpp
```diff
@@ -59,83 +65,338 @@ void convertCRLFtoLF(std::string& str)
...
+    std::vector<char*> processArguments;
+    std::vector<std::string> environmentStrings;
+    std::vector<char*> environmentVarString;
+
+    ProcessHandle(lua_State* L, ProcessOptions& opts, std::vector<std::string>& args, int waitForN, std::string context = "Process Spawn")
+        : loop(getRuntimeLoop(L))
+        , resumeToken(getResumeToken(L))
+        , stdinPipe(std::make_unique<uvutils::PipeStream>(loop, false, context + " stdin pipe"))
+        , stdoutPipe(std::make_unique<uvutils::PipeStream>(loop, false, context + " stdout pipe"))
+        , stderrPipe(std::make_unique<uvutils::PipeStream>(loop, false, context + " stderr pipe"))
```
vrn-sn: In the case of `none` and `inherit`, can we avoid these heap allocations? On my machine, looks like each one is 400+ bytes.

vrn-sn (commented):
A few comments, but my big takeaway: I think we've hit critical mass for process.cpp in terms of complexity. I had a really hard time reading the diff and had to just open the old and new file split in my editor.

Can we spend some time pulling each struct into its own .h/.cpp files? Maybe some simple ones could be grouped, but I'd prefer if the structs that have large comment blocks above them get their own .h/.cpp files in the `process` folder.

Otherwise, approving - similar to @Nicell, I don't have a great sense for whether it's the correct abstraction, but it definitely looks cleaner. Trusting that we can just keep iterating from here to get it to where we want it.

vrn-sn (approved):

lute/runtime/include/lute/UVStream.h
```diff
@@ -0,0 +1,143 @@
...
+        , handleContext(handleContext)
+        , closed(false)
+    {
+        stream.data = this;
+    }
+
+    UVStream(const UVStream&) = delete;
+    UVStream& operator=(const UVStream&) = delete;
+    UVStream(UVStream&&) = delete;
+    UVStream& operator=(UVStream&&) = delete;
```
vrn-sn, line 45: Just curious, what's the harm in enabling move semantics? Saw it was set up like this in UVRequest as well.

lute/runtime/include/lute/UVStream.h
```diff
@@ -0,0 +1,143 @@
...
+    {
+        return static_cast<T*>(&stream);
+    }
+
+    void read(OnRead&& onRead, OnStreamEnd&& onStreamEnd)
+    {
+        callbacks.onRead = onRead;
+        callbacks.onStreamEnd = onStreamEnd;
+        uv_read_start((uv_stream_t*)&stream, allocBuffer, readStream);
+    }
```
vrn-sn, line 57:
```suggestion
    void read(OnRead&& onRead, OnStreamEnd&& onStreamEnd)
    {
        callbacks.onRead = std::move(onRead);
        callbacks.onStreamEnd = std::move(onStreamEnd);
        uv_read_start((uv_stream_t*)&stream, allocBuffer, readStream);
    }
```
Currently doing a copy, probably not intended.

Also, if the signature were `OnRead` instead of `OnRead&&`, we could also accept lvalues in addition to rvalues, which makes this API a bit more flexible. Although, maybe that was an intentional restriction, not sure.

lute/runtime/include/lute/UVStream.h
Vighnesh-V: no harm, I want to err on the side of more restrictive until we have a good reason to loosen the restriction

lute/runtime/include/lute/UVStream.h
```diff
@@ -0,0 +1,143 @@
...
+    OnClose onClose;
+};
+
+template<typename T>
+struct UVStream
+{
+
+    UVStream(uv_loop_t* loop, std::string handleContext = "")
+        : loop(loop)
+        , handleContext(handleContext)
```
vrn-sn:
```suggestion
        , handleContext(std::move(handleContext))
```

lute/runtime/include/lute/UVStream.h
```diff
@@ -0,0 +1,143 @@
...
+    {
+        return closed;
+    }
+
+    void close(OnClose&& onClose)
+    {
+        if (isClosed())
+            return;
+        callbacks.onClose = std::move(onClose);
+        if (!uv_is_closing((uv_handle_t*)&stream))
```
vrn-sn: Is there potentially a subtle bug here? We get an OnClose callback that the caller expects to be called (in your process.cpp code, it must be called so that `onHandleClose` can decrement `pendingCloses`).

However, if uv is already closing the stream, then it will never be called. We store it but never call it.

lute/runtime/src/UVStream.cpp
Vighnesh-V: probably not, I created it for convenience, but I can get rid of this.

lute/runtime/CMakeLists.txt
Vighnesh-V: I'm going to do this in a separate PR if that's okay.

lute/process/src/process.cpp
Vighnesh-V: I think that's fine - the ResumeToken is owned by the ProcessHandle - when the handle is complete, it should release the things it owns :)

lute/process/src/process.cpp
Vighnesh-V: I've gone ahead and heap allocated the actual stream structure because it's quite large and that has reduced the actual profile of PipeStream. I am also setting this code up to avoid allocating these pipes when the mode is not kStdioKindDefault.

lute/runtime/include/lute/UVStream.h
Vighnesh-V: Good catch! This is what I think this bug might be:
If you call close a second time when the stream is closing but not closed yet, you clobber the callback.

The fix here is to assign this callback iff uv_is_closing is false. So the caller who calls close is guaranteed that the callback they registered the first time will be called when the callback passed to uv_close is invoked.

If you do that, it is possible for a callback to be silently ignored. Like you pointed out, the caller might have done something that depends on their callback being called. E.g. if they incremented an 'outstanding' counter and then decremented it in the callback. If we don't return any indication of if their callback was successfully registered then they will unconditionally do something that depends on successfully invocation of their callback. I made this return a bool that indicates whether or not the close was successfully registered.

PR #898, opened 2026-03-25
Enables CLI commands to use the lute version string defined in C++

Vighnesh-V:
We've defined a version string in CMake, which gets turned into a header that C++ commands can consume. However, the code in lute uses a hardcoded version of this string. We need to expose this information to the cli commands so that they can use the correct version when setting up new projects or generating type definition files.

This PR sets up a global table called `version` with shape: `{version: string, versionFull : string, versionSuffix : string}` for consumption _exclusively_ by the CLI commands. User code and code loaded by various lute apis in the tests cannot use this global (as those are sandboxed environments).

### DIFF
--- a/lute/cli/src/climain.cpp
+++ b/lute/cli/src/climain.cpp
@@ -630,7 +630,7 @@ int handleCompileCommand(int argc, char** argv, int argOffset, LuteReporter& rep
 
 void setupVersionLibrary(lua_State* L)
 {
-    lua_checkstack(L, 3);
+    lua_checkstack(L, 2);
 
     lua_createtable(L, 0, 3);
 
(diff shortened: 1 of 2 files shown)


lute/cli/src/climain.cpp
```diff
@@ -628,10 +628,28 @@ int handleCompileCommand(int argc, char** argv, int argOffset, LuteReporter& rep
     return 0;
 }
 
+void setupVersionLibrary(lua_State* L)
+{
+    lua_checkstack(L, 3);
```
vrn-sn:
```suggestion
    lua_checkstack(L, 2);
```
I believe we only use 2 scratch slots here?

aatxe (approved):

lute/cli/src/climain.cpp
Vighnesh-V: yup, that's my bad. Fixing now. (The first draft of this added a single global per lute version string instead of using a table).

PR #900, opened 2026-03-25
Changes the standard library implementation of `@std/time` to be backed by `@lute/time`

aatxe:
We've had a long-standing split between `@std/time` and `@lute/time` where the former provided an entirely userland implementation of duration, but the latter provided both durations and instants from C++. In the interest of getting the standard library generally into shape and avoiding these confusing duplications (with differences!) between the runtime and the standard library, this PR eliminates the `@std/time` that already exists, and replaces it with one that re-exports the runtime functionality. Making this a proper replacement required both implementing `__mul` and `__div`, as well as fixing a number of smaller bugs that were scattered throughout the runtime implementation of duration.

### DIFF
--- a/definitions/time.luau
+++ b/definitions/time.luau
@@ -14,6 +14,10 @@ end
 
 time.duration = {}
 
+function time.duration.create(seconds: number, subsecnanos: number): Duration
+	error("not implemented")
+end
+
 function time.duration.nanoseconds(nanoseconds: number): Duration
 	error("not implemented")
 end
@@ -100,6 +104,8 @@ export type Duration = setmetatable<Frozen, {
 	read __index: typeof(table.freeze(duration_ops)),
 	read __add: (Duration, Duration) -> Duration,
 	read __sub: (Duration, Duration) -> Duration,
+	read __mul: (Duration, number) -> Duration,
+	read __div: (Duration, number) -> Duration,
 	read __eq: (Duration, Duration) -> boolean,
 	read __lt: (Duration, Duration) -> boolean,
 	read __le: (Duration, Duration) -> boolean,
--- a/lute/time/src/time.cpp
+++ b/lute/time/src/time.cpp
@@ -98,7 +98,7 @@ int createDurationFromSeconds(lua_State* L, double seconds)
 static int duration_tonanoseconds(lua_State* L)
 {
     uv_timespec64_t timespec = getTimespecFromDuration(L, 1);
-    lua_pushnumber(L, static_cast<double>(timespec.tv_sec) * NANOSECONDS_PER_SECOND + timespec.tv_nsec);
+    lua_pushnumber(L, static_cast<double>(timespec.tv_sec * NANOSECONDS_PER_SECOND) + timespec.tv_nsec);
     return 1;
 }
 
--- a/lute/time/src/time.cpp
+++ b/lute/time/src/time.cpp
@@ -12,15 +12,15 @@
 #include <cstdint>
 #include <iterator>
 
-const int64_t NANOSECONDS_PER_SECOND = 1000000000;
-const int64_t MICROSECONDS_PER_SECOND = 1000000;
+const int64_t NANOSECONDS_PER_SECOND = 1'000'000'000;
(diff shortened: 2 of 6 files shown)


Vighnesh-V:
Title nit: 
Changes the standard library implementation of @std/time ....

lute/time/src/time.cpp
Vighnesh-V: Small nit - for some of the larger numbers, can you re-write using `'` to separate out groups of three zeroes?

lute/time/src/time.cpp
```diff
@@ -72,28 +91,28 @@ int createDurationFromTimespec(lua_State* L, uv_timespec64_t timespec)
...
-    return createDurationFromTimespec(L, {static_cast<int64_t>(seconds), static_cast<int32_t>(fmod(seconds, 1) * NANOSECONDS_PER_SECOND)});
+    return createDurationFromTimespec(L, getTimespecFromSeconds(seconds));
 }
 
 // Duration methods
 static int duration_tonanoseconds(lua_State* L)
 {
     uv_timespec64_t timespec = getTimespecFromDuration(L, 1);
-    lua_pushnumber(L, static_cast<double>(timespec.tv_sec * NANOSECONDS_PER_SECOND) + timespec.tv_nsec);
+    lua_pushnumber(L, static_cast<double>(timespec.tv_sec) * NANOSECONDS_PER_SECOND + timespec.tv_nsec);
```
Vighnesh-V: Why do you have to move the multiplication outside the cast here?

lute/time/src/time.cpp
```diff
@@ -121,35 +140,35 @@ static int duration_tohours(lua_State* L)
 static int duration_todays(lua_State* L)
 {
     uv_timespec64_t timespec = getTimespecFromDuration(L, 1);
-    lua_pushnumber(L, getSecondsFromTimespec(timespec) / SECONDS_PER_DAY);
+    lua_pushinteger(L, getSecondsFromTimespec(timespec) / SECONDS_PER_DAY);
```
vrn-sn, line 143: Wanted to check in about why we're pushing these as integers since `getSecondsFromTimespec` should be returning a double. (Applies to all the similar functions below as well.)

aatxe: The spec is not necessarily super well-defined here, but the tests seem to expect that none of these functions return doubles, e.g. 
```luau
		local mix = duration.milliseconds(1234) + duration.microseconds(567_000) + duration.nanoseconds(890)
		-- 1234ms + 567ms + 0.000890ms = 1801.00089ms
		assert.eq(mix:tomilliseconds(), 1801)
```

We could go the other way and change the tests, but it feels like people would "expect" to get the results at the precision of the method's unit.

lute/time/src/time.cpp
aatxe: We don't, we can leave that bit unchanged actually.

definitions/time.luau
Vighnesh-V, line 85: does this need to be `toWeeks` instead?

tests/std/time.test.luau
```diff
@@ -51,18 +51,15 @@ test.suite("TimeSuite", function(suite)
...
 
-		-- if nanoseconds exceed a second, create still allows it
 		local d2 = duration.create(1, 1_500_000_000)
 		assert.eq(d2:toseconds(), 2.5)
-		assert.eq(d2:subsecnanos(), 1_500_000_000)
+		assert.eq(d2:subsecnanos(), 500_000_000)
 
-		-- operations like addition normalize nanoseconds
 		local normalized = d2 + duration.seconds(0)
 		assert.eq(normalized:toseconds(), 2.5)
```
Vighnesh-V, line 64: This should also be pascalCase right?

aatxe: I was going to do camelCasing stuff as a second PR (and probably do it to more than just this one API).

Vighnesh-V: sure, not a big deal. I'll rereview.

Vighnesh-V (approved):

PR #901, opened 2026-03-25
Fix `task.awaitall`

wmccrthy:
If you try to run the `process` example, an error throws:
```
0
Hello, lute!

@std/task.luau:60: attempt to index number with 'success'
stacktrace:
@std/task.luau:60 function awaitall
./examples/process.luau:15
```

This occurs as we iterate over `tasks`, which is equal to a `table.pack` return value. This means `tasks` has an `n` property and is not entirely array-like. We can use `ipairs` when iterating over `tasks` to avoid the bug

### DIFF
--- a/lute/std/libs/task.luau
+++ b/lute/std/libs/task.luau
@@ -45,6 +45,7 @@ end
 
 function tasklib.awaitall(...: Task): ...unknown
 	local tasks = table.pack(...)
+	tasks.n = nil :: never
 
 	for i, v in ipairs(tasks) do
 		if not v.co then
--- a/lute/std/libs/task.luau
+++ b/lute/std/libs/task.luau
@@ -45,7 +45,6 @@ end
 
 function tasklib.awaitall(...: Task): ...unknown
 	local tasks = table.pack(...)
-	tasks.n = nil :: never
 
 	for i, v in ipairs(tasks) do
 		if not v.co then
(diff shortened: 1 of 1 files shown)


lute/std/libs/task.luau
Vighnesh-V, line 47: Can you just explicitly nil the `n` field here instead? That's what we've done previously.

Vighnesh-V (changes requested):

lute/std/libs/task.luau
Nicell: Why is overriding a prop preferred over the built-in `ipairs`? I just told wyatt to switch from `n = nil` to ipairs, and `ipairs` was originally used in this function it looks like

Nicell: https://github.com/luau-lang/lute/commit/e4ed4e10e4a009c647114c9613e10f1ba62de128 looks like it was originally using `n = nil`, I just don't understand how modifying the table is preferable to using `ipairs` for the only reason it still might be useful 🤔

Vighnesh-V: Lmao, I think I made the subsequent change in this file to move from nilling out the .n property to using ipairs 😆  because it played better with the typechecker. I'm got it wrong here - ipairs is better :)

Vighnesh-V: https://github.com/luau-lang/lute/commit/53411325b493bf904344e7204b99ffb708c932b8 - yeah here. We should be using ipairs instead.

Vighnesh-V (approved):

lute/std/libs/task.luau
Vighnesh-V: @wmccrthy is this good to merge?

wmccrthy: yup

PR #903, opened 2026-03-26
Adds support for bare paths in .luaurc

### LINKED ISSUE #853: Aliases in `.luaurc` are not implicitly relative when consuming packages
I had to make the following change to allow my package to be consumed by another project. Without this I was getting require resolution errors.

```diff
{
  "languageMode": "strict",
  "aliases": {
    "lint": "~/.lute/typedefs/0.1.0/lint",
    "std": "~/.lute/typedefs/0.1.0/std",
    "lute": "~/.lute/typedefs/0.1.0/lute",
    "repo": ".",
-   "root": "src",
+   "root": "./src",
-   "scripts": "scripts"
+   "scripts": "./scripts"
  }
}
```

I'm using Loom but I believe this is generally true for any nested `.luaurc`. 

For reference:

1. Package: https://github.com/flipbook-labs/flipbook-batteries
2. Consumer: https://github.com/flipbook-labs/flipbook

tyulpan:
`jumpToAlias` implementations now detect implicitly relative paths. Bare names like `"src"` that lack the trailing slash required by Luau's classifier are processed component by component via `toChild` / `toParent` from the current navigator position. Since the navigator is already positioned at the directory containing the config file when `jumpToAlias` is called, this resolves correctly regardless of the working directory.

Resolves #853.

P.S.
If such support is needed at the language level, [here](https://github.com/luau-lang/luau/blob/b37af212cb60366043c93c5a4acd085ab29c8d7a/CLI/src/AnalyzeRequirer.cpp#L45) would be the best entry point for it. Right now `luau` rejects all non-absolute alias values, so bare paths like `"src"` fail with `NotFound`.

### DIFF
--- a/lute/luau/src/resolvemodule.cpp
+++ b/lute/luau/src/resolvemodule.cpp
@@ -86,6 +86,9 @@ NC::NavigateResult FileVfsContext::resetToRequirer()
 
 NC::NavigateResult FileVfsContext::jumpToAlias(const std::string& path)
 {
+    if (isBarePath(path))
+        return convert(walkBarePath(path, vfs));
+
     return convert(vfs.resetToPath(path));
 }
 
--- a/lute/require/include/lute/modulepath.h
+++ b/lute/require/include/lute/modulepath.h
@@ -3,6 +3,7 @@
 #include <functional>
 #include <optional>
 #include <string>
+#include <string_view>
 
 enum class NavigationStatus
 {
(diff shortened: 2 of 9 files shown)


vrn-sn:
Hey, thanks so much for working on this! I read through your PR and wanted to let you know we went with the implementation in #902. Both take a similar approach using `jumpToAlias`, but #902 avoids having to reimplement path walking (`walkBarePath`) by calling back into the Luau.Require library.

Really appreciate you working on this! Thank you, and really sorry about the confusion.

tyulpan:
@vrn-sn No problem! Learned a few new things while working on this.

PR #904, opened 2026-03-26
Consolidates `assert` equality checks into `assert.eq`, unexposing `tableeq` and `buffereq`

annieetang:
Updates `assert.eq` to perform structural equality, calling `assert.tableeq` and `assert.buffereq` to simplify user experience for the assertion library. We also unexpose `tableeq`, `buffereq` so users only need to use `assert.eq` to test equality

Updates tests and snapshots to reflect changes


lute/std/libs/test/assert.luau
```diff
@@ -146,6 +138,30 @@ local function erroreq<A...>(func: (A...) -> ...unknown, expectedErrorMessage: s
 	return nil
 end
 
+local function eq<T>(lhs: unknown, rhs: unknown, msg: string?): Failure?
```
Vighnesh-V: eq no longer uses this generic parameter, so we can get rid of it here.

tests/std/snapshots/buffereq_failure_suite.snap
```diff
@@ -6,7 +6,7 @@ TEST RESULTS
 Failed Tests (1):
 
 ❌ buffereq_failure_suite.buffereq_error
-	<ROOT>/tests/std/snapshots/buffereq_failure_suite.snap.luau:9
+	@std/test/assert.luau:200
```
aatxe: Is this misattributing the error to the testing library itself?

annieetang: hm let me rerun the snapshot test I might've ran it wrong

annieetang: looks like it's because `failures.assertion` in `failure.luau` is explicitly checking 5 function calls up for the assertion:

```lua
function failures.assertion(msg: string): Failure
	-- we need to go 5 function calls up to get to the actual assertion that invoked this
	local filename, linenumber = debug.info(4, "sl")
...
```

so `assert.eq` calling `buffereq` and `tableeq` breaks that. I could introduce an optional arg to `assertion` to specify the depth to find the assertion? or should we find a better fix

Vighnesh-V: I think you probably want to thread the depth through somehow, so that the error messages are accurate. When we raise the error, we probably want to grab the depth off the error, instead of it being hardcoded to 4.

skberkeley (approved):

Vighnesh-V (approved):

PR #907, opened 2026-03-27
Adds `argnames` field in serialization of function Type

annieetang:
When working on cleaning up `lute doc` with `typeofmodule` API integration, I noticed that function type serialization was missing the `argNames` fields, so we weren't preserving the argument names of the parameters in function types. This makes it hard for users (aka the `lute doc` command) to use the type serialization on function types because our output would be missing the argument names and only have the types, which isn't very useful to understand the function signature.

Before this change, using the type serializer to generate docs would result in (the wip) `fs.link` serialization appearing as:

```luau
fs.link

(string, string) -> ()
```

But after this change, we have:

```luau
fs.link

(src: string, dest: string) -> ()
```

which better matches the actual function signature of of fs.link, which is 

```luau
function fs.link(src: string, dest: string): ()
```

Pls see more examples of this specific change's additions (in my [WIP branch](https://github.com/luau-lang/lute/compare/annietang/lute_doc_integration_with_typeofmodule) of `lute doc`) if curious: https://github.com/luau-lang/lute/commit/94880094b613962d4aa24d566ff6953fc4318110

### DIFF
--- a/tests/src/typeserializer.test.cpp
+++ b/tests/src/typeserializer.test.cpp
@@ -274,12 +274,17 @@ TEST_CASE_FIXTURE(TypeSerializeFixture, "serialize_function_type")
     TypePackId retTypes = arena.addTypePack(TypePack{{}, std::nullopt});
 
     FunctionType ftv{{genericTy}, {}, argTypes, retTypes};
+
+    FunctionArgument arg;
+    arg.name = "x";
+    ftv.argNames.push_back(arg);
+
     TypeId ty = arena.addType(ftv);
 
     lua_checkstack(L, 3);
 
-    // <T>(number) -> ()
-    // { tag: "function", parameters: { head: { { tag: "number" } }, tail: nil }, returns: { head: nil, tail: nil }, generics: { { tag: "generic", name: "T", ispack: false } }, genericpacks: {} }
+    // <T>(x: number) -> ()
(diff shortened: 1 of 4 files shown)


tests/src/typeserializer.test.cpp
```diff
@@ -274,12 +274,17 @@ TEST_CASE_FIXTURE(TypeSerializeFixture, "serialize_function_type")
...
+
     TypeId ty = arena.addType(ftv);
 
     lua_checkstack(L, 3);
 
-    // <T>(number) -> ()
-    // { tag: "function", parameters: { head: { { tag: "number" } }, tail: nil }, returns: { head: nil, tail: nil }, generics: { { tag: "generic", name: "T", ispack: false } }, genericpacks: {} }
+    // <T>(x: number) -> ()
+    // { tag: "function", parameters: { head: { { tag: "number" } }, tail: nil }, argnames: { "x" }, returns: { head: nil, tail: nil }, generics: { { tag: "generic", name: "T", ispack: false } }, genericpacks: {} }
     REQUIRE_EQ(Luau::serializeType(L, ty), 1);
```
hgoldstein: Can you add more tests here? I would like to see:
- Testing more than one argument
- Testing a function where some parameters have names and others don't

hgoldstein (approved):
Seems reasonable to me

PR #911, opened 2026-03-28
Adds `Runtime::runSource` method

### LINKED ISSUE #285: `Runtime::runSource` method to support running source code directly.
This would help supplement testing Luau source code in the runtime; we could load in source code, then compare expected stdout, implementing "golden file" testing.

It would also help with Luau-based CLI commands - we could invoke `Runtime::runSource()` and have two header files containing source code for the CLI commands.

tyulpan:
Adds `Runtime::runSource` method to support running Luau source code directly.

Resolves #285.

To avoid duplication and unnecessary dependencies between the `Runtime` and `CLI`, I decided to move main logic of `runBytecode` to the `Runtime`, which feels like its natural place. If that's not the case and main logic of `runBytecode` should stay in `climain`, I can revert the changes. But the solution will be more verbose.

The `CLI` layer's `runBytecode` now delegates to `Runtime::runBytecode` with an `onLoaded` callback for codegen and profiling, eliminating duplication within the `Runtime` implementation.

~~Additionally, I moved the `profiler options` from the `CLI` directly into the `profiler`. Currently, the sampling `frequency` in Hz duplicates its default value in two places: [here](https://github.com/luau-lang/lute/blob/6ed3142c96d109a7ac94ec7c9a7d8dc86a56ac69/lute/cli/src/profiler.cpp#L53) and [here](https://github.com/luau-lang/lute/blob/6ed3142c96d109a7ac94ec7c9a7d8dc86a56ac69/lute/cli/include/lute/climain.h#L13). 
It has no impact at the moment, but having a single source will help avoid potential bugs in the future.~~
Moved to PR #918.


tests/src/runsource.test.cpp
```diff
@@ -0,0 +1,121 @@
...
+
+static std::vector<std::string> capturedOutputs;
+
+static int capture(lua_State* L)
+{
+    const char* str = luaL_tolstring(L, 1, nullptr);
+    capturedOutputs.emplace_back(str);
+    return 0;
+}
+
```
Vighnesh-V: This shouldn't be done using a static variable at the top of the test. We actually already have a mechanism to do this. You can thread a `TestReporter` through (see CliRuntimeFixture for an example of how this is done). Then the `capture` method will write to the errors / output vector. You can augment this reporter to work with captured outputs if doesn't already.

lute/cli/include/lute/profiler.h
```diff
@@ -1,8 +1,16 @@
...
 
+#include <string>
+
 struct lua_State;
 
+struct ProfileOptions
+{
+    std::string filename;
+    int frequency = 10000;
+};
```
Vighnesh-V, line 12: Can you make this profiler change in a separate PR?

lute/runtime/CMakeLists.txt
```diff
@@ -15,5 +15,5 @@ target_sources(Lute.Runtime PRIVATE
 
 target_compile_features(Lute.Runtime PUBLIC cxx_std_17)
 target_include_directories(Lute.Runtime PUBLIC "include" ${LIBUV_INCLUDE_DIR})
-target_link_libraries(Lute.Runtime PRIVATE Luau.Require Luau.VM Lute.Common uv_a)
+target_link_libraries(Lute.Runtime PRIVATE Luau.Require Luau.Compiler Luau.VM Lute.Common uv_a)
```
Vighnesh-V: If the runtime now links in the VM and the Compiler, does the CMakeLists for lute/cli/ have to be updated to remove this dependency?

Vighnesh-V (changes requested):

tests/src/runsource.test.cpp
tyulpan: Ah, right! Initially, I was using custom setup arguments to pass the reporter and wanted to test `Runtime::runSource` without relying on `runCode`, so the fixture wasn't working for me. But by the end, I should have remembered it, yeah.

lute/runtime/CMakeLists.txt
tyulpan: The `CLI` is not yet entirely free of these dependencies. 
`Lute.CLI.lib` still depends on `Luau.Compiler`. For example, it is still used for `lute compile` subcommand and [here](https://github.com/luau-lang/lute/blob/23498af01f5a09b820a794ba4f164ba93ffd616b/lute/cli/src/climain.cpp#L217). This is based on a quick look at the code. 
The same applies to `VM` API calls for [setupVersionLibrary](https://github.com/luau-lang/lute/blob/23498af01f5a09b820a794ba4f164ba93ffd616b/lute/cli/src/climain.cpp#L631) and other code.

Probably these dependencies can be removed completely, but such a refactor is clearly out of scope for this PR.

Vighnesh-V (approved):
@tyulpan you'll need to merge in latest first, but the changes here look good.

tyulpan:
@Vighnesh-V done.

PR #915, opened 2026-03-30
Refactor AST expr and singleton string type types 

skberkeley:
Part of a larger refactor of AST types so property names are camelCased


lute/luau/src/luau.cpp
```diff
@@ -1148,7 +1148,7 @@ struct AstSerialize : public Luau::AstVisitor
             serializeToken(cstNode->varargAnnotationColonPosition, ":");
         else
             lua_pushnil(L);
-        lua_setfield(L, -2, "varargcolon");
+        lua_setfield(L, -2, "varargColon");
```
annieetang, line 1151: is this `varargColon` or `varArgColon`

annieetang (approved):
quick question and can you add to description that we're camelCasing for property names

lute/luau/src/luau.cpp
skberkeley: keeping it vararg, since it's treated it as one word elsewhere

PR #916, opened 2026-03-30
Refactor stat AST types

skberkeley:
Part of a larger refactor of AST types so that property names are camelCased


annieetang (approved):
ty! can you mention we're using camelCase for property names

Issue #919, opened 2026-04-01
Renaming Lute API export surface


annieetang:
Rename exposed functions, method, and property names in the following `@std/` Lute API export surface to be camelCase

- [x] path @skanosue 
- [x] syntax @skanosue 
- [x] system @skanosue 
- [x] test @Vighnesh-V 
- [x] time @Vighnesh-V 
- [x] fs @skanosue 
- [x] io @annieetang 
- [x] json @annieetang 
- [x] luau @annieetang 
- [x] net @annieetang 
- [x] process @Vighnesh-V 
- [x] stringext @Vighnesh-V 
- [x] tableext @annieetang 
- [x] task @Vighnesh-V

PR #922, opened 2026-04-01
Refactors stringext library to match conventions

Vighnesh-V:
- camelCases the members of the stringext library.
- `local stringext = require("@std/stringext') -> local stringExt = require("@std/stringext")` - not strictly required, but I just feel like it's better.


annieetang (approved):
🤠
hm not sure about `stringExt` vs `stringext` tho - @skanosue thoughts?

annieetang:
also see https://github.com/luau-lang/lute/issues/919

PR #930, opened 2026-04-02
Moves test.run() into @std/test/runner

Vighnesh-V:
We should still support writing tests without using `lute test`, but now the method for doing this is explicitly moved into @std/test/runner.

### DIFF
--- a/examples/testing.luau
+++ b/examples/testing.luau
@@ -1,4 +1,5 @@
 local test = require("@std/test")
+local runner = require("@std/test/runner")
 
 test.case("foo", function(assert)
 	assert.eq(3, 3)
@@ -67,4 +68,4 @@ test.suite("MySuite", function(suite)
 	end)
 end)
 
-require("@std/test/runner").runRegisteredTests()
+runner.runRegisteredTests()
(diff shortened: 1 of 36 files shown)


examples/testing.luau
```diff
@@ -67,4 +67,4 @@ test.suite("MySuite", function(suite)
 	end)
 end)
 
-test.run()
+require("@std/test/runner").runRegisteredTests()
```
annieetang: could we still follow the requiring conventions where the `require("@std/test/runner")` part is at the top of the file? 


```suggestion
local runner = require("@std/test/runner")

-- ...

runner.runRegisteredTests()
```

lute/std/libs/test/init.luau
```diff
@@ -49,7 +44,7 @@ function TestSuite:afterAll(hook: () -> ())
 	self._afterAll = hook
 end
 
-local env: TestEnvironment = { anonymous = {}, suites = {}, suiteindex = {}, caseindex = {} }
+local env = require("@self/env")
```
annieetang: same with this, can requires be at the top of the file?

examples/testing.luau
annieetang: for all of the snapshot tests as well

annieetang (approved):
approving! but noting that we'll follow up with calling `stylua` with require sorting enabled for the whole codebase afterwards for code consistency & quality 😄

PR #934, opened 2026-04-02
Codemods for sorting module requires

Vighnesh-V:
This codemod groups requires by `@lute, @std, @batteries, @self, .... all the other imports`. Requires within each group are sorted lexicographically by require path. Each group of imports renders with a new line.

The current transform library doesn't offer a way to replace without trivia, which ends up duplicating comments at the top of files, for pretty much every file in this repo. I've just returned no replacements, and manually `writeStringToFile` until we can fix that issue.

### DIFF
--- a/lute/std/libs/fs.luau
+++ b/lute/std/libs/fs.luau
@@ -1,13 +1,15 @@
+local fs = require("@lute/fs")
+
+local pathlib = require("@std/path")
+local win32paths = require("@std/path/win32")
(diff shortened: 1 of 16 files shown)


Vighnesh-V:
Setting this back to draft for now - I ran into issues with the way codemods handle replacing w/ trivia, which means that this codemod would produce insane duplication of comments.

Vighnesh-V:
Closing now for the moment.

PR #935, opened 2026-04-03
Contributing glob library to batteries

dphblox:
As part of internal explorations, we built a library that reimplements the Linux `glob` functionality in Lute. Since we anticipate this being widely useful for ecosystem tooling (particularly for defining configuration formats for common tools like package managers, unit test dispatchers, etc.), we want to propose moving our implementation into batteries so that it can be shared and improved communally, rather than fragmenting this functionality across the ecosystem.

The new library has two APIs: `glob.matches()`, which returns a boolean if a path matches a glob pattern, and `glob.discover()`, which efficiently walks the filesystem with pruning to discover all paths that match a glob pattern.

---

**Note:** at lute/lute/cli/commands/lib/ignore.luau there are some similar globbing facilities, but they're pretty divergent in implementation and in API surface. It's nothing we can't expand this new glob surface to standardise; I got Claude to summarise:

> **What they share** — both implement the core glob metacharacters (*, ?, **) with the same semantic meaning (don't cross /, match any char, recurse into subdirectories).
>
> **Where they diverge:**
>
> - **Purpose:** batteries/glob is a filesystem glob that walks the tree and returns matching paths. parseIgnores is gitignore matching that tests whether a path matches an ignore rule.
> - **Matching strategy:** batteries/glob does recursive backtracking, char-by-char, per segment. parseIgnores compiles globs into Lua patterns and uses string.match on full relative paths.
> - **Filesystem:** batteries/glob walks directories via fs.listDirectory. parseIgnores has no filesystem interaction — it just tests a string.
Returns: batteries/glob returns { string } (list of matching paths). parseIgnores returns boolean.
> - **Character classes:** batteries/glob has full [...], [!...], [a-z] range support. parseIgnores doesn't support them — [ passes through as a literal.
> - **Gitignore semantics:** batteries/glob has none. parseIgnores supports ! negation/whitelists, trailing / for directory-only matching, anchoring rules (patterns with / anchor to the .gitignore location, without / they match anywhere), and hierarchical .gitignore chain resolution.
> - **Separator handling:** batteries/glob handles this implicitly — it matches one segment name at a time, so separators never appear in the match input. parseIgnores does it explicitly, baking the platform separator into compiled patterns ([^\]* vs [^/]*).

I imagine we can introduce an `options` table for enabling/disabling features like character classes and gitignore patterns, and the choice between full-string vs per-segment processing, and the idea of whether to compile to regex or just directly ingest the string, is basically just a performance tradeoff.


dphblox:
Looks like I'll need to fix the path separator on Windows, whoops

batteries/glob.luau
```diff
@@ -0,0 +1,212 @@
...
+			inputIndex += 1
+		end
+	end
+	if inputIndex > #inputSegments then
+		return "match"
+	end
+	return "no_match"
+end
+
+local function walk(directory: string, segments: { string }, segmentIndex: number, results: { string })
```
Vighnesh-V: The filesystem library provides a walk method that you might be able to leverage here to simplify this.

tests/batteries/glob.test.luau
```diff
@@ -0,0 +1,285 @@
+local fs = require("@std/fs")
+local glob = require("@batteries/glob")
+local path = require("@std/path")
+local test = require("@std/test")
+
+type Folder = { [string]: Hierarchy }
+type Hierarchy = "file" | Folder
+
+local root = "./test-glob-hierarchy"
```
Vighnesh-V: The testing helpers for creating directory structures are dope. Can you create them in a tmpdir (system.tmpdir() is the call you want here). Then, you can do setup and tear down using `suite:beforeEach` to ensure the directory is sanitized before each test.

tests/batteries/glob.test.luau
```diff
@@ -0,0 +1,285 @@
...
+		assert.eq(glob.match("src/main.lua", "src/main.lua"), true)
+		assert.eq(glob.match("src/main.lua", "lib/main.lua"), false)
+	end)
+
+	-- no match
+
+	suite:case("mismatchedDepth", function(assert)
+		assert.eq(glob.match("a/b", "a/b/c"), false)
+		assert.eq(glob.match("a/b/c", "a/b"), false)
+	end)
```
Vighnesh-V, line 287: for `assert.eq(x, bool)` style comparisons, we have a `that` assertion. You could do: `assert.that(cond)` or `assert.that(not cond)`

batteries/glob.luau
```diff
@@ -0,0 +1,212 @@
...
+
+-- Returns true if the given path matches the glob pattern, without touching the filesystem.
+function glob.match(pattern: path.Pathlike, input: path.Pathlike): boolean
+	local patternParts = path.parse(pattern).parts
+	local inputParts = path.parse(input).parts
+	return matchStep(patternParts, inputParts, 1, 1) == "match"
+end
+
+-- Walks the filesystem and returns all paths matching the glob pattern.
+function glob.discover(input: path.Pathlike): { string }
```
Vighnesh-V: should this return a `{path.Path}`? Most of the std library tries to consume a path.

Vighnesh-V (changes requested):
Some small tidying changes!

batteries/glob.luau
dphblox: The documentation on the walk method is sparse, so I got Claude to summarise how it works; from what I understand, the walk iterator would visit all recursive children with no ability to prune paths that don't match. Implementing the walk ourselves here allows glob to prune away sub-trees which are guaranteed never to match which reduces the number of FS queries needed overall, so I'm inclined to keep this for now. We could consider how to add pruning to fs.walk in the future but seems like a non-trivial design problem; best to keep simple for now IMO.

Vighnesh-V: Cool, that makes sense to me

Vighnesh-V (approved):

PR #945, opened 2026-04-07
Split net into client and server

Nicell:
We want to split net between client and server as they grow in complexity and this is a clear split we can make.

API summary:
- `@lute/net` no longer provides top-level `request` or `serve` and instead exposes `client` and `server`
- `@lute/net/client` and `@lute/net/server` are now separate modules that expose `request` and `serve` respectively
- `@std/net` still provides the top-level request helper.


definitions/net/init.luau
```diff
@@ -14,6 +16,10 @@ export type Response = {
 	ok: boolean,
 }
 
+function client.request(url: string, metadata: Metadata?): Response
+	error("not implemented")
+end
+
 function net.request(url: string, metadata: Metadata?): Response
```
Nicell: do we still want to maintain `net.request` and `net.serve` or should net just export client and server? I'm happy to change either way but left it with both for now

aatxe: I think we should just move them into the separate submodules for the lute builtins. The standard library may re-export them at the top-level of `net` as well if we're so inclined (I think we _would_ for `net.request`, but _wouldn't_ for `net.serve`, at least right now), but I don't think we need to do that kind of thing for the runtime builtin parts.

aatxe (commented):
Looks good to me besides the comment about the re-exports.

definitions/net/init.luau
```diff
@@ -0,0 +1,18 @@
+-- lute-lint-global-ignore(unused_variable)
```
Vighnesh-V: which var is unused here?

Vighnesh-V (approved):

PR #946, opened 2026-04-07
Adds support for a variadic argument to the CLI battery

skberkeley:
This PR adds support for a variadic argument to the CLI parsing battery. Any arguments not consumed by an option or positional argument which come before a `--` is seen will now be consumed by the variadic argument if present. If a variadic argument isn't specified, then arguments before a `--` that aren't otherwise consumed will still be considered forwarded.

I also added a test suite for the CLI battery.


batteries/cli.luau
```diff
@@ -133,16 +138,24 @@ function cli.parse(self: Parser, args: { string }): ()
 			continue
 		end
 
+		-- if we have a variadic argument, we can take this argument as one
```
aatxe:
```suggestion
		-- if we have a variadic argument, we can take this argument as part of it
```

aatxe (approved):

Issue #873, opened 2026-03-12
add io.flush


AsynchronousAI:
io.flush is a needed feature.

aatxe:
@AsynchronousAI Do you want to make a case for _why_ this is a necessary feature? What are you doing in the first place to produce a bunch of buffered writes to stdout that need to be flushed?

AsynchronousAI:
@aatxe I made a chess engine in Luau (10m moves/s movegen :fire:) and was trying to get it to work as a UCI both for stress testing and using as a bot outside of Roblox (LiChess for example). 

For this usecase I went with using Lute for input reading.

It is not possible to have a UCI communication with Python for example unless the output can be flushed manually after input read.

aatxe:
Oh, you're wanting to flush standard _in_? This is incredibly non-obvious, and is hopefully a lesson in why you should actually explain the ticket in the ticket.

AsynchronousAI:
> Oh, you're wanting to flush standard _in_? This is incredibly non-obvious, and is hopefully a lesson in why you should actually explain the ticket in the ticket.

No I mean flush for stdout, not stdin.

Sorry I don't know much about flush for IO and was unaware that `io.flush` wasn't enough context, I was just able to diagnose that as why the communication wasn't working.

Nicell:
@AsynchronousAI are you running lute as a subprocess where the io is piped so your output is buffered? I added an unbuffered force env var in #953. Try setting the environment variable in a newer version of lute and see if that helps.

If this isn't the case, can you please explain what you're doing a bit further?

PR #947, opened 2026-04-07
runtime: add support for `io.flush`

aatxe:
Implements #873 because it's very easy to implement, but when I went to write tests for it, it occurred to me that the use case for this is currently non-obvious since the only mechanisms for writing to stdout are println which flushes the buffer already because of the newline.

I'll leave this as a draft until I see some justification for why this is needed.


aatxe:
The feature request was for flushing `stdin` which is... a fair bit more complicated than this since you cannot fflush an input stream.

PR #948, opened 2026-04-07
Adds a walkthrough for `lute transform`

skberkeley:


docs/guide/code-transforms.md
```diff
@@ -0,0 +1,166 @@
+---
+order: 6
+---
+
+# Code Transformations
+One of the coolest parts of Lute is that we've exposed many of the Luau internals programmatically. That means that things like types, require-resolution, the Luau AST(Abstract Syntax Tree), Luau bytecode, and so on are exposed with public facing apis that you can use. Of these features, one of the slickest things we've exposed is the ability to write automatic code transformations. Using Lute libraries, you can now programmatically  describe how to change Luau source code and `lute transform` will handle running these transformations for you. In the following sections, we'll walk through writing a simple code transformation which doubles any numbers in the source program (we assume some basic level of familiarity with what an [AST](https://en.wikipedia.org/wiki/Abstract_syntax_tree) is). As in the previous library, we'll be creating a new directory, with one file - `transform.luau`. Make sure to run `lute setup` in this directory!
```
hgoldstein:
```suggestion
One of the coolest parts of Lute is that we've exposed parts of Luau like types, require-resolution, the Luau AST (abstract syntax tree), and Luau bytecode through public facing APIs. We've built a tool to write automatic, and deterministic, code transformations on top of these APIs. Using Lute libraries, you can now programmatically describe how to change Luau source code and `lute transform` will handle running these transformations for you. In the following sections, we'll walk through writing a simple code transformation which doubles any numbers in the source program (we assume some basic level of familiarity with what an [AST](https://en.wikipedia.org/wiki/Abstract_syntax_tree) is). As in the previous library, we'll be creating a new directory, with one file - `transform.luau`. Make sure to run `lute setup` in this directory!
```

docs/guide/code-transforms.md
```diff
@@ -0,0 +1,166 @@
...
+```luau
+export type Context<Options = { [any]: any }> = {
+	path: string,
+	source: string,
+	parseresult: syntax.ParseResult,
+	options: Options,
+}
+```
+
+Note that AST's are immutable - rather than mutating them in place, we specify the changes we would like to make using a table of replacements and `lute transform` applies them for us.
```
hgoldstein, line 25: Hm, I would love to embed a note here about string replacements versus node replacements. Is there a way to do footnotes?

docs/guide/code-transforms.md
```diff
@@ -0,0 +1,166 @@
...
+
+Lute provides two ways of writing transforms - a declarative, query based approach, and a visitor based approach. Since we recommend using the query based approach, we'll focus on that, and show how to write a visitor based approach at the end.
```
hgoldstein:
```suggestion
Lute provides two ways of writing transforms: a query based approach and a visitor based approach. Most transformations can be done with queries. We'll start with that and later write a visitor based transformation.
```

docs/guide/code-transforms.md
```diff
@@ -0,0 +1,166 @@
...
+
+### Query Based AST transforms
+
+In this section, we build up an example of a declarative 'query' based implementation which doubles any numbers found in the source program. We begin by selecting every AST node we're interested in - in this case, every number:
```
hgoldstein:
```suggestion
In this section, we build up an example of a query based implementation which doubles any numbers found in the source program. We begin by selecting every AST node we're interested in - in this case, every number:
```

hgoldstein (commented):
Meta note: let's make sure all of the examples type check.

docs/guide/code-transforms.md
Vighnesh-V: You can do:
```
:::info
Your info block here
:::
```

If you want to highlight this fact inline? We removed node replacements.

hgoldstein: Oh I meant: most people when they look at this API think "hey what if I want to build a new node" because that's what you do in other libraries of this kind. Adding a note of why that's _not_ the API and that we may add something like that in the future, but it would be different than actually building the node by hand.

docs/guide/dev-tooling/index.md
```diff
@@ -14,3 +14,9 @@ to consider using on a day to day basis.
 written using the `@std/test` library in `.spec.luau` or `.test.luau` files, in
 a `tests/` directory and `lute test` will handle discovering and running these
 tests for you.
+
+
+
+## [Linting](../../cli/lint/index.md)
+
+lute lint is a programmable linter for Luau code, shipped as part of Lute. It is programmable - you can write custom rules for your repository in Luau. It ships with a number of builtin rules, which you can run on your current directory with `lute lint`. For more information, check out the [lint documentation](../../cli/lint/index.md)!
```
aatxe:
```suggestion
`lute lint` is a programmable linter for Luau code, shipped as part of Lute. It is programmable - you can write custom rules for your repository in Luau. It ships with a number of builtin rules, which you can run on your current directory with `lute lint`. For more information, check out the [lint documentation](../../cli/lint/index.md)!
```

aatxe (approved):

PR #949, opened 2026-04-07
Improves ModuleResolver's `resolveModule` by using `LuteReporter` and adds tests

annieetang:
Removes the use of `printf` in `LuteTypeCheckModuleResolver`'s `resolveModule` and uses `reporter.reportError` instead. To do this, I created a `TCReporter` struct that mimics how `CLIReporter` works.
Additionally, finally adding tests for `resolveModule`


tests/src/moduleresolver.test.cpp
```diff
@@ -5,17 +5,20 @@
...
 #include "luteprojectroot.h"
+#include "testreporter.h"
 
 TEST_CASE("moduleresolver_read_source")
 {
     std::string root = getLuteProjectRootAbsolute();
     std::string file = joinPaths(root, "tests/src/resolver/mainmodule.luau");
-    Luau::LuteTypeCheckModuleResolver resolver;
+    TestReporter reporter;
+    Luau::LuteTypeCheckModuleResolver resolver{reporter};
```
Vighnesh-V: Can you instead create a fixture for this? Take a look at CliRunFixture for an example of how to do this. Then future tests can simply to `getreporter()`.

lute/luau/src/luau.cpp
```diff
@@ -3032,7 +3033,8 @@ int typeofModule_luau(lua_State* L)
 {
     std::string modulePath = luaL_checkstring(L, 1);
 
-    Luau::LuteTypeCheckModuleResolver moduleResolver;
+    TCReporter tcReporter;
+    Luau::LuteTypeCheckModuleResolver moduleResolver{tcReporter};
```
Vighnesh-V: Can you make a constructor for this and pipe this data in? In this situation, there is a reporter on the runtime, so you can do `getRuntime(L)->reporter`

lute/luau/include/lute/tcmoduleresolver.h
```diff
@@ -12,7 +14,7 @@ namespace Luau
 // Based on CliFileResolver in Analyze.cpp.
 struct LuteTypeCheckModuleResolver : Luau::FileResolver
 {
-    LuteTypeCheckModuleResolver() = default;
+    LuteTypeCheckModuleResolver(LuteReporter& reporter) : reporter(reporter) {}
```
Vighnesh-V: Can you move this to the cpp file :)

tests/src/moduleresolver.test.cpp
```diff
@@ -98,3 +101,76 @@ TEST_CASE("moduleresolver_typecheck_resolve")
...
+    Luau::ModuleInfo context;
+    context.name = mainLuau;
+
+    Luau::TypeCheckLimits limits;
+
+    SUBCASE("resolves_std_module")
+    {
+        std::string path = "@std/process";
+        Luau::AstArray<char> value{path.data(), path.size()};
+        Luau::AstExprConstantString strNode(Luau::Location{}, value, Luau::AstExprConstantString::QuotedSimple);
```
skberkeley: is it worth adding a helper function to make all these?

lute/luau/include/lute/tcreporter.h
```diff
@@ -0,0 +1,19 @@
...
+    void reportError(const std::string& message) override
+    {
+        std::fprintf(stderr, "%s\n", message.c_str());
+    }
+
+    void reportOutput(const std::string& message) override
+    {
+        std::fprintf(stdout, "%s\n", message.c_str());
+    }
+};
```
Vighnesh-V: Note, we don't need to create a separate reporter here! The LuteReporter is the abstract reporting class. Tests will initialize it with TestReporter, and the cli will initialize it with the `CliReporter`.

skberkeley (approved):

tests/src/moduleresolver.test.cpp
annieetang: ya prob! will do

Vighnesh-V (changes requested):

lute/luau/src/luau.cpp
annieetang: ya! wasn't sure we wanted to use runtime / cli here, but making the changes now

Vighnesh-V (approved):

PR #950, opened 2026-04-07
Update lute transform to use the CLI battery

skberkeley:
This results in a slight API change for the CLI. Previously, options forwarded on to the code transformation scripts could be passed anywhere, not just after a `--`, but now we only forward them if they come after a `--`. This is because we now use the CLI battery's shiny new variadic argument feature to parse file paths to be transformed.


lute/cli/commands/transform/init.luau
```diff
@@ -142,33 +143,92 @@ local function applyMigration(
 	end
 end
 
+local USAGE = "Usage: lute transform [options] <migration-path> [files...] [-- [--option value...]]"
+
+local function printHelp()
```
aatxe, line 148: I'm fine with this re-implementation, but just wanted to make sure you saw the battery _does_ have a "generate help dialogue" function too.

Vighnesh-V (approved):
my goat

aatxe (approved):

lute/cli/commands/transform/init.luau
skberkeley: Yeah I've seen it, but think it's also the case that none of our commands use it lol. tbf doing it manually lets us present the information how we want

PR #952, opened 2026-04-08
Add support to `lute compile` for bundling .config.luau

jkelaty-rbx:
Adds support for the compile command to support bundling .config.luau for require alias resolution.


lute/cli/src/staticrequires.cpp
```diff
@@ -90,11 +91,18 @@ void StaticRequireTracer::trace(const std::string& entryPoint)
...
-                    luaurcAbsolutePaths.insert(luaurcPath);
+                std::string luauConfigPath = dir + "/" + Luau::kLuauConfigName;
+
+                bool luaurcExists = isFile(luaurcPath);
+                bool luauConfigExists = isFile(luauConfigPath);
+
+                if (luaurcExists)
+                    configAbsolutePaths.insert(luaurcPath);
+                if (luauConfigExists)
+                    configAbsolutePaths.insert(luauConfigPath);
```
Vighnesh-V: Should we be inserting both a .luaurc and a config.luau?

Vighnesh-V: Imo, we should insert 1 (probably preferring config.luau), and warn them if they have two. If the staticrequires file doesn't already have access to a LuteReporter, we should thread one through and report a warning here.

jkelaty-rbx: Done!

aatxe (approved):

PR #954, opened 2026-04-08
Move lint and transform types into std, add chained aliases pointing to them

skberkeley:
Makes `@lint` and `@transform` generally available to consumers of lute by moving the types they refer to into `std` and modifying `lute setup` to generate chained aliases pointing to them. Also:
- Restricts how much of the lute codebase within which they're available by modifying .luaurc files
- Updates relevant requires to use `@lint` and `@transform`
- Makes the needed change in `clifvs` to support this


lute/require/src/clivfs.cpp
```diff
@@ -76,12 +76,17 @@ std::optional<std::string> CliVfs::getContents(const std::string& path) const
...
 {
-    // Currently, we do not support .luaurc files in CLI commands.
-    return std::nullopt;
+    return R"(
+{
+    "aliases": {
+        "lint": "@std/commands/lint/types",
+        "transform": "@std/commands/transform/types"
+    }
+}
```
Vighnesh-V: nit: constexpr allocate this and then return kCliConfig

aatxe (approved):

PR #957, opened 2026-04-08
Support yielding in net serve handler

Nicell:
net's server handler would just immediately return a response even if the lua thread yielded, so yielding was basically broken. by adding a completion hook to `runOnce` we can properly wait and notify the server when the lua thread yields

### DIFF
--- a/lute/runtime/include/lute/runtime.h
+++ b/lute/runtime/include/lute/runtime.h
@@ -31,9 +31,7 @@ struct ThreadToContinue
 // they finish or error after one or more yields.
 struct ThreadCompletionHandler
 {
-    void (*onFinish)(lua_State* L, int status, void* userdata) = nullptr;
-    void (*destroy)(void* userdata) = nullptr;
-    void* userdata = nullptr;
+    std::function<void(lua_State* L, int status)> onFinish;
 };
 
 struct StepErr
--- a/lute/runtime/include/lute/runtime.h
+++ b/lute/runtime/include/lute/runtime.h
@@ -27,6 +27,15 @@ struct ThreadToContinue
     std::function<void()> cont;
 };
 
+// Optional completion hook for threads that need native follow-up work once
+// they finish or error after one or more yields.
+struct ThreadCompletionHandler
+{
+    void (*onFinish)(lua_State* L, int status, void* userdata) = nullptr;
+    void (*destroy)(void* userdata) = nullptr;
+    void* userdata = nullptr;
+};
+
 struct StepErr
 {
     lua_State* L;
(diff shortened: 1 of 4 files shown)


lute/runtime/include/lute/runtime.h
```diff
@@ -27,6 +27,15 @@ struct ThreadToContinue
...
 };
 
+// Optional completion hook for threads that need native follow-up work once
+// they finish or error after one or more yields.
+struct ThreadCompletionHandler
+{
+    void (*onFinish)(lua_State* L, int status, void* userdata) = nullptr;
+    void (*destroy)(void* userdata) = nullptr;
+    void* userdata = nullptr;
+};
```
Vighnesh-V: Is there a reason these are C style pointers instead of std::function?

Nicell: I'm a C++ noob. Updated to use std::function

lute/runtime/include/lute/runtime.h
```diff
@@ -98,8 +110,12 @@ struct Runtime
     std::vector<std::string> args;
 
 private:
+    bool runThreadCompletionHandler(lua_State* L, int status);
+    void clearThreadCompletionHandler(lua_State* L);
+
     std::mutex continuationMutex;
     std::vector<std::function<void()>> continuations;
+    std::unordered_map<lua_State*, ThreadCompletionHandler> threadCompletionHandlers;
```
Vighnesh-V, line 143: Q should there be multiple thread completion handlers for a given lua_State? This assumes that there's exactly one right.

lute/runtime/include/lute/runtime.h
```diff
@@ -73,6 +81,10 @@ struct Runtime
     // argPusher should push arguments onto the stack and return the count.
     void scheduleLuauCallback(std::shared_ptr<Ref> callbackRef, std::function<int(lua_State*)> argPusher);
 
+    // Attach native follow-up work to a specific Luau thread. The hook will run
+    // once that thread eventually returns or errors after any number of yields.
+    void setThreadCompletionHandler(lua_State* L, ThreadCompletionHandler completion);
```
Vighnesh-V:
```suggestion
    void addThreadCompletionHandler(lua_State* L, ThreadCompletionHandler completion);
```

lute/runtime/src/runtime.cpp
```diff
@@ -110,6 +110,7 @@ RuntimeStep Runtime::runOnce()
     int co_status = lua_costatus(GL, L);
     if (co_status == LUA_COFIN)
     {
+        clearThreadCompletionHandler(L);
```
Vighnesh-V, line 117: Is this the only case in which we need to clean up a thread completion handler?

Vighnesh-V (approved):
left some comments, but not blocking :)

lute/runtime/src/runtime.cpp
Nicell: it was also getting cleaned up in `runThreadCompletionHandler` but it was .erasing directly. I updated it to call this clear method, too, so it's clearer

Vighnesh-V: ok awesome, thank you!

PR #959, opened 2026-04-08
runtime: introduce `LuteLibrary` shared interface for runtime modules.

aatxe:
There's a lot of inconsistency between the different runtime library definitions, and it's bothered me for a while. To try to make things more structured, this PR introduces a shared interface `LuteLibrary` and refactors each library namespace to instead be a `struct` that implements that `LuteLibrary` interface. The goal here is to make it more difficult to author a library that isn't structured how we would expect, and to bring all of the existing libraries into that uniform structure.

---

I'm not necessarily 100% convinced that this is worth it, but I think it came together well enough that I want to solicit other opinions on it. Does this look and feel better to work on going forward? It definitely looks to make headers, especially, a lot more uniform. The static_asserts in particular are a bit janky, but it seems arguably nicer than the alternative of doing an abstract class with all of these fields virtual.

### DIFF
--- a/lute/common/include/lute/library.h
+++ b/lute/common/include/lute/library.h
@@ -1,0 +1,20 @@
+#pragma once
+
+#include "lua.h"
+#include "lualib.h"
+
+// CRTP base for all lute runtime libraries.
+// Each Derived must provide:
+//   static constexpr const char kName[];   -- the Lua global name (e.g. "task")
+//   static int pushLibrary(lua_State* L);  -- push the library table onto the stack
+//   static const luaL_Reg lib[];           -- function registration table (where applicable)
+template<typename Derived>
+struct LuteLibrary
+{
+    static int openAsGlobal(lua_State* L)
+    {
+        Derived::pushLibrary(L);
+        lua_setglobal(L, Derived::kName);
+        return 1;
+    }
+};
--- a/lute/common/include/lute/library.h
+++ b/lute/common/include/lute/library.h
@@ -5,9 +5,10 @@
 
 // CRTP base for all lute runtime libraries.
 // Each Derived must provide:
-//   static constexpr const char kName[];   -- the Lua global name (e.g. "task")
-//   static int pushLibrary(lua_State* L);  -- push the library table onto the stack
-//   static const luaL_Reg lib[];           -- function registration table (where applicable)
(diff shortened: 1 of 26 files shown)


lute/common/include/lute/library.h
```diff
@@ -0,0 +1,52 @@
...
+        );
+    }
+
+    static int openAsGlobal(lua_State* L)
+    {
+        verifyInterface();
+        Derived::pushLibrary(L);
+        lua_setglobal(L, Derived::kName);
+        return 1;
+    }
```
Vighnesh-V, line 51: I have a small q - this function uses `::pushLibrary` and `::kName` so it makes sense to assert for them. Does it still make sense to assert for lib and properties, since these can just be implementation details wherever they are needed, and the derived classes don't have to use properties if they aren't exported?

lute/common/include/lute/library.h
```diff
@@ -0,0 +1,52 @@
...
+            std::is_array_v<decltype(Derived::properties)> &&
+                std::is_same_v<std::remove_extent_t<decltype(Derived::properties)>, const char* const>,
+            "LuteLibrary must declare: static const char* const properties[]"
+        );
+    }
+
+    static int openAsGlobal(lua_State* L)
+    {
+        verifyInterface();
+        Derived::pushLibrary(L);
```
Vighnesh-V, line 48: Do you think it's possible for this to exist as an implementation on the `LuteLibrary`? I'm looking at the existing `pushLibraries` we have and it looks like they're all doing pretty much the same thing? I guess metatable initialization is the main difference here right?

Vighnesh-V (commented):
I think this is really good - I'm just curious if we can move even more stuff into generic lute initialization.

lute/common/include/lute/library.h
aatxe: I wanted to assert for all of them just for the sake of having all of them be present in all libraries, essentially.

lute/common/include/lute/library.h
aatxe: Initializing the metatables or any other initialization that needs to happen is a difference, and then we'd also need to turn properties into a paired thing, similar to the libraries, which would require us to move the setup code for each individual properties values into functions. That could very well be alright though.

vrn-sn (approved):
Looks good from my end! CRTP's the best

PR #960, opened 2026-04-08
Sets up a release job for Lute

Vighnesh-V:
The job that will be used for automating releases and the job that will be used for nightlies share a lot of common actions. This PR separates the common chunk out into a re-usable workflow, called shared-build which is parameterized on the branch name.

1) Nightlies can be manually kicked off, but will continue to run with the current cadence. These will automatically run of the 'primary' branch.
2) Releases will be kicked off on release/v{Major}.{Minor}.x branches. 

In both cases the job will check out the branch and derive the tag name directly from the get_version.cmake command. If the branch is primary, we'll get 'nightly.YYYMMDD' appended.

### DIFF
--- a/.github/workflows/shared-build.yml
+++ b/.github/workflows/shared-build.yml
@@ -1,0 +1,140 @@
+name: _shared_release
+
+on:
+  workflow_call:
+    inputs:
+      branch:
+        description: "Branch to use for the release. 'primary' triggers a nightly build."
+        required: false
+        default: "primary"
+        type: string
+
+env:
+  IS_NIGHTLY: ${{ inputs.branch == 'primary' }}
+
+jobs:
+  check:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+
+    steps:
+      - name: Validate CI status
+        id: check
+        run: |
+          LAST_STATUS=$(curl -s -H "Authorization: Bearer ${{ secrets.GITHUB_TOKEN }}" \
+            "https://api.github.com/repos/luau-lang/lute/actions/workflows/ci.yml/runs?branch=${{ inputs.branch }}&status=completed&per_page=1" \
+            | jq -r '.workflow_runs[0].conclusion')
+          if [[ "$LAST_STATUS" != "success" ]]; then
+            echo "The latest CI run on '${{ inputs.branch }}' branch did not succeed. Aborting release."
+            exit 1
+          fi
+
+  build:
+    strategy:
+      matrix:
+        include:
+          - os: ubuntu-22.04
+            artifact_name: lute-linux-x86_64
+            options: --c-compiler clang --cxx-compiler clang++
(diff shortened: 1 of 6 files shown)


.github/workflows/#shared-build.yml#
```diff
@@ -0,0 +1,136 @@
+name: _shared_release
```
aatxe: This file is an editor artifact, right?

.github/workflows/.#shared-build.yml
```diff
@@ -0,0 +1 @@
+vvijay@HQ-KNWWYRX7WQ.712
```
aatxe: Ditto about this being an editor artifact

.github/workflows/shared-build.yml
```diff
@@ -0,0 +1,140 @@
+name: _shared_release
```
aatxe: Feel like this should probably be _called_ `shared-release` then? Or maybe `release-backend` or something of that sort? `shared-build` feels a little odd since it does do release-specific workflow tasks, not just building the artifact.

RELEASE.md
```diff
@@ -0,0 +1,23 @@
+# Release Process
+
+Lute uses branches with the format: `release/v{Major}.{Minor}.x` to store changes associated with a given release.
+We tag commits with the format `v{Major}.{Minor}.{Patch}` to indicate that a commit should be build as that release version.
+Changes like security updates, hotfixes, etc, occur on these branches and go through the code review process in order to be merged.
```
aatxe: There's probably some nuance here. I think changes that would affect all future versions should probably always be against `primary` (or at least generally be?), with the idea that we'll, as maintainers, cherry-pick them to any supported release branches as-needed if there's a high degree of urgency. If the issue _only_ exists in that version, then yeah, it should definitely just be against the version. I think a workflow that has most contributors mostly working against `primary` is probably the easiest to support.

aatxe (commented):
Mostly looks good, but some comments.

.github/workflows/shared-build.yml
Vighnesh-V: I think shared-release is a better name for this file!

.github/workflows/#shared-build.yml#
Vighnesh-V: Yeah, I've gone ahead and deleted these.

.github/workflows/shared-build.yml
Vighnesh-V: on second thought, I like release-common?

RELEASE.md
Vighnesh-V: I'll expand on this a little in the documentation.

.github/workflows/shared-build.yml
aatxe: yeah, `release-common` is a good name

aatxe (approved):

PR #965, opened 2026-04-09
Fix net server query parsing and empty response bodies

Nicell:
Fixes two issues in `@lute/net/server`:
- query parsing handles empty queries and parses the final key/value pair correctly
- response tables without a body no longer assign a null pointer

### DIFF
--- a/lute/net/src/server.cpp
+++ b/lute/net/src/server.cpp
@@ -48,11 +48,15 @@ struct ServerLoopState
 static void parseQuery(const std::string_view& query, lua_State* L)
 {
     lua_createtable(L, 0, 0);
-    size_t start = 1; // Skip the '?'
-    size_t end = query.find('&');
-    while (end != std::string::npos)
+    size_t start = (!query.empty() && query[0] == '?') ? 1 : 0;
+    if (start >= query.size())
+        return;
+
+    while (start < query.size())
     {
-        std::string_view pair = std::string_view(query.data() + start, end - start);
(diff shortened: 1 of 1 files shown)


lute/net/src/server.cpp
```diff
@@ -162,10 +159,14 @@ static void handleResponse(auto* res, lua_State* L, int responseIndex)
...
-    std::string body = "";
-    size_t bodyLength;
-    const char* bodyData = lua_tolstring(L, -1, &bodyLength);
-    body.assign(bodyData, bodyData + bodyLength);
+    std::string body;
+    if (!lua_isnil(L, -1))
+    {
+        size_t bodyLength = 0;
+        const char* bodyData = lua_tolstring(L, -1, &bodyLength);
+        if (bodyData)
```
Vighnesh-V, line 167: should this be a length check instead?

Vighnesh-V (commented):
Left a comment, but otherwise looks good, thanks for fixing these bugs.

lute/net/src/server.cpp
Nicell: I don't think so. An empty string is an ok body (`""`). Maybe something more clear would be only allowing the string type with `lua_type(L, -1) == LUA_TSTRING`? but right now I believe this is correctly checking that bodyData was turned into a string (so if it were a boolean/table it'd fail the if check).

Vighnesh-V (approved):

PR #968, opened 2026-04-10
Support requiring native modules

jkelaty-rbx:
Note: I'm mostly leaving this open for historical context, though I am also happy to revive this PR if there's interest in moving forward with it.

## Motivation
Enables users to write C/C++ libraries that extend Lute with native functionality (e.g., bindings to libgit2, SQLite, image processing) without forking Lute. Native modules follow the same `luteopen_*` convention as built-in `@lute/` modules and are loaded through the standard `require()` path.

## Summary
- Extends `require()` to load native shared libraries (`.dylib`, `.so`, `.dll`) from disk via `uv_dlopen`/`uv_dlsym`
- Entry point convention: `libfoo.dylib` → `extern "C" int luteopen_foo(lua_State* L)`
- Library handles are anchored in a `_NATIVELIBS` registry table and closed when the `lua_State` is torn down
- Includes a working example at `examples/native-extension/`
- Type checking is supported via a types file embedded in the shared object


jkelaty-rbx:
This PR has gotten pretty stale, and I'd imagine we're going to continue exploring alternative FFIs in lieu of this, so gonna close this for now!

PR #976, opened 2026-04-14
Orders API Reference tab by std/ then lute/

Vighnesh-V:
This change just re-orders the automatically generated api reference so that the doc generator command uses the argument order of docs to set their order in the reference tab.


docs/README.md
```diff
@@ -11,4 +11,4 @@ The terminal output should then tell you which `localhost` port the docsite is r
 
 ## Regenerating documentation files
 
-If you want to regenerate the documentation files in `docs/reference`, you will need to run `<path-to-lute> doc -o ./ ../definitions ../lute/std/libs` from the `docs` folder using a version of `lute` which supports `lute doc`.
+If you want to regenerate the documentation files in `docs/reference`, you will need to run `<path-to-lute> doc -o ./ ../lute/std/libs ../definitions` from the `docs` folder using a version of `lute` which supports `lute doc`. The documentation will be ordered in the sidebar in the order you pass the source folders.
```
aatxe: It is slightly cursed that this is true, but fair enough lol

aatxe (approved):
Looks good for just the ordering bits, guess we're going to split it up visually a bit more in a follow up PR.

Vighnesh-V:
I'm going to split it up right now I think - lemme put this back in draft.

docs/lute/index.md
```diff
@@ -0,0 +1,5 @@
+---
+order: 4
+---
+
+# (@lute) Lute Libraries
```
aatxe:
```suggestion
# (@lute) Lute Runtime Builtins
```

Maybe a better heading? I don't quite like `Lute Libraries`, I think.

docs/.gitignore
```diff
@@ -2,4 +2,10 @@ node_modules
...
 reference
-docs/
\ No newline at end of file
+docs/
+
+# Auto-generated API docs (keep root index.md as table of contents)
+lute/*
+!lute/index.md
+std/*
+!std/index.md
```
aatxe:
```suggestion
!std/index.md

```

aatxe (approved):

docs/lute/index.md
Vighnesh-V: I set it to Lute Libraries!

docs/.gitignore
Vighnesh-V: added!

annieetang (approved):
ty!

PR #981, opened 2026-04-15
`lute pkg`: Github releases

jkelaty-rbx:
Enables loom to support fetching and installing native modules: https://github.com/luau-lang/lute/pull/968


jkelaty-rbx:
This isn't needed (by me) until something like this lands: https://github.com/luau-lang/lute/pull/968 closing for now but will definitely revisit this once we make movement on lute's FFI

PR #982, opened 2026-04-16
docs: capture exported types in current doc generation

annieetang:
Captures exported types and displays them on the doc page now through same string pattern parsing as functions. Also adds a summary table for the functions/properties that link to the relevant section since the docs are getting longer. 

Could probably be written better but this was written a bit hastily since we're trying to cut a release today!

see https://github.com/luau-lang/lute/pull/983 `TESTDIR` for new docs output

example:
https://github.com/luau-lang/lute/blob/annietang/exported_types_doc_preview/TESTDIR/reference/std/luau.md

image
image


lute/cli/commands/doc/init.luau
```diff
@@ -207,57 +408,71 @@ local function generateMarkdown(
...
 
-	-- LUAUFIX: we should not need to explicitly instantiate the type of `table.sort` here
-	table.sort<<Definition>>(definitions, function(a, b)
-		return a.name < b.name
-	end)
+	-- Summary table
+	if #definitions > 0 or #typeDefinitions > 0 then
+		table.insert(lines, "## Summary")
+		table.insert(lines, "")
+		table.insert(lines, "| Function / Property | Description |")
```
aatxe:
```suggestion
		table.insert(lines, "| Entry | Description |")
```

This table also includes types, right? I think we should probably just call it an entry or something else generic since it's basically "anything we might be documenting."

annieetang: oo yes ty for the catch

aatxe (approved):

PR #984, opened 2026-04-16
Update the website homepage.

aatxe:
The website homepage needs a bunch of work to be production-ready IMO. So, let's do that! We should actually explain what Lute is and what we're doing here.

The result:

image


annieetang (approved):
ty!!!

PR #988, opened 2026-04-16
website: write more extensive installation instructions for lute

aatxe:
Update the installation instructions to include instructions for Foreman and also to mention stable releases now, and that we follow semver.


annieetang (approved):
WOO!!!

PR #994, opened 2026-04-17
TOML flat arrays

Shchvova:
# Add inline array support to TOML battery

## Problem

Flat arrays of primitives (strings, numbers, booleans) were not handled as a distinct case in either the serializer or deserializer.

- **Serializer**: a value like `{ "--stdio" }` was treated as an array of tables and emitted as `[[key]]` sections instead of `key = ["--stdio"]`
- **Deserializer**: a line like `args = ["--stdio"]` fell through all value checks and was stored as the raw string `'["--stdio"]'`
- **Side effect**: because `hasNestedTables` counted flat arrays as nested, tables containing them never got a `[section]` header, producing invalid TOML

## Changes

**`batteries/toml.luau`**

- Added `isFlatArray` helper — returns true for non-empty sequence tables whose values are all non-table
- Updated `hasNestedTables` to treat flat arrays as scalar values, so `[section]` headers are emitted correctly for tables that contain them
- Serializer now emits flat primitive arrays as TOML inline arrays (`key = ["a", "b"]`) instead of `[[key]]` array-of-table sections; array-of-table behavior is unchanged
- Deserializer now parses inline arrays into Lua sequences; handles strings, numbers, and booleans

**`tests/batteries/toml.test.luau`**

Added three new suites (`TOML inline array deserialization`, `TOML inline array serialization`, `TOML inline array round-trip`) covering: single and multi-element string arrays, number and boolean arrays, mixed primitive arrays, correct `[section]` header generation when a table contains an inline array, array-of-tables still using `[[...]]`, and a full round-trip of a real-world MCP server config.

## Example

```toml
[mcp_servers.Roblox_Studio]
args = ["--stdio"]
command = "/Applications/RobloxStudio.app/Contents/MacOS/StudioMCP"
enabled = true
```

This previously could not be serialized or deserialized correctly. Both directions now work and round-trip cleanly.

<details>
  <summary>problematic results</summary>

#### `test.luau`
```luau

local var = [[
[mcp_servers.Roblox_Studio]
args = ["--stdio"]
command = "/Applications/RobloxStudio.app/Contents/MacOS/StudioMCP"
enabled = true
]]

local toml = require("./batteries/toml")
local json = require "@std/json"

local parsed = toml.deserialize(var)
print(json.serialize(parsed, true))
```  

#### current `toml.luau`

```
❯ lute run test.luau
{
    "mcp_servers": {
        "Roblox_Studio": {
            "command": "/Applications/RobloxStudio.app/Contents/MacOS/StudioMCP",
            "enabled": true,
            "args": "[\"--stdio\"]"
        }
    }
}
```
#### patched `toml.luau`
```
❯ lute run test.luau
{
    "mcp_servers": {
        "Roblox_Studio": {
            "command": "/Applications/RobloxStudio.app/Contents/MacOS/StudioMCP",
            "enabled": true,
            "args": [
                "--stdio"
            ]
        }
    }
}
```

</details>


tests/batteries/toml.test.luau
```diff
@@ -74,3 +74,137 @@ test.suite("TOML serialization", function(suite)
 		assert.eq(second.images["icon.png"].hash, "abc")
 	end)
 end)
+
+test.suite("TOML inline array deserialization", function(suite)
```
Vighnesh-V: PascalCase suite names - `TOMLInlineArrayDeserialization`. Ditto for the cases below.

Shchvova: sure

batteries/toml.luau
```diff
@@ -47,9 +61,18 @@ local function tableToToml(tbl: { [any]: any }, parent: string?): string
...
+				if isFlatArray(v) then
+					local items = {}
+					for _, item in v do
+						table.insert(items, serializeValue(item))
+					end
+					hasDirectValues = true
+					result ..= quoteKey(k) .. " = [" .. table.concat(items, ", ") .. "]\n"
+				else
+					for _, entry in v do
+						result ..= "\n[[" .. (parent and parent .. "." or "") .. quoteKey(k) .. "]]\n"
```
Vighnesh-V, line 73: Might be clearer to use string interpolation here:
`\n[[{...}{quoteKey(k)}...`

Shchvova: this is pre-existing code moved to the `if` statement, should I refactor the whole file to use format strings? I would rather make it a separate follow-up tbh

edit: diff view with spaces ignored:

Image

batteries/toml.luau
```diff
@@ -190,6 +213,26 @@ local function deserialize(input: string): { [string]: unknown }
...
+						if item ~= "" then
+							if string.match(item, '^"(.*)"$') or string.match(item, "^'(.*)'$") then
+								table.insert(arr, string.sub(item, 2, -2))
+							elseif tonumber(item) then
+								table.insert(arr, tonumber(item) :: number)
+							elseif item == "true" then
+								table.insert(arr, true)
+							elseif item == "false" then
+								table.insert(arr, false)
+							end
```
Vighnesh-V, line 231: Should there be a fallback / error here if the item doesn't match string/number/bool?

batteries/toml.luau
```diff
@@ -190,6 +213,26 @@ local function deserialize(input: string): { [string]: unknown }
 				value = string.gsub(value, "\\\\", "\\")
 				value = string.gsub(value, "\\n", "\n")
 				value = string.gsub(value, "\\t", "\t")
+			elseif string.sub(value, 1, 1) == "[" then
+				local content = string.match(value, "^%[(.*)%]$")
+				if content then
+					local arr: { boolean | string | number } = {}
+					for item in string.gmatch(content .. ",", "([^,]+),") do
```
Vighnesh-V, line 220: I'm guessing this matches things that end with comma right? What happens if the actual toml string has a ',' in the string?

batteries/toml.luau
```diff
@@ -190,6 +213,26 @@ local function deserialize(input: string): { [string]: unknown }
 				value = string.gsub(value, "\\\\", "\\")
 				value = string.gsub(value, "\\n", "\n")
 				value = string.gsub(value, "\\t", "\t")
+			elseif string.sub(value, 1, 1) == "[" then
+				local content = string.match(value, "^%[(.*)%]$")
```
Vighnesh-V, line 217: It would be helpful to get a comment explaining these matches - my lua regex is very rusty.

Vighnesh-V (commented):
Just a few comments.

batteries/toml.luau
aatxe: yeah, it's fine to leave that for a follow-up!

batteries/toml.luau
Shchvova: Ah, sure - `^` and `$` to match beginning and end of string to avoid getting partial match, `%` is escape symbol in Lua regex, so `%[` and `%]` just mean literal `[` and `]`, and `(.*)` is a capture group matching anything in between.

batteries/toml.luau
Shchvova: good spot. I think I will give up since there is way too much to fix though... It seems that for my needs I need to support way bigger set of TOML than just this

Shchvova:
I realized that for my needs we would have to support much bigger part of TOML spec than Lute does now, like now it don't support embedded dicts. For exmaple

```toml
[mcp_servers.Roblox.Studio]
version = { version.number = "=1.2.3"}
```

is now deserialized as 

```json
{
    "mcp_servers": {
        "Roblox_Studio": {
            "version": "{ version.number = \"=1.2.3\"}"
        }
    }
}
```

rather than
```json
{
  "mcp_servers": {
    "Roblox_Studio": {
      "version": {
        "version": {
          "number": "=1.2.3"
        }
      }
    }
  }
}
```

Basically current implementation works exclusively for small subset of TOML and it doesn't work for my purpose (modifying some 3rd party configs)

PR #1002, opened 2026-04-18
Parallel test runners

Nicell:
Creates a worker for each system thread using the VM lib and then workers pull test files as they become free to distribute test load.

Tests ran in ~8.0s on 1.0.0. With #1001, we got down to ~5.4s. With parallel execution, we reach ~2.8s. 🎉 

<details>
<summary>Details</summary>

```
Benchmark 1: lute test (1.0.0)
  Time (mean ± σ):      7.967 s ±  0.084 s    [User: 3.547 s, System: 2.175 s]
  Range (min … max):    7.813 s …  8.093 s    10 runs
 
Benchmark 2: lute test (primary, has discovery fixes)
  Time (mean ± σ):      5.386 s ±  0.052 s    [User: 2.281 s, System: 0.690 s]
  Range (min … max):    5.312 s …  5.473 s    10 runs
 
Benchmark 3: lute test (parallel execution)
  Time (mean ± σ):      2.807 s ±  0.050 s    [User: 2.608 s, System: 1.104 s]
  Range (min … max):    2.750 s …  2.909 s    10 runs
```

</details>

### DIFF
--- a/lute/cli/commands/test/init.luau
+++ b/lute/cli/commands/test/init.luau
@@ -66,15 +66,16 @@ local function runtests(
 		local done = 0
 		for _ = 1, nworkers do
 			task.spawn(function()
+				local current: path.Path? = nil
 				xpcall(function()
 					local worker = vm.create("@self/worker")
 					while true do
-						local p = takenext()
-						if not p then
+						current = takenext()
+						if not current then
 							break
 						end
 						local r = worker.runfile({
-							file = path.format(p),
+							file = path.format(current),
 							suite = suiteName,
 							case = caseName,
 							verbose = runOptions.verbose,
@@ -87,18 +88,18 @@ local function runtests(
 						end
 					end
 				end, function(err: any)
+					local where = if current then path.format(current) else "<worker setup>"
 					result.total += 1
 					result.failed += 1
 					table.insert(result.failures, {
-						test = "<worker>",
+						test = where,
 						failure = {
 							__tag = "runtime_error",
-							msg = `worker errored: {tostring(err)}`,
+							msg = `worker errored on {where}: {tostring(err)}`,
 							stacktrace = debug.traceback(nil, 2),
 						},
 					})
 				end)
-
 				done += 1
 			end)
 		end
--- a/lute/cli/commands/test/init.luau
+++ b/lute/cli/commands/test/init.luau
@@ -1,11 +1,12 @@
 local cli = require("@batteries/cli")
 local finder = require("@self/finder")
-local filter = require("@self/filter")
 local luau = require("@std/luau")
 local path = require("@std/path")
 local ps = require("@std/process")
+local system = require("@std/system")
+local task = require("@std/task")
+local vm = require("@lute/vm")
 local testtypes = require("@std/test/types")
-local runner = require("@std/test/runner")
 local reporter = require("@self/reporter")
 local test = require("@std/test")
 local snaprunner = require("@self/snap/runner")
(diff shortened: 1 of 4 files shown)


Vighnesh-V:
my goat

lute/cli/commands/test/worker.luau
```diff
@@ -0,0 +1,40 @@
...
+						__tag = "runtime_error",
+						msg = `failed to load test file: {tostring(err)}`,
+						stacktrace = debug.traceback(nil, 2),
+					},
+				},
+			},
+			total = 1,
+			passed = 0,
+			failed = 1,
+		}
```
Vighnesh-V: Can you make it so that this just returns a TestRunResult?, then the coordinator has to check the optional and yield an appropriate error? It feels weird to fake the test structure here.

lute/cli/commands/test/init.luau
```diff
@@ -41,14 +42,74 @@ local function listtests()
...
+				return nil
+			end
+			local p = testfiles[cursor]
+			cursor += 1
+			return p
+		end
+
+		local done = 0
+		for _ = 1, nworkers do
+			task.spawn(function()
```
Vighnesh-V, line 73: i'm slightly curious - if you replace this task.spawn w/ defer does the performance change at all?

lute/cli/commands/test/init.luau
```diff
@@ -41,14 +42,74 @@ local function listtests()
...
+				local current = nil
+				xpcall(function()
+					local worker = vm.create("@self/worker")
+					while true do
+						current = takenext()
```
Vighnesh-V, line 78: is this safe to do without atomics / synchronization? If multiple threads can access this value, then you could be cursed with a bad interleaving?

lute/cli/commands/test/init.luau
```diff
@@ -41,14 +42,74 @@ local function listtests()
...
+			return p
+		end
+
+		local done = 0
+		for _ = 1, nworkers do
+			task.spawn(function()
+				local current = nil
+				xpcall(function()
+					local worker = vm.create("@self/worker")
+					while true do
```
Vighnesh-V, line 77: The nested `while true` inside the for each worker is a bit confusing to me. It looks like we're doing:
```
for i=1,numWorkers:
    spawn(
          local worker = vm.create...
          while true do
                cursor := takenext()
                worker.runfile(cursor)
          end
     )
```

In this case, the spawned task gets executed immediately and you get stuck in the while loop, with only a single thread spawned that just sequentially keeps taking the next cursor right? Then once you've loaded all files, you move on to the next iteration of the for loop?

Vighnesh-V: Just looked at vm.create's implementation - I see that it detaches the thread in parallel. In that case, I think these threads are probably racing on takeNext. You probably need to shard the data up front e.g numWorkers / numFiles, so each thread gets a non overlapping range.

lute/cli/commands/test/init.luau
Nicell: perf seems relatively similar on my machine

lute/cli/commands/test/init.luau
Nicell: takenext() is always called on the same system thread, vm.create makes a new system thread, but we don't call takenext() on it. the message passing between threads happens on the line `worker.runfile(cursor)` by the vm.

lute/cli/commands/test/init.luau
Nicell: multiple system threads cannot access this value, this is multiple lua threads on the main system thread 👍

lute/cli/commands/test/init.luau
Vighnesh-V: ahhhh okay. Cool, that makes more sense to me.

Vighnesh-V (approved):

PR #1006, opened 2026-04-18
Fix yielding in vm lib

Nicell:
Adjusts vm library to fully support multiple yields in its scheduled work.

### DIFF
--- a/lute/vm/src/spawn.cpp
+++ b/lute/vm/src/spawn.cpp
@@ -1,5 +1,6 @@
 #include "lute/vm.h"
 
+#include "lute/clivfs.h"
 #include "lute/ref.h"
 #include "lute/require.h"
 #include "lute/requirevfs.h"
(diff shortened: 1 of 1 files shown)


vrn-sn:
I'm not sure I understand the piece about the CliVfs here. CLI commands should already be able to use the VM library, I would expect the changes here making it so that spawned VMs can run CLI commands - which doesn't really make sense to me.

Nicell:
@vrn-sn In `vm.create`, the bootstrap require is executed in the child VM, but the requirer source it's resolved against is the parent's chunkname. When the parent is a CLI-embedded module that's `@cli/`, and the default `RequireVfs::isRequireAllowed` will reject it. 

Maybe instead of giving VM's CLI generally, we could instead forward the parent capabilities to the VM? Might be the more correct fix.

vrn-sn:
I'd say the simplest solution is probably to allow the `isRequireAllowed` to pass regardless of the type of VFS we're coming from:
```cpp
bool RequireVfs::isRequireAllowed(lua_State* L, std::string_view requirerChunkname) const
{
    bool isStdin = (requirerChunkname == "=stdin");
    bool isFile = (!requirerChunkname.empty() && requirerChunkname[0] == '@');
    bool isStdLibFile = (requirerChunkname.size() >= 6 && requirerChunkname.substr(0, 6) == "@@std/");
    bool isCliFile = (requirerChunkname.size() >= 6 && requirerChunkname.substr(0, 6) == "@@cli/");
    bool isBundleFile = (requirerChunkname.size() >= 9 && requirerChunkname.substr(0, 9) == "@@bundle/");
    return isStdin || isFile || isStdLibFile || isCliFile || isBundleFile;
}
```

i.e. "even though this require VFS doesn't support being _in_  one of the optional VFSs (CLI or bundle), it does support requiring _from_ one, into a supported one"

Nicell:
Chatted offline, going to look into forwarding require capabilities to VM in a separate PR

aatxe (approved):

PR #1010, opened 2026-04-18
Fix typos and inconsistencies with transform docs

Nicell:


docs/guide/code-transforms.md
```diff
@@ -66,16 +66,14 @@ The manual work of collecting these replacements into a single table is handled
...
 
 ```luau
-local transform = require("@transform")
-
 local ast = require("@std/syntax")
 
-local function transformQuery(ctx: transform.Context)
-	return query.findAllFromRoot(ctx.parseResult, utils.isExprConstantNumber)
```
skberkeley, line 76: thanks for the fixups! can you keep the `@transform` require and type annotations in? we intend to export `@transform` as a requireable alias as soon as `luau-lsp` supports chained aliases (hopefully sometime this week)

Nicell: Can we do that as a follow up? This guide was reported as broken by a community member, and I'd prefer this matches ground truth.

aatxe (approved):

PR #1011, opened 2026-04-18
Fix luthier --clean on Windows

OMouta:
## Summary
Replace the `rm -rf` clean step in `tools/luthier.luau` with a recursive delete implemented through Lute's filesystem API.

## Why
`lute tools/luthier.luau build --clean lute` fails on Windows because `luthier` tries to spawn `rm`, which is a Unix command and may not exist in a normal Windows shell.

## Changes
- add a small recursive delete helper in `tools/luthier.luau`
- use that helper for the `--clean` configure path
- make cleaning a no-op when the build directory does not exist


Nicell (approved):
Thanks!

PR #1021, opened 2026-04-21
Migrate `luthier` to `@std` wrappers

### LINKED ISSUE #1017: Make luthier use @std
Right now luthier uses the `@lute` namespace. Generally we recommend writing writing with `@std`, which wasn't ready at the time of making `luthier`. This should be possible now and a nice chore. https://github.com/luau-lang/lute/blob/primary/tools/luthier.luau

tyulpan:
Migrates `luthier` from the low-level `@lute` namespace to the `@std` wrappers.

Resolves #1017.

### DIFF
--- a/tools/luthier.luau
+++ b/tools/luthier.luau
@@ -1,9 +1,10 @@
 --!strict
 
 local crypto = require("@lute/crypto")
-local fs = require("@lute/fs")
-local process = require("@lute/process")
-local system = require("@lute/system")
+local fs = require("@std/fs")
+local pathlib = require("@std/path")
+local process = require("@std/process")
+local system = require("@std/system")
 local cli = require("@batteries/cli")
 local richterm = require("@batteries/richterm")
 local toml = require("@batteries/toml")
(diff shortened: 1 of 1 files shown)


vrn-sn:
Thanks for the contribution! Unfortunately, Luthier intentionally uses the low-level versions of the `@lute` APIs because we want to support bootstrapping builds from scratch. Maybe we could add a comment to the top of this file explaining this - I'd be happy to approve that.

Luthier is responsible for embedding `@std` into Lute, so it cannot itself depend on `@std` being present. It'll work for non-bootstrapped builds (for example, in this PR's CI checks) because an existing Lute executable already supports `@std`, but a bootstrapped build of Lute will be unable to require the `@std` versions of these libraries.

If you want to see this in action, you can try running [tools/bootstrap.sh](https://github.com/luau-lang/lute/blob/primary/tools/bootstrap.sh) with these changes.

tyulpan:
@vrn-sn Oh, it really does break everything. 
Alternatively, to avoid duplicating the logic from `@std`, `luthier` could be split into two files/phases.

For example, we could keep a minimal `@lute` only generator script (just for the `generateEmbeddedModules` and `generateTypeDefinitionFiles` paths), run it with `lute0`, re-link it to produce a `lute1` with `@std` embedded, and then let a fully migrated `luthier.luau` (using `@std`) take over the configure/build steps.

On the other hand, it's an extra step during the build process and yet another script. 
It really is simpler to just leave a comment at the top.

Should I leave this in the current `PR` or open a new one?

aatxe:
We probably ought to close this, and then add the comment separately.

PR #1023, opened 2026-04-21
Add HTTP routes backed by uWS

Nicell:
Adds a minimal `routes` API for `@lute/net/server`, backed by uWebSockets route matching.

Routes can map paths to handler functions, static responses, or per-method handler tables, with support for exact paths, `:params`, wildcards, and fallback to the existing top-level `handler`. Route handlers receive matched params on `request.params`.


Nicell:
This needs some extra work to get server:upgrade working in the route handlers. #1115 should help with that, so pending that change.

PR #1028, opened 2026-04-21
runtime: clean up a few instances of memory leaks and other undefined behavior

aatxe:
There's some outstanding undefined behavior in a few of the runtime libraries that I was able to identify. There's a double-free in some of the error-path code for the process library, some missing clean up calls leaking memory, and two remaining instances of possible integer overflows for timespec code in durations (with very large durations, they're something like ~290 years). All of these are pretty easy to fix: remove the offending code, add the missing clean up calls, and cast to the larger type _earlier_ in the computation respectively.

### DIFF
--- a/lute/process/src/process.cpp
+++ b/lute/process/src/process.cpp
@@ -257,11 +257,10 @@ int executionHelper(lua_State* L, std::vector<std::string> args, ProcessOptions 
         uv_env_item_t* currentEnvItems;
         int currentEnvCount;
         int err = uv_os_environ(&currentEnvItems, &currentEnvCount);
+        // if error is non-zero, no allocation happened
         if (err != 0)
-        {
-            uv_os_free_environ(currentEnvItems, currentEnvCount);
             luaL_error(L, "Failed to get current environment: %s", uv_strerror(err));
-        }
+
         for (int i = 0; i < currentEnvCount; i++)
         {
             if (currentEnvItems[i].name && currentEnvItems[i].value && opts.env.find(currentEnvItems[i].name) == opts.env.end())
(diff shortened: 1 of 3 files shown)


annieetang (approved):

skberkeley (approved):
do you have test cases for nay of these?

lute/process/src/process.cpp
```diff
@@ -257,32 +257,36 @@ int executionHelper(lua_State* L, std::vector<std::string> args, ProcessOptions
...
         for (int i = 0; i < currentEnvCount; i++)
         {
             if (currentEnvItems[i].name && currentEnvItems[i].value && opts.env.find(currentEnvItems[i].name) == opts.env.end())
             {
                 opts.env[currentEnvItems[i].name] = currentEnvItems[i].value;
             }
         }
+
+        // Free the environment items allocated by uv_os_environ
         uv_os_free_environ(currentEnvItems, currentEnvCount);
```
Vighnesh-V, line 273: Up to your judgement, but I think it would be nice for this to be a helper function - something like:
```
std::optional<int> getEnvironmentVariables(std::map<std::string, std::string>& map)
{
    uv_env_item_t* currentEnvItems;
    int currentEnvCount;
    int err = uv_os_environ(&currentEnvItems, &currentEnvCount);
    if (err != 0)
    {
        uv_os_free_environ(currentEnvItems, currentEnvCount);
        return err;
    }

    for (int i = 0; i < currentEnvCount; i++)
    {
        auto var = currentEnvItems[i];
        if (var.name && var.value && map.find(var.name) == map.end())
            map[var.name] = var.value;
    }
    uv_os_free_environ(currentEnvItems, currentEnvCount);

    return std::nullopt;
}
```

in uvutils. I also don't mind doing this in a separate PR, but it would make doing this PR: https://github.com/luau-lang/lute/pull/897 a touch easier.

Vighnesh-V (commented):
left one comment, but it's mostly a nit

lute/process/src/process.cpp
aatxe: Yeah, we can extract this into a helper in another PR. o7

PR #1029, opened 2026-04-21
`@lute/io`: fix a possible hang in `io.read` when reading returns no bytes read

aatxe:
I'm not sure what conditions would cause libuv to give us no bytes read (if I did, I'd try to add a regression test for it), but in that situation, we'll leave the resume token in neither a completed nor failed state, but otherwise close the stream. This means that the runtime would ultimately fail to ever close correctly since the number of active tokens would always be non-zero.


Vighnesh-V:
I see this in the libuv docs:
image
nread can be 0 but not EOF?

aatxe:
Yeah, it's something like "no bytes were ready to read, but the stream is still open."

Vighnesh-V (approved):

PR #1030, opened 2026-04-21
Comment on `luthier` migration to `@std` wrappers

tyulpan:
Related to `PR` #1021.

### DIFF
--- a/tools/luthier.luau
+++ b/tools/luthier.luau
@@ -1,5 +1,9 @@
 --!strict
 
+-- Luthier intentionally uses low-level @lute APIs instead of @std wrappers to support bootstrapping from scratch.
+-- Since Luthier is responsible for embedding @std into Lute, it cannot depend on @std being available.
+-- Avoid using @std here to prevent bootstrap issues.
+
 local crypto = require("@lute/crypto")
 local fs = require("@lute/fs")
 local process = require("@lute/process")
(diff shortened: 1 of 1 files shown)


tools/luthier.luau
```diff
@@ -1,5 +1,9 @@
 --!strict
 
+-- Luthier intentionally uses low-level @lute APIs instead of @std wrappers to support bootstrapping from scratch.
+-- Since Luthier is responsible for embedding @std into Lute, it cannot depend on @std being available.
```
```suggestion
-- Since Luthier is responsible for generating the files that embed `@std` into Lute, it cannot depend on `@std` being available.
```

tools/luthier.luau
```diff
@@ -1,5 +1,9 @@
...
+-- Avoid using @std here to prevent bootstrap issues.
```
```suggestion
-- Avoid using `@std` here to prevent bootstrap issues.
```

tools/luthier.luau
```diff
@@ -1,5 +1,9 @@
 --!strict
 
+-- Luthier intentionally uses low-level @lute APIs instead of @std wrappers to support bootstrapping from scratch.
```
```suggestion
-- Luthier intentionally uses the runtime `@lute` APIs, instead of the standard library in order to support bootstrapping from scratch.
```

aatxe (approved):

PR #1039, opened 2026-04-22
Adds a dotenv battery with tests and example usage

OMouta:
## Summary

Adds a new `@batteries/env` module for dotenv-style `.env` loading.

## Included

- parsing for common `.env` syntax
- variable expansion support
- `parse`, `load`, `apply`, `serialize`, and `config` helpers
- tests covering parsing, expansion, serialization, and file loading
- an example under `examples/env`

## Usage

`env.config()` loads `.env` from the current working directory, applies values to `process.env`, and returns the parsed table.

### DIFF
--- a/batteries/env.luau
+++ b/batteries/env.luau
@@ -1,0 +1,515 @@
+local fs = require("@std/fs")
+local path = require("@std/path")
+local process = require("@std/process")
+
+local env = {}
+
+export type Environment = { [string]: string }
+export type Pathlike = path.Pathlike
+
+export type ParseOptions = {
+	expand: boolean?,
+	strict: boolean?,
+	env: Environment?,
+}
+
+export type ApplyOptions = {
+	target: Environment?,
+	override: boolean?,
+}
+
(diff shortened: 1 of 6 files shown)


OMouta:
Just saw that one of my commits changed rokit.toml by accident, will revert.

Nicell:
Thanks for the contribution @OMouta! `@batteries`'s current purpose is to support us authoring `lute`. I don't think we need this battery to support `lute` at the moment, but this would be a great package to exist for the community. Our package manager "loom" is still quite young, but you could set up your own repository with a `loom.config.luau` and give directions for people to install it with `lute pkg install` like some are already doing with our batteries.

OMouta:
Fair enough, any reference on how to set up the package correctly and on the structure of loom.config.luau? Thanks.

Nicell:
@OMouta Here's one example that may be useful: https://github.com/grilme99/unicode-segmentation. We don't have any docs yet afaik

PR #1040, opened 2026-04-22
Snapshot tests now support a configuration file format to specify additional arguments.

Vighnesh-V:
This PR adds:
1) A configuration file format to snapshot tests - if you write a snap.config.luau file in the snapshot assets directory, we will look at that for additional args / per test overrides.
2) Ports the lint tests over and removes them from the corresponding directory
3) Runs these tests in CI so that we are dogfooding snapshot testing.

In the next series of PR's:
- style fixes - we ignored this chunk of the code base because this feature was way more experimental, but I'm going to do a separate pass that fixes the casing here.
- documentation about snapshot testing
- Separate out stderr and stdout as separate assets.
- porting more tests


lute/cli/commands/test/snap/runner.luau
```diff
@@ -47,10 +97,33 @@ function runner.runsnapshots(files: { path.Path }): snapty.snapresult
...
-		local artifactpath = artifactpath(filepath)
+		local configuration: snapty.SnapshotConfiguration? = nil
+		local directory = path.dirname(filepath)
+		local maybeConfig = configCache[directory]
+		if maybeConfig == nil then
+			local ok, conf = pcall(retrieveConfiguration, directory)
+			if not ok then
+				-- Something went wrong loading this module
+				break
+			end
```
annieetang, line 133: do we want to print/warn here?

annieetang (approved):
looks good to me thanks for adding additional support! just one edge case q
will also defer to @skanosue for linting stuff

lute/cli/commands/test/snap/runner.luau
Vighnesh-V: I could add a print statement if we're not able to retrieve configuration here.

lute/cli/commands/test/snap/types.luau
```diff
@@ -26,4 +26,18 @@ export type snapresult = {
 	new: { snapnew },
 }
 
+--- Configuration for an individual test
+export type TestConfiguration = {
+	--- Arguments to pass `lute` before the snapshot file
+	args: { string },
```
skberkeley: if we want external users to adopt snapshot testing maybe we don't prefix `lute` every time?

lute/cli/commands/test/snap/runner.luau
```diff
@@ -1,25 +1,75 @@
+local fs = require("@std/fs")
 local process = require("@std/process")
 local path = require("@std/path")
-local fs = require("@std/fs")
-local difftext = require("@batteries/difftext")
+local luau = require("@std/luau")
```
skberkeley: l < p

lute/cli/commands/test/snap/runner.luau
annieetang: sgtm! would be nice for logging

lute/cli/commands/test/snap/runner.luau
```diff
@@ -63,7 +136,8 @@ function runner.runsnapshots(files: { path.Path }): snapty.snapresult
 			continue
 		end
 
-		local expected = fs.readFileToString(artifactpath)
+		local rawExpected = fs.readFileToString(artifactpath)
+		local expected, _ = string.gsub(rawExpected, "\r\n", "\n")
```
skberkeley: is this bc we run ci on windows?

lute/cli/commands/test/snap/runner.luau
```diff
@@ -1,25 +1,75 @@
...
+
+local function getCommandArgs(
+	execpath: string,
+	filepath: string,
+	configuration: snapty.SnapshotConfiguration?,
+	artifactPath: string
+)
+	local args = { execpath }
+	local testname = strext.removeSuffix(artifactPath, ".snap")
+	if not configuration then
```
skberkeley, line 50: can you add a comment here or something? it's just a little confusing having two variables names `config` and `configuration`

skberkeley (changes requested):
talked about this offline, but dym laying what the concrete plan is for checking exit codes? otherwise we're loosening the lint tests you ported

lute/cli/commands/test/snap/runner.luau
Vighnesh-V: yup

lute/cli/commands/test/snap/runner.luau
Vighnesh-V: I'll annotate this a bit!

lute/cli/commands/test/snap/runner.luau
Vighnesh-V: will fix!

lute/cli/commands/test/snap/runner.luau
annieetang: is having a helper function for this useful? no strong opinions but i do see you use `string.gsub(..., "\r\n", "\n")` twice

lute/cli/commands/test/snap/types.luau
Vighnesh-V: The intention here is explicitly that users use `luau` to write scripts that produce output that they could snapshot against. The prior art here is expect tests in dune - you can write ocaml code that writes to stdout, and that gets captured inline.

Vighnesh-V:
@skberkeley Discussed offline, but the plan here is to make the expected exit code configurable. I've added it as a configuration option.

lute/cli/commands/test/snap/runner.luau
Vighnesh-V: I think I can just fold this into normalize snapshot output and omit this use.

lute/cli/commands/test/snap/runner.luau
Vighnesh-V: instead of this, I've tweaked this code so it's less of an issue. I think it makes sense to rename the `config` property to `default` for readability.

lute/cli/commands/test/snap/runner.luau
```diff
@@ -1,7 +1,7 @@
 local fs = require("@std/fs")
+local luau = require("@std/luau")
 local process = require("@std/process")
 local path = require("@std/path")
```
skberkeley: oh wait i missed `path < process` sorry

lute/cli/commands/test/snap/runner.luau
```diff
@@ -63,9 +165,13 @@ function runner.runsnapshots(files: { path.Path }): snapty.snapresult
 			continue
 		end
 
-		local expected = fs.readFileToString(artifactpath)
+		local rawExpected = fs.readFileToString(artifactpath)
+		local expected = normalizeSnapshotOutput(rawExpected, cwd)
+		-- A test passes when a) no exit code checking is requested and b) actual == expected
+		-- It also passes when a) exit code checking is request and b) actual == expected
```
skberkeley: this comment seems wrong?
shouldn't it just be 
a) no exit code checking is requested or b) actual == expected

skberkeley (approved):

lute/cli/commands/test/snap/runner.luau
Vighnesh-V: Comment is wrong on the second line. Should be:
1. no exit code checking is requested -> check just the snapshot contents
2. exit code checking is requested -> check snapshot contents AND exit code.

PR #1043, opened 2026-04-23
Enable codegen

Nicell:
Enables codegen with `CodeGen_OnlyNativeModules` to match Roblox's current behavior. Also updates compile options (see review comment on why).

I'll follow up with some targeted `--!native` opt-ins in a future PR.

### DIFF
--- a/lute/require/src/options.cpp
+++ b/lute/require/src/options.cpp
@@ -1,24 +1,12 @@
 #include "lute/options.h"
 
-static bool codegen = false;
-
 Luau::CompileOptions copts()
 {
     Luau::CompileOptions result = {};
     result.optimizationLevel = 2;
-    result.debugLevel = 2;
-    result.typeInfoLevel = 1;
+    result.debugLevel = 1;
+    result.typeInfoLevel = 0;
     result.coverageLevel = 0;
 
     return result;
-}
-
-bool getCodegenEnabled()
-{
-    return codegen;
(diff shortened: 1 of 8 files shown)


tyulpan:
Maybe make `codegen` available via an optional `--codegen`/`--native` flag?

lute/require/src/options.cpp
```diff
@@ -1,20 +1,12 @@
...
-static bool codegen = false;
-
 Luau::CompileOptions copts()
 {
     Luau::CompileOptions result = {};
     result.optimizationLevel = 2;
-    result.debugLevel = 2;
-    result.typeInfoLevel = 1;
+    result.debugLevel = 1;
+    result.typeInfoLevel = 0;
```
Nicell, line 8: we currently don't make use of level 2 debug, so dropped to 1. If we have a use in the future we can re-evaluate? type info level is automatically set by codegen and afaik is otherwise useless to us?

Nicell:
> Maybe make `codegen` available via an optional `--codegen`/`--native` flag?

Updated the approach of the PR. Sounds like Luau is going in the direction of better automatic codegen, so we're preferring always on with the caveat of using `CodeGen_OnlyNativeModules` to not impact start up times for now.

I think we may want to add a `--no-codegen` flag in the future though 🤔

aatxe (approved):

PR #1052, opened 2026-04-27
Add a new AST replacement variant which can control whether or not trivia is preserved

skberkeley:
This PR adds a new replacement variant which lets users trying to modify ASTs specify on a per-replacement basis whether they want trivia from the original AST node to be preserved. As the unit tests demonstrate, reasoning about which AST nodes trivia is attached to isn't the most straightforward thing in the world, so I've added some explanatory documentation


lute/std/libs/syntax/types.luau
```diff
@@ -161,7 +161,18 @@ export type AstNode = luau.AstNode
 
 export type ParseResult = luau.ParseResult
 
-export type Replacements = { [AstNode]: string }
+--- The type of a replacement for an AST node.
+--- A replacement can be:
+--- 1. A string, which will be used verbatim as the replacement text for the node.
```
Vighnesh-V: I think this line is a bit confusing for me. When I read this, I think that the `string` I supply will directly replace the node, which means we would abandon trivia. 

However `printer.luau` does:
```
local preserveTrivia = if typeof(replacement) == "table" then replacement.preserveTrivia ~= false else true
```

which means that in the string case, we would preserve trivia.

The original issue I raised with codemods was that for something like:
```
--This is a thing
local foo = badrequire("evil")
```

If I replace that node with the exact string `local foo = goodrequire("good")`, would this duplicate the comments?

skberkeley: Oh yeah good point, I'll clarify that trivia is preserved with string replacements

skberkeley: For your second point, I'm not entirely sure which node you're replacing, but as long as you don't explicitly specify that you want to discard trivia, then trivia (including comments) should be preserved

Vighnesh-V: Oh gotcha. I think as long as we clarify the actual semantics, that's good enough.

tests/std/syntax/printer.test.luau
```diff
@@ -1000,4 +1000,106 @@ function foo() return end
 				end)
 		end, assert)
 	end)
+
+	suite:case("structuredReplacementPreservesTrivia", function(assert)
```
Vighnesh-V: Are any of these tests amenable to snapshot testing (only asking because these now run in CI, so we are dogfooding it).? It looks like you're diffing a checkReplacement(sourceFile)?

Vighnesh-V: (obv, you don't have to port every test)

Vighnesh-V (approved):
Left one comment, but the PR looks good, thank you for fixing this! Can you let me know how the snapshot tests look/whether or not they are easy to make?

You can just make a .snap.luau file that prints the replacement to stdout. If the snapshot doesn't exist, you'll be prompted to make one.

tests/std/syntax/printer.test.luau
hgoldstein: I would really appreciate these as snapshot tests!

PR #1053, opened 2026-04-28
@batteries/toml: implement the TOML 1.1 specification

### LINKED ISSUE #1024: TOML Specification Compliance
The `@batteries/toml` library is not currently specification-compliant and this has come up for a few different people. We did a project already of getting a compliance test suite for JSON and bringing the JSON library up to spec, we should do the same for TOML, even if it's not part of the standard library.

aatxe:
This PR essentially rewrites the entire implementation of the TOML battery using the specification and test suite defined in [toml-test](https://github.com/toml-lang/toml-test), which we add as an extern dependency for the purpose of pulling in its test suite. The implementation here passes the entire test suite for the TOML 1.1.0 specification, and yields a value of the following Luau type:

```luau
export type TomlValue = string | number | boolean | { TomlValue } | { [string]: TomlValue }
```

It has an associated benchmark script in `tools/` that benchmarks the performance which has the following behavior on my machine (an M4 Max MacBook Pro):

```
Corpus: 745 files, 52713 bytes total

Passes:      20
Total time:  0.455 s
Per pass:    22.8 ms
Per file:    0.031 ms avg
Throughput:  2.21 MB/s
```

I don't know how this compares to other implementations, but 2 MB/s seems like a reasonable throughput for our needs today (which are largely just going to be processing fairly small TOML files used for configuring legacy package managers).
This fixes #1024.

### DIFF
--- a/batteries/toml.luau
+++ b/batteries/toml.luau
@@ -166,6 +166,7 @@ local function deserialize(input: string): { [string]: unknown }
 	local src = input
 	local pos = 1
 	local len = #src
+	local srcBuf = buffer.fromstring(src)
 
 	-- Table metadata: maps table reference -> { kind, keys }
 	-- kind values:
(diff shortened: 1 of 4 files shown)


vrn-sn (approved):
Admitting that I have not carefully read through, but considering this is basically verified by the test suite and nothing looks obviously wrong, still approving it. Glad to see some `math.nan`s!

PR #1056, opened 2026-04-28
Fix a few http server bugs

Nicell:
- Server closing is now idempotent
- Preserve embedded `\0` bytes in plain string server responses (table responses already did)
- Make server WebSocket upgrades safer by retaining and invalidating upgrade context instead of exposing raw request/response pointers as light userdata


lute/net/src/server.cpp
```diff
@@ -926,7 +959,7 @@ static void listenApp(AppT* app, const std::shared_ptr<ServerLoopState>& state,
 
 static bool closeServer(int serverId)
 {
-    if (!serverInstances.contains(serverId) || !serverStates.contains(serverId))
+    if (!serverInstances.contains(serverId) || !serverStates.contains(serverId) || !serverStates[serverId])
```
Vighnesh-V, line 921: I'm assuming that `server:close()` writes a 'closed' boolean to serverStates, which is what makes this idempotent right?

lute/net/src/server.cpp
```diff
@@ -66,6 +66,17 @@ struct ServerLoopState
...
+template <bool SSL>
+struct RequestUpgradeContext
+{
+    uWS::HttpResponse<SSL>* res = nullptr;
+    uWS::HttpRequest* req = nullptr;
+    us_socket_context_t* socketContext = nullptr;
+    bool pointersValid = true;
+    bool attempted = false;
+    bool upgraded = false;
+};
```
Vighnesh-V, line 82: Should we be using smart pointers here? Can these be made uniqueptr?

lute/net/src/server.cpp
```diff
@@ -380,19 +393,29 @@ static int server_upgrade(lua_State* L)
 template <bool SSL>
 static int server_upgrade_do(lua_State* L)
 {
-    auto* res = static_cast<uWS::HttpResponse<SSL>*>(lua_touserdata(L, lua_upvalueindex(1)));
-    auto* req = static_cast<uWS::HttpRequest*>(lua_touserdata(L, lua_upvalueindex(2)));
-    auto* context = static_cast<us_socket_context_t*>(lua_touserdata(L, lua_upvalueindex(3)));
-    auto* upgradedPtr = static_cast<bool*>(lua_touserdata(L, lua_upvalueindex(4)));
+    using UpgradeContext = RequestUpgradeContext<SSL>;
```
Vighnesh-V: Is this just because the static_cast got too long?

Vighnesh-V: You could also extract this and make it templated:

```
template<bool T>
using UpgradeCtx = RequestUpgradeContext<T>
```

and then reference this type inside of `server_upgrade_do`

lute/net/src/server.cpp
```diff
@@ -380,19 +393,29 @@ static int server_upgrade(lua_State* L)
...
+    auto* contextPtr = static_cast<std::shared_ptr<UpgradeContext>*>(lua_touserdata(L, lua_upvalueindex(1)));
```
Vighnesh-V: shared_ptr\<UpgradeContext\>* is crazy 😆

Vighnesh-V: more concretely, why does this live as a captured value on this function?

lute/net/src/server.cpp
```diff
@@ -380,19 +393,29 @@ static int server_upgrade(lua_State* L)
...
+
+    if (!contextPtr || !(*contextPtr))
+    {
+        lua_pushboolean(L, 0);
+        return 1;
+    }
```
Vighnesh-V: This looks like error management code - would it be possible to use std::optional here instead if that is the case? It's a little tricky for me to parse the condition here:
e.g. (if we don't get a valid pointer to the context, and the context(which itself is a pointer) is nil, return false).

lute/net/src/server.cpp
```diff
@@ -380,19 +393,29 @@ static int server_upgrade(lua_State* L)
...
         return 1;
     }
 
-    bool upgraded = performWebSocketUpgrade<SSL>(res, req, context);
-    *upgradedPtr = upgraded;
+    upgradeContext.attempted = true;
+    bool upgraded = performWebSocketUpgrade<SSL>(upgradeContext.res, upgradeContext.req, upgradeContext.socketContext);
+    upgradeContext.upgraded = upgraded;
+    if (upgraded)
+        upgradeContext.pointersValid = false;
```
Vighnesh-V: this feels a little strange to me, and I'm not sure why? Is there a reason we have to track the pointer validity in a boolean?

Also:
```suggestion
        upgradeContext.pointersValid = !upgraded;
```
maybe?

lute/net/src/server.cpp
```diff
@@ -616,33 +639,37 @@ static HandlerThread prepareHttpHandlerThread(
...
     LUTE_ASSERT(state->handlerRef);
 
-    auto pushUpgradeUpvalues = [res, req, context, &upgraded](lua_State* L)
+    auto pushUpgradeUpvalues = [upgradeContext](lua_State* L)
     {
-        lua_pushlightuserdata(L, res);
-        lua_pushlightuserdata(L, req);
-        lua_pushlightuserdata(L, context);
-        lua_pushlightuserdata(L, &upgraded);
+        using UpgradeContext = RequestUpgradeContext<SSL>;
```
Vighnesh-V: If you're using the same `using` multiple times, probably worth extracting. You can give it a terser name like `UC<SSL>`

lute/net/src/server.cpp
```diff
@@ -616,33 +639,37 @@ static HandlerThread prepareHttpHandlerThread(
...
 
     HandlerThread thread = createHandlerThread(state->runtime);
     lua_State* L = thread.L;
 
     // `lua_resume(L, nullptr, 2)` expects the stack shape `[handler, request, server]`.
     state->handlerRef->push(L);
-    pushRequestTable(L, headers, route, std::string_view(""), server_upgrade_do<SSL>, 4, pushUpgradeUpvalues);
-    pushServerTable(L, state->serverRef, server_upgrade_do<SSL>, 4, pushUpgradeUpvalues);
+    pushRequestTable(L, headers, route, std::string_view(""), server_upgrade_do<SSL>, 1, pushUpgradeUpvalues);
+    pushServerTable(L, state->serverRef, server_upgrade_do<SSL>, 1, pushUpgradeUpvalues);
```
Vighnesh-V: some magic numbers/strings here. Does it make sense for `pushRequestTable` or `pushServerTable` to set these or nah?

Vighnesh-V: ATM i don't know why we changed the 4 -> 1. It would be helpful for other contributors if we could explain why that's the case.

Vighnesh-V (commented):
Left this as a comment since I'm getting more clarification on some of the code structures. Otherwise the fixes all make sense. One thing I feel is that the:
`shared_ptr<T>*` stuff is extremely hard to parse, we should either be doing `shared_ptr<T>&` or `optional<shared_ptr<T>>`

lute/net/src/server.cpp
Nicell: added RequestUpgradeContextPtr<SSL> so the repeated using UpgradeContext blocks are gone.

lute/net/src/server.cpp
Nicell: Removed pointersValid. the code now invalidates by clearing the raw uWS pointers and checks canAttempt().

lute/net/src/server.cpp
Nicell: The uWS res/req/socketContext fields should stay raw pointers because they’re non-owning and only valid during the synchronous upgrade callback.

lute/net/src/server.cpp
Nicell: closeServer idempotence comes from setting serverStates[serverId] = nullptr, not a separate closed boolean. The added test covers first close true, second close false.

Vighnesh-V (approved):

PR #1058, opened 2026-04-30
Add a little disclaimer banner to the API documentation.

andyfriesen:
Just a little something to help make sure that people don't treat these APIs as a stable contract yet.

### DIFF
--- a/lute/cli/commands/doc/init.luau
+++ b/lute/cli/commands/doc/init.luau
@@ -405,6 +405,9 @@ local function generateMarkdown(
 		"```luau",
 		`local {moduleName} = require("@{requireLibraryAlias}/{requirePath}")`,
 		"```",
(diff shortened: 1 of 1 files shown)


lute/cli/commands/doc/init.luau
```diff
@@ -405,6 +405,9 @@ local function generateMarkdown(
 		"```luau",
 		`local {moduleName} = require("@{requireLibraryAlias}/{requirePath}")`,
 		"```",
+		":::warning",
+		"These apis are still a work in progress.  They may change in backwards incompatible ways.",
```
aatxe:
```suggestion
		"These APIs are still open to future evolution.  In new major versions, they may change in backwards incompatible ways.",
```

aatxe (approved):

PR #1060, opened 2026-05-01
Tidies up some code duplication in the websockets library

Vighnesh-V:
The `extractWebsocketPayload` function and `WebsocketPayload` struct is duplicated between the websocket server and client implementation. I've just pulled them out into a common file.

In a follow up, I'll do some style tidying (these files have to be clang-formatted)


vrn-sn (approved):

vrn-sn:
Another style note for the follow-up, noticed that these files use underscores to separate words, but we haven't done that everywhere in the repo (e.g. requirevfs). I personally like the underscores, but regardless of which we choose, should probably keep it consistent.

PR #1063, opened 2026-05-02
`assert.never` simple impl

wmccrthy:
Implements a simple `assert.never` utility, so folks can easily negate any assertion. This is generally useful, a familiar convention for Jest Luau users, and offers some assertions the existing lib fails to directly expose i.e asserting that a function does not throw.

We modify each assertion function's signature to: `(negate: boolean, ...)` such that `reqassert` can easily construct negated assertions, without needing a separate implementation for each negation. I found this minimally intrusive, but open to ideas to slim down the surface even more if folks have any

Added snapshot tests for new negated assertions


lute/std/libs/test/types.luau
```diff
@@ -34,6 +34,14 @@ export type Failure =
 	}
 
 --- The assertion functions passed to each test case. Each assertion returns a `Failure` on failure, or `nil` on success.
+export type AssertNegations = {
```
aatxe:
```suggestion
export type NegatedAsserts = {
```

Probably a better name?

lute/std/libs/test/assert.luau
```diff
@@ -203,26 +278,32 @@ local function strContains(haystack: string, needle: string, instances: number?,
 	return nil
 end
 
-local function strNotContains(haystack: string, needle: string, msg: string?): Failure?
+local function strNotContains(_: boolean, haystack: string, needle: string, msg: string?): Failure?
```
Vighnesh-V: Should this assertion still exist if we have assert.never.strcontains? (same with `neq`.

wmccrthy: Nope! But removing it would be breaking 🫠

wmccrthy: Though actually, i guess I can just map the exported `strNotContains` to `never.strContains`

Vighnesh-V: Breaking changes are okay fwiw, but for now we can do it like this.

Vighnesh-V (approved):

PR #1065, opened 2026-05-05
loom: update the vendored version of luau-unzip and add more tests for `pkg install`

aatxe:
This PR adds some additional tests for `lute pkg` to be able to test `install` more directly, and upgrades the version of `luau-unzip` to v0.2.1 which has some nice performance improvements. We also update `luthier` to ignore a lot of files for the embedding since I noticed e.g. embedding `.DS_Store` on macOS was causing build problems.

### DIFF
--- a/lute/cli/commands/pkg/loom-core/extern/unzip/utils/path.luau
+++ b/lute/cli/commands/pkg/loom-core/extern/unzip/utils/path.luau
@@ -44,7 +44,7 @@ local function replaceBackslashes(input: string, replacement: "/"): string
 		return "\\\\" .. input:sub(3):gsub("\\", replacement)
 	else
 		-- Replace all single backslashes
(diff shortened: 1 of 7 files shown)


lute/cli/commands/pkg/loom-core/extern/unzip/utils/path.luau
```diff
@@ -44,7 +44,7 @@ local function replaceBackslashes(input: string, replacement: "/"): string
 		return "\\\\" .. input:sub(3):gsub("\\", replacement)
 	else
 		-- Replace all single backslashes
-		return (input:gsub("\\", replacement))
+		return input:gsub("\\", replacement)
```
aatxe:
```suggestion
		return (input:gsub("\\", replacement))
```

This one stays.

Vighnesh-V (approved):

Issue #993, opened 2026-04-17
Loom: utilize a shared package store


aatxe:
(Part of) the intended goal of leveraging Luau require aliases as the foundation of the package manager is to ensure that we don't need to duplicate every dependency into every project on the system as some of the pre-existing tools in this space do. We should instead have a shared system store (or in a Roblox environment, a single store per place) that has all of the dependencies in it to be referenced by any projects depending on them as needed.

PR #1066, opened 2026-05-05
loom: implement a shared store for placing package content.

aatxe:
Currently, `lute pkg install` extracts all GitHub dependencies into a project-local `Packages/` directory and wipes it clean on every run. This means the same package is re-downloaded for every project, and every reinstall re-fetches everything from the network even if nothing changed. This also replicates one of the clunkier things about working with the existing package management tools for Luau, and it was my intention to try to fix that clunkiness.

This PR implements #993, introducing a shared store for dependencies at `~/.loom/store/`. GitHub-source packages are extracted there once, named as `{name}@{rev}`, and reused across all local projects without redownloading them. We recommend only doing this with projects that have enabled immutable releases (which GitHub is pushing for generally), otherwise it may lead to some weird behavior. We also add some functionality to `packagerun` to be aware of a `@pkg` alias that resolves to the store's location so that the lockfiles needn't specify the user's home directory path.

### DIFF
--- a/lute/cli/commands/pkg/loom-core/src/packageresolution.luau
+++ b/lute/cli/commands/pkg/loom-core/src/packageresolution.luau
@@ -4,6 +4,7 @@ local lockfile = require("./lockfile")
 local path = require("@std/path")
 local pp = require("@batteries/pp")
 local projectmanifest = require("./projectmanifest")
+local store = require("./store")
 
 local function getManifestPath(directory: string): string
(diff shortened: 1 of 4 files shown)


vrn-sn (approved):
Thanks for implementing this - hoping that the package-aware VFS handles everything correctly from here. We haven't yet tested out multiple projects using the same underlying package as a dependency, but the package-aware VFS was built with that use case in mind 🤞.

PR #1067, opened 2026-05-05
`@std/fs`: add an `fs.move` operation to move/rename files and directories.

aatxe:
I was surprised to see that we hadn't actually authored a move operation of any sort for the filesystem library, but this is obviously useful. This PR implements it analogously to the other filesystem operations, using libuv for the underlying core implementation. When the rename fails due to being to an external device, our rename operation automatically falls back to a copy followed by an unlink.

### DIFF
--- a/definitions/fs.luau
+++ b/definitions/fs.luau
@@ -123,9 +123,9 @@ function fs.copy(src: string, dest: string): ()
 	error("not implemented")
 end
 
---- Renames (moves) the file or directory at `src` to `dest`.
+--- Moves the file or directory at `src` to `dest`.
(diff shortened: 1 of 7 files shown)


aatxe:
`fs.move` might be a better name given the copy-unlink behavior across devices? I'm conflicted.

Vighnesh-V (approved):

Vighnesh-V:
Both fs.rename and fs.move are fine! I think fs.move is more descriptive given the cross device handling.

aatxe:
`move` is also the name Lune used, so I think that's enough to settle the conflict for me.

PR #1069, opened 2026-05-06
loom: unzip into a temporary directory first

aatxe:
There's a bug possible today with the package manager where you could interrupt the execution during unzipping (or encounter an exception) and we'd leave the packages store in an incorrect state. We can fix this by first unzipping everything into a temporary directory, and then moving that temporary directory into the store after the fact.


vrn-sn (approved):
Separately - would be nice if we had some signals-like interface for Luau scripts to be able to handle `SIGINT`s that get sent to Lute.

PR #1070, opened 2026-05-06
Add `no_any` lint

jkelaty-rbx:
I think the world would be better with a bit more `unknown`s

Adds a new ~default~ lint rule that warns on explicit uses of the any type and suggests unknown as a type-safe alternative

~Also happy to not make it default :^)~

### DIFF
--- a/lint.config.luau
+++ b/lint.config.luau
@@ -2,6 +2,11 @@ return {
 	lute = {
 		lint = {
 			ignores = { "**/*.snap.luau" },
+			ruleconfigs = {
+				no_any = {
+					off = true,
+				},
+			},
 		},
 	},
 }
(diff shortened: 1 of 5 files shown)


lint.config.luau
```diff
@@ -2,6 +2,11 @@ return {
 	lute = {
 		lint = {
 			ignores = { "**/*.snap.luau" },
+			ruleconfigs = {
+				no_any = {
+					off = true,
+				},
+			},
```
jkelaty-rbx: for another time

aatxe:
I think it's a reasonable lint, but I fear it must be non-default given a lot of the limitations of Luau's type system today. There's too many places even just in Lute where we must use `any` because refinements won't work well enough.

jkelaty-rbx:
> I think it's a reasonable lint, but I fear it must be non-default given a lot of the limitations of Luau's type system today. There's too many places even just in Lute where we must use `any` because refinements won't work well enough.

That seems reasonable, updated the PR

aatxe (approved):

PR #1078, opened 2026-05-07
`lute format`

Nicell:


lute/std/libs/syntax/format/init.luau
```diff
@@ -0,0 +1,1556 @@
...
+	["+"] = true,
+	["-"] = true,
+	["*"] = true,
+	["/"] = true,
+	["//"] = true,
+	["%"] = true,
+	["^"] = true,
+	[".."] = true,
+	["::"] = true,
+	["->"] = true,
```
Vighnesh-V, line 63: this is a binary operator in luau?

Vighnesh-V:
Possible subfeatures: `lute fmt` as an alias for `lute format`, interactive / patch based mode in a followup?

lute/std/libs/syntax/format/init.luau
aatxe: ai hallucination

aatxe: it came to him in a dream

Vighnesh-V: we should just add it to luau ;)

aatxe: it, `&` and `|` are probably from later lua versions and/or garry's mod's weird lua fork

Nicell: pretty sure it added these because it worked well for the rules that used this list, maybe the list should just be renamed

aatxe: Ah, I see what you mean. I guess they're all binary operators on types, so it's fine by that definition, it's just a little non-obvious that this includes type operators as well.

PR #1079, opened 2026-05-08
`lute format` results example

Nicell:


aatxe:
Honestly, the difference here against stylua is a lot smaller than I thought it'd be.

PR #1081, opened 2026-05-08
loom: implement a parser and constraint solver for semver

aatxe:
Implements a semver library for loom, covering the two concerns needed for dependency resolution: version parsing and constraint satisfaction. Integration of this is left to follow-up work.

## Parser (semver/parser.luau)
  - Strict SemVer 2.0.0 parsing with full validation (leading zeros, empty identifiers, malformed pre-release/build metadata)
  - Pre-release identifier comparison per the spec (numeric vs. string precedence, field-by-field)
  - Constraint syntax includes: bare versions, comparison operators (>=, <=, >, <, =), caret (^), tilde (~), wildcards (*/x/X), and space-separated AND ranges
  - Pre-release opt-in: pre-release versions are excluded from range matches unless the constraint explicitly names a pre-release on the same M.m.p tuple

## Solver (semver/solver.luau)
  - Greedy BFS dependency resolver: selects the highest satisfying version for each package, propagates transitive constraints, and retracts stale constraints when a selection changes
  - Conflict detection with a human-readable error message listing all conflicting requirements and their sources
  - Circular dependency detection
  - Deterministic resolution order (root requirements sorted by alias)


lute/cli/commands/pkg/loom-core/src/semver/solver.luau
```diff
@@ -0,0 +1,262 @@
...
+}
+
+-- All known versions per package name, sorted newest-first.
+export type Universe = { [string]: { PackageVersion } }
+
+-- One selected PackageVersion per package name.
+export type Resolution = { [string]: PackageVersion }
+
+-- Find the highest-precedence version in `versions` satisfying all requirements.
+-- `versions` must already be sorted newest-first.
```
Nicell, line 28: Do we want to do a pass asserting that they're sorted first? Or are we cool with this being documentation only check since it's internal?

Nicell (approved):
Parser looks good and I think this is a good start for the solver. We'll likely want some form of backtracking if the "best" (newest) version doesn't work under transitive deps. We also may want more complex OR operations in constraints at some point. (packages that support multiple majors of a dep)

lute/cli/commands/pkg/loom-core/src/semver/solver.luau
aatxe: Seems reasonable enough to add the assert until such a time as we know it's a performance problem to include.

lute/cli/commands/pkg/loom-core/src/semver/solver.luau
```diff
@@ -0,0 +1,262 @@
...
+local function findBest(versions: { PackageVersion }?, requirements: { Requirement }): PackageVersion?
+	if versions == nil then
+		return nil
+	end
+
+	for _, pv in versions :: { PackageVersion } do
```
Vighnesh-V: can we not figure out the refinement here? :(

lute/cli/commands/pkg/loom-core/src/semver/solver.luau
```diff
@@ -0,0 +1,262 @@
...
+		local ok = true
+		for _, req in requirements do
+			if not parser.satisfies(pv.version, req.constraint) then
+				ok = false
+				break
+			end
+		end
```
Vighnesh-V: I think tablext has an all predicate!

Vighnesh-V (changes requested):

lute/cli/commands/pkg/loom-core/src/semver/solver.luau
aatxe: we totally can, actually! good catch!

lute/cli/commands/pkg/loom-core/src/semver/solver.luau
```diff
@@ -0,0 +1,262 @@
...
+	local inQueue: { [string]: boolean } = {}
+
+	local function enqueue(name: string)
+		if inQueue[name] then
+			return
+		end
+
+		inQueue[name] = true
+		table.insert(queue, name)
+	end
```
Vighnesh-V, line 74: don't want to be a nerd, but it would make this function a lot more readable if you pulled out the queuing behaviour. Would using @batteries/deque be easier?

lute/cli/commands/pkg/loom-core/src/semver/solver.luau
```diff
@@ -0,0 +1,259 @@
...
+
+	local function addConstraint(targetName: string, sourceKey: string, req: Requirement)
+		if not constraintsBySource[targetName] then
+			constraintsBySource[targetName] = {}
+		end
```
Vighnesh-V: `constraintsBySource[targetName] = constraintsBySource[targetName] or {}` ?

Vighnesh-V (changes requested):
some baby nits, but I think this is good to go!

Vighnesh-V (approved):
approved

PR #1084, opened 2026-05-09
Migrate net.request to curl multi

### LINKED ISSUE #974: Migrate net.request to curl multi
Using curl easy is nice and easy as the name would suggest, but we submit the work to the libuv threadpool, which by default only has four threads. Instead we should probably introduce the complexity of curl multi so we can do proper async curl code (or switch to something not curl built on libuv for example).

We should consider adding support for http2 via nghttp2 when this is implemented. Should be straightforward and give a nice perf bump on same origin requests. multi + nghttp2 should allow us to mostly match browser perf. Long horizon we can look at nghttp3.

For example this code sending a request to a local server that waits 1 second to reply:

```luau
local net = require("@std/net")
local time = require("@std/time")

local url = "http://127.0.0.1:8000"

local function run(n: number): number
	local done = 0
	local start = time.now()
	for _ = 1, n do
		task.spawn(function()
			net.request(url)
			done += 1
		end)
	end
	while done < n do
		task.wait()
	end
	return time.since(start)
end

local t4 = run(4)
local t5 = run(5)

print("4 requests:", t4)
print("5 requests:", t5)
print("ratio:", t5 / t4)

```

Result:
```
4 requests:     1.006098985671997
5 requests:     2.0047099590301514
ratio:  1.9925573801181786
```

Nicell:
This branch moves HTTP client requests from blocking `curl_easy_perform` on libuv’s worker pool to a runtime-scoped libcurl multi manager integrated with the runtime’s libuv loop. That removes the 4-thread bottleneck for concurrent `net.request` calls while preserving the existing Luau API.

It also adds some minor fixes. Final-response headers after redirects (was a bug), request/header validation (also a bug), default request/connect timeouts (good safety), large body size handling (more safety), GET/HEAD body rejection (correctness), and stronger net tests including a >4 concurrent request regression. The HTTP implementation is split into `httpclient.cpp`, alongside the `wsclient.cpp`.

Closes #974


lute/net/src/httpclient.cpp
```diff
@@ -0,0 +1,727 @@
...
+#include <vector>
+
+#include "system_ca.h"
+
+namespace net::client
+{
+
+static const std::string kEmptyHeaderKey = "";
+static constexpr long kDefaultRequestTimeoutMs = 5 * 60 * 1000;
+static constexpr long kDefaultConnectTimeoutMs = 30 * 1000;
```
Nicell, line 25: these seem to be strong defaults from other http clients

lute/net/src/httpclient.cpp
```diff
@@ -0,0 +1,725 @@
...
+    }
+
+    void drainCompletions()
+    {
+        if (closing || !multi)
+            return;
+
+        int messagesLeft = 0;
+        CURLMsg* message = nullptr;
+        while ((message = curl_multi_info_read(multi, &messagesLeft)))
```
aatxe:
```suggestion
        while (!closing && multi && (message = curl_multi_info_read(multi, &messagesLeft)))
```

I think there's a use-after-free here if the loop itself doesn't check that we're not closing since we call `maybeDestroyWhenIdle` in `completeRequest`? With the change above, it should be safe. We could also cut the early exit since it'd be redundant with this, but I'm also fine with leaving it.

lute/net/src/httpclient.cpp
```diff
@@ -0,0 +1,725 @@
...
+    auto state = std::make_unique<HttpRequestState>();
+    state->easy = curl;
+    state->url = std::move(url);
+    state->method = std::move(method);
+    state->body = std::move(body);
+    state->token = std::move(token);
+
+    curl_easy_setopt(curl, CURLOPT_URL, state->url.c_str());
+    curl_easy_setopt(curl, CURLOPT_ERRORBUFFER, state->errorBuffer);
+    curl_easy_setopt(curl, CURLOPT_PRIVATE, state.get());
```
aatxe:
```suggestion
```

I don't think we ever read this again, which makes it feel unnecessary. `CURLOPT_PRIVATE` is stated in the spec to not be used by libcurl internally.

lute/net/src/httpclient.cpp
```diff
@@ -0,0 +1,725 @@
...
+
+static bool isValidHeaderValue(const std::string& value)
+{
+    for (unsigned char c : value)
+    {
+        if (c == '\t')
+            continue;
+
+        if (c >= 0x20 && c != 0x7f)
+            continue;
```
aatxe, line 559: This allows through values that aren't visible ASCII + HTAB.

From the HTTP spec (RFC 9110):

> Field values are usually constrained to the range of US-ASCII characters [[USASCII](https://www.rfc-editor.org/rfc/rfc9110.html#USASCII)]. Fields needing a greater range of characters can use an encoding, such as the one defined in [[RFC8187](https://www.rfc-editor.org/rfc/rfc9110.html#RFC8187)]. Historically, HTTP allowed field content with text in the ISO-8859-1 charset [[ISO-8859-1](https://www.rfc-editor.org/rfc/rfc9110.html#ISO-8859-1)], supporting other charsets only through use of [[RFC2047](https://www.rfc-editor.org/rfc/rfc9110.html#RFC2047)] encoding. Specifications for newly defined fields SHOULD limit their values to visible US-ASCII octets (VCHAR), SP, and HTAB. A recipient SHOULD treat other allowed octets in field content (i.e., [obs-text](https://www.rfc-editor.org/rfc/rfc9110.html#fields.values)) as opaque data.

My understanding here is that you can support more for, e.g. UTF-8, since the language is using _should_, but we should probably at least leave a comment explicitly saying this is the case here if so.

aatxe (commented):
Some small comments, and I added some tests.

lute/net/src/httpclient.cpp
Nicell: Stricter policy makes more sense since this is for high-level network requests sent from Luau, not a lower-level library processing arbitrary HTTP headers. Updated to allow visible ASCII + SP + HTAB. Thanks!

aatxe (approved):
Looks good, but we need to update a definition file as well, I think?

PR #1087, opened 2026-05-09
Compress embedded Luau source

### LINKED ISSUE #1016: Compress Luau source embedded in binary
Right now we embed almost 700kb of Luau in the binary. We could gzip this for example and save ~500kb on the binary size. Prototyping this locally shows less than a millisecond added to startup cost. Alternatively maybe shipping bytecode instead of source code in the binary would be better as it would remove a minimal amount of startup cost, too.

Nicell:
Compress generated embedded Luau source with gzip and inflate it when loaded.

Measurements:
- Release binary: 7,520,904 -> 6,992,520 bytes (-528,384 bytes, -7.03%)
- Startup: `lute --help` unchanged at ~2.6ms
- Embedded-source paths: `examples/parsing.luau` +0.3ms, `check all_builtins` +1.3ms

Closes #1016


lute/batteries/src/clibatteries.cpp
```diff
@@ -2,21 +2,26 @@
...
 
-        if (contents == "#directory")
+        if (module.isDirectory)
             return {BatteryModuleType::Directory};
-        else
-            return {BatteryModuleType::Module, contents};
+
+        std::optional<std::string> contents = decompressEmbeddedModule(module);
+        if (contents)
+            return {BatteryModuleType::Module, std::move(*contents)};
```
Vighnesh-V, line 24: Should we need to cache the ungzipped module here, or are we actually good because the require already caches the modules? Could you run a quick test where you implement the deflate caching, and then run a script that requires something like @std/fs and then maybe @std/path afterwards and see if we hit the decompressed cache?

Issue #1080, opened 2026-05-08
Cross-VM type support


tyulpan:
Values passed across a `vm.create` boundary (arguments into a worker function or it's return value) are re-created in the destination `VM` by [copyLuauObject](https://github.com/luau-lang/lute/blob/b2b700f6a3307ade9b50260afce7901725466913/lute/vm/src/spawn.cpp#L24). 
That function only handles:
- [X] `nil`
- [X] `boolean`
- [X] `number`
- [X] `string`
- [X] `table` (recursively)

Anything else falls through to the `default` branch and triggers `Failed to copy arguments between VMs`.

It would be nice to have support for the remaining types:
- [ ] `function`
- [ ] `thread`
- [ ] `userdata`
- [x] `vector` (PR #1100)
- [x] `buffer` (PR #1111)

aatxe:
I'm going to close this issue as complete because we will not be supporting reference types across VMs since it would mean breaking identity when passing a reference from one VM to the other, and when passing that same reference back. The semantics of doing this would be surprising and hard to work with in the best case, and discussions with folks on our VM team suggested to me that, for closures especially, it simply won't work at all. e.g. Consider the case where a function has a closure with references in it that it refers to statefully, like with a counter function. The function will have divergent behavior in every instance of it being marshalled.

tyulpan:
@aatxe yeah, I had similar concerns about reference types -- the identity and state divergence issues make transparent marshalling impractical. In #1127, I described the idea I wanted to propose after PR #1113 got accepted or rejected.

PR #1088, opened 2026-05-09
parallelize lute lint

wmccrthy:
Distribute walking input paths and linting files across workers. Number of workers capped by the minimum of available system threads and # of lintee files.

side note: it'd be _reallyyyy_ sweet if we had documentation on the `vm` lib and how to use it, as it took me a while to figure out what it's constraints were and how to work around them. If i'm missing something, pls lmk 😄

## Testing
All tests passing.

Also manually sanity checked on some local repos with the script in the collapsable below. Verified that every linted file had same the # of violations when linted with old / new build. Seeing pretty big improvements.
**Comparing 5/08 nightly release (default) against release build from this branch (parallel)**

| Repo | Old | New | Diff | Diff % | Old Violations | New Violations | Violation Diff |
| --- | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| lua-apps | 46.12s | 7.75s | 38.38s | 83.2% | 22943 | 22943 | 0 |
| foundation | 2.47s | 0.56s | 1.90s | 77.2% | 855 | 855 | 0 |
| lute | 0.65s | 0.26s | 0.39s | 60.0% | 23 | 23 | 0 |

<details>
	<summary>Test script</summary>
 
```
local json = require("@std/json")
local types = require("./lute/cli/commands/lint/types")

local SEVERITY_MAP: { [number]: types.severity } = {
	[1] = "error",
	[2] = "warning",
	[3] = "info",
	[4] = "hint",
}

local function convertRange(range: any): { beginLine: number, beginColumn: number, endLine: number, endColumn: number }
	return {
		beginLine = range.start.line + 1,
		beginColumn = range.start.character + 1,
		endLine = range["end"].line + 1,
		endColumn = range["end"].character,
	}
end

local function convertDiagnostic(item: any, sourcepath: string): types.LintViolation
	local violation: types.LintViolation = {
		lintname = item.code,
		message = item.message,
		severity = (SEVERITY_MAP[item.severity] or "warning") :: types.severity,
		location = convertRange(item.range) :: any,
		sourcepath = path.parse(sourcepath),
		target = item.codeDescription,
	}

	if item.suggestedfix then
		violation.suggestedfix = {
			fix = item.suggestedfix.fix,
			location = if item.suggestedfix.range then convertRange(item.suggestedfix.range) :: any else nil,
		}
	end

	return violation
end

local function parseLintJsonOutput(jsonString: string): { [string]: { types.LintViolation } }
	local jsonStart = string.find(jsonString, "{")
	if not jsonStart then
		return {}
	end

	local parsed = json.deserialize(jsonString:sub(jsonStart, #jsonString)) :: any
	if not parsed or not parsed.items then
		return {}
	end

	local violationsByFile: { [string]: { types.LintViolation } } = {}

	for _, fileEntry in parsed.items do
		local filePath = fileEntry.uri :: string
		local violations: { types.LintViolation } = {}

		for _, item in fileEntry.items do
			table.insert(violations, convertDiagnostic(item, filePath))
		end

		violationsByFile[filePath] = violations
	end

	return violationsByFile
end

local function countViolations(results: { [string]: { types.LintViolation } }): number
	local count = 0
	for _, violations in results do
		count += #violations
	end
	return count
end

local function formatSeconds(value: number): string
	return string.format("%.2fs", value)
end

local function formatPercent(value: number): string
	return string.format("%.1f%%", value)
end

-- local build = path.join(".", "build", "xcode", "debug", "lute", "cli", "lute")

local releaseBuild = path.join(".", "build", "xcode", "release", "lute", "cli", "lute")

local matrix: { [string]: string } = {
	lute = "./",
	["lua-apps"] = "../roblox/lua-apps",
	["foundation"] = "../roblox/foundation",
}

type Result = {
	repo: string,
	old: number,
	new: number,
	diffSeconds: number,
	diffPercent: number,
	oldViolations: number,
	newViolations: number,
	violationDiff: number,
}

local results: { Result } = {}

for repo, repoPath in matrix do
	local start = os.clock()

	local normalLute = process.run({ "lute", "lint", "-j", repoPath })
	local normalTime = os.clock() - start

	start = os.clock()
	local parallelLint = process.run({ path.format(releaseBuild), "lint", "-j", repoPath })
	local parallelTime = os.clock() - start

	local specialResults = parseLintJsonOutput(parallelLint.stdout)

	local normalResults = parseLintJsonOutput(normalLute.stdout)

	for file, violations in specialResults do
		local normalViolations = normalResults[file]
		assert(normalViolations ~= nil, `Missing normal lint results for {file}`)
		assert(#violations == #normalViolations, `Violation count mismatch for {file}`)
	end

	for file, violations in normalResults do
		local specialViolations = specialResults[file]
		assert(specialViolations ~= nil, `Missing parallel lint results for {file}`)
		assert(#violations == #specialViolations, `Violation count mismatch for {file}`)
	end

	local normalViolationCount = countViolations(normalResults)
	local parallelViolationCount = countViolations(specialResults)
	local diffSeconds = normalTime - parallelTime
	local diffPercent = if normalTime > 0 then (diffSeconds / normalTime) * 100 else 0

	table.insert(results, {
		repo = repo,
		old = normalTime,
		new = parallelTime,
		diffSeconds = diffSeconds,
		diffPercent = diffPercent,
		oldViolations = normalViolationCount,
		newViolations = parallelViolationCount,
		violationDiff = parallelViolationCount - normalViolationCount,
	})
end

-- pretty print results in markdown table
print("| Repo | Old | New | Diff | Diff % | Old Violations | New Violations | Violation Diff |")
print("| --- | ---: | ---: | ---: | ---: | ---: | ---: | ---: |")

for _, result in results do
	print(
		`| {result.repo} | {formatSeconds(result.old)} | {formatSeconds(result.new)} | {formatSeconds(
			result.diffSeconds
		)} | {formatPercent(result.diffPercent)} | {result.oldViolations} | {result.newViolations} | {result.violationDiff} |`
	)
end
```

</details>


lute/cli/commands/lint/init.luau
```diff
@@ -438,32 +72,96 @@ local function lintPaths(
...
+		end)
+	end
 
-			curr = walker()
+	for _ = 1, #inputFiles do
+		if curWorkers < maxWorkers then
+			spawnWorker()
+			curWorkers += 1
 		end
 	end
```
wmccrthy:
```suggestion
	for _ = 1, maxWorkers do
			spawnWorker()
	end
```
do this and drop `curWorkers`, just use `maxWorkers` everywhere

skberkeley:
This is a pretty large change and difficult to review at the moment. It seems like this is currently a refactor combined with the functional changes to parallelize linting. If that's the case, could I ask that you split out the refactor part into a separate PR to make it easier to review?

wmccrthy:
> This is a pretty large change and difficult to review at the moment. It seems like this is currently a refactor combined with the functional changes to parallelize linting. If that's the case, could I ask that you split out the refactor part into a separate PR to make it easier to review?

Sure! Some of the refactor is tightly coupled to the constraints of using `vm`, so even with an additional PR I will have to move a couple things around from the refactor PR -> parallel PR, but it should be more digestible regardless

wmccrthy:
> This is a pretty large change and difficult to review at the moment. It seems like this is currently a refactor combined with the functional changes to parallelize linting. If that's the case, could I ask that you split out the refactor part into a separate PR to make it easier to review?

@skberkeley refactor up [here](https://github.com/luau-lang/lute/pull/1092), i'll rebase and clean up this branch once that is merged

lute/cli/commands/lint/init.luau
```diff
@@ -11,6 +11,9 @@ local types = require("@self/types")
 local internalTypes = require("@self/internaltypes")
 local parseIgnores = require("./lib/parseIgnores")
 local lintCore = require("@self/lintCore")
+local task = require("@std/task")
+local system = require("@std/system")
+local vm = require("@lute/vm")
```
```suggestion
local cli = require("@batteries/cli")
local files = require("./lib/files")
local fs = require("@std/fs")
local internalTypes = require("@self/internaltypes")
local json = require("@std/json")
local lintCore = require("@self/lintCore")
local lsp = require("@self/lsp")
local parseIgnores = require("./lib/parseIgnores")
local pathLib = require("@std/path")
local printer = require("@self/printer")
local process = require("@std/process")
local system = require("@std/system")
local tableext = require("@std/tableext")
local task = require("@std/task")
local types = require("@self/types")
local vm = require("@lute/vm")
```

lute/cli/commands/lint/types.luau
```diff
@@ -3,6 +3,10 @@ local syntax = require("@std/syntax")
 
 export type severity = "error" | "warning" | "info" | "hint"
 export type tag = "unnecessary" | "deprecated"
+export type fix = {
+	location: syntax.Span,
+	replacement: string,
+}
```
skberkeley:
```suggestion
```

lute/cli/commands/lint/init.luau
```diff
@@ -63,30 +73,90 @@ local function lintPaths(
...
-					continue
+	local function spawnWorker()
+		task.spawn(function()
+			xpcall(function()
+				local worker = vm.create("@self/worker")
+				worker.initialize({
+					rulePath = rulePath,
+					noDefaults = noDefaults,
+					configPath = passedConfigVal,
+					verbose = VERBOSE,
```
skberkeley: why are we passing these things around in a table?

lute/cli/commands/lint/init.luau
```diff
@@ -63,30 +73,90 @@ local function lintPaths(
...
+				})
+
+				local curJob = lintNext()
```
skberkeley: let's try not to shadow variables

lute/cli/commands/lint/init.luau
skberkeley: also why does every worker need to do initialize?

lute/cli/commands/lint/init.luau
```diff
@@ -63,30 +73,90 @@ local function lintPaths(
...
+						end)
+						if success then
+							allViolations[curr] = res :: { types.LintViolation }
+						else
+							print(`(worker) Error linting file '{curr}': {res}`)
+						end
+
+						curr = walker()
+					end
+					curJob = lintNext()
```
skberkeley: this feels like a strange way to balance tasks across workers; if we give one worker a folder with many files and subfolders, we'll just end up waiting for that one. can you add a comment explaining that we don't have thread safe data structures so this is kind of the best we can do?

skberkeley (changes requested):
this change feels quite risky to me; can you gate it behind a command line flag?

lute/cli/commands/lint/init.luau
wmccrthy: We can't pass functions as args to the separate VM instance, so the best workaround I know of is re-initializing based on the location of relevant files (rules and config). If there's a better alternative, I'm all ears

Passing the args as a table is just a stylistic choice

lute/cli/commands/lint/init.luau
wmccrthy: Yea that's fair; I also explored keeping the `inputFilePaths` walking sequential, queueing jobs as files were walked, and having workers pull from the queue, but found it was empirically slower. The scenario you call out definitely makes sense in theory but also seems impossible given that `getSourceFiles` returns exclusively files and no directories. I kept the fs walk logic because that is how `lintPaths` worked previously, but maybe it's worth dropping 🤷

wmccrthy:
> this change feels quite risky to me; can you gate it behind a command line flag?

Yes, my only concern is duplicating test logic as I like the safety of having every test case exercise the parallel codepath

lute/cli/commands/lint/init.luau
skberkeley: i see, it sounds like we should tease apart the parts of `initialize` which load the rules and the parts that load the config, and do the config loading before spawning the workers then

creating and reading from tables is more costly, and this isn't an option table, so we should pass the arguments around unwrapped

lute/cli/commands/lint/init.luau
```diff
@@ -63,30 +74,78 @@ local function lintPaths(
...
+				while activeJob ~= nil do
+					local job = activeJob :: Job
+					local inputFilePathString = pathLib.format(activeJob.inputFile)
+
+					if
+						not hasIgnores
+						or not parseIgnores.isIgnored(
+							ignoreData,
+							inputFilePathString,
+							fs.type(inputFilePathString) == "dir"
```
skberkeley: isn't this always `false`?

skberkeley: nit: can you put this whole condition into it's own varaible?

lute/cli/commands/lint/init.luau
wmccrthy: Re passing as tables:
Sounds good, will change.

Re only initializing rules:
I think we still need to initialize both, since rule configs have arbitrary typing; it's totally possible, for instance, that a custom rule has an option that expects a function.

It looks like there's an [issue](https://github.com/luau-lang/lute/issues/1080) up regarding support for passing all primitive types through the VM barrier, so we likely won't need `initialize` in the long-run. I can add a comment calling this out

lute/cli/commands/lint/init.luau
```diff
@@ -63,29 +77,91 @@ local function lintPaths(
...
-	for _, inputPath in inputFiles do
-		local walker = fs.walk(inputPath, { recursive = true })
-
-		local curr = walker()
-		while curr ~= nil do
-			if hasIgnores then
-				local inputFilePathString = pathLib.format(curr)
+	if not PARALLEL then
+		for _, inputPath in inputFiles do
+			local walker = fs.walk(inputPath, { recursive = true })
```
wmccrthy: I removed fs walking in the parallel path since we identified it's redundant given how `getSourceFiles` works. Kept it in non-parallel to keep changes as minimal as possible, but maybe worth removing too? Will defer to @skberkeley

tests/cli/lint.test.luau
```diff
@@ -569,6 +647,41 @@ lintee.luau': @std/syntax/parser.luau:12: parsing failed:
...
+	suite:case("luteLintParallelMultipleFilesWithErrorsRecovers", function(assert)
+		-- Setup
+		-- Create multiple files where one fails to parse and another violates a rule
+		local lintee = path.format(path.join(tmpDir, "parallel_lintee.luau"))
+		fs.writeStringToFile(lintee, "bogus")
+
+		local violatorPath = path.format(path.join(tmpDir, "parallel_violator.luau"))
+		fs.writeStringToFile(violatorPath, "local _x = 1 / 0")
+
+		-- Do
```
wmccrthy:
```suggestion
```

lute/cli/commands/lint/init.luau
skberkeley: Yeah probably worth refactoring in a separate PR?

lute/cli/commands/lint/init.luau
```diff
@@ -74,78 +77,96 @@ local function lintPaths(
...
+				if success then
+					-- Violations are returned as the second return value from pcall
+					allViolations[curr] = err
+				else
+					print(`Error linting file '{curr}': {err}`)
+				end
+
+				curr = walker()
+			end
 		end
```
skberkeley: if we early return here we don't need to wrap the parallel part in an else block

skberkeley (approved):
Looks good, thanks for taking my feedback into consideration! I'm going to ask @~Vighnesh-V to take a look too since he understands the VM/task library the best

skberkeley:
Oh, also maybe worth updating our CI to run `lute lint` in parallel?

docs/cli/lint/index.md
```diff
@@ -71,6 +71,10 @@ Output lint violations in JSON format matching the LSP diagnostic spec.
 
 Lint the provided string input instead of reading from files.
 
+### `-p, --parallel`
+
+Lint files in parallel instead of sequentially.
```
Nicell: why isn't this just the default? why wouldn't you want this? tests already run them in parallel by default

Nicell:
> this change feels quite risky to me; can you gate it behind a command line flag?

I think a flag means that no one will use it. lute should be fast by default, if we're not confident enough to have it on by default, why are we confident enough to add it behind a CLI flag that people are much less likely to use?

lute/cli/commands/lint/init.luau
Nicell: I'd recommend scanning files and then dispatching them in groups afterwards. see what lute test/ the format PR does.

we should probably abstract this worker pool soon so we can share it between test, lint, transform, format, etc. it's all very similar

Vighnesh-V: Hi! I'll put up my draft for this sometime this week. When that happens I can refactor this code during my PR :)

docs/cli/lint/index.md
Vighnesh-V: I agree with Nick here. This should be parallel by default. Can you add the inverse (-s, --sequentially) instead, so users can opt in to determinism?

lute/cli/commands/lint/init.luau
Vighnesh-V: I'm going to be a bit annoying and ask that we extract the code for 'running' lints into its own file. Probably something that exposes runSequential and runParallel as part of its export surface.

lute/cli/commands/lint/init.luau
```diff
@@ -63,29 +77,95 @@ local function lintPaths(
...
+		for _, inputPath in inputFiles do
+			local walker = fs.walk(inputPath, { recursive = true })
+
+			local curr = walker()
+			while curr ~= nil do
+				if hasIgnores then
+					local inputFilePathString = pathLib.format(curr)
+
+					if
+						parseIgnores.isIgnored(ignoreData, inputFilePathString, fs.type(inputFilePathString) == "dir")
```
Vighnesh-V: is this actually what stylua does if the condition is too long?

lute/cli/commands/lint/init.luau
```diff
@@ -63,29 +77,95 @@ local function lintPaths(
...
+			curJob += 1
+
+			return nextJob
+		end
+
+		local function spawnWorker()
+			task.spawn(function()
+				xpcall(function()
+					local worker = vm.create("@self/worker")
+					-- Initialize will be UNNECESSARY once we can pass functions to separate VM
```
Vighnesh-V: what does this mean?

Vighnesh-V: oh, serializing functions across vm bodies. Could I suggest instead that we add support for adding arguments to `vm.create` calls, and the module can initialize state based on that?

Vighnesh-V: e.g. `vm.create(.., rulesPath, noDefaults, passedConfigVal, VERBOSE)`?

lute/cli/commands/lint/init.luau
```diff
@@ -120,6 +201,7 @@ local function main(...: string)
 	end
 
 	VERBOSE = args:has("verbose")
+	PARALLEL = args:has("parallel")
```
Vighnesh-V: Capital case is usually used for constants. Can we stick these options in a table and then read them in here please?

lute/cli/commands/lint/init.luau
```diff
@@ -165,8 +247,16 @@ local function main(...: string)
...
+		local allViolations = lintPaths(
+			inputFiles :: { string },
+			lintRules,
+			configContextData,
+			rulesPath,
+			noDefaultLints,
+			passedConfigVal,
+			autofixEnabled,
+			ignoreData
+		) -- LUAUFIX: type refinement on line 305 should mean that cast on inputFiles isn't needed
```
Vighnesh-V: Isn't this file only 282 lines? Where does this point to now?

tests/cli/lint.test.luau
```diff
@@ -531,6 +531,84 @@ violator2.lua:5:1-6:6 ──
 		fs.removeDirectory(modulePath, { recursive = true })
 	end)
 
+	suite:case("luteLintParallelMultipleInputFiles", function(assert)
```
Vighnesh-V: Can these be expressed as a snapshot test?

Vighnesh-V (changes requested):

docs/cli/lint/index.md
wmccrthy: Sure

lute/cli/commands/lint/init.luau
wmccrthy: seems that way

tests/cli/lint.test.luau
wmccrthy: If we're going to default to parallel, I'll swap these cases for `--sequential`. We can use snapshots for those

lute/cli/commands/lint/init.luau
wmccrthy: Good question, from a quick look at commit history it looks like this comment's been bouncing around for a while. Maybe we just remove it

lute/cli/commands/lint/init.luau
wmccrthy: That makes sense. Probably out of the scope of this PR though? Happy to handle in a separate PR and update this accordingly

lute/cli/commands/lint/init.luau
wmccrthy: As in have a module-scoped table like `{
	verbose: boolean,
	parallel: boolean,
}`? If so, why is that better than just changing the casing?

tests/cli/lint.test.luau
wmccrthy: On second thought, it doesn't really make sense to have dedicated snapshot tests (or explicit tests) for sequential linting. For the sake of safety I'd prefer if all test cases exercised both sequential and parallel code paths. This may be a bit redundant, but casts the widest net for catching regressions

lute/cli/commands/lint/init.luau
wmccrthy: Also, just curious, is serializing function across the VM boundary _not_ something we want in Lute? Initializing module-state is totally unnecessary if we can do that, and seems like an extra step

wmccrthy: ^asking this naively knowing nothing abt the vm lib

aatxe: "once we can pass functions to separate VM"

I don't imagine we will do ever do this realistically. Functions have environments (closures) that are local to the VM they're in, and those environments can contain things that are themselves references. If you tried to copy the function (and transitively its whole environment along with all the ref-y things), you end up in this very muddy state where you have "the same function" in two places, but it actually does different things and refers to different states. We could, in principle, allow marshaling functions with only value-closures (i.e. closures without tables and other ref-y things in them), or functions without closures at all, but then we have made something with unclear behavior from a user perspective (and I _think_ the bulk of the use cases people would have would actually be _wanting_ to marshal functions with references to things).

Vighnesh-V: My wishlist for luau language features is `closed function foo() ... end`, where the runtime will yell at you if you've captured anything ...

lute/cli/commands/lint/init.luau
Vighnesh-V: we could also change the casing, but I figured it might be easier for developers to do `opts.|` and get a list of options that they can check via autocomplete.

tests/cli/lint.test.luau
Vighnesh-V: sure - we should just be asserting output. For determinism, you should just run them sequentially.

Vighnesh-V (approved):
Thank you for adding this! The PR looks good to me.

Issue #1090, opened 2026-05-11
`process.run` corrupts binary stdout/stderr


Nicell:
`process.run()` currently calls `convertCRLFtoLF` on captured stdout and stderr. That’s fine for text, but it corrupts binary output: any `0x0d 0x0a` byte sequence becomes just `0x0a`.

This broke gzip output captured from `process.run({ "gzip", "-c", path })`, producing invalid compressed data.

## Repro
```luau
local process = require("@lute/process")

local result = process.run({
	"python3",
	"-c",
	"import sys; sys.stdout.buffer.write(bytes([0x41, 0x0d, 0x0a, 0x42]))",
})

print(#result.stdout) -- expected 4, currently 3
```

We probably shouldn't call `convertCRLFtoLF` on captured stdout/stderr. Callers that want text normalization can do it themselves.

aatxe:
This was fixed in #1091.

PR #1091, opened 2026-05-11
Remove crlf to lf on process.run

Nicell:
See #1090, fixes bug with binary data corrupting.


aatxe:
Looks like the change is currently load-bearing for the linter tests. Can we add the normalization there?

aatxe (approved):

aatxe (approved):

PR #1094, opened 2026-05-12
runtime: change how we handle argument validation to match Luau builtins

aatxe:
In #1073, the biggest change was transforming argument validation to correspond to how Luau builtins generally do things. This PR  applies the same refactoring to the remaining runtime libraries: `@lute/crypto`, `@lute/luau`, `@lute/net`, and `@lute/task`.

The pattern removed is the `lua_gettop` / `nArgs` check that errors on too-few or too-many arguments. Luau's built-in libraries don't do this — they rely on `luaL_check*` to surface missing-argument errors, and silently ignore extra arguments. With this PR, Lute's entire API is now consistent with that convention as well.

### DIFF
--- a/lute/crypto/src/crypto.cpp
+++ b/lute/crypto/src/crypto.cpp
@@ -58,7 +58,7 @@ int makeHashFunctionMap(lua_State* L)
     return 1;
 }
 
-const env_md_st* getHashFunction(lua_State* L, int idx)
+const env_md_st* checkHashFunction(lua_State* L, int idx)
 {
     if (auto typ = static_cast<const env_md_st*>(lua_tolightuserdatatagged(L, idx, kHashFunctionTag)))
         return typ;
@@ -72,7 +72,7 @@ struct BinaryData
     size_t length;
 };
 
-BinaryData extractData(lua_State* L, int idx)
+BinaryData checkBinaryData(lua_State* L, int idx)
 {
     if (!lua_isstring(L, idx) && !lua_isbuffer(L, idx))
         luaL_typeerrorL(L, idx, "string or buffer");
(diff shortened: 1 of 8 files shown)


lute/crypto/src/crypto.cpp
```diff
@@ -130,10 +122,7 @@ int lua_secretbox_keygen(lua_State* L)
 // seal(message: string | buffer, key: buffer?): { ciphertext: buffer, key: buffer, nonce: buffer }
 int lua_secretbox_seal(lua_State* L)
 {
-    int argumentCount = lua_gettop(L);
-    if (argumentCount != 1 && argumentCount != 2)
-        luaL_error(L, "%s: expected 1 or 2 arguments, but got %d", kSealName, argumentCount);
-
+    bool hasKey = !lua_isnoneornil(L, 2);
```
Vighnesh-V, line 125: Do you want to assert anything on the first argument besides its existence (or does extractData do that already)?

Vighnesh-V (changes requested):
small nit, but otherwise it looks good.

lute/crypto/src/crypto.cpp
aatxe:
```cpp
BinaryData extractData(lua_State* L, int idx)
{
if (!lua_isstring(L, idx) && !lua_isbuffer(L, idx))
luaL_typeerrorL(L, idx, "string or buffer");

if (lua_isstring(L, idx))
{
size_t length = 0;
const char* data = lua_tolstring(L, idx, &length);

return BinaryData{data, length};
}


if (lua_isbuffer(L, idx))
{
size_t length = 0;
void* data = lua_tobuffer(L, idx, &length);

return BinaryData{data, length};
}

luaL_error(L, "failed to extract binary data from stack: %d", idx);
}
```

it sure does! we could change the name perhaps to make it more obvious that it mirrors the `check*` functions.

Vighnesh-V: pog! Can you change the name then (checkBinaryData maybe?)

Vighnesh-V (approved):

lute/crypto/src/crypto.cpp
aatxe: did it for that and the other ones named `get*` and `extract*`

PR #1099, opened 2026-05-13
infra: add cmake option for bootstrapping lute without template files

### LINKED ISSUE #362: Add a build system option for building without the standard library embedding

aatxe:
Previously, building `lute0` during bootstrapping required copying stub files from `tools/templates/` into the generated source directories before running `cmake configure`. This process was a bit brittle to changes to those generated files, and requires more work on the part of someone choosing to not use `luthier`. In #362, I had proposed another approach that we implement here.

This PR replaces the template copying mechanism with a `LUTE_STDLESS` option in the build system. When set to `ON`, each of the four embedded-module targets (`Lute.Std`, `Lute.CLI.Commands`, `Lute.Batteries`, `Lute.Lute`) compiles a checked-in stub .cpp that implements its lookup function to return `NotFound` directly. This gives us a version of `lute` with _no_ embedding Luau code, and a stripped down CLI surface area as a result. We update the bootstrapping script and associated documentation to reflect the usage of this flag, rather than copying the templates.

Closes #362


max-au (approved):
I'd probably put this method at the top of README, because those cloning the repo for the first time are unlikely to have `lute` preinstalled. At least I didn't have it, and I'd really appreciate the STDLESS build to be mentioned first.

PR #1100, opened 2026-05-14
Add cross-VM `vector` type support

tyulpan:
Extend cross-VM marshalling with `vector` type support. 
Related to #1080.


aatxe (approved):
Thanks!

PR #1104, opened 2026-05-16
Even faster JSON

Nicell:
Optimizes `@std/json` serialization/deserialization by reducing repeated string and parsing work in hot paths:

- Adds a per-serialize-call cache for escaped object keys
- Caches hot serializer buffer/cursor locals in string write paths
- Rewrites number parsing to a byte-scanning path with a fast integer case
- Removes redundant whitespace scans in object parsing
- Caps cached keys at 64 bytes to avoid retaining oversized key strings
- Changes checkState so it lowers (thanks @nwinn-student!)

| Dataset | Operation | Before | After | Speedup |
|---|---|---:|---:|---:|
| `tools/dataset.json` | Deserialize | 33.42 MB/s | 36.86 MB/s | 1.10x |
| `tools/dataset.json` | Serialize | 31.39 MB/s | 59.71 MB/s | 1.90x |
| `tools/dataset-large.json` | Deserialize | 27.15 MB/s | 31.34 MB/s | 1.15x |
| `tools/dataset-large.json` | Serialize | 31.80 MB/s | 59.09 MB/s | 1.86x |

Deserialization improved by roughly 10-15%, serialization by roughly 86-90%.

I didn't commit the json bench but may add it on, just didn't want to make the diff look huge with JSON lol.


nwinn-student:
For `serializeAny`:

You could cache the type of the value to avoid having to call it many times in worst-case scenerios.  `table` is already at 4 calls and `string` is at 3.

There is a low likelihood of `json.null` being used enough to outweight the performance benefits this slight change brings.  An early return would ensure `json.null` doesn't slow down (or if it does, only marginally).

Nicell:
> You could cache the type of the value to avoid having to call it many times in worst-case scenerios. `table` is already at 4 calls and `string` is at 3.

Locally I'm seeing equal or worse performance with this change. I've tried it before and saw similar results. `typeof` is already a FASTCALL and really cheap. As to why it seems slightly worse even? I'm guessing it's because the typeof local takes up an extra register, but I'm not quite sure.

After this PR I'm hitting quite a wall. There's some inlining opportunities, but I'd rather avoid making the code too ugly. I think potentially some conditionals could be reordered, but that'd also be workload dependent. After that I think wins will have to be on the Luau side, or we'd have to move to a native module like simdjson/yyjson, but that seems unlikely to happen for now.

nwinn-student:
I just looked at the prior JSON performance PR and noticed you had exactly what I was implying prior.  Whoops, should have checked before I said anything.  I checked this time 😃.

For serialization, we can assume the size will be 2^n.  We could have a fastpath to avoid `buffer.len` since we know the starting size, but it causes regressions for larger datasets.  We can choose to have `bufferSize` set to `0` to use `bit32.countlz(cursor + len) == bit32.countlz(cursor)` instead of `buffer.len`, similar performance.

```luau
-- checkState
local cursor = state.cursor

local bits = bit32.countlz(cursor)
if bit32.countlz(cursor + len) == bits then
  return
end

local curLen = 2 ^ (32 - bits)

-- local newBuffer...
```

Forcing `checkState` to inline helps performance too.  I got the above to inline, or at least improve performance noticably, via the below code.  Hopefully it doesn't look too bad.

```luau 

local function checkStateSlow(state: SerializerState, curLen: number): ()
  -- local newBuffer...
end

-- assumptions: 0 starting size and scales 2^n
local function checkState(state: SerializerState, len: number): ()
  -- local cursor ...
  local curLen = 2 ^ (32 - bits)
  -- see later note for #1105 fix
  checkStateSlow(state, curLen)
end
```

I tried a similar approach with the `buffer.len` variant, but it didn't improve performance so I gave up.  Maybe it needs special treatment.  Although, using the similar approach negates the performance hit caused by looping in #1105.

Anyways.  I intentionally didn't fix the bug addressed in #1105, the bug is out of scope and already addressed there.  The `bit32` approach meshes well with the loop in #1105.  Just swap the inputs to the `bit32` calls (and keep the change of `curLen * 2` to just `curLen`).

lute/std/libs/json.luau
```diff
@@ -102,10 +109,34 @@ local function serializeString(state: SerializerState, str: string): ()
...
+		writeString(state, cached)
+		return
+	end
+
+	local escaped: string
+	if string.find(str, STRING_NEEDS_ESCAPE) then
+		escaped = string.gsub(str, STRING_NEEDS_ESCAPE, STRING_ESCAPES)
+	else
+		escaped = str
+	end
```
aatxe, line 135:
```suggestion
	local escaped =
		if string.find(str, STRING_NEEDS_ESCAPE) then string.gsub(str, STRING_NEEDS_ESCAPE, STRING_ESCAPES)
		else str
```

I don't know about formatting preferences for stylua here, but feels like an if-else expression would be better.

aatxe (approved):

PR #1105, opened 2026-05-16
Improve JSON correctness for math.huge, nan, and massive serialization parts

Nicell:
Improves `@std/json` correctness around two serialization edge cases:

- Rejects non-finite numbers (`math.huge`, `-math.huge`, `math.nan`) instead of emitting invalid JSON tokens
  - We could alternatively emit `null` if we don't want to be as annoying with errors and match what some serializers do, curious if there are any opinions on this
- Grows the serialization buffer until the pending write fits, fixing failures for very large strings or object keys

### DIFF
--- a/lute/std/libs/json.luau
+++ b/lute/std/libs/json.luau
@@ -196,7 +196,7 @@ serializeAny = function(state: SerializerState, value: Value): ()
 	elseif typeof(value) == "boolean" then
 		writeString(state, if value then "true" else "false")
 	elseif typeof(value) == "number" then
-		if math.isnan(value) or value == math.huge or value == -math.huge then
+		if not math.isfinite(value) then
 			error("Cannot serialize non-finite number", 2)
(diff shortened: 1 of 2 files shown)


lute/std/libs/json.luau
```diff
@@ -191,6 +196,10 @@ serializeAny = function(state: SerializerState, value: Value): ()
 	elseif typeof(value) == "boolean" then
 		writeString(state, if value then "true" else "false")
 	elseif typeof(value) == "number" then
+		if math.isnan(value) or value == math.huge or value == -math.huge then
```
nwinn-student:
```suggestion
		if not math.isfinite(value) then
```

`math.isfinite(x)` is the same as `not ( math.isnan(x) or value == math.huge or value == -math.huge )`.  It is easier on the eyes too.

aatxe (approved):

PR #1106, opened 2026-05-17
Use string_view for server responses to remove an extra copy

Nicell:
Use `std::string_view` for server response bodies and headers to avoid an extra `std::string` allocation/copy before sending the response.


Vighnesh-V (approved):
good catch!

PR #1107, opened 2026-05-17
Remove unnecessary wrapper on req server param

Nicell:
This wrapper is unnecessary given the type contract we offer and intend to give resulting in unnecessary overhead.

### DIFF
--- a/lute/net/src/server.cpp
+++ b/lute/net/src/server.cpp
@@ -517,27 +517,17 @@ static void pushRequestTable(
     lua_setmetatable(L, requestIndex);
 }
 
-template<typename PushUpvalues>
-static void pushServerTable(lua_State* L, const std::shared_ptr<Ref>& serverRef, lua_CFunction upgradeFn, int nUpvalues, PushUpvalues pushUpvalues)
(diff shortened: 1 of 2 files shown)


lute/net/src/server.cpp
```diff
@@ -517,27 +517,17 @@ static void pushRequestTable(
...
-    lua_pushvalue(L, serverBaseIndex);
-    lua_setfield(L, -2, "__index");
-    lua_setmetatable(L, -2);
+        return;
+    }
 
-    pushUpvalues(L);
-    lua_pushcclosure(L, upgradeFn, "server.upgrade", nUpvalues);
+    lua_newtable(L);
+    lua_pushcfunction(L, server_upgrade, "server_upgrade");
```
aatxe:
```suggestion
    lua_pushcfunction(L, server_upgrade, "server.upgrade");
```

No reason to change the debug name, right?

Nicell: they didn't match before 🤦 updated to have the two pushes match under `server.upgrade`

Nicell: although we should probably just assert on `serverRef` and then remove this fallback table + upgrade? I think?

aatxe (approved):

PR #1110, opened 2026-05-20
cli: `lute run` should be able to list scripts in `.lute`

### LINKED ISSUE #1101: Add a way to print available scripts under `.lute`
Right now I find myself forgetting available commands and having to manually look through my `.lute` folder (along with directory and sometimes test file clutter) is tedious. I would love something built into the CLI to handle this for me.

As an example of what I'm looking for, here's the contents of Flipbook's `.lute` directory:

```
.lute
├── analyze.luau
├── build.luau
├── bump-version.luau
├── clean.luau
├── install.luau
├── lib
├── lint.luau
├── serve-docs.luau
├── tasks
└── test.luau
```

Then when running any of the following:

```bash
lute run --list # available scripts
lute run --help # help text _and_ available scripts
lute run # same as the last
```

I would expect to get some output like:

```
Available commands: 
- analyze
- build
- bump-version
- clear
- install
- lint
- serve-docs
- test
Run any of these with `lute run <scriptName>`
```

Anecdotally I also see my agents get confused and start running commands like `lute run .lute/build.luau` and, while this executes just fine, its clear the agent doesn't fully understand how `lute run` works and that it's nicer for a human to see `lute run build` fly by than a file path. Having output like the above would help steer them (and myself) in the right direction.

aatxe:
`lute run` supports discovering scripts located in the `.lute` folder of your workspace as an affordance for "extending" the lute executable. For example, scripts like our example `luthier` and `bootstrap` ones let you invoke commands like `lute bootstrap` or `lute luthier`. #1101 raised the issue of discoverability of the scripts in this folder, and proposed a `--list` flag that we could provide to list them out. This PR implements exactly that. The output is fairly straightforward, just an alphabetized list of the commands under a commands header, e.g. for the `lute` repository itself right now:

```
Commands
        bootstrap
        luthier
```

Closes #1101.

### DIFF
--- a/.lute/bootstrap.luau
+++ b/.lute/bootstrap.luau
@@ -5,7 +5,7 @@ local process = require("@lute/process")
 -- LUAUFIX: should `...` actually have the type `string...` for command-line environments
 local args: { [number]: string, n: number } = table.pack(...) :: any
 
-local cmd: { string } = { process.execPath(), "tools/luthier.luau", table.unpack(args, 2, args.n) }
+local cmd: { string } = { "tools/bootstrap.sh", table.unpack(args, 2, args.n) }
 
 local result = process.run(cmd, { stdio = "inherit" })
 process.exit(result.exitcode)
--- a/lute/cli/src/climain.cpp
+++ b/lute/cli/src/climain.cpp
@@ -29,6 +29,7 @@
 #include <cstdlib>
 #include <ctime>
 #include <optional>
+#include <set>
 #include <string>
 #include <utility>
 #include <vector>
@@ -87,6 +88,7 @@ static const char* VERSION_STRING = LUTE_VERSION_FULL;
 static const char* RUN_HELP_STRING = R"(Usage: lute run <script.luau> [args...]
 
 Run Options:
+	--list                  List all available scripts in the nearest .lute folder.
 	--profile               Enable profiling for the script.
 	--profile-output <path> Output file for the profile (default: <datetime>_<filename>.json).
 	--frequency <Hz>        Profiler sampling frequency in Hz (default: 10000).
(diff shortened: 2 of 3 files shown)


.lute/bootstrap.luau
```diff
@@ -0,0 +1,11 @@
+--!strict
+
+local process = require("@lute/process")
+
+-- LUAUFIX: should `...` actually have the type `string...` for command-line environments
+local args: { [number]: string, n: number } = table.pack(...) :: any
```
hgoldstein, line 6: It has `any` right now no?

lute/cli/src/climain.cpp
```diff
@@ -255,6 +258,55 @@ static std::pair<bool, std::string> getValidPath(std::string filePath)
...
+    std::set<std::string> names;
+    std::string normalizedPrefix = normalizePath(luteFolderPath);
+    if (!normalizedPrefix.empty() && normalizedPrefix.back() != '/')
+        normalizedPrefix += '/';
+
+    traverseDirectory(luteFolderPath, [&](const std::string& rawFilePath)
+    {
+        std::string filePath = normalizePath(rawFilePath);
+        if (filePath.size() <= normalizedPrefix.size() || filePath.substr(0, normalizedPrefix.size()) != normalizedPrefix)
+            return;
```
hgoldstein: What is this guarding against?

hgoldstein (approved):

.lute/bootstrap.luau
aatxe: i can't tell? it looks like it's `nil` from what I'm seeing, but it's kinda stubborn? this comment was just copied from `.lute/luthier.luau` anyway.

lute/cli/src/climain.cpp
aatxe: The AI hallucinated situations that could not happen because we live in a wonderful world.

Issue #1096, opened 2026-05-13
An API to respond to signals


jackTabsCode:
It would be useful for me to respond to a SIGINT in our build script to run some code on Ctrl+C of the current process.

aatxe:
Yeah, @Vighnesh-V and I were talking about this the other day. Definitely worth doing.

PR #1112, opened 2026-05-20
runtime: implement an API for signal handling

aatxe:
It's generally useful for command-line programs to be able to respond to various OS signals, especially interrupt or quit signals. This comes up trivially with user inputs to command-line programs, anything that runs continuously/interactively/etc. We can expose a fairly straightforward API here to deal with this, mirroring roughly what's available in node. `process.onSignal` will take a signal by name and a callback to register for responding to that signal, and give you a handle for the signal registration. If you wish to later unregister that signal, you can do so by calling `close` on the handle. This is consistent with our existing filesystem API, for instance, which produces file handles. We also include `process.pid` to give you the current process ID, which enables us to write some tests sending signals to the process we're currently running.

Implements #1096.


lute/process/src/process.cpp
```diff
@@ -25,6 +28,196 @@
...
+
+    // Hardware synchronous signals — unsafe to handle from Luau
+    if (strcmp(name, "SIGBUS") == 0)
+        luaL_errorL(L, "SIGBUS is a synchronous hardware signal and cannot be safely handled");
+    if (strcmp(name, "SIGFPE") == 0)
+        luaL_errorL(L, "SIGFPE is a synchronous hardware signal and cannot be safely handled");
+    if (strcmp(name, "SIGSEGV") == 0)
+        luaL_errorL(L, "SIGSEGV is a synchronous hardware signal and cannot be safely handled");
+    if (strcmp(name, "SIGILL") == 0)
+        luaL_errorL(L, "SIGILL is a synchronous hardware signal and cannot be safely handled");
```
Vighnesh-V: `luaL_errorL` supports a format string, so you could just do:

```
luaL_errorL(L, "%s is a synchronous hardware signal and cannot be safely handled", name);
```

It might be easier to just sprintf this error message once and then re-use it in each branch also.

lute/process/src/process.cpp
```diff
@@ -25,6 +28,196 @@
...
+}
+
+struct SignalHandle
+{
+    SignalHandle(lua_State* L, int callbackIdx, int signum)
+        : runtime(getRuntime(L))
+        , callbackReference(std::make_shared<Ref>(L, callbackIdx))
+    {
+        if (signum < 0) // no-op: signal not available on this platform
+            return;
```
Vighnesh-V: How come this doesn't raise an error?

Vighnesh-V: Oh I see, you opted to say binding to a signal that doesn't exist does nothing.

lute/std/libs/process.luau
```diff
@@ -62,6 +63,18 @@ function processlib.execPath(): Path
 	return pathlib.parse(process.execPath())
 end
 
+--- Registers `callback` to be called whenever `signal` is delivered to this process.
+--- Suppresses the default OS behavior (e.g. `"SIGINT"` will no longer terminate the process).
+--- Returns a handle; call `handle:close()` to deregister.
+function processlib.onSignal(signal: string, callback: () -> ()): SignalHandle
```
Vighnesh-V: might be a cool opportunity to define:
```
type SignalTypes = "SIG1" | "SIG2" | "SIG3" ...
```
that way you can get editor autocomplete on these constants

lute/process/src/process.cpp
```diff
@@ -368,6 +561,8 @@ const luaL_Reg Process::lib[] = {
     {"cwd", process::cwd},
     {"exit", process::exitFunc},
     {"execPath", process::execPath},
+    {"onSignal", process::signalFunc},
+    {"pid", process::pidFunc},
     {nullptr, nullptr},
```
Vighnesh-V, line 380: we probably also want to implement `uv_kill(pid, signal)` at some point.

Vighnesh-V (changes requested):
Couple small changes, the only other thing I would suggest is maybe moving the signal stuff from process.cpp -> signal.cpp?

lute/process/src/process.cpp
aatxe: Yeah, I think it would be alright to error as well, I leaned on node.js behavior here though since I didn't see a strong reason to deviate.

lute/process/src/process.cpp
aatxe: Yeah, probably, I just didn't wanna scope creep this too much all at once. I think the obvious follow up is definitely stuff to _send_ signals to processes.

Vighnesh-V (approved):

PR #1113, opened 2026-05-21
Convert `Runtime` to shared ownership for cross-VM marshalling

tyulpan:
Related to #1080.
All three remaining types (`function` / `thread` / `userdata`) are reference types that can't be cloned across states. 

They're proxied via foreign handles that hold the source value's `Ref` together with a `shared_ptr<Runtime>` to keep the source runtime alive while the handle exists. 

`copyLuauObject` only sees the source state as a `lua_State*`, so it needs a `lua_State* -> shared_ptr<Runtime>` lookup -- which doesn't exist today because `Runtime` is not currently shared-owned. 

This PR adds the lookup via `enable_shared_from_this<Runtime>` plus a `Runtime::create` factory, so any future stack-allocated `Runtime` becomes a compile error rather than a runtime `bad_weak_ptr` deep inside marshalling.

### DIFF
--- a/lute/cli/src/climain.cpp
+++ b/lute/cli/src/climain.cpp
@@ -415,7 +415,8 @@ int handleRunCommand(int argc, char** argv, int argOffset, bool packageAwareness
         return 1;
     }
 
-    Runtime runtime{reporter};
+    auto runtimeOwner = Runtime::create(reporter);
+    Runtime& runtime = *runtimeOwner;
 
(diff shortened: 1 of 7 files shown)


aatxe:
We cannot reasonably support reference types across VMs since they'll always have a surprising semantics between VM instances, and when round-tripped. So, I don't see a reason to proceed with this refactoring at this time.

PR #1116, opened 2026-05-23
Cleanup uWS loop on runtime destruction

Nicell:
uWS holds onto its loop after creation of a net server. We weren't testing net in c++ tests until #1115 which shows the sanitization failure. This adds a runtime hook on destruction so modules like net server can handle frees related to the runtime. I also tried this on the server close, but a true hook on runtime is probably more correct since net server's loop comes from the runtime.

First commit shows the sanitization failure, second commit fixes it.

### DIFF
--- a/lute/runtime/include/lute/runtime.h
+++ b/lute/runtime/include/lute/runtime.h
@@ -89,6 +89,10 @@ struct Runtime
     // once that thread eventually returns or errors after any number of yields.
     void addThreadCompletionHandler(lua_State* L, ThreadCompletionHandler completion);
 
+    // Register native cleanup to run before the runtime closes its libuv loop.
+    // Reusing the same key replaces the previous cleanup hook.
+    void addCleanupHook(void* key, std::function<void()> cleanup);
+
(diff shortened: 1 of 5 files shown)


lute/runtime/include/lute/runtime.h
```diff
@@ -89,6 +89,10 @@ struct Runtime
     // once that thread eventually returns or errors after any number of yields.
     void addThreadCompletionHandler(lua_State* L, ThreadCompletionHandler completion);
 
+    // Register native cleanup to run before the runtime closes its libuv loop.
+    // Reusing the same key replaces the previous cleanup hook.
+    void addCleanupHook(void* key, std::function<void()> cleanup);
```
Vighnesh-V, line 94: This seems like an ownership smell. Instead of doing this, would it make more sense to have the `Runtime` own this state? It seems like logically, the lifetime of the `Runtime` must be >= server state runtime. If this is the case, this problem goes away right?

Vighnesh-V (changes requested):

PR #1120, opened 2026-05-26
Introduces a tty library in lute

Vighnesh-V:
Adds support for a basic tty library in lute. I've tried to keep the api surface small, but we will probably need to iterate on this a bit.

Some things to note:
- The tests only run on posix environments. This is because script(1) ships by default on these platforms, which lets us simulate running a subcommand as if it was in the terminal, which makes tty.open(in/out/err) work correctly.
- There isn't an equivalent on windows afaict, and doing so would require writing more platform specific C++ to get pseudoconsoles.

### DIFF
--- a/definitions/tty.luau
+++ b/definitions/tty.luau
@@ -1,0 +1,37 @@
+-- lute-lint-global-ignore(unused_variable)
+--- The size of the terminal window.
+export type Winsize = {
+	rows: number,
+	cols: number,
+}
+
+--- Options for `tty.open`.
+--- - `raw`: put the terminal into raw mode (no line buffering, no signal generation).
+export type OpenOptions = {
+	raw: boolean?,
+}
+
+--- A handle to a TTY file descriptor. Created via `tty.open`.
+export type TTY = {
+	--- Reads the next chunk of input from the terminal. Yields until input is
+	--- available. Returns `nil` once the stream reaches EOF or is closed.
+	read: (self: TTY) -> string?,
+	--- Writes `bytes` to the terminal. Buffered: returns immediately without
+	--- yielding for the write to drain.
+	write: (self: TTY, bytes: string) -> (),
+	--- Closes the terminal handle. After `close`, `read` returns `nil` and
+	--- `write` errors.
+	close: (self: TTY) -> (),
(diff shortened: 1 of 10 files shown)


lute/tty/src/tty.cpp
```diff
@@ -0,0 +1,455 @@
...
+    return chunk;
+}
+
+void TTYHandle::beginRead(ResumeToken token)
+{
+    startReading();
+    pendingRead = std::move(token);
+}
+
+void TTYHandle::close()
```
Nicell, line 153: if raw mode was enabled we should make sure we turn it off in close

definitions/tty.luau
```diff
@@ -0,0 +1,37 @@
...
+export type TTY = {
+	--- Reads the next chunk of input from the terminal. Yields until input is
+	--- available. Returns `nil` once the stream reaches EOF or is closed.
+	read: (self: TTY) -> string?,
+	--- Writes `bytes` to the terminal. Buffered: returns immediately without
+	--- yielding for the write to drain.
+	write: (self: TTY, bytes: string) -> (),
+	--- Closes the terminal handle. After `close`, `read` returns `nil` and
+	--- `write` errors.
+	close: (self: TTY) -> (),
```
Nicell, line 24: given write doesn't yield, will close yield until the buffer of writes is complete? if I wrote

```luau
tty:write("stuff")
tty:close()
```

it's not clear here if this guarantees the write completes before we close. took a peek and looks like close doesn't wait? maybe the underlying stream guarantees this, but wanted to raise this as 🤨

definitions/tty.luau
```diff
@@ -0,0 +1,37 @@
...
+	--- Registers `callback` to be invoked whenever the terminal receives
+	--- a `SIGWINCH`(window changed) event
+	onResize: (self: TTY, callback: (Winsize) -> ()) -> (),
```
Nicell, line 27: how do you get the window size if it never resizes? I think we'll need a `getSize`/`getWindowSize` API to start, right?

Vighnesh-V: that's easy enough to add, I'll pop it in.

PR #1121, opened 2026-05-26
Update CI so we actually typecheck

skberkeley:
CI seemingly hasn't been running typechecking because we've been using globs with `**`, which apparently doesn't work on GHA's Ubuntu because globstar is off by default. I've updated our `lute check` invocations to instead use `find`, and fixed some legitimate type errors that snuck their way in since we got rid of the type checking faillist.


.github/workflows/ci.yml
```diff
@@ -187,28 +187,37 @@ jobs:
         uses: ./.github/actions/setup-and-build
         with:
           target: Lute.CLI
-          config: debug
+          config: release
```
aatxe: Don't think we should change it to release builds.

lute/cli/commands/pkg/loom-core/extern/.luaurc
```diff
@@ -0,0 +1,3 @@
+{
+  "languageMode": "nocheck"
+}
```
aatxe: We don't need this now since I did a PR to fix the type checking for unzip.

Vighnesh-V (approved):

PR #1122, opened 2026-05-26
Refactors signal handling to take a generic lambda, not just a luau callback reference.

Vighnesh-V:
I want to re-use the Signal Handling code written here to augment a `TTYHandle` with the ability to respond to SIGWINCH events. 
I might want to run regular C++ code in response to a signal, or even supply a different Luau Callback Ref at a later time, unrelated to the current context the Signal Handle was constructed in.

This PR exposes the signal handle in a header file, makes it dispatch a regular C++ lambda, and tweaks the constructor so it's not tied to a lua_State* directly.


Vighnesh-V:
This isn't quite right - I'm going to go back to the drawing board on this one.

aatxe (approved):

PR #1123, opened 2026-05-27
`@batteries/toml`: clean up a bunch of the TOML library code.

aatxe:
I wanted to improve a lot of the obvious tech debt-y kinds of things in the TOML library code. Most of this is transforming a lot of the mapping processes that were big series of branches into singular table lookups to resolve them, there's also some clean up of whitespace, conditional logic, etc. There's a negligible performance impact measured on my machine (within noise), but I think maintainability is reasonably improved with the tables.

### DIFF
--- a/batteries/toml.luau
+++ b/batteries/toml.luau
@@ -319,11 +319,11 @@ local function deserialize(input: string): { [string]: TomlValue }
 		end
 
 		local c = byteAt()
-		if c == 10 then
-			pos += 1
-		else
+		if c ~= 10 then
 			tomlParseError(`expected newline or EOF, got '${string.char(c :: number)}'`)
 		end
+
+		pos += 1
 	end
 
 	-- string escape helpers
--- a/batteries/toml.luau
+++ b/batteries/toml.luau
@@ -900,6 +900,23 @@ local function deserialize(input: string): { [string]: TomlValue }
 
 	local parseValue: () -> TomlValue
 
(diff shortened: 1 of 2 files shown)


batteries/toml.luau
```diff
@@ -921,6 +906,23 @@ local function deserialize(input: string): { [string]: TomlValue }
 
 	local parseValue: () -> TomlValue
 
+	local numberPrefixTable: { [number]: (() -> number)? } = {
+		-- 0x
+		[120] = function()
+			return tonumber(parseDigits(isHexDigit), 16) :: number
```
Vighnesh-V, line 912: Do you have to do the cast here, or can you just annotate the function return `number?`

Vighnesh-V (approved):
🙏 dispatch tables 🙏 

One note: Does it make sense to move the definition of the dispatch tables to the top level scope for clarity?

batteries/toml.luau
aatxe: `tonumber` returns an optional, so we need to cast. The type system can't know that the logic has proven that `tonumber` _will_ parse.

aatxe:
> Does it make sense to move the definition of the dispatch tables to the top level scope for clarity?

I wanted to keep them close to the code that dispatches on them personally, rather than treat them as constants at the top of the program.

batteries/toml.luau
Vighnesh-V: gotcha, that makes sense.

PR #1124, opened 2026-05-27
Renames Ast* node types to Cst*

skberkeley:
As part of the work to split out AST nodes vs CST nodes in Lute's syntax library, this PR renames AST node types and references to ASTs generally to refer to CSTs, since that's what we actually operate on today.


lute/cli/commands/transform/init.luau
```diff
@@ -217,6 +217,7 @@ local function main(...: string)
 	if not migrationPath then
 		print(`Error: No code transformation script specified.\n\n{USAGE}\nUse --help for more information.`)
 		process.exit(1)
+		-- error("unreachable") -- for type checking
```
aatxe:
```suggestion
		error("unreachable") -- for type checking
```

aatxe (approved):

PR #1126, opened 2026-05-27
cli: fix some typechecker issues that snuck into the CLI commands.

aatxe:
I don't know what went wrong that enabled this to happen (@skberkeley is looking into it), but there's some type errors in a few of the CLI commands. We can fix them pretty easily, they're mostly just refinements not quite working when we'd like them to.


aatxe:
Underlying cause seems to be globbing failures in CI, Sora will actually put up a PR with more of these type fixes soon.

PR #1130, opened 2026-05-28
infra: bootstrap into a separate bootstrap directory

### LINKED ISSUE #1128: Typechecking with a debug build built using bootstrap's lute0 raises a bunch of erroneous type errors
Repro path:
1. `rm -rf build`
2. `./build/lute0 tools/luthier.luau build --config debug Lute.CLI`
3. `./build/xcode/debug/lute/cli/lute check lute/std`

aatxe:
Our docs builds have been broken, and we discovered #1128 recently, which both share the same underlying cause. The initial debug build of `lute0` during bootstrapping uses the same folder as the "luthier managed" build folders we have. While it does a clean luthier build after, it does so in a release configuration. So, on our CI machines (and in a local setup where you bootstrap and then start developing), you end up in a bad state for your debug builds. The situation would be resolved by doing a clean debug build, but it's easy to forget that, and in the case of our CI environment, we likely don't want to force all of the builds into being clean ones. This PR changes `bootstrap.sh` to bootstrap into a separate folder, so it will never be reused by debug builds, and prevent running into the situation where you incrementally rebuild into a bad state for your `debug` build.

Fixes #1128.

### DIFF
--- a/tools/bootstrap.sh
+++ b/tools/bootstrap.sh
@@ -76,17 +76,20 @@ else
   BUILD_ROOT=build
 fi
 
-BUILD_PATH=$BUILD_ROOT/debug
+# Use a dedicated directory for the STDLESS bootstrap build so that the
+# cmake cache with LUTE_STDLESS=ON never contaminates the real debug build
+# directory that luthier will configure afterwards.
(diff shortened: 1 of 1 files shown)


annieetang (approved):
ty!

PR #1139, opened 2026-05-29
Refactors syntax.Punctuated

skberkeley:
As part of the planned refactor to add AST parsing to `@std/syntax` (not just CST parsing), I would like CST nodes to subtype covariantly against AST nodes. That is, any program that deals with ASTs should be able to run even if at runtime it receives a CST instead of an AST. In parallel, I would like `Punctuated` to be more ergonomic to program with when dealing with just an AST. In particular, someone programming an the AST version of `Punctuated` will not interact with `separators`, and neither should they have to index into `nodes` every time they see `Punctuated`. This PR makes these changes to the current version of the `Punctuated`, in advance of adding an AST-specific version.

Specifically, it moves `Punctuated`'s subnodes from an array stored in the `nodes` field to the array entries in `Punctuated` itself, and adds an `__iter` override to support iterating over those nodes. Finally, we also rename `Punctuated` to `CstPunctuated`.

### DIFF
--- a/definitions/luau.luau
+++ b/definitions/luau.luau
@@ -56,8 +56,16 @@ export type CstToken<Kind = string> = {
 --- The end-of-file token.
 export type CstEof = CstToken<""> & { read tag: "eof" }
 
-export type Pair<T, Separator> = { read node: T, read separator: CstToken<Separator>? }
-export type Punctuated<T, Separator = ","> = { Pair<T, Separator> }
+type PunctuatedData<T, Separator = ","> = { read [number]: T, read separators: { CstToken<Separator> } }
+
+-- __len not needed because # already returns number of array entries in table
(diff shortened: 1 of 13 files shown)


definitions/luau.luau
```diff
@@ -742,7 +753,6 @@ export type CstTypeTable = {
 
 --- A parameter in a function type annotation.
 export type CstFunctionTypeParameter = {
-	read location: Span,
```
aatxe, line 745: Seems unrelated? Why is this being cut?

aatxe (approved):
Looks good besides the comment.

definitions/luau.luau
skberkeley: it's not referenced anywhere and technically not even serialized at runtime

PR #1141, opened 2026-05-30
Add client tcp stream to `@lute/net`

Gargafield:
Implements the basics of a client tcp stream using libuv with the following api:
```luau
export type TCPStreamHandle = {
	write: (self: TCPStreamHandle, data: buffer | string) -> (),
	read: ((self: TCPStreamHandle, amount: number?, timeout: time.Duration?) -> buffer)
		& ((self: TCPStreamHandle, timeout: time.Duration) -> buffer),
	handshake: (self: TCPStreamHandle) -> (),
	close: (self: TCPStreamHandle) -> (),
}

client.connect = ((function()
	error("not implemented")
end) :: any) :: (
	((hostname: string, port: number) -> TCPStreamHandle)
	& (({ hostname: string, port: number, tls: boolean? }) -> TCPStreamHandle)
)
```
A `:read()` with no amount will try to return the entire buffer, or block until something is read.

The API isn't final and will probably be changed as more discussion about buffer vs. string and so on...

I've added a small test in `tests/lute/net.test.luau` that uses the HTTP serve functionality since there isn't any TCP server to test on.

Also, this code is based on wsclient and processhandle because that was my entry into the codebase and what seemed most similar.

### DIFF
--- a/definitions/net/client.luau
+++ b/definitions/net/client.luau
@@ -41,4 +41,14 @@ function client.websocket(url: string, options: WebSocketOptions?): WebSocket
 	error("not implemented")
 end
 
+export type TCPStreamHandle = {
+	write: (self: TCPStreamHandle, data: buffer | string) -> (),
+	read: (self: TCPStreamHandle, amount: number?) -> buffer,
+	close: (self: TCPStreamHandle) -> (),
+}
+
+function client.connect(host: string, port: number): TCPStreamHandle
+	error("not implemented")
+end
+
 return client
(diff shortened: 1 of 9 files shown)


Gargafield:
I added a timeout parameter to `:read()` pr. `bottersnike`'s wish. A zero duration timeout fill fail immediately if the request would block. Im not entirely sure if using a `uv_timer_t` is the best approach, but couldn't find anything similarly in the codebase.

Gargafield:
So I managed to add TLS support. It's kinda messed up but it's what you get when trying to integrate libuv and boringssl. I tried uSockets, but it simply couldn't do everything I wished for, notably opportunistic TLS, which is added through the `handshake` method. This commit also makes it possible to connect through a domain name.

Issue #1147, opened 2026-06-01
Hitting "illegal operation on a directory" when attempting to install packages on v1.0.1 nightly


vocksel:
When installing packages through `lute pkg install` we're seeing workflow failures on at least `v1.0.1-nightly.20260523`. This is only happening on our GitHub runners and not locally on macOS so presumably it's something to do with our Linux runners.

Here's the error message we're getting:

```
$ lute run install
Writing definitions at ~/.lute/typedefs/1.0.1
Successfully wrote type definition files to ~/.lute/typedefs/1.0.1
move: Error copying /tmp/lute-pkg-install-v1.0.1-nightly.20260523 to /github/home/.loom/store/Lute@v1.0.1-nightly.20260523: illegal operation on a directory
stacktrace:
[C] function move
@std/fs.luau:160 function move
@cli/pkg/loom-core/src/github.luau:112 function installPackage
@cli/pkg/loom-core/src/resolution.luau:62 function fetchSource
@cli/pkg/loom-core/src/resolution.luau:126 function resolve
@cli/pkg/loom-core/src/commands/install.luau:27 function install
@cli/pkg/init.luau:33 function main
@cli/pkg/init.luau:43

/__w/foundation/foundation/.lute/install.luau:9: 
stacktrace:
[C] function assert
/__w/foundation/foundation/.lute/install.luau:9 function installLutePackages
/__w/foundation/foundation/.lute/install.luau:24
```

Relevant job failure: https://github.com/Roblox/foundation/actions/runs/26525389034/job/78127524412?pr=1753

And PR: https://github.com/Roblox/foundation/pull/1753

aatxe:
Fixed in #1148, thanks for the report!

PR #1148, opened 2026-06-02
runtime: fix the external device fallback path for `fs.move`

aatxe:
We recently got a report of an illegal move operation happening only in CI in #1147 when trying to use `lute pkg install`. I investigated the issue, and the underlying cause here is not that the `move` itself is failing. In GitHub Actions, `/tmp` and `/github/home` are different devices, and the underlying move (it's called "rename" in the syscalls) only works within a device. I had written fall-back code for this case that has us do a copy and then remove, but the fallback code is bugged. It doesn't happen locally since you don't have the same filesystem split, and so you don't hit the fallback. The fix is to make the fallback path deal properly with directories (and symbolic links), which we do here in this PR. We also add an explicit test case that does a move from the temporary directory into the home directory, just like `lute pkg install` does, which _should_ mean that our CI flows for this function now explicitly test the fallback case. It's harder to have a local test available for this since we can't really guarantee the presence or absence of external devices to test with, but you can follow the procedures in the CI script locally to set up such a device locally to test with.

### DIFF
--- a/lute/fs/src/fs_impl.cpp
+++ b/lute/fs/src/fs_impl.cpp
@@ -292,7 +292,7 @@ struct FSRename : FSPathPairRequest
     using FSPathPairRequest::FSPathPairRequest;
 
     static void renameCallback(uv_fs_t* req);
-    static void statCallback(uv_fs_t* req);
+    static void crossDeviceFallback(uv_fs_t* req);
 };
 
 // Handles a recursive cross-filesystem directory move as a self-managed async state machine.
@@ -1056,10 +1056,10 @@ void FSRename::renameCallback(uv_fs_t* req)
     // since the fallback copy strategy differs between the two.
     uv_fs_req_cleanup(&r->req);
     uvutils::ScopedUVRequest<FSRename> statReq{std::move(r)};
-    uv_fs_lstat(statReq->getLoop(), &statReq->req, statReq->src.c_str(), FSRename::statCallback);
+    uv_fs_lstat(statReq->getLoop(), &statReq->req, statReq->src.c_str(), FSRename::crossDeviceFallback);
 }
 
-void FSRename::statCallback(uv_fs_t* req)
+void FSRename::crossDeviceFallback(uv_fs_t* req)
 {
     auto r = uvutils::retake<FSRename>(req);
     auto result = req->result;
--- a/lute/fs/src/fs_impl.cpp
+++ b/lute/fs/src/fs_impl.cpp
@@ -101,7 +101,8 @@ struct FSPathPairRequest : FSRequest
     const std::string dest;
 };
 
-// Heap-managed primitive: creates a symlink at `dest` pointing to `target`.
+// Creates a symlink at `dest` pointing to `target`.
+//
 // On Win32, stats `target` first to determine whether UV_FS_SYMLINK_DIR is needed.
 // Calls onDone(0) on success, onDone(negative_uv_error) on failure. Self-deletes.
 struct FSCreateSymlink
@@ -157,7 +158,8 @@ private:
     }
 };
 
-// Heap-managed primitive: moves a symlink from src to dest.
+// Moves a symlink from src to dest.
+//
 // readlink(src) -> FSCreateSymlink(linkTarget, dest) -> unlink(src).
 // Calls onDone(0) on success, onDone(negative_uv_error) on failure. Self-deletes.
 struct FSMoveSymlink
@@ -230,7 +232,8 @@ private:
     }
 };
 
-// Heap-managed primitive: moves a regular file from src to dest.
+// Moves a regular file from src to dest.
+//
 // copyfile(src, dest) -> unlink(src).
 // Calls onDone(0) on success, onDone(negative_uv_error) on failure. Self-deletes.
 struct FSMoveSingleFile
--- a/lute/fs/src/fs_impl.cpp
+++ b/lute/fs/src/fs_impl.cpp
@@ -152,9 +152,8 @@ private:
         auto* self = static_cast<FSCreateSymlink*>(req->data);
         int result = req->result;
         uv_fs_req_cleanup(req);
-        auto done = std::move(self->onDone);
+        self->onDone(result);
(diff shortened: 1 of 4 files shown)


lute/fs/src/fs_impl.cpp
```diff
@@ -99,13 +101,506 @@ struct FSPathPairRequest : FSRequest
...
+#else
+        doSymlink(this, 0);
+#endif
+    }
+
+    uv_loop_t* loop;
+    uv_fs_t req;
+    std::string target;
+    std::string dest;
+    std::function<void(int)> onDone;
```
Vighnesh-V, line 132: Do any of these fields need to be private?

lute/fs/src/fs_impl.cpp
```diff
@@ -99,13 +101,506 @@ struct FSPathPairRequest : FSRequest
...
+#endif
+
+    static void symlinkCallback(uv_fs_t* req)
+    {
+        auto* self = static_cast<FSCreateSymlink*>(req->data);
+        int result = req->result;
+        uv_fs_req_cleanup(req);
+        auto done = std::move(self->onDone);
+        delete self;
+        done(result);
```
Vighnesh-V: Is there a reason you can't do:
```
self->onDone()
delete self;
```
instead? That feels like it flows more logically.

lute/fs/src/fs_impl.cpp
```diff
@@ -99,13 +101,506 @@ struct FSPathPairRequest : FSRequest
...
 {
     using FSPathPairRequest::FSPathPairRequest;
 
     static void renameCallback(uv_fs_t* req);
-    static void copyCallback(uv_fs_t* req);
-    static void unlinkCallback(uv_fs_t* req);
+    static void crossDeviceFallback(uv_fs_t* req);
+};
+
+// Handles a recursive cross-filesystem directory move as a self-managed async state machine.
```
Vighnesh-V: Is it more correct to say this is a cross-directory move instead (all moves in a file system are cross file system)?

lute/fs/src/fs_impl.cpp
```diff
@@ -99,13 +101,506 @@ struct FSPathPairRequest : FSRequest
...
+    {
+        req.data = this;
+        pendingDirs.push_back({std::move(src), std::move(dest)});
+    }
+
+    struct DirPair
+    {
+        std::string src;
+        std::string dest;
+    };
```
Vighnesh-V, line 318: why not just use:
```
using DirPair = std::pair<string, string>;
```?

File: lute/fs/src/fs_impl.cpp [hunk 5]
```diff
@@ -99,13 +101,506 @@ struct FSPathPairRequest : FSRequest
...
+    {
+        if (unknownIdx < pendingUnknown.size())
+        {
+            uv_fs_lstat(loop, &req, pendingUnknown[unknownIdx].src.c_str(), FSCopyDirectory::unknownLstatCallback);
+            return;
+        }
+
+        pendingUnknown.clear();
+        unknownIdx = 0;
+        scannedDirs.push_back(pendingDirs.front().src);
```
Vighnesh-V: any value in doing emplace_back here?

lute/fs/src/fs_impl.cpp
```diff
@@ -99,13 +101,506 @@ struct FSPathPairRequest : FSRequest
...
+        {
+            std::string childSrc = srcDir + "/" + name;
+            std::string childDest = destDir + "/" + name;
+            if (type == UV_DIRENT_DIR)
+            {
+                self->pendingDirs.push_back({childSrc, childDest});
+            }
+            else if (type == UV_DIRENT_FILE)
+            {
+                self->pendingFiles.push_back({childSrc, childDest});
```
Vighnesh-V: emplace_back here, and the other places too

Vighnesh-V (changes requested):
Mostly small comments for my own understanding and small nits (using emplace_back instead of push_back), but otherwise good.

The only thing that feels like a bit weird is that we're creating a bunch of these different request structs that look pretty much like the other fs requests (except for owning a resume token),but we're doing mostly manual memory management. It feels like we should be able to write these like the other FsRequest structs, but I see why we aren't(we don't need a resume token). It feels like UVRequest probably doesn't have the right shape. Regardless, that's out of scope for this PR, but I'm curious about what you think.

lute/fs/src/fs_impl.cpp
aatxe: They could be made private, but I think we've generally worked under the assumption that we don't stand too much to gain from private visibility?

lute/fs/src/fs_impl.cpp
aatxe: We definitely could.

lute/fs/src/fs_impl.cpp
aatxe: Yeah, we could just say it implements a recursive move of a directory, it's not specific to cross-filesystem/cross-device except when used as a fallback for `FSRename`.

lute/fs/src/fs_impl.cpp
aatxe: I think it copies either way, but we can change it.

lute/fs/src/fs_impl.cpp
aatxe: The benefit of a named struct is that the fields are named, instead of having to know whether it's src first or dst first, which is... the footgun I always run into with unix commands and syscalls at least.

lute/fs/src/fs_impl.cpp
Vighnesh-V: yeah this one will still be a copy, but the other places where we're constructing a DirPair to push back will benefit.

lute/fs/src/fs_impl.cpp
Vighnesh-V: we don't really have any explicit standards here. I think there's less point in making these private since they are in the implementation only.

Vighnesh-V (approved):
Ty!

PR #1153, opened 2026-06-03
Enables ccache in CI to speed up builds

Vighnesh-V:
This PR attempts to speed up bootstrap builds for lute CI by doing the following:
1. Try to fetch dependencies in extern/ in parallel - there is no reason we need to do these serially.
2. After fetching dependencies, compute the same hash produced by luthier. This avoids us having to refetch those dependencies when building the actual `lute` binary that bootstrap produces.
3. Do a bit less work in cmake for dependencies.

In addition to this, I've also setup ccache for us. We'll use `actions/cache@v4` to cache the .ccache directory. We invalidate and recompute the cache on merges to `primary`, but changes on your branch will just load the cache from primary.

When testing locally, I've brought bootstrap down to about a minute from two minutes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,4 @@ extern/*/
 generated
 **/generated-types.luau
 WARP.md
+.ccache
--- a/README.md
+++ b/README.md
@@ -57,3 +57,14 @@ This will have you performing the steps of the bootstrap script yourself. A manu
 - Configure with `cmake -G=Ninja -B build -DCMAKE_BUILD_TYPE=Debug -DCMAKE_EXPORT_COMPILE_COMMANDS=1 -DLUTE_STDLESS=ON` to build a version with no embedded Luau functionality.
 - Build a `lute` executable with `ninja -C build lute/cli/lute` or our test suite with `ninja -C build tests/lute-tests`.
 - Optionally, use this version to run `luthier generate` to generate the embedded Luau source files, then reconfigure without `-DLUTE_STDLESS=ON` and rebuild to get a fully-featured `lute`.
+
+### CCache support
+Lute uses ccache in CI to speed up builds. Both the `bootstrap` and `luthier` build scripts support a `--with-ccache` option, which proxies all compilation through ccache. To benefit from
+caching during local development, download ccache using your systems package manager (apt/brew/choco/etc), and perform a clean configure with `--with-ccache` passed. Subsequent builds
+will then use the cached build artifacts in the .ccache directory.
+Additionally, you'll need to set the following variables:
+```
+CCACHE_DIR=path/to/store/ccachedir
+CCACHE_MAXSIZE=maximum size of the cache
+CCACHE_COMPRESS=true
+```
--- a/tools/bootstrap.sh
+++ b/tools/bootstrap.sh
@@ -2,17 +2,25 @@
 
 set -e
 install_requested=false
+use_ccache=false
+build_config="release"
 
 # simple argument parsing
 for arg in "$@"; do
   if [[ "$arg" == "--install" ]]; then
     install_requested=true
-    break
+  elif [[ "$arg" == "--with-ccache" ]]; then
+    use_ccache=true
+  elif [[ "$arg" == "--build-debug" ]]; then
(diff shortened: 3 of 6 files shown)


tools/bootstrap.sh
```diff
@@ -24,10 +32,10 @@ fetch_dependency() {
...
 
   # extract each field by key
-  name=$(grep '^name' "$file" | sed -E 's/^name *= *"(.*)"/\1/')
-  remote=$(grep '^remote' "$file" | sed -E 's/^remote *= *"(.*)"/\1/')
-  branch=$(grep '^branch' "$file" | sed -E 's/^branch *= *"(.*)"/\1/')
-  revision=$(grep '^revision' "$file" | sed -E 's/^revision *= *"(.*)"/\1/')
+  local name=$(grep '^name' "$dep_file" | sed -E 's/^name *= *"(.*)"/\1/')
+  local remote=$(grep '^remote' "$dep_file" | sed -E 's/^remote *= *"(.*)"/\1/')
+  local branch=$(grep '^branch' "$dep_file" | sed -E 's/^branch *= *"(.*)"/\1/')
+  local revision=$(grep '^revision' "$dep_file" | sed -E 's/^revision *= *"(.*)"/\1/')
```
skberkeley, line 38: why did `$file` become `$dep_file`?

skberkeley (approved):
thanks for looking into this! should we document somewhere that people will need to install `ccache` to be able to use the options you added or would they already have it by default?

tools/bootstrap.sh
Vighnesh-V: Our good friend Claude decided that this needed a rename. I don't hate it so I left it around.

.gitignore
```diff
@@ -20,3 +20,4 @@ extern/*/
 generated
 **/generated-types.luau
 WARP.md
+.ccache
```
aatxe, line 23:
```suggestion
.ccache

```

tools/bootstrap.sh
Vighnesh-V: Actually, Claude fixed a bug!

README.md
```diff
@@ -57,3 +57,14 @@ This will have you performing the steps of the bootstrap script yourself. A manu
 - Configure with `cmake -G=Ninja -B build -DCMAKE_BUILD_TYPE=Debug -DCMAKE_EXPORT_COMPILE_COMMANDS=1 -DLUTE_STDLESS=ON` to build a version with no embedded Luau functionality.
 - Build a `lute` executable with `ninja -C build lute/cli/lute` or our test suite with `ninja -C build tests/lute-tests`.
 - Optionally, use this version to run `luthier generate` to generate the embedded Luau source files, then reconfigure without `-DLUTE_STDLESS=ON` and rebuild to get a fully-featured `lute`.
+
+### CCache support
+Lute uses ccache in CI to speed up builds. Both the `bootstrap` and `luthier` build scripts support a `--with-ccache` option, which proxies all compilation through ccache. To benefit from
```
aatxe, line 62:
```suggestion
Lute supports the use of ccache to speed up builds. Both the `bootstrap` and `luthier` build scripts support a `--with-ccache` option, which proxies all compilation through ccache. To benefit from
```

tools/bootstrap.sh
aatxe: It was actually a bug, no? The variable was already named `$dep_file` if you look at the existing code. These references were, I think, to nothing.

aatxe (approved):

.gitignore
Vighnesh-V: I'll make these changes in a follow up!!

tools/bootstrap.sh
Vighnesh-V: they weren't to nothing - $file gets initialized when we loop through the .tune files in the extern directory, so this code just happened to be reading that! That's why this still worked. Bash is a bad language

aatxe: Yeah, okay, accidentally picking it up from a larger, non-lexically-enclosing scope. Literally worked by coincidence, fun.

PR #1164, opened 2026-06-09
`lute pkg`: introduce semver-versioned dependencies as a concept

aatxe:
In #1081, we added an implementation of semantic versioning with a simplistic BFS constraint solving algorithm. This implementation is not attached to anything yet, but was tested in isolation. This PR adds a `registry` dependency kind (we may want to bikeshed the name) for a versioned dependency that uses this semver implementation. The resolution logic for the package manager currently takes a universe of packages as a data structure to work against, and uses that with the constraint solving algorithm to determine which packages to resolve theses versioned registry packages as.


lute/cli/commands/pkg/loom-core/src/resolution.luau
```diff
@@ -165,7 +213,68 @@ local function resolve(rootDirectory: string): lockfile.Lockfile
...
+	if next(registryRequirements) ~= nil then
+		assert(universe ~= nil, "Registry dependencies require a universe to be provided")
+
+		local resolution = semver.solve(universe, registryRequirements)
+
+		for name, pkgVersion in resolution do
+			local conflictKey = nameIndex[name]
+			if conflictKey then
+				error(
+					`Package '{name}' is defined with two sources in the manifest:\n`
```
afujiwara-roblox: Do we need to error here? It should be fine to have a registry dependency + a git dependency as long as they have different aliases

lute/cli/commands/pkg/loom-core/src/commands/install.luau
```diff
@@ -24,7 +26,18 @@ end
 
 local function install()
 	local rootDirectory = path.format(process.cwd())
-	local lock = resolution.resolve(rootDirectory)
+	local manifestPath = `{rootDirectory}/loom.config.luau`
+
+	if fs.exists(manifestPath) then
```
afujiwara-roblox, line 31: I know this is temporary and you mentioned refactoring, but probably still a good idea to give an error message here if the manifest path is not found

lute/cli/commands/pkg/loom-core/src/manifest.luau
```diff
@@ -78,6 +85,8 @@ local function parseManifest(t: unknown): ProjectManifest
 					sourceKind = "path",
 					source = spec.source,
 				}) :: PathDependencySpec
+
+				continue
```
afujiwara-roblox: Any reason not to use `elseif` instead of `continue` inside every `if` block?

lute/cli/commands/pkg/loom-core/src/resolution.luau
```diff
@@ -91,19 +92,53 @@ local function resolve(rootDirectory: string): lockfile.Lockfile
...
-		dependencies[depAlias] = getKey(depSpec)
-		table.insert(queue, depSpec)
+	-- All registry solver inputs: root-level dependencies keyed by alias, transitive by "{pkgKey}:{dependencyAlias}".
+	local registryRequirements: { [string]: semver.Requirement } = {}
+	local registrySources: { [string]: string } = {}
+
+	-- stores a solved source for a particular name, erroring if there's a conflict
+	local function registerSource(name: string, source: string)
+		local existing = registrySources[name]
+		if existing and existing ~= source then
```
afujiwara-roblox: Why does this conflict error? As long as the aliases are different, it should be fine to allow multiple packages from different sources.

aatxe: Maybe I confused myself with the names vs aliases? I was thinking about it as "we're trying to define the same alias to be two different things," but I think you're probably right that this is the proper names, and if the aliases are different it'd work just fine.

lute/cli/commands/pkg/loom-core/src/resolution.luau
```diff
@@ -91,19 +92,53 @@ local function resolve(rootDirectory: string): lockfile.Lockfile
...
+	for dependencyAlias, dependencySpec in rootManifest.package.dependencies do
+		if dependencySpec.sourceKind == "registry" then
+			registryRequirements[dependencyAlias] = {
+				name = dependencySpec.name,
+				constraint = semver.parseConstraint(dependencySpec.version),
+				requiredBy = dependencyAlias,
+			}
+
+			registerSource(dependencySpec.name, dependencySpec.source)
+			table.insert(rootRegistryAliases, dependencyAlias)
```
afujiwara-roblox, line 121: It looks like we should be handling conflicting aliases. I wonder if we could also do that when parsing/validating a valid manifest file.

lute/cli/commands/pkg/loom-core/src/resolution.luau
```diff
@@ -91,19 +92,53 @@ local function resolve(rootDirectory: string): lockfile.Lockfile
...
 
-	local nameIndex: { [string]: string } = {}
+	local rootManifest = manifest.parseFile(rootManifestPath)
+
+	for dependencyAlias, dependencySpec in rootManifest.package.dependencies do
+		if dependencySpec.sourceKind == "registry" then
+			registryRequirements[dependencyAlias] = {
+				name = dependencySpec.name,
+				constraint = semver.parseConstraint(dependencySpec.version),
+				requiredBy = dependencyAlias,
```
jkelaty-rbx: Should this be `dependencyAlias`, or `rootManifest.package.name`? I'm not sure if I'm interpreting "requiredBy" correctly here though

lute/cli/commands/pkg/loom-core/src/manifest.luau
```diff
@@ -88,6 +97,20 @@ local function parseManifest(t: unknown): ProjectManifest
...
+			end
+
+			if spec.sourceKind == "registry" then
+				assert(typeof(spec.version) == "string", `Expected 'version' field for registry dependency '{name}'`)
+				dependencies[name] = table.freeze({
+					name = spec.name or name,
+					sourceKind = "registry",
+					source = spec.source,
+					version = spec.version,
+				}) :: RegistryDependencySpec
```
jkelaty-rbx: nit:
```suggestion
					sourceKind = "registry" :: "registry",
					source = spec.source,
					version = spec.version,
				})
```

lute/cli/commands/pkg/loom-core/src/resolution.luau
aatxe: `rootManifest.package.name` is definitely the most right for `requiredBy`, but `requiredBy` is used in the error messages. I think I'll change the error to include the alias itself, e.g. `required by <root.package.name> as <alias>`, and that should fix that without misusing `requiredBy`.

lute/cli/commands/pkg/loom-core/src/resolution.luau
aatxe: There's not _really_ a good way to do this right now since `loadModule` is ultimately going to just evaluate the manifest file as luau and get our result, similar to `.config.luau` generally. This means we inherit the semantics of luau tables where duplicate keys prefer the last entry. If we decided to use toml or json for the manifest format, we'd be able to do this kind of validation though.

lute/cli/commands/pkg/loom-core/src/resolution.luau
```diff
@@ -122,37 +143,39 @@ local function resolve(rootDirectory: string): lockfile.Lockfile
...
-				transitiveDependencies[depAlias] = getKey(normalizedSpec)
+			local dependencyManifest = manifest.parseFile(nestedManifestPath)
+			for dependencyAlias, dependencySpec in dependencyManifest.package.dependencies do
+				local normalizedSpec = normalizePathDependency(dependencySpec, installPath, rootDirectory)
+
+				if normalizedSpec.sourceKind == "registry" then
+					registryRequirements[`{key}:{dependencyAlias}`] = {
+						name = normalizedSpec.name,
+						constraint = semver.parseConstraint(normalizedSpec.version),
+						requiredBy = key,
```
jkelaty-rbx, line 161:
```suggestion
						requiredBy = key,
						aliasedAs = if dependencyAlias ~= normalizedSpec.name then dependencyAlias else nil,
```

lute/cli/commands/pkg/loom-core/src/resolution.luau
```diff
@@ -165,7 +188,58 @@ local function resolve(rootDirectory: string): lockfile.Lockfile
...
+			end
+
+			packages[key] = table.freeze({
+				name = name,
+				rev = pkgVersion.versionString,
+				source = registrySources[name] or name,
+				sourceKind = "registry",
+				installPath = store.toStorePath(key),
+				dependencies = table.freeze(transitiveDependencies),
+			} :: lockfile.LockfileDependency)
```
jkelaty-rbx: same nit:
```suggestion
				sourceKind = "registry" :: "registry",
				installPath = store.toStorePath(key),
				dependencies = table.freeze(transitiveDependencies),
			})
```

jkelaty-rbx (approved):
epic

afujiwara-roblox (approved):

PR #1166, opened 2026-06-10
Add `mapConcat` + `deepClone` to `tableext`

wmccrthy:
Title is self-explanatory.

I feel like I've ended up implementing bespoke versions of these quite often in side projects, so I think they'd be useful to have. Open to name-change ideas


lute/std/libs/tableext.luau
```diff
@@ -107,4 +107,35 @@ function tableext.all<T>(arr: { T }, predicate: (T) -> boolean): boolean
 	return true
 end
 
+-- I want the method to take a generic s.t it returns an identical shape (deepclone(object) -> same object structure)
+-- using K, V generics and { [K]: V } does not yield thise behavior
+-- only using T (afaik)
+-- This is why I must cast on line 119
+function tableext.deepClone<T>(tbl: T): T
```
aatxe: This is true, but the comment should probably be written more as documentation, and less as you pondering. It's true that using a table with a generic indexer and index result would not yield the same output type as you're likely taking in, and that's fine. The idealized version of this function would have a bound on the generic `T` to say that it's a table, but that would be the best type, i.e. `deepClone: <T>(T) -> T where T <: {}`. We haven't implemented bounded polymorphism, so this signature is the best we can do right now.

lute/std/libs/tableext.luau
```diff
@@ -107,4 +107,35 @@ function tableext.all<T>(arr: { T }, predicate: (T) -> boolean): boolean
...
+function tableext.mappedConcat<T>(arr: { T }, map: (T) -> string, sep: string?): string
+	local separator = sep or ""
+	local buf = ""
+	for i, item in arr do
+		if i > 1 then
+			buf ..= separator
+		end
+		buf ..= map(item)
+	end
+	return buf
```
Vighnesh-V: This can be implemented by doing table.concat(tableext.map(arr, map), sep or "") (which can avoid the O(n^2) string concatenation.

Vighnesh-V: mapConcat is probably a terser name as well.

lute/std/libs/tableext.luau
```diff
@@ -107,4 +107,35 @@ function tableext.all<T>(arr: { T }, predicate: (T) -> boolean): boolean
 	return true
 end
 
+-- I want the method to take a generic s.t it returns an identical shape (deepclone(object) -> same object structure)
+-- using K, V generics and { [K]: V } does not yield thise behavior
+-- only using T (afaik)
+-- This is why I must cast on line 119
+function tableext.deepClone<T>(tbl: T): T
+	local clone = {}
+	assert(typeof(tbl) == "table", `{tbl} is not a table`)
```
Vighnesh-V: Should this be an error instead of an assertion?

lute/std/libs/tableext.luau
```diff
@@ -107,4 +107,35 @@ function tableext.all<T>(arr: { T }, predicate: (T) -> boolean): boolean
 	return true
 end
 
+-- I want the method to take a generic s.t it returns an identical shape (deepclone(object) -> same object structure)
```
Vighnesh-V: comment says deepclone, but the method is deepClone.

lute/std/libs/tableext.luau
```diff
@@ -107,4 +107,35 @@ function tableext.all<T>(arr: { T }, predicate: (T) -> boolean): boolean
...
+-- using K, V generics and { [K]: V } does not yield thise behavior
```
Vighnesh-V: nit: type on `thise` -> `this`

lute/std/libs/tableext.luau
```diff
@@ -107,4 +107,35 @@ function tableext.all<T>(arr: { T }, predicate: (T) -> boolean): boolean
...
+-- only using T (afaik)
+-- This is why I must cast on line 119
```
Vighnesh-V: there is no cast on line 119? Did you mean the `tbl::any` cast? What about unknown?

tests/std/tableext.test.luau
```diff
@@ -139,4 +139,87 @@ test.suite("TableExt", function(suite)
 		end)
 		assert.that(allInEmpty)
 	end)
+
+	suite:case("mappedConcat", function(assert)
```
Vighnesh-V: Can each of these sub cases be their own test?

lute/std/libs/tableext.luau
```diff
@@ -107,4 +107,35 @@ function tableext.all<T>(arr: { T }, predicate: (T) -> boolean): boolean
...
+function tableext.deepClone<T>(tbl: T): T
+	local clone = {}
```
Vighnesh-V: You could do `clone = table.clone(tbl)` to start with. Then if the deeply cloned item is an array of non-tables, you already have the cheap deep clone. Then, the body of the loop becomes;
```
if typeof(_) == table then
     clone[k] = deepClone(v)
end
```

lute/std/libs/tableext.luau
Vighnesh-V: small question as well - lets say you have:
```
local t = {}
local e = {x = 1}

t[a] = e
t[b] = t[a]
```

The current deepClone will return a table:
```
t'[a] = { x = 1 }
t'[b] = { x = 1 }
where
t'[a] != t'[b]
```
Moreover, in the presence of cycles - t[u] -> b, b[v] -> c, c[w] -> b, this function will infinite loop. I think you have cache previously cloned tables to avoid this.

Vighnesh-V (dismissed):

lute/std/libs/tableext.luau
wmccrthy: assert will throw if it's not a table, is there something else `error` offers here?

lute/std/libs/tableext.luau
wmccrthy: o dam, didn't realize `table.concat` was faster like that. nice

lute/std/libs/tableext.luau
wmccrthy: i figured we have to iterate over the the input table anyway, so cloning at the start would be a redundant pass. Is implementing as you described cheaper? and if so, what about it makes it cheaper?

tests/std/tableext.test.luau
wmccrthy: i'm fine w/ doing this, but the rest of the suite uses a single case for each `tableext` method

lute/std/libs/tableext.luau
wmccrthy: `unknown` still yields a type error as it could be nil: `TypeError: Value of type 'unknown' could be nilLuau[1032](https://luau.org/types)`

lute/std/libs/tableext.luau
wmccrthy: Thanks for the explanation, I just dropped the comment entirely for now.

lute/std/libs/tableext.luau
```diff
@@ -107,4 +107,31 @@ function tableext.all<T>(arr: { T }, predicate: (T) -> boolean): boolean
...
+	assert(typeof(tbl) == "table", `deepClone only supports tables, you passed: {tbl}`)
+
+	seen[tbl] = clone
+	for k, v in tbl :: any do
+		if typeof(v) == "table" then
+			clone[k] = if seen[v] then seen[v] else deepClone(v, seen)
+		else
+			clone[k] = v
+		end
+	end
```
aatxe: This is a bit of a weird thing, and probably a big part of why Luau itself doesn't have a deep clone, but I'll note here that the keys are not cloned deeply if they're tables. This might be intended, or might be an accident, but it should definitely be documented either way.

lute/std/libs/tableext.luau
```diff
@@ -107,4 +107,31 @@ function tableext.all<T>(arr: { T }, predicate: (T) -> boolean): boolean
...
 end
 
+function deepClone<T>(tbl: T, seen: { [T]: T }): T
+	local clone = {}
+	assert(typeof(tbl) == "table", `deepClone only supports tables, you passed: {tbl}`)
+
+	seen[tbl] = clone
+	for k, v in tbl :: any do
+		if typeof(v) == "table" then
+			clone[k] = if seen[v] then seen[v] else deepClone(v, seen)
```
aatxe: We also never look at metatables here, which means that this deep clone does not preserve metatables. This again might be intended or might not be, but we should definitely document the behavior at minimum.

lute/std/libs/tableext.luau
```diff
@@ -107,4 +107,31 @@ function tableext.all<T>(arr: { T }, predicate: (T) -> boolean): boolean
 	return true
 end
 
+function deepClone<T>(tbl: T, seen: { [T]: T }): T
+	local clone = {}
+	assert(typeof(tbl) == "table", `deepClone only supports tables, you passed: {tbl}`)
```
```suggestion
	assert(typeof(tbl) == "table", `deepClone expects a table, but got a {typeof(tbl)}`)
```

lute/std/libs/tableext.luau
```diff
@@ -107,4 +107,31 @@ function tableext.all<T>(arr: { T }, predicate: (T) -> boolean): boolean
...
+			clone[k] = if seen[v] then seen[v] else deepClone(v, seen)
+		else
+			clone[k] = v
+		end
+	end
+
+	return clone
+end
+
+function tableext.deepClone<T>(tbl: T): T
```
aatxe: Speaking of documentation, this function needs a doc comment explaining how it works, and it should include the things above.

lute/std/libs/tableext.luau
```diff
@@ -107,4 +107,31 @@ function tableext.all<T>(arr: { T }, predicate: (T) -> boolean): boolean
...
+	local cloneCache = {}
+
+	return deepClone(tbl, cloneCache)
+end
+
+--- Maps each element of `arr` through `refine` and joins the resulting strings with `sep` (default "").
```
```suggestion
--- Maps each element of `arr` through `map` and joins the resulting strings with `sep` (default "").
```

tests/std/tableext.test.luau
```diff
@@ -139,4 +139,106 @@ test.suite("TableExt", function(suite)
...
+		local cyclicClone: typeof(a)
+		assert.never.throws(function()
+			cyclicClone = tableext.deepClone(a)
+		end)
+		assert.eq(cyclicClone.arbitrary, 1)
+		assert.eq(cyclicClone.next.prev.arbitrary, 1)
+		assert.that(cyclicClone.next.prev ~= a)
+		assert.that(cyclicClone.next.prev == cyclicClone) -- cyclic edge correctly points to clone of a, not original
+		assert.that(cyclicClone ~= a)
+
```
aatxe, line 263:
```luau
  local shared = { x = 1 }
  local t = { a = shared, b = shared }
  local clone = tableext.deepClone(t)
  assert.that(clone.a == clone.b)  -- shared structure preserved
  assert.that(clone.a ~= shared)   -- but it's still a new table
```

It seems reasonable to explicitly test that we're preserving reference structures in the tables since that's a significant feature of deep cloning correctly.

tests/std/tableext.test.luau
```diff
@@ -139,4 +139,106 @@ test.suite("TableExt", function(suite)
...
+		assert.eq(emptyResult, "")
+
+		local single = { 42 }
+		local singleResult = tableext.mapConcat(single, function(v: number)
+			return tostring(v)
+		end, ",")
+		assert.eq(singleResult, "42")
+	end)
+
+	suite:case("deepClone", function(assert)
```
aatxe, line 199: We should probably also have tests for the things I called out as needing to be documented, i.e. support or non-support for key cloning and metatables.

lute/std/libs/tableext.luau
```diff
@@ -62,6 +62,54 @@ function tableext.keys<K, V>(tbl: { [K]: V }): { K }
...
+	end
+
+	if opts.preserveFrozen and table.isfrozen(tbl) then
+		table.freeze(clone)
+	end
+
+	return clone
+end
+
+--- Recursively deep-clones `tbl`. By default, metatables are stripped, table-keys are kept as-is (same reference), and frozen tables are not preserved.
```
```suggestion
--- Recursively deep-clones `tbl`. By default, metatables are stripped, keys are kept as-is without any cloning, and frozenness of tables is not preserved.
```

aatxe (approved):

PR #1170, opened 2026-06-12
Add `find` to `tableext`

sbi-n:
Returns the first key-value pair in the table which the predicate returns true. it seems like it would b a useful addition to the std library

### DIFF
--- a/lute/std/libs/tableext.luau
+++ b/lute/std/libs/tableext.luau
@@ -31,6 +31,16 @@ function tableext.filter<K, V>(table: { [K]: V }, predicate: (V) -> boolean): { 
 	return new
 end
 
+--- Returns a key-value pair from `table` for which `predicate` returns `true`. key-value pair is not guaranteed.
+function tableext.find<K, V>(table: { [K]: V }, predicate: (V) -> boolean): (V?, K?)
+	for k, v in table do
+		if predicate(v) then
+			return v, k
+		end
+	end
+	return nil, nil
+end
+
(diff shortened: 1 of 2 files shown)


lute/std/libs/tableext.luau
```diff
@@ -31,6 +31,16 @@ function tableext.filter<K, V>(table: { [K]: V }, predicate: (V) -> boolean): {
 	return new
 end
 
+--- Returns a key-value pair from `table` for which `predicate` returns `true`. key-value pair is not guaranteed.
```
```suggestion
--- Returns the first key-value pair from `table` for which `predicate` returns `true`. Returns `nil, nil` if the predicate does not hold for any value.
```

tests/std/tableext.test.luau
```diff
@@ -19,6 +19,19 @@ test.suite("TableExt", function(suite)
...
+		local b: { number } = { 1, 2, 3 }
+		local foundA = tableext.find(a, function(v: number)
+			return v % 2 == 0
+		end)
+		assert.eq(foundA, 2)
+		local foundB = tableext.find(b, function(v: number)
+			return v == 3
+		end)
+		assert.eq(foundB, 3)
+	end)
```
aatxe, line 45: A test for not finding something? A test where the predicate holds on multiple values?

aatxe (changes requested):

tests/std/tableext.test.luau
sbi-n: Then I'll add test cases for when TableExt.find doesn't find anything and when predicate holds multiple values

aatxe (approved):

PR #1173, opened 2026-06-15
lute pkg: update version constraint solving to add backtracking

afujiwara-roblox:
This PR replaces the greedy BFS resolver with backtracking solver supporting multi-major versions
  
# Core loop
The solver repeatedly picks the next unresolved slot, selects the highest satisfying candidate, propagates its dependencies as new constraints, and advances. When a slot has zero satisfying candidates, it backtracks up the decision stack trying alternative versions at earlier decision points until it finds a viable path or exhausts all options.

## Behaviors
- Backtracking. When no candidate satisfies accumulated constraints, the solver unwinds the decision stack and tries alternatives at earlier decision levels.
- Multi-major coexistence. Versions are grouped into compatibility slots (pkg@1, pkg@2, pkg@0.1). Only one version per slot; different slots resolve independently. ^1.0.0 and ^2.0.0 of the same package no longer conflict.
- Allow circular dependencies. Resolver successfully resolves circular dependencies and does not report an error
- Conflict reporting. Names the conflicting package and lists all unsatisfiable constraints with their sources.

## Optimizations
- MRV heuristic. The slot with the fewest viable candidates is resolved first to fail fast and reduce backtracking.

### DIFF
--- a/lute/cli/commands/pkg/loom-core/src/semver/solver.luau
+++ b/lute/cli/commands/pkg/loom-core/src/semver/solver.luau
@@ -57,7 +57,7 @@ local function slotFromConstraint(name: string, constraint: parser.Constraint): 
 		end
 	end
 
-	-- split nil checks for better type inference.
+	-- FIXME(Luau): we had to split nil checks for better type inference, refinements _should_ still work if it was one check
 	if lower == nil then
 		return nil
 	end
--- a/lute/cli/commands/pkg/loom-core/src/semver/solver.luau
+++ b/lute/cli/commands/pkg/loom-core/src/semver/solver.luau
@@ -191,48 +191,6 @@ local function propagateDeps(
 	return out
 end
 
-local function checkAcyclic(selected: Resolution, universe: Universe)
-	local path: { string } = {}
-	local onPath: { [string]: number } = {}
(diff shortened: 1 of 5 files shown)


lute/cli/commands/pkg/loom-core/src/semver/solver.luau
```diff
@@ -22,233 +21,331 @@ export type Requirement = {
...
+	local upper: parser.Version? = nil
+	for _, comp in constraint do
+		if (comp.op == ">=" or comp.op == ">") and lower == nil then
+			lower = comp.version
+		elseif comp.op == "<" and upper == nil then
+			upper = comp.version
 		end
 	end
+
+	-- split nil checks for better type inference.
```
aatxe:
```suggestion
	-- FIXME(Luau): we had to split nil checks for better type inference, refinements _should_ still work if it was one check
```

aatxe: We try to label these sorts of "known type system issues" with `FIXME(Luau)` (or I think sometimes `LUAUFIX`) so that we can check in the future that we're actually solving these problems in the type system.

aatxe (approved):

PR #1177, opened 2026-06-17
Update actions to support Node.js 24 following Node.js 20 deprecation

annieetang:
Fixing warnings in github actions (see [nightly release](https://github.com/luau-lang/lute/actions/runs/27659029838)) following the Node.js 20 deprecation. The only remaining is `Roblox/setup-foreman@v3` because there isn't a newer version available 

Tested by running a nightly release! https://github.com/luau-lang/lute/actions/runs/27712327753


aatxe (approved):

aatxe:
The final release failed because of release immutability (expected), but otherwise looks like everything ran as expected!

PR #1181, opened 2026-06-18
Combine CstTableTypeItemStringProperty and CstTableTypeItemProperty

skberkeley:
As part of introducing AST types to Lute, we would like AST types to subtype structurally against CST types. However, the AST returned by the Luau parser provides no way to distinguish between table type properties that are bracketed and those that aren't. This means that the eventual Lute AST will have a single type for bracketed and non-bracketed table type properties, so the Lute CST should too.


definitions/syntax/cst.luau
```diff
@@ -714,30 +714,21 @@ export type CstTableTypeItemIndexer = {
...
-	read colon: CstToken<":">,
-	read value: CstType,
-	read separator: CstToken<"," | ";">?,
-}
-
 --- A named property entry in a table type (e.g., `key: T`).
 export type CstTableTypeItemProperty = {
 	read kind: "property",
 	read access: CstToken<"read" | "write">?,
+	read indexerOpen: CstToken<"[">?,
```
Vighnesh-V, line 721: tiny nit:

If you're talking about properties that are defined like:
```
local t = { ["$$Illegal identifier character"] = ... }
```

Maybe `nonIdentKeyOpen

aatxe (approved):

PR #1182, opened 2026-06-18
CST Fixups

skberkeley:
Small fixups to the CST:
- Renames `CstStatCompoundAssign.operand` to be operator. This was a misnomer before.
- Makes `CstExprFunction`'s `generics` and `genericPacks` nonoptional and just empty tables if a function doesn't have generics.
- Adds `hasVararg` to `CstExprFunction` in preparation for adding `AstExprFunction`. Currently, vararg presence is indicated by a nilable `...` token, but this field won't be present in `AstExprFunction`.
- Adds `isExported` to `CstStatTypeAlias` and `CstStatTypeFunction`. Like `CstExprFunction`, exported-ness is indicated by nilable `export` token which will be absent from their AST counterparts.
- Use `Luau::CstTypeGroup` and `Luau::CstExprGroup` which were recently added to correctly encode the position of closing parentheses.


lute/syntax/src/parser.cpp
```diff
@@ -1074,10 +1084,21 @@ struct AstSerialize : public Luau::AstVisitor
...
+            lua_setfield(L, -2, "vararg");
+        }
         else
+        {
+            lua_pushboolean(L, 0);
+            lua_setfield(L, -2, "hasVararg");
+
             lua_pushnil(L);
-        lua_setfield(L, -2, "vararg");
+            lua_setfield(L, -2, "vararg");
```
annieetang, line 1100: you could probably keep the `lua_setfield(L, -2, "vararg");` out of the if else block right?

lute/syntax/src/parser.cpp
```diff
@@ -1740,17 +1761,28 @@ struct AstSerialize : public Luau::AstVisitor
...
+            lua_setfield(L, -2, "export");
+        }
         else
+        {
+            lua_pushboolean(L, 0);
+            lua_setfield(L, -2, "isExported");
+
             lua_pushnil(L);
-        lua_setfield(L, -2, "export");
+            lua_setfield(L, -2, "export");
```
annieetang, line 1784: ditto above comment (and same with the export set field below)

lute/syntax/src/parser.cpp
skberkeley: i considered it, but this feels more readable

annieetang: hm ok, as long as this file's style follows that? i just know some other files do the other style

skberkeley: yeah i'll usually factor the last `setfield` out if we're only setting one property, but in this case we're setting two

annieetang (approved):

PR #1183, opened 2026-06-19
Set CURL_USE_SCHANNEL on windows

jkelaty-rbx:
Switches the Windows libcurl build to Schannel. On our Windows CI runners, `lute pkg install` hangs for 30 seconds and times out at libcurl's connect-phase limit, and the current suspicion is that using the native backend would resolve the issue, though the specifics of how/why are a bit out of my reach.


skberkeley:
Have you been able to verify that this change fixes the issue?

aatxe (approved):

jkelaty-rbx:
@skberkeley Thank you for the nudge on this! I tested this more thoroughly and I found this PR would not have fixed my issue, so I'm going to close this. I wasn't able to fix this in the most correct/ideal way (yet), but I found a simple enough short-term fix, that at least helps diagnose the issue further: https://github.com/luau-lang/lute/pull/1195

PR #1184, opened 2026-06-20
Add `@lute/keyring` and remove plaintext tokens from `lute pkg auth`

jkelaty-rbx:
`~/.loom/auth.luau` currently stores tokens in plaintext on disk. This PR replaces it with the OS-native credential store via a new `@lute/keyring` module.

I chose to just completely remove the old plaintext authentication code (aside from a `migrate` subcommand), but we could also choose to leave it. I personally see no point in maintaining a less-secure authentication method, but I'm open to changing this.

`@lute/keyring` api:
```luau
function keyring.set(service: string, account: string, password: string): ()
function keyring.get(service: string, account: string): string?
function keyring.delete(service: string, account: string): ()
function keyring.isSupported(): boolean
```

`lute pkg auth` api changed from:
```
Usage: `lute pkg auth --token <token> [--domain <domain>]`
```
to:
```
Usage: lute pkg auth <action> [options]
Manage loom's authentication credentials in the OS-native credential
store (Keychain on macOS, Credential Manager on Windows).
Actions:
  add --token <token> [--domain <domain>]   Store a credential
  remove              [--domain <domain>]   Remove a stored credential
  migrate                                   Move legacy plaintext credentials
                                            from ~/.loom/auth.luau into the keyring
```

Inspired by: https://github.com/Roblox/foreman/pull/108 :-)


jkelaty-rbx:
I'm exploring a different direction that's a bit simpler and safer, I think

PR #1187, opened 2026-06-23
Update create-pr-request version

skberkeley:
Our update luau/lute job started failing with 
```
remote: Duplicate header: "Authorization"
fatal: unable to access 'https://github.com/luau-lang/lute/': The requested URL returned error: 400
```

After some Googling, it turns out that `create-pull-request@v7` is incompatible with `actions/checkout@v6`, and that we should use at least [v7.0.9](https://github.com/peter-evans/create-pull-request/releases/tag/v7.0.9) instead. I synced our fork of `create-pull-request` to the upstream, including tags, and bumped the version we use in the create Luau/Lute job. While testing I had to add a base branch argument which I kept since it doesn't hurt.

### DIFF
--- a/.github/workflows/update-prs.yml
+++ b/.github/workflows/update-prs.yml
@@ -1,13 +1,9 @@
 name: Update Luau/Lute
 
-# on:
-#   schedule:
-#     - cron: "0 0 * * 6"
-#   workflow_dispatch:
-
 on:
-  pull_request:
-    branches: ["primary"]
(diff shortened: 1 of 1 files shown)


annieetang (approved):
ty! i guess the nightly test didn't catch this?

skberkeley:
> ty! i guess the nightly test didn't catch this?

the update ci job failed and nobody noticed lol

Issue #1178, opened 2026-06-17
Windows: net.request hangs on versions > 1.0.1-nightly.20260513


fossinating:
Example hanging code:

```
local net = require("@std/net")

net.request("https://captive.apple.com")
```

Upgrading beyond this version causes the request to hang and fail after 30 seconds:

```
network request failed: Resolving timed out after 30007 milliseconds
stacktrace:
[C] function request
@std/net.luau:19 function request
.lute/example.luau:3
```

I've tested this on my windows machine and had a teammate test on their mac, does not seem to happen for them on mac, so leads me to believe it is likely a windows issue

aatxe:
This issue was reported as resolved by #1197 on the thread for #1195, so I'm gonna close it out. Don't hesitate to reopen it if you continue to encounter the issue again.

PR #1195, opened 2026-06-24
Disable threaded resolver

jkelaty-rbx:
On Windows, `lute pkg install` (and any `@std/net.request` call) hangs for exactly 30 seconds before failing with `network request failed: Resolving timed out after 30000 milliseconds`. This PR disables libcurl's threaded resolver so DNS falls back to the synchronous code path, which experimentally fixes the issue I'm observing. I'm still unsure of the actual issue myself, but it seems to potentially be related to DNS resolution. One theory is that our http client doesn't properly handle observing the async DNS resolution step, but I'm not familiar enough with curl to really understand the details.

This is a short-term fix, I think. Assuming the theory I stated is right, then this should be fixable in our http client implementation.

This also seems directly related to: https://github.com/luau-lang/lute/issues/1178


aatxe (approved):

aatxe:
I think I'm okay with this change (the alternative seems to be switching curl to use c-ares as a separate DNS resolver), but I am wondering if we've tried just upgraded curl first. There's been a number of fixes in recent curl versions related to the DNS resolver from what I can tell, so might just be worth upgrading to 8.21.0 and seeing if that fixes it.

jkelaty-rbx:
> so might just be worth upgrading to 8.21.0 and seeing if that fixes it.

I'll try that next and report back here!

jkelaty-rbx:
So far this seems to be resolved for me with the curl upgrade: https://github.com/luau-lang/lute/pull/1197 I'll close this once the nightly build lands tonight and I can validate it e2e

aatxe:
Awesome, glad to hear it!

jkelaty-rbx:
Confirmed this seems to be resolved now. Thank you for your help with this @aatxe!

PR #1007, opened 2026-04-18
`lute pkg`: workspaces

jkelaty-rbx:
Adds workspace support to loom:
* Root manifests can declare `workspace = { ... }` member paths, which appear under `members` in the lockfile
* Member names participate in cross-source name conflict detection alongside dependencies

The core resolution step read pretty well (not accounting for all the positional args)
```luau
	local queue: { manifest.DependencySpec } = {}
	local registry = createRegistryContext()

	local rootManifest = manifest.parseFile(rootDirectory)
	registry.names[rootManifest.package.name] = rootManifest.package.name

	local dependencies = enqueueDependencies(
		queue,
		registry,
		rootManifest.package.dependencies,
		rootDirectory,
		rootDirectory,
		rootManifest.package.name
	)
	local members = seedWorkspaceMembers(queue, registry, rootDirectory, rootManifest)
	local packages = processSpecQueue(queue, registry, rootDirectory)

	if next(registry.requirements) ~= nil then
		assert(universe ~= nil, "Registry dependencies require a universe to be provided")
		solveRegistryRequirements(packages, registry, universe)
	end
```

PR #1199, opened 2026-06-24
`lute pkg`: refactor resolution in terms of modular helpers

jkelaty-rbx:
Refactors lute's dependency resolution implementation to be slightly more modular with semantically named functions. My intention/hope is to make the code easier to read and extend.

One issue that this refactor caught was https://github.com/luau-lang/lute/pull/1194 did not use the same path-spec normalization helper, which was caught with unit tests after the refactor correct this.

### DIFF
--- a/lute/cli/commands/pkg/loom-core/src/resolution.luau
+++ b/lute/cli/commands/pkg/loom-core/src/resolution.luau
@@ -199,9 +199,6 @@ local function processSpecQueue(
 	return packages
 end
 
--- Solve all collected registry requirements in one pass and stitch the resolved
--- versions into the package graph: emit one packages entry per resolved version,
--- and back-patch every fixup's alias table with the concrete `name@version` key.
 local function solveRegistryRequirements(
 	packages: { [string]: lockfile.LockfileDependency },
 	registry: RegistryContext,
--- a/lute/cli/commands/pkg/loom-core/src/resolution.luau
+++ b/lute/cli/commands/pkg/loom-core/src/resolution.luau
@@ -230,7 +230,7 @@ local function solveRegistryRequirements(
 			source = registry.sources[name] or name,
 			sourceKind = "registry" :: "registry",
 			installPath = store.toStorePath(key),
-			dependencies = transitiveDependencies,
+			dependencies = table.freeze(transitiveDependencies),
 		})
 	end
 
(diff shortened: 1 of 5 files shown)


lute/cli/commands/pkg/loom-core/src/resolution.luau
```diff
@@ -1,18 +1,29 @@
...
-local pp = require("@batteries/pp")
 local manifest = require("./manifest")
+local path = require("@std/path")
 local semver = require("./semver")
 local store = require("./store")
 
-local function getManifestPath(directory: string): string
-	return `{directory}/loom.config.luau`
-end
+type RegistryFixup = {
```
aatxe: Can we call this `RegistryDiff`, `RegistryPatch`, `RegistryUpdate`, or `RegistryDelta` or something of that sort? `fixup` is confusing/surprising language to me.

lute/cli/commands/pkg/loom-core/src/resolution.luau
```diff
@@ -215,75 +192,102 @@ local function resolve(rootDirectory: string, universe: semver.Universe?, option
...
+			transitiveDependencies[alias] = `{dependency.name}@{resolvedVersion.versionString}`
 		end
+
+		packages[key] = table.freeze({
+			name = name,
+			rev = pkgVersion.versionString,
+			source = registry.sources[name] or name,
+			sourceKind = "registry" :: "registry",
+			installPath = store.toStorePath(key),
+			dependencies = transitiveDependencies,
```
aatxe: Do we mean to unfreeze the transitiveDependencies with this change?

aatxe (commented):
Overall, looks like a solid improvement to me. Just left a few comments. Thanks! 😄

lute/cli/commands/pkg/loom-core/src/resolution.luau
jkelaty-rbx: Good catch! I guess I did this to treat dependency freezing uniformly as a post-processing step:
https://github.com/luau-lang/lute/blob/28617f5938a02510cb3ea7d9084692752b9a1271/lute/cli/commands/pkg/loom-core/src/resolution.luau#L274-L281
but I see we can freeze them at this point since they should be unaffected by the resolution patching step below

lute/cli/commands/pkg/loom-core/src/resolution.luau
jkelaty-rbx: Done! Renamed to RegistryPatch

afujiwara-roblox (approved):

jkelaty-rbx:
See: https://github.com/luau-lang/lute/pull/1007#issuecomment-4949650192

PR #1201, opened 2026-06-25
lute pkg: Use lockfile for dependency resolution

afujiwara-roblox:
Summary
  
  - Make lute pkg install read the existing loom.lock.luau before resolving. If the lockfile satisfies the current manifest and path dependencies haven't changed, skip resolution entirely.
  - When re-resolution is needed, the solver prefers previously-locked registry versions to minimize version churn
  - Add --locked flag that errors if the lockfile is missing or doesn't match the manifest

### DIFF
--- a/tests/cli/pkg/install.test.luau
+++ b/tests/cli/pkg/install.test.luau
@@ -203,6 +203,5 @@ test.suite("LutePkgInstall", function(suite)
 
 		assert.eq(r1.exitcode, 0)
 		assert.eq(r2.exitcode, 0)
-		assert.eq(r1.stdout, r2.stdout)
 	end)
 end)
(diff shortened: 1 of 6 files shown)


jescherob (commented):
Code looks good, however the old test `reinstallIsIdempotent` in install.test.luau now seems to fail the last condition `assert.eq(r1.stdout, r2.stdout)` as while the second install was the same before, it now prints the additional line of Lockfile up to date. Additionally, think the stylua/lint need to be rerun!

tests/cli/pkg/install.test.luau
```diff
@@ -203,6 +203,5 @@ test.suite("LutePkgInstall", function(suite)
 
 		assert.eq(r1.exitcode, 0)
 		assert.eq(r2.exitcode, 0)
-		assert.eq(r1.stdout, r2.stdout)
```
afujiwara-roblox, line 206: @aatxe I'm removing the stdout check for the "reinstallIsIdempotent" test. The second run prints "lockfile up to date" to signify that the existing resolution matches the manifest and a resolve was not necessary. I believe this is still considered idempotent behavior since the exit code and disk state are the same across runs. Maybe we could harden this test to verify the lockfiles and packages installed are the same too. Let me know if you think we should handle this differently

jescherob (approved):

PR #1202, opened 2026-06-26
Add ability to set breakpoints along with a new `lute/debug` library

wizhao96:
Adds the basic ability to set breakpoint to a target file. Introduces a `lute/debug` library that will be contributed to for future debugging features.

Notes
* hit breakpoints don't actually stop execution currently
* debugging files that are loaded from `require` is not yet supported
* while the library will eventually be exposed to Luau scripts, right now it is C++ only


lute/debug/src/debugger.cpp
```diff
@@ -0,0 +1,116 @@
...
+    if (bp.status == BreakpointStatus::Installed)
+        installedBreakpoints.push_back(bp);
+    else
+        pendingBreakpoints.push_back(bp);
+    return bp;
+}
+
+std::vector<Breakpoint> Target::getBreakpoints() const
+{
+    std::vector<Breakpoint> all;
```
Vighnesh-V, line 74: Does `std::vector<_>::insert` do a reserve() on the input iterators size? If so, no worries, if not, can you do an all.reserve(installedBreakpoints.size() + pendingBreakpoints.size())?

lute/debug/src/debugger.h
```diff
@@ -0,0 +1,62 @@
...
+
+private:
+    std::string executablePath;
+    Runtime& runtime;
+
+    int currentBreakpointId = 0;
+    std::vector<Breakpoint> pendingBreakpoints;
+    std::vector<Breakpoint> installedBreakpoints;
+
+    std::map<std::string, std::shared_ptr<Ref>> loadedSources;
```
Vighnesh-V: Do you need an ordered map here? If possible you should avoid std::map because insertions and deletions are O(log n), and Red Black Tree operations don't really explit spatial locality, so they can be a bit slower. We have some more performance versions of this implemented - if you need a unordered_map, you can use `Luau::DenseMap`, the only annoying thing is it requires a sentinel tombstone.

lute/debug/src/debugger.h
```diff
@@ -0,0 +1,62 @@
...
+
+    bool launch(std::vector<std::string> args);
+
+private:
+    std::string executablePath;
+    Runtime& runtime;
+
+    int currentBreakpointId = 0;
+    std::vector<Breakpoint> pendingBreakpoints;
+    std::vector<Breakpoint> installedBreakpoints;
```
Vighnesh-V: Can you leverage a map data structure here to make lookups faster?

tests/src/debug.test.cpp
```diff
@@ -0,0 +1,67 @@
+#include "debugfixture.h"
+#include "doctest.h"
+
+TEST_CASE_FIXTURE(DebugFixture, "Debug_addBreakpoint")
```
Vighnesh-V: giganit: might be good to stick this in a TEST_SUITE to group similar sets of tests.

tests/src/debug.test.cpp
```diff
@@ -0,0 +1,67 @@
...
+{
+    std::string source = "local x = 1\nlocal y = 2\n\nlocal z = x + y\n";
```
Vighnesh-V: In the tests/ folder, can you add this source file as a real fixture on disk and write some helpers to open these up? This will make testing requires much easier.

lute/debug/src/debugger.h
```diff
@@ -0,0 +1,62 @@
...
+
+    std::map<std::string, std::shared_ptr<Ref>> loadedSources;
+
+    bool installBreakpoint(Breakpoint& bp, lua_State* L);
+    void installPendingBreakpoints(lua_State* L);
```
Vighnesh-V: As a convention, we try to make the lua_state* come first here!

lute/debug/src/debugger.h
```diff
@@ -0,0 +1,62 @@
...
+    Invalid,
+};
+
+struct Breakpoint
+{
+    int id;
+    std::string sourcePath;
+    int line;
+    BreakpointStatus status;
+    Breakpoint(int id, std::string sourcePath, int line, BreakpointStatus status);
```
Vighnesh-V: nit: mark as explicit if this is the only constructor type.

lute/debug/src/debugger.h
```diff
@@ -0,0 +1,62 @@
...
+    bool operator==(const Breakpoint& other) const
+    {
+        return id == other.id;
+    }
+};
+
+struct Target
+{
+    Target(std::string sourcePath, Runtime& runtime);
```
Vighnesh-V: nit: flip the order of arguments :)

lute/debug/src/debugger.h
```diff
@@ -0,0 +1,62 @@
...
+    // involve a source that has already been loaded by the VM, we attempt to install that
+    // breakpoint. Otherwise, it exists as a pending breakpoint until new sources are loaded.
+    // We do this because clients may 1) configure breakpoints before launch executables
+    // 2) we load sources dynamically with @require that a client may want to debug.
+    // TODO: implement 2 and add some callback when breakpoints get installed
+    Breakpoint addBreakpoint(std::string sourcePath, int line);
+    std::vector<Breakpoint> getBreakpoints() const;
+    std::optional<Breakpoint> getBreakpointById(int breakpointId) const;
+
+    bool launch(std::vector<std::string> args);
```
Vighnesh-V:
```suggestion
    bool launch(const std::vector<std::string>& args);
```

lute/debug/src/debugger.h
```diff
@@ -0,0 +1,62 @@
...
+
+private:
+    std::string executablePath;
```
Vighnesh-V: naming nit: technically a script path :)

lute/debug/src/debugger.cpp
```diff
@@ -0,0 +1,116 @@
...
+{
+}
+
+Target::Target(std::string executablePath, Runtime& runtime)
+    : executablePath(executablePath)
+    , runtime(runtime)
+{
+}
+
+Breakpoint Target::addBreakpoint(std::string sourcePath, int line)
```
afujiwara-roblox, line 33: This is still a draft, but we should make sure to add a way to remove breakpoints in the same PR.

tests/src/debug.test.cpp
```diff
@@ -0,0 +1,67 @@
...
+    CHECK(postLaunch->status == debug::BreakpointStatus::Invalid);
+    CHECK(postLaunch->line == -1);
+
+    postLaunch = target.getBreakpointById(2);
+    REQUIRE(postLaunch.has_value());
+    CHECK(postLaunch->status == debug::BreakpointStatus::Installed);
+    CHECK(postLaunch->line == 4);
+
+    // check that adding breakpoints after launch should be immediately installed
+    debug::Breakpoint bp4 = target.addBreakpoint(scriptPath, 2);
```
afujiwara-roblox: nit: unused variable

lute/debug/src/debugger.cpp
```diff
@@ -0,0 +1,116 @@
...
+{
+    int id = currentBreakpointId;
+    currentBreakpointId++;
+    Breakpoint bp(id, sourcePath, line, BreakpointStatus::Pending);
+    installBreakpoint(bp, runtime.GL);
+    if (bp.status == BreakpointStatus::Installed)
+        installedBreakpoints.push_back(bp);
+    else
```
afujiwara-roblox: What if bp.status is BreakpointStatus::Invalid? I don't think we're handling invalid breakpoints anywhere. The current behavior is just polluting pendingBreakpoints with invalid breakpoints that never get cleaned. It might be good to store them separately so we can use them for error messages

lute/debug/src/debugger.cpp
```diff
@@ -0,0 +1,116 @@
...
+            return bp;
+    for (const Breakpoint& bp : pendingBreakpoints)
+        if (bp.id == breakpointId)
+            return bp;
+    return std::nullopt;
+}
+
+bool Target::installBreakpoint(Breakpoint& bp, lua_State* L)
+{
+    if (loadedSources.find(bp.sourcePath) == loadedSources.end())
```
afujiwara-roblox: nit: we're looking up bp.sourcePath in loadedSources twice

tests/src/debugfixture.cpp
```diff
@@ -0,0 +1,30 @@
+#include "debugfixture.h"
+
+#include "lute/runtime.h"
+
+#include <unistd.h>
```
afujiwara-roblox: Can we use this library? Lute (including tests) needs to run on windows

Vighnesh-V (changes requested):
Looks generally good - mostly questions about the data structures you're using.

lute/debug/src/debugger.cpp
wizhao96: i don't think it's necessary to call `reserve` according to https://stackoverflow.com/questions/2208293/what-is-the-most-efficient-way-to-append-one-stdvector-to-the-end-of-another.

lute/debug/src/debugger.h
wizhao96: i just renamed this `sourcePath` to be consistent with the breakpoints naming

lute/debug/src/debugger.h
wizhao96: i've generally combined all breakpoints into one `unordered_map` for fast lookup (this is an unordered_map instead of a `Luau::DenseHashMap` to support deleting breakpoints)

lute/debug/src/debugger.cpp
```diff
@@ -0,0 +1,144 @@
...
+    if (bp.status == BreakpointStatus::Installed)
+    {
+        auto chunkRef = loadedSources.find(bp.sourcePath);
+        if (chunkRef)
+        {
+            (*chunkRef)->push(runtime.GL);
+            int removed_line = lua_breakpoint(runtime.GL, -1, bp.line, 0);
+            lua_pop(runtime.GL, 1);
+            if (removed_line == -1)
+                runtime.reporter.reportError("installed breakpoint is invalid in source");
```
wizhao96: errors like this returns `true` because we have successfully removed the bp from our internal map representation. but perhaps someway to note that the actual bp removal on the source failed would be useful

lute/debug/src/debugger.cpp
wizhao96: ok generally i've coalesced all of the bp maps into one map that we can then use `getBreakpointsByStatus` to filter (so this resolves this concern)

tests/src/debugfixture.cpp
wizhao96: ok yep this dependency has been removed now given that we are now reading test scripts from disk

lute/debug/src/debugger.h
wizhao96: `loadedSources` is now using `Luau::DenseHashMap`

lute/debug/src/debugger.cpp
afujiwara-roblox: The error can be a little more descriptive. Let's include at least the source path and line, maybe the id too.

tests/src/debug.test.cpp
```diff
@@ -0,0 +1,89 @@
...
+
+        // check that adding breakpoints after launch should be immediately installed
+        debug::Breakpoint bp4 = target.addBreakpoint(fixturePath, 2);
+        CHECK(target.getBreakpoints().size() == 4);
+        CHECK(bp4.status == debug::BreakpointStatus::Installed);
+        CHECK(bp4.line == 2);
+        CHECK(bp4.id == 3);
+    }
+
+    TEST_CASE_FIXTURE(DebugFixture, "Debug_removeBreakpoint")
```
afujiwara-roblox, line 72: Can you add a test to remove an installed breakpoint

tests/src/debug.test.cpp
```diff
@@ -0,0 +1,89 @@
...
+    {
+        std::string fixturePath = getDebugFixturePath("simple.luau");
+        debug::Target target(*runtime, fixturePath);
+
+        // check removing breakpoints
+        debug::Breakpoint bp = target.addBreakpoint(fixturePath, 2);
+        CHECK(target.getBreakpoints().size() == 1);
+
+        bool ok = target.removeBreakpoint(0);
```
afujiwara-roblox:
```suggestion
        bool ok = target.removeBreakpoint(bp.id);
```
nit: If you ever change how breakpoint ids are generated, this test will break. Unless there is a specific reason the id here needs to be 0, it's probably better to test "removing the same breakpoint I added" instead of "adding a breakpoint with id 0 and removing a breakpoint with id 0"

lute/debug/src/debugger.cpp
```diff
@@ -0,0 +1,144 @@
...
+{
+}
+
+Breakpoint Target::addBreakpoint(std::string sourcePath, int line)
+{
+    int id = currentBreakpointId;
+    currentBreakpointId++;
+    Breakpoint bp(id, sourcePath, line, BreakpointStatus::Pending);
+    installBreakpoint(runtime.GL, bp);
+    breakpoints.insert_or_assign(id, bp);
```
afujiwara-roblox:
```suggestion
auto [it, _] = breakpoints.insert_or_assign(id, Breakpoint(id, sourcePath, line, BreakpointStatus::Pending));
installBreakpoint(runtime.GL, it->second);
return it->second;
```
Not sure if the syntax of this suggestion is right, but ensuring there is a single source of truth for the breakpoint can help avoid subtle bugs in the future. It would be a bug if the order was reversed. `breakpoints` stores Breakpoint values, so `installBreakpoint` would be mutating a different copy than the one you're storing.
```
breakpoints.insert_or_assign(id, bp);
installBreakpoint(runtime.GL, bp);
```

lute/debug/src/debugger.h
```diff
@@ -0,0 +1,64 @@
...
+    // 2) we load sources dynamically with @require that a client may want to debug.
+    // TODO: implement 2 and add some callback when breakpoints get installed
+    Breakpoint addBreakpoint(std::string sourcePath, int line);
+    bool removeBreakpoint(int bpId);
+
+    std::vector<Breakpoint> getBreakpoints() const;
+    std::vector<Breakpoint> getBreakpointsByStatus(BreakpointStatus status) const;
+    std::optional<Breakpoint> getBreakpointById(int breakpointId) const;
+
+    bool launch(const std::vector<std::string>& args);
```
wizhao96, line 49: `launch` will return a `Process` structure similar to https://lldb.llvm.org/python_api/lldb.SBProcess.html. change should come in the next PR.

Vighnesh-V (approved):
Approved! The only consideration I have for you is that the DAP server has to run in the main runtime, which means this structure probably needs to create it's own child runtime within which to execute the program. I don't think this is a blocking issue though, and you'll probably end up working through this as you go.

wizhao96:
yea I've actually been thinking about this. spawning a child process will probably go in the next PR.

> Approved! The only consideration I have for you is that the DAP server has to run in the main runtime, which means this structure probably needs to create it's own child runtime within which to execute the program. I don't think this is a blocking issue though, and you'll probably end up working through this as you go.

PR #1205, opened 2026-06-29
`lute self`: Host lute installation manager

jkelaty-rbx:
Adds a `lute self` subcommand for installing, updating, and uninstalling lute from a well-known system location (~/.lute/bin). This gives users a first-class way to bootstrap and maintain their lute installation without relying on external toolchain managers.

### Subcommands

- `lute self install`: Copies the running lute binary to `~/.lute/bin` and adds it to PATH via shell profile modifications (Unix) or the user registry (Windows)
- `lute self update`: Queries the latest GitHub release (or a specific `--version`) and replaces the installed binary
- `lute self uninstall`: Removes the binary directory and strips PATH entries from all known shell profiles. Supports `--remove-data` to also delete cached type definitions under `~/.lute`


lute/cli/commands/self/update.luau
```diff
@@ -0,0 +1,74 @@
...
+	local current: string? = nil
+	local ok, result = pcall(process.run, { destination, "--version" }, nil)
+	if ok and result.ok then
+		current = release.normalizeTag(result.stdout)
+	else
+		print(`warning: could not read version from {destination}; proceeding with update`)
+	end
+
+	if requested == nil and current ~= nil and string.find(current, "-nightly.", 1, true) ~= nil then
+		error("Automatic updates are not supported for nightly releases. Use --version to select a nightly release.")
```
jkelaty-rbx, line 58: planned follow-up

lute/cli/commands/self/release.luau
```diff
@@ -0,0 +1,100 @@
+local json = require("@std/json")
+local net = require("@std/net")
+local path = require("@std/path")
+local process = require("@std/process")
+local stringext = require("@std/stringext")
+local system = require("@std/system")
+local unzip = require("../pkg/loom-core/extern/unzip")
```
aatxe, line 7: I think we should probably leave this for now, but it kinda sucks that this is depending on a submodule of a different subcommand. A part of me looks at it and wants something like `lute/cli/extern` and the ability to require `@extern/unzip`, but I'm also wary of just piling on more special aliases into lute. Especially since the ideal is that we would use `lute pkg` for our luau dependencies, at which point we could grab `unzip` as a package itself.

tools/install.sh
```diff
@@ -0,0 +1,59 @@
...
+asset="lute-${os_name}-${arch_name}.zip"
+url="https://github.com/${REPO}/releases/latest/download/${asset}"
+
+tmp="$(mktemp -d)"
+trap 'rm -rf "$tmp"' EXIT
+
+echo "lute: downloading $asset"
+curl --proto '=https' --tlsv1.2 -fsSL "$url" -o "$tmp/lute.zip"
+
+unzip -q "$tmp/lute.zip" -d "$tmp"
```
aatxe: I don't think Linux machines will _always_ have `unzip` (thinking particularly about alpine and other minimal container environments). Can we add an error message here if unzip isn't present?

aatxe (commented):
Looks good, would just like to add the error for missing `unzip` to `install.sh` (or if you're feeling particularly spicy, python does have a deflate implementation in its standard library, and I'd feel comfortable relying on the assumption that python is present on every unix machine).

tools/install.sh
jkelaty-rbx: Added!

lute/cli/commands/self/release.luau
jkelaty-rbx: > Especially since the ideal is that we would use lute pkg for our luau dependencies, at which point we could grab unzip as a package itself.

Yeah thats my thinking as well. It's tempting to tidy some of this up in the meantime which I don't think would be a bad thing. Agreed on leaving this for now

aatxe (approved):

Issue #1076, opened 2026-05-07
parser.parse records attribute groups incorrectly


sircfenner:
For example, the relevant part of ...
```Lua
@[deprecated]
local function x()
end
```
... is recorded as an attribute node with `text="@deprecated"`, a `"@["` leading trivia item, and no record of the closing `"]"` in the trailing trivia of the attribute node or in the leading trivia of the following "local" node. 

As a result, using std/printer to transform the parse result into source text results in:
```Lua
@[@deprecated
local function x()
end
```
***
For groups with more than one attribute, e.g.
```Lua
@[deprecated, native]
```
... the relevant part of the output is ...
```Lua
@[@deprecated @native
```
Note also the missing comma.
***
Attribute parameters work fine, though the original issue persists. However, they seem to bring in the missing closing `"]"`:
```Lua
@[deprecated { use = "another" }]
```
... outputs to ...
```Lua
@[@deprecated{ use = "another" }]
```
... though there is some whitespace loss between the attribute name and the parameter.
***
Tested on v1.0.0

aatxe:
Thanks for the report! This is definitely a bug, I'll take a look.

aatxe:
I investigated this, and the underlying issue is actually in Luau itself and how Luau's parser deals with attributes. @skberkeley is going to go ahead and add proper CST support for the bracketed attributes, and we'll be able to fix this properly once that's in.

skberkeley:
Update: The fix on the Luau side has been implemented and will be available in the 726 Luau release, after which we can do the work needed in Lute.

PR #1210, opened 2026-06-30
Correctly serialize parametrized attributes and attribute lists

skberkeley:
This PR renames the existing `CstAttribute` node type to `CstAttributeSimple`, and adds `CstAttributeParametrized` and `CstAttributeList`.

Addresses #1076.

### DIFF
--- a/definitions/syntax/cst.luau
+++ b/definitions/syntax/cst.luau
@@ -543,8 +543,18 @@ export type CstAttributeParametrized = {
 	read closeParens: CstToken<")">,
 }
 
+-- A (possibly bracketed) list of comma-separated attributes
+export type CstAttributeList = {
+	read location: Span,
+	read kind: "attribute",
+	read tag: "list",
+	read atBracket: CstToken<"@[">,
+	read attributes: CstPunctuated<CstAttribute, ",">,
+	read closeBracket: CstToken<"]">,
+}
+
 --- Union of all attribute node types.
-export type CstAttribute = CstAttributeSimple | CstAttributeParametrized
+export type CstAttribute = CstAttributeSimple | CstAttributeParametrized | CstAttributeList
 
 --- A function declaration statement node.
 export type CstStatFunction = {
--- a/definitions/syntax/cst.luau
+++ b/definitions/syntax/cst.luau
@@ -524,12 +524,27 @@ export type CstStatCompoundAssign = {
 	read value: CstExpr,
 }
 
---- A function attribute node (`@checked`, `@native`, or `@deprecated`).
-export type CstAttribute = {
+--- A simple function attribute node (e.g., `@checked`, `@native`, `@deprecated`).
+export type CstAttributeSimple = {
 	read location: Span,
 	read kind: "attribute",
-	read name: CstToken<"@checked" | "@native" | "@deprecated">,
+	read tag: "simple",
+	read name: CstToken,
 }
+
+--- A parametrized function attribute node (e.g., `@[deprecated("reason")]`).
+export type CstAttributeParametrized = {
+	read location: Span,
+	read kind: "attribute",
+	read tag: "parametrized",
+	read name: CstToken,
+	read openParens: CstToken<"(">,
+	read arguments: CstPunctuated<CstExpr>,
+	read closeParens: CstToken<")">,
+}
+
+--- Union of all attribute node types.
+export type CstAttribute = CstAttributeSimple | CstAttributeParametrized
 
 --- A function declaration statement node.
 export type CstStatFunction = {
--- a/lute/syntax/src/parser.cpp
+++ b/lute/syntax/src/parser.cpp
@@ -618,15 +618,99 @@ struct AstSerialize : public Luau::AstVisitor
         }
     }
 
-    void serializeAttributes(Luau::AstArray<Luau::AstAttr*>& attrs, size_t nrec = 0)
+    void serializeAttributeList(const Luau::AstArray<Luau::AstAttr*>& attrs, const Luau::CstAttrList& attrList, size_t& attrIndex)
     {
-        lua_rawcheckstack(L, 2);
-        lua_createtable(L, attrs.size, nrec);
+        LUTE_ASSERT(attrIndex < attrs.size);
 
-        for (size_t i = 0; i < attrs.size; i++)
+        lua_rawcheckstack(L, 4);
+        lua_createtable(L, 0, preambleSize + 3); // atBracket, attributes, closeBracket
+
+        withLocation(Luau::Location{attrList.atBracketPosition, attrList.closeBracketPosition});
+
+        lua_pushstring(L, "attribute");
+        lua_setfield(L, -2, "kind");
+
+        lua_pushstring(L, "list");
+        lua_setfield(L, -2, "tag");
+
+        LUTE_ASSERT(attrList.atBracketPosition.hasValue());
+        serializeToken(attrList.atBracketPosition, "@[");
+        lua_setfield(L, -2, "atBracket");
+
+        size_t lastAttrIndex = attrIndex + attrList.commaPositions.size + 1;
+        LUTE_ASSERT(lastAttrIndex <= attrs.size);
+
+        lua_createtable(L, attrList.commaPositions.size + 1, 1);
+
+        // separators
+        lua_createtable(L, attrList.commaPositions.size, 0);
+
+        size_t numAttrs = attrList.commaPositions.size + 1;
+        for (size_t i = 0; i < numAttrs; i++)
         {
-            serializeAttribute(attrs.data[i]);
-            lua_rawseti(L, -2, i + 1);
+            serializeAttribute(attrs.data[attrIndex]);
+            attrIndex++;
+            lua_rawseti(L, -3, static_cast<int>(i) + 1);
+
+            if (i < attrList.commaPositions.size)
+            {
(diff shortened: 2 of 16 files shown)


lute/syntax/src/parser.cpp
```diff
@@ -720,16 +804,41 @@ struct AstSerialize : public Luau::AstVisitor
...
 
-        lua_pushstring(L, "attribute");
-        lua_setfield(L, -2, "kind");
+            serializeNodePreamble(node, "parametrized", "attribute");
 
-        withLocation(node->location);
+            serializeToken(node->location.begin, (std::string(node->name.value)).c_str());
+            lua_setfield(L, -2, "name");
+
+            serializeToken(cstParametrized->openParenPosition, "(");
```
ActualMasterOogway, line 820: Parentheses aren't guaranteed to exist since `{ ... }` call sugar is valid attribute syntax. Upstream only sets paren positions for the `(` form (https://github.com/luau-lang/luau/blob/86d2a9dcd7cef396b73b1585371723e169e69a41/Ast/include/Luau/Cst.h#L71). For the table form, they are `Position::missing()`

```luau
-- parser.cpp(226): ASSERTION FAILED: newPos.line < lineOffsets.size()
require("@lute/syntax/parser").parse('@[deprecated { reason = "x" }]\nfunction foo() end')
```

Guarding both tokens with `hasValue()` (similar to upstream's `maybeAdvanceAndWrite` logic) fixes this

```cpp
if (cstParametrized->openParenPosition.hasValue())
{
    serializeToken(cstParametrized->openParenPosition, "(");
    lua_setfield(L, -2, "openParens");
}
```

And update the type definitions to make these fields optional

```luau
read openParens: CstToken<"(">?,
read closeParens: CstToken<")">?,
```

lute/syntax/src/parser.cpp
```diff
@@ -720,16 +804,41 @@ struct AstSerialize : public Luau::AstVisitor
...
+            lua_setfield(L, -2, "openParens");
+
+            serializePunctuated(node->args, cstParametrized->argsCommaPositions, ",");
+            lua_setfield(L, -2, "arguments");
+
+            serializeToken(cstParametrized->closeParenPosition, ")");
```
ActualMasterOogway, line 826: same guard needed here

tests/parserExamples/attributes-2.luau
ActualMasterOogway, line 1: Test table call sugaring, see comment https://github.com/luau-lang/lute/pull/1210/changes#r3525591380
```luau
@[deprecated { reason = "use foo instead" }]
function quux() end
```

lute/syntax/src/parser.cpp
```diff
@@ -618,15 +618,99 @@ struct AstSerialize : public Luau::AstVisitor
...
 
-    void serializeAttributes(Luau::AstArray<Luau::AstAttr*>& attrs, size_t nrec = 0)
+    void serializeAttributeList(const Luau::AstArray<Luau::AstAttr*>& attrs, const Luau::CstAttrList& attrList, size_t& attrIndex)
     {
+        LUTE_ASSERT(attrIndex < attrs.size);
+
+        lua_rawcheckstack(L, 4);
+        lua_createtable(L, 0, preambleSize + 3); // atBracket, attributes, closeBracket
+
+        withLocation(Luau::Location{attrList.atBracketPosition, attrList.closeBracketPosition});
```
ActualMasterOogway, line 628: Because `closeBracketPosition` points to the start of `]`, the span ends one character short of the node's actual closing bracket (other nodes end one past their last character). Slicing the source by the list's span drops the `]`

```luau
local parser = require("@lute/syntax/parser")
local res = parser.parse("@[native, checked]\nfunction f() end")
local list = res.root.statements[1].func.attributes[1]
print(list.location.endColumn, list.closeBracket.location.endColumn) --> 18 19
```

Bumping the end column by 1 fixes this

```cpp
withLocation(Luau::Location{
    attrList.atBracketPosition,
    Luau::Position{attrList.closeBracketPosition.line, attrList.closeBracketPosition.column + 1}
});
```

lute/syntax/src/parser.cpp
```diff
@@ -720,16 +804,41 @@ struct AstSerialize : public Luau::AstVisitor
...
+        {
+            const Luau::CstAttr* cstSimple = (*cstNode)->as<Luau::CstAttr>();
+            LUTE_ASSERT(cstSimple);
+
+            lua_rawcheckstack(L, 2);
+            lua_createtable(L, 0, preambleSize + 1); // location, kind, tag, name
+
+            serializeNodePreamble(node, "simple", "attribute");
+
+            serializeToken(node->location.begin, (cstSimple->hasAt ? "@" + std::string(node->name.value) : std::string(node->name.value)).c_str());
```
ActualMasterOogway, line 839: Two different name spellings for the same node shape depending on where the attribute is written is in my opinion a bad design

```luau
local res = require("@lute/syntax/parser").parse("@native function a() end\n@[native]\nfunction b() end")
print(res.root.statements[1].func.attributes[1].name.text) --> @native
print(res.root.statements[2].func.attributes[1].attributes[1].name.text) --> native
```

Since both use `tag = "simple"`, matching an attribute currently requires checking for both `"@native"` and `"native"` (like the tests in this PR do).

Serializing the name without the `@` in both cases and exposing the "at-sign" separately (like a `read atSign: boolean` field or an `at` token) would make name matching much easier

lute/syntax/src/parser.cpp
```diff
@@ -618,15 +618,99 @@ struct AstSerialize : public Luau::AstVisitor
...
+            {
+                serializeToken(attrList.commaPositions.data[i], ",");
+                lua_rawseti(L, -2, static_cast<int>(i) + 1);
+            }
+        }
+
+        lua_setfield(L, -2, "separators");
+
+        luaL_getmetatable(L, kCstPunctuatedType);
+        lua_setmetatable(L, -2);
```
ActualMasterOogway, line 665: Can `serializePunctuated` not be reused here with a `visit(AstAttr*)` override?

lute/syntax/src/parser.cpp
```diff
@@ -618,15 +618,99 @@ struct AstSerialize : public Luau::AstVisitor
...
+        lua_setfield(L, -2, "closeBracket");
+    }
+
+    void serializeAttributes(Luau::AstArray<Luau::AstAttr*>& attrs, const Luau::AstArray<Luau::CstAttrList*>& attrLists)
+    {
+        LUTE_ASSERT(attrs.size >= attrLists.size);
+
         lua_rawcheckstack(L, 2);
-        lua_createtable(L, attrs.size, nrec);
+        lua_createtable(L, attrLists.size, 0);
```
ActualMasterOogway, line 679: attrLists.size is 0 for plain `@checked @native`, old hint was attrs.size. Intentional?

definitions/syntax/cst.luau
```diff
@@ -524,13 +524,38 @@ export type CstStatCompoundAssign = {
...
 
+-- A (possibly bracketed) list of comma-separated attributes
+export type CstAttributeList = {
+	read location: Span,
+	read kind: "attribute",
+	read tag: "list",
+	read atBracket: CstToken<"@[">,
+	read attributes: CstPunctuated<CstAttribute, ",">,
+	read closeBracket: CstToken<"]">,
+}
```
ActualMasterOogway, line 554: Because the list shares `kind = "attribute"` with its members `utils.isAttribute` matches the container too. This results in 3 matches for 2 attributes, and the container lacks a `name` field (which all other attribute nodes have). The current typing also theoretically allows nested lists

```luau
local root = parser.parseBlock("@[native, checked]\nfunction f() end")
local list = root.statements[1].func.attributes[1]

print(utils.isAttribute(list) ~= nil)               --> true (container)
print(utils.isAttribute(list.attributes[1]) ~= nil) --> true
print(list.name)                                    --> nil
```

Giving the list its own kind and narrowing the element types fixes both issues

```luau
export type CstAttributeList = {
    read location: Span,
    read kind: "attributelist",
    read atBracket: CstToken<"@[">,
    read attributes: CstPunctuated<CstAttributeSimple | CstAttributeParametrized, ",">,
    read closeBracket: CstToken<"]">,
}
```

ActualMasterOogway (commented):
Queries now return every attribute node twice because the new dispatch fires both `visitAttribute` and the tag specific callback on the same node. Currently, `newSelectVisitor` in `query.luau` only suppresses the generic `visitExpr`:

```luau
local parser = require("@std/syntax/parser")

local root = parser.parseBlock("@[native]\nfunction foo() end")
local lists = query.findAllFromRoot(root, utils.isAttributeList):mapToArray(function(n) return n end)
print(#lists, lists[1] == lists[2]) --> 2 true
```

Printer snapshots miss this since `Replacements` is keyed by node, but `:replace(fn)` still runs twice, and `forEach`/`mapToArray` yield duplciates.

Suppressing `visitAttribute` in `newSelectVisitor` just like done for `visitExpr` would fix this

```luau
selectVisitor.visitAttribute = function(_)
    return true
end
```

skberkeley:
> Queries now return every attribute node twice because the new dispatch fires both `visitAttribute` and the tag specific callback on the same node. Currently, `newSelectVisitor` in `query.luau` only suppresses the generic `visitExpr`:

This actually raises a separate point I missed, `visitor` should have `visitStat`, `visitType`, and `visitTypePack` callbacks too

PR #1213, opened 2026-07-03
Add option to disable the use of vendored libraries

lukadev-0:
Adds a `LUTE_NO_VENDOR` build option that will use the system libraries instead of statically linking to vendored libraries in `extern`, except for uSockets and uWebSockets which will still be vendored and built with Lute.

This is useful for packaging Lute for Linux package repositories, as they generally prefer to avoid vendoring dependencies. (e.g. https://github.com/NixOS/nixpkgs/pull/511134#discussion_r3198005901)

When the option is enabled, OpenSSL will be used instead of BoringSSL, however this required some changes to Lute's crypto library because OpenSSL does not have `EVP_blake2b256`.

### DIFF
--- a/CMakeBuild/uSockets.cmake
+++ b/CMakeBuild/uSockets.cmake
@@ -7,13 +7,6 @@ set(CMAKE_C_STANDARD_REQUIRED ON)
 
 set(USOCKETS_SOURCE_DIR ${CMAKE_CURRENT_SOURCE_DIR}/extern/uSockets)
 
-# Include directories
-include_directories(${USOCKETS_SOURCE_DIR}/src)
-include_directories(${USOCKETS_SOURCE_DIR}/src/eventing)
-include_directories(${USOCKETS_SOURCE_DIR}/src/crypto)
-include_directories(${USOCKETS_SOURCE_DIR}/src/io_uring)
-include_directories(${LIBUV_INCLUDE_DIR})
-
 # Source files
 file(GLOB USOCKETS_SOURCES
     ${USOCKETS_SOURCE_DIR}/src/*.c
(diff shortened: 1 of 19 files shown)


aatxe:
Is it actually necessary to cease using BoringSSL to remove the reliance on vendoring? It doesn't feel like it should be necessary at all, and the nixpkgs link seems to include mention of a BoringSSL nix package. The use of BoringSSL over OpenSSL is an intentional choice, and I don't feel good about making it trivial to package and ship Lute with OpenSSL as the default.

lukadev-0:
BoringSSL doesn't have a stable API and ABI and is therefore not a good fit for dynamic linking, as mentioned in their [documentation](https://github.com/google/boringssl/blob/main/PORTING.md):
> Note: BoringSSL does not have a stable API or ABI. It must be updated with its consumers. It is not suitable for, say, a system library in a traditional Linux distribution. For instance, Chromium statically links the specific revision of BoringSSL it was built against. Likewise, Android's system-internal copy of BoringSSL is not exposed by the NDK and must not be used by third-party applications.

If adding support for OpenSSL is unacceptable, I think it is fine if we continue to vendor BoringSSL.

aatxe:
I'm aware it doesn't make stability guarantees, but I thought I chased through some conversations from the Nix PR and saw mention of people using BoringSSL as a pinned package on the Nix side. I failed to find it again though, so oh well.

I would rather vendor BoringSSL than set things up in a way where any Linux distributions picking Lute up are liable to end up using OpenSSL over BoringSSL. It also feels like it makes the option itself much more non-obvious, since I don't imagine most people would expect `LUTE_NO_VENDOR` to suddenly change the security profile of the software. If there's strong interest for using OpenSSL itself, I'd expect that to at least be its own configuration option that makes explicit the choice you're making.

lukadev-0:
I have removed the OpenSSL support. It will now always vendor BoringSSL.

PR #1217, opened 2026-07-06
Add breakpoints actually pausing execution

wizhao96:
This PR makes breakpoints actually functional but in order to do we need to add many features under the hood
* as mentioned here https://github.com/luau-lang/lute/pull/1202#pullrequestreview-4603646302, we need to run the child debuggee process in a separate VM.
* as a side effect of the above change, we need to instead schedule operations (such as adding and removing breakpoints) onto the child VM, which effectively makes them async
* we add some callbacks when breakpoints are installed or un-installed, and when program execution terminates, and when a breakpoint is hit
* we add a `continue()` to resume program execution

Caveats:
* most of this code relies on the debugged script (`exit` and `continue`) being single coroutine. Multi-coroutines execution will be handled in future PRs.

### DIFF
--- a/lute/debug/src/debugger.cpp
+++ b/lute/debug/src/debugger.cpp
@@ -214,7 +214,11 @@ void Target::installPendingBreakpoints(lua_State* L)
         {
             bool installed = installBreakpoint(L, bp);
             if (installed && launchConfig.onBreakpointInstall)
+            {
+                lock.unlock();
                 launchConfig.onBreakpointInstall(bp);
+                lock.lock();
+            }
         }
     }
 }
@@ -222,11 +226,10 @@ void Target::installPendingBreakpoints(lua_State* L)
 std::shared_ptr<Process> Target::launch(const std::vector<std::string>& args, LaunchConfig config)
 {
     // launch() cannot be called twice from the same target.
-    if (childRuntime != nullptr)
+    if (activeProcess != nullptr)
         return nullptr;
     childRuntime = std::make_shared<Runtime>(parentRuntime.reporter);
     setupState(*childRuntime, nullptr);
-
     launchConfig = config;
 
     std::ifstream file(sourcePath);
@@ -278,35 +281,36 @@ void Process::installBpHitCallback()
     cb->userdata = this;
     cb->debugbreak = [](lua_State* L, lua_Debug* ar)
     {
-        int line = ar->currentline;
-        lua_Debug info = {};
-        lua_getinfo(L, 0, "s", &info);
-
-        // TODO: this pause/resume mechanism assumes single co-routine runtime
         Process* process = static_cast<Process*>(lua_callbacks(L)->userdata);
+        // TODO: this pause/resume mechanism assumes single co-routine runtime
         process->resumeToken = getResumeToken(L);
+
+        lua_Debug info = {};
+        lua_getinfo(L, 0, "sl", &info);
+        int line = info.currentline;
+        if (!info.source)
+            process->runtime.reporter.reportError(Luau::format("breakpoint hit at line %d could not be find a runtime source", line));
         std::string source = info.source;
         std::optional<Breakpoint> bp = process->parentTarget.getBreakpointBySourceLine(source, line);
-        if (bp)
+        if (bp && bp->status == BreakpointStatus::Installed)
         {
-            if (bp->status == BreakpointStatus::Installed)
+            if (process->config.onBreakpointHit)
             {
-                // notify caller that we stopped
-                if (process->config.onBreakpointHit)
-                {
-                    process->config.onBreakpointHit(*process, bp.value());
-                }
-            }
-            else if (bp->status == BreakpointStatus::PendingUninstall)
-            {
-                process->continueProcess();
+                process->config.onBreakpointHit(*process, bp.value());
             }
         }
         else
         {
-            process->runtime.reporter.reportError(
-                Luau::format("breakpoint hit at line %d in %s could not be found in breakpoint map", line, source.c_str())
-            );
+            // it is normal to hit breakpoints that are pending uninstall but not normal
+            // to hit any other type of breakpoint
+            if (!bp || bp->status != BreakpointStatus::PendingUninstall)
+            {
+                process->runtime.reporter.reportError(
+                    Luau::format("breakpoint hit at line %d in %s could not be found in breakpoint map", line, source.c_str())
+                );
+            }
+            // this prevents us from hanging forever
+            process->continueProcess();
         }
     };
 }
--- a/lute/debug/src/debugger.cpp
+++ b/lute/debug/src/debugger.cpp
@@ -298,9 +298,9 @@ void Process::installBpHitCallback()
         Process* process = static_cast<Process*>(lua_callbacks(L)->userdata);
         // TODO: this pause/resume mechanism assumes single co-routine runtime
         // We land on the same instruction after a continue() after hitting a bp so we basically don't do anything
-        if (process->continueRequested)
+        if (process->continueRequestedBp)
         {
-            process->continueRequested = false;
+            process->continueRequestedBp = false;
             return;
         }
         lua_Debug info = {};
@@ -316,6 +316,7 @@ void Process::installBpHitCallback()
         // Only stop execution on installed breakpoints; otherwise, don't stop.
         if (bp && bp->status == BreakpointStatus::Installed)
         {
+            process->bpHit = *bp;
             process->resumeToken = getResumeToken(L);
             lua_break(L);
             if (process->config.onBreakpointHit)
@@ -348,7 +349,16 @@ bool Process::continueProcess()
 {
     if (!resumeToken)
         return false;
-    continueRequested = true;
+    // we are continuing on a breakpoint and so might need to flag continueRequestedBp.
+    if (bpHit)
+    {
+        // we need to check if our breakpoint is still currently installed after
+        // onBreakpointHit() callback
+        std::optional<Breakpoint> currentBp = parentTarget.getBreakpointById(bpHit->id);
+        if (currentBp && currentBp->status == BreakpointStatus::Installed)
+            continueRequestedBp = true;
+        bpHit = std::nullopt;
+    }
     resumeToken->complete(
         [](lua_State* L)
         {
--- a/lute/debug/src/debugger.cpp
+++ b/lute/debug/src/debugger.cpp
@@ -23,9 +23,10 @@ Breakpoint::Breakpoint(int id, std::string sourcePath, int line, BreakpointStatu
 {
 }
 
-Target::Target(Runtime& runtime, std::string sourcePath)
-    : runtime(runtime)
+Target::Target(Runtime& parentRuntime, std::string sourcePath, std::function<void(const Breakpoint& bp)> onBreakpointInstall)
+    : parentRuntime(parentRuntime)
     , sourcePath(sourcePath)
+    , onBreakpointInstall(onBreakpointInstall)
     , loadedSources("")
 {
 }
@@ -35,7 +36,20 @@ Breakpoint Target::addBreakpoint(std::string sourcePath, int line)
     int id = currentBreakpointId;
     currentBreakpointId++;
     auto [it, _] = breakpoints.insert_or_assign(id, Breakpoint(id, sourcePath, line, BreakpointStatus::Pending));
-    installBreakpoint(runtime.GL, it->second);
+    if (childRuntime)
+        childRuntime->schedule(
+            [this, id]() mutable
+            {
+                auto it = breakpoints.find(id);
+                // check breakpoint has not been erased when this scheduled install has been fired
+                if (it != breakpoints.end())
+                {
+                    bool installed = installBreakpoint(childRuntime->GL, it->second);
+                    if (installed && onBreakpointInstall)
+                        onBreakpointInstall(it->second);
+                }
+            }
+        );
     return it->second;
 }
 
@@ -48,19 +62,25 @@ bool Target::removeBreakpoint(int bpId)
     if (bp.status == BreakpointStatus::Installed)
     {
         auto chunkRef = loadedSources.find(bp.sourcePath);
-        if (chunkRef)
+        if (chunkRef && childRuntime)
         {
-            (*chunkRef)->push(runtime.GL);
-            int removed_line = lua_breakpoint(runtime.GL, -1, bp.line, 0);
-            lua_pop(runtime.GL, 1);
+            (*chunkRef)->push(childRuntime->GL);
+            int removed_line = lua_breakpoint(childRuntime->GL, -1, bp.line, 0);
+            lua_pop(childRuntime->GL, 1);
             if (removed_line == -1)
-                runtime.reporter.reportError(
+                parentRuntime.reporter.reportError(
                     Luau::format("breakpoint %d installed at line %d in %s could not be removed", bp.id, bp.line, bp.sourcePath.c_str())
                 );
         }
+        else if (!childRuntime)
+        {
+            parentRuntime.reporter.reportError(
+                Luau::format("breakpoint %d installed at line %d in %s is missing a runtime", bp.id, bp.line, bp.sourcePath.c_str())
+            );
+        }
         else
         {
-            runtime.reporter.reportError(
+            parentRuntime.reporter.reportError(
                 Luau::format("breakpoint %d installed at line %d in %s is missing a loaded source", bp.id, bp.line, bp.sourcePath.c_str())
             );
         }
@@ -120,12 +140,19 @@ void Target::installPendingBreakpoints(lua_State* L)
     for (auto& [_, bp] : breakpoints)
     {
         if (bp.status == BreakpointStatus::Pending)
-            installBreakpoint(L, bp);
+        {
+            bool installed = installBreakpoint(L, bp);
+            if (installed && onBreakpointInstall)
+                onBreakpointInstall(bp);
+        }
     }
 }
 
 bool Target::launch(const std::vector<std::string>& args)
 {
+    childRuntime = std::make_shared<Runtime>(parentRuntime.reporter);
+    setupState(*childRuntime, nullptr);
+
     std::ifstream file(sourcePath);
     if (!file.is_open())
         return false;
@@ -134,15 +161,19 @@ bool Target::launch(const std::vector<std::string>& args)
     debugOptions.optimizationLevel = 1;
     debugOptions.debugLevel = 2;
     std::string bytecode = Luau::compile(source, debugOptions);
-    lua_State* thread = lua_newthread(runtime.GL);
+    lua_State* thread = lua_newthread(childRuntime->GL);
     luaL_sandboxthread(thread);
     luau_load(thread, sourcePath.c_str(), bytecode.c_str(), bytecode.size(), 0);
     loadedSources[sourcePath] = std::make_shared<Ref>(thread, -1);
     installPendingBreakpoints(thread);
     for (const std::string& arg : args)
         lua_pushstring(thread, arg.c_str());
-    runtime.runningThreads.push_back({true, getRefForThread(thread), static_cast<int>(args.size())});
-    lua_pop(runtime.GL, 1);
+    childRuntime->runningThreads.push_back({true, getRefForThread(thread), static_cast<int>(args.size())});
+    lua_pop(childRuntime->GL, 1);
+
+    // the runtime provides the guarantees that C++ continuations (such as already enqueued breakpoints to install)
+    // happen before the first Luau thread runs
+    childRuntime->runContinuously();
     return true;
 }
 } // namespace debug
--- a/lute/debug/src/debugger.cpp
+++ b/lute/debug/src/debugger.cpp
@@ -280,7 +280,6 @@ Target& Process::getTarget()
 {
     return parentTarget;
 }
-
 void Process::installBpHitCallback()
 {
     lua_Callbacks* cb = lua_callbacks(runtime.GL);
@@ -289,33 +288,37 @@ void Process::installBpHitCallback()
     {
         Process* process = static_cast<Process*>(lua_callbacks(L)->userdata);
         // TODO: this pause/resume mechanism assumes single co-routine runtime
-        process->resumeToken = getResumeToken(L);
-
+        // We land on the same instruction after a continue() after hitting a bp so we basically don't do anything
+        if (process->continueRequested)
+        {
+            process->continueRequested = false;
+            return;
+        }
         lua_Debug info = {};
         lua_getinfo(L, 0, "sl", &info);
         int line = info.currentline;
         if (!info.source)
         {
-            process->runtime.reporter.reportError(Luau::format("breakpoint hit at line %d could not be find a runtime source", line));
+            process->runtime.reporter.reportError(Luau::format("breakpoint hit at line %d could not find a runtime source", line));
             return;
         }
         std::string source = info.source;
         std::optional<Breakpoint> bp = process->parentTarget.getBreakpointBySourceLine(source, line);
+        // Only stop execution on installed breakpoints; otherwise, don't stop.
         if (bp && bp->status == BreakpointStatus::Installed)
         {
+            process->resumeToken = getResumeToken(L);
+            lua_break(L);
             if (process->config.onBreakpointHit)
                 process->config.onBreakpointHit(*process, bp.value());
         }
-        else
+        else if (!bp || bp->status != BreakpointStatus::PendingUninstall)
         {
-            // it is normal to hit breakpoints that are pending uninstall but not normal
-            // to hit any other type of breakpoint
-            if (!bp || bp->status != BreakpointStatus::PendingUninstall)
-                process->runtime.reporter.reportError(
-                    Luau::format("breakpoint hit at line %d in %s could not be found in breakpoint map", line, source.c_str())
-                );
-            // this prevents us from hanging forever
-            process->continueProcess();
+            // It is normal to hit breakpoints that are pending uninstall but not normal
+            // to hit any other type of breakpoint so we error in those cases.
+            process->runtime.reporter.reportError(
+                Luau::format("breakpoint hit at line %d in %s could not be found in breakpoint map", line, source.c_str())
+            );
         }
     };
 }
@@ -336,6 +339,7 @@ bool Process::continueProcess()
 {
     if (!resumeToken)
         return false;
+    continueRequested = true;
     resumeToken->complete(
         [](lua_State* L)
         {
--- a/lute/debug/src/debugger.cpp
+++ b/lute/debug/src/debugger.cpp
@@ -1,5 +1,7 @@
 #include "debugger.h"
 
+#include "lute/common.h"
+
 #include "Luau/Compiler.h"
 #include "Luau/DenseHash.h"
 #include "Luau/StringUtils.h"
@@ -7,10 +9,12 @@
 #include "lua.h"
 #include "lualib.h"
 
+#include <cstddef>
 #include <fstream>
 #include <mutex>
 #include <optional>
 #include <string>
+#include <tuple>
 #include <unordered_map>
 #include <vector>
 
@@ -117,7 +121,7 @@ std::vector<Breakpoint> Target::getBreakpoints() const
     std::vector<Breakpoint> all;
     all.reserve(breakpoints.size());
     for (auto& [_, bp] : breakpoints)
-        all.push_back(bp);
+        all.emplace_back(bp);
     return all;
 }
 
@@ -128,7 +132,7 @@ std::vector<Breakpoint> Target::getBreakpointsByStatus(BreakpointStatus status) 
     statusBps.reserve(breakpoints.size());
     for (auto& [_, bp] : breakpoints)
         if (bp.status == status)
-            statusBps.push_back(bp);
+            statusBps.emplace_back(bp);
     return statusBps;
 }
 
@@ -218,28 +222,27 @@ bool Target::uninstallBreakpoint(lua_State* L, Breakpoint& bp)
     return true;
 }
 
-// This returns the list of installed and unisntalled vectors for use in callbacks
+// This returns the list of installed and uninstalled vectors for use in callbacks
 // later.
 std::pair<std::vector<Breakpoint>, std::vector<Breakpoint>> Target::modifyPendingBreakpoints(lua_State* L)
 {
-    std::vector<Breakpoint> installedBpsCallback, uninstalledBpsCallback;
+    std::vector<Breakpoint> installedBpsCallback;
+    std::vector<Breakpoint> uninstalledBpsCallback;
     std::vector<int> toErase;
     for (auto& [id, bp] : breakpoints)
     {
         if (bp.status == BreakpointStatus::PendingInstall)
         {
-            bool installed = installBreakpoint(L, bp);
-            if (installed && launchConfig.onBreakpointInstall)
-                installedBpsCallback.push_back(bp);
+            if (installBreakpoint(L, bp) && launchConfig.onBreakpointInstall)
+                installedBpsCallback.emplace_back(bp);
         }
         if (bp.status == BreakpointStatus::PendingUninstall)
         {
-            bool uninstalled = uninstallBreakpoint(L, bp);
-            if (uninstalled)
+            if (uninstallBreakpoint(L, bp))
             {
                 if (launchConfig.onBreakpointUninstall)
-                    uninstalledBpsCallback.push_back(bp);
-                toErase.push_back(bp.id);
+                    uninstalledBpsCallback.emplace_back(bp);
+                toErase.emplace_back(bp.id);
             }
         }
     }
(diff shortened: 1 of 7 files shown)


lute/debug/src/debugger.cpp
```diff
@@ -23,54 +24,130 @@ Breakpoint::Breakpoint(int id, std::string sourcePath, int line, BreakpointStatu
...
-    : runtime(runtime)
+Target::Target(Runtime& parentRuntime, std::string sourcePath)
+    : parentRuntime(parentRuntime)
     , sourcePath(sourcePath)
     , loadedSources("")
 {
 }
 
-Breakpoint Target::addBreakpoint(std::string sourcePath, int line)
+Breakpoint Target::setBreakpoint(std::string sourcePath, int line)
```
wizhao96, line 45: this is `setBreakpoint` now to try to better reflect the added behavior that we can't add duplicate breakpoints at the same line (instead we return the preexisting bp at that line)

lute/debug/src/debugger.h
```diff
@@ -4,18 +4,22 @@
...
 struct lua_State;
+struct lua_Debug;
 
 namespace debug
 {
 enum class BreakpointStatus
 {
-    Pending,
+    PendingInstall,
+    PendingUninstall,
```
wizhao96, line 22: Because of the way we trigger removals asynchronously we need a new `PendingUninstall` state. Otherwise, if we remove the bp from the bp map immediately, we can still hit it because the actual uninstall is scheduled asynchronously, resulting in an error.

lute/debug/src/debugger.cpp
```diff
@@ -23,54 +24,129 @@ Breakpoint::Breakpoint(int id, std::string sourcePath, int line, BreakpointStatu
 {
 }
 
-Target::Target(Runtime& runtime, std::string sourcePath)
-    : runtime(runtime)
-    , sourcePath(sourcePath)
+Target::Target(Runtime& parentRuntime)
```
wizhao96, line 31: in reality we only need to pass in a sourcePath on `target::launch()` to match with what DAP expects

lute/debug/src/debugger.cpp
```diff
@@ -117,32 +206,135 @@ bool Target::installBreakpoint(lua_State* L, Breakpoint& bp)
...
+        {
+            process->runtime.reporter.reportError(Luau::format("breakpoint hit at line %d could not be find a runtime source", line));
+            return;
+        }
+        std::string source = info.source;
+        std::optional<Breakpoint> bp = process->parentTarget.getBreakpointBySourceLine(source, line);
+        if (bp && bp->status == BreakpointStatus::Installed)
+        {
+            if (process->config.onBreakpointHit)
+                process->config.onBreakpointHit(*process, bp.value());
```
afujiwara-roblox: do we need to run `process->continueProcess();` in an `else` block here?

wizhao96: i don't think so. i want to make sure that even if no callback is set, you still block (since there still is a breakpoint placed there). although i understand there's some concern that you hitting the breakpoint can not actually be communicated to a debugger script. ig you can imagine somebody polling whether the process is paused before making some sort of action.

tests/src/debug.test.cpp
```diff
@@ -61,59 +82,130 @@ TEST_SUITE("Debug")
...
 
         // check cannot remove never added breakpoint
         bool cannotRemove = target.removeBreakpoint(100);
         CHECK(!cannotRemove);
+
+        // wait until script is finished to call destructors
+        REQUIRE(exitFuture.wait_for(std::chrono::seconds(5)) == std::future_status::ready);
+    }
+
+    TEST_CASE_FIXTURE(DebugFixture, "Debug_hitBreakpoint")
```
raymondnumbergenerator, line 155: can we add a test that actually verifies that execution is paused at a breakpoint and verify the state before and after the breakpoint

lute/debug/src/debugger.cpp
```diff
@@ -117,32 +214,135 @@ bool Target::installBreakpoint(lua_State* L, Breakpoint& bp)
...
+        Process* process = static_cast<Process*>(lua_callbacks(L)->userdata);
+        // TODO: this pause/resume mechanism assumes single co-routine runtime
+        process->resumeToken = getResumeToken(L);
+
+        lua_Debug info = {};
+        lua_getinfo(L, 0, "sl", &info);
+        int line = info.currentline;
+        if (!info.source)
+        {
+            process->runtime.reporter.reportError(Luau::format("breakpoint hit at line %d could not be find a runtime source", line));
```
raymondnumbergenerator: typo?

lute/debug/src/debugger.cpp
raymondnumbergenerator: shouldn't there be a lua_break here to pause execution when an installed breakpoint is hit

raymondnumbergenerator: i think the continueProcess would be unnecessary too, we simply just don't call lua_break if the breakpoint isn't installed
in that case `process->resumeToken = getResumeToken(L);` should also just be moved into the installed condition so we don't need to grant a resume token when a break doesn't happen

lute/debug/src/debugger.cpp
wizhao96: if this is in reference to line 297, primarily this is to check if info.source is not null (since it's a info.source is actually a c-string when defined in https://github.com/luau-lang/luau/blob/ddcea05e1cc6f534e5eaac33325690c12f1ed274/VM/include/lua.h#L492)

wizhao96: ok nvm this is prob in reference to the fact i can't spell :( so I resolved it

lute/debug/src/debugger.cpp
wizhao96: ok yea both of these have been fixed.

tests/src/debug.test.cpp
wizhao96: it's kind of hard to check process state reliably right now without any primitives to do so, so basically the added test right now checks
1. we place a bp at lines 1 and 2
2. we hit bp 1 and make sure we don't hit bp 2 after 100 ms. then, we continue
3. we hit bp 2 and make sure we don't exit the program after 100 ms. then, we continue

this provides a relatively correct understanding that we're at the right lines within the debuggee and that we are actually pausing the program. 

i tried to fetch the program lines directly but there's some finicky/unintuitive behavior with how the program counter is reset to reexecute the current instruction after a breakpoint

raymondnumbergenerator: that's alright, this looks good
we can revisit it with a better test later, just felt like there was a small gap where we were only really testing that the breakpoint was being hit and not testing the actual breaking

lute/debug/src/debugger.h
```diff
@@ -29,36 +33,74 @@ struct Breakpoint
...
+    std::optional<Breakpoint> getBreakpointBySourceLine(std::string source, int line) const;
 
-    bool launch(const std::vector<std::string>& args);
+    std::shared_ptr<Process> launch(std::string sourcePath, const std::vector<std::string>& args, LaunchConfig config = {});
 
 private:
-    Runtime& runtime;
-    std::string sourcePath;
+    Runtime& parentRuntime;
+    std::shared_ptr<Runtime> childRuntime;
```
Vighnesh-V: If this class is the only thing that owns this Runtime, can this be made unique_pointer?

lute/debug/src/debugger.cpp
```diff
@@ -23,54 +24,137 @@ Breakpoint::Breakpoint(int id, std::string sourcePath, int line, BreakpointStatu
...
+            {
+                std::unique_lock lock(breakpointsMutex);
+                auto it = breakpoints.find(id);
+                // check breakpoint has not been erased when this scheduled install has been fired
+                if (it != breakpoints.end())
+                {
+                    bool installed = installBreakpoint(childRuntime->GL, it->second);
+                    lock.unlock();
+                    if (installed && launchConfig.onBreakpointInstall)
+                        launchConfig.onBreakpointInstall(it->second);
```
Vighnesh-V: Not a 100% sure, but this accesses the iterator after the lock is released. The iterator was acquired during the lock, so is this safe to do?

lute/runtime/src/runtime.cpp
```diff
@@ -126,7 +126,7 @@ RuntimeStep Runtime::runOnce()
     else
         status = lua_resume(L, nullptr, next.argumentCount);
 
-    if (status == LUA_YIELD)
+    if (status == LUA_YIELD || status == LUA_BREAK)
```
Vighnesh-V, line 129: I think this is okay for now, but I would like to think about the constructor for the Runtime having a 'debug' parameter.

lute/debug/src/debugger.cpp
```diff
@@ -117,32 +214,139 @@ bool Target::installBreakpoint(lua_State* L, Breakpoint& bp)
...
+    childRuntime->schedule([]() {});
+    childRuntime->runContinuously();
+    return process;
+}
+
+Process::Process(lua_State* thread, Target& parentTarget, LaunchConfig config)
+    : thread(thread)
+    , runtime(*getRuntime(thread))
+    , parentTarget(parentTarget)
+    , config(std::move(config))
```
Vighnesh-V: Are you able to initialize the resumeToken here instead of by setting it on the Process struct at a delayed time? Or is thread != the thread you install the Bp callback on?

lute/debug/src/debugger.cpp
```diff
@@ -23,54 +24,137 @@ Breakpoint::Breakpoint(int id, std::string sourcePath, int line, BreakpointStatu
...
+    : parentRuntime(parentRuntime)
     , loadedSources("")
 {
 }
 
-Breakpoint Target::addBreakpoint(std::string sourcePath, int line)
+Target::~Target()
+{
+    // We want to stop the runtime so nothing runs while we are destroying the target but first
+    // we need to clear all sources (which are stored as refs in the runtme).
```
Vighnesh-V:
```suggestion
    // we need to clear all sources (which are stored as refs in the runtime).
```

lute/debug/src/debugger.h
```diff
@@ -29,36 +33,74 @@ struct Breakpoint
...
     bool installBreakpoint(lua_State* L, Breakpoint& bp);
     void installPendingBreakpoints(lua_State* L);
 };
 
+struct Process
+{
+    explicit Process(lua_State* thread, Target& parentTarget, LaunchConfig config);
+    Target& getTarget();
+    bool continueProcess();
+    int getCurrentLine() const;
```
Vighnesh-V: i dont see this implemented anywhere ...

lute/debug/src/debugger.h
```diff
@@ -29,36 +33,74 @@ struct Breakpoint
...
+    Breakpoint setBreakpoint(std::string sourcePath, int line);
     bool removeBreakpoint(int bpId);
 
     std::vector<Breakpoint> getBreakpoints() const;
     std::vector<Breakpoint> getBreakpointsByStatus(BreakpointStatus status) const;
     std::optional<Breakpoint> getBreakpointById(int breakpointId) const;
+    std::optional<Breakpoint> getBreakpointBySourceLine(std::string source, int line) const;
 
-    bool launch(const std::vector<std::string>& args);
+    std::shared_ptr<Process> launch(std::string sourcePath, const std::vector<std::string>& args, LaunchConfig config = {});
```
Vighnesh-V: is there a reason you picked shared_ptr here?

lute/debug/src/debugger.h
```diff
@@ -29,36 +33,74 @@ struct Breakpoint
...
 
 private:
-    Runtime& runtime;
-    std::string sourcePath;
+    Runtime& parentRuntime;
+    std::shared_ptr<Runtime> childRuntime;
 
     int currentBreakpointId = 0;
     std::unordered_map<int, Breakpoint> breakpoints; // breakpoint id -> breakpoint object (this is unordered_map to support erase)
+    mutable std::mutex breakpointsMutex;
```
Vighnesh-V: Code Organization: Can you group breakPointsMutex next to all the fields that need to be thread synchronized and leave a comment? That way, I can easily tell where you're missing a lock in review.

tests/src/debug.test.cpp
```diff
@@ -1,48 +1,70 @@
...
 
         CHECK(bp3.id == 2);
         CHECK(bp3.sourcePath == fixturePath);
         CHECK(bp3.line == 3);
-        CHECK(bp3.status == debug::BreakpointStatus::Pending);
+        CHECK(bp3.status == debug::BreakpointStatus::PendingInstall);
 
         std::optional<debug::Breakpoint> found = target.getBreakpointById(0);
         REQUIRE(found.has_value());
         CHECK(found->id == 0);
```
Vighnesh-V: Can you write a single macro for testing this? 
CHECK_BREAKPOINT(target, id) ....

tests/src/debug.test.cpp
```diff
@@ -61,59 +83,175 @@ TEST_SUITE("Debug")
...
+        CHECK(installedBp4.id == 3);
+
+        // check that setting breakpoints at same breakpoint returns same id
+        debug::Breakpoint bp5 = target.setBreakpoint(fixturePath, 2);
+        CHECK(bp5.status == debug::BreakpointStatus::Installed);
+        CHECK(bp5.id == 0);
+        CHECK(bp5.line == 2);
+
+        // wait until script is finished to call destructors
+        REQUIRE(exitFuture.wait_for(std::chrono::seconds(5)) == std::future_status::ready);
```
Vighnesh-V, line 85: note, this introduces a real time delay in tests

lute/debug/src/debugger.h
```diff
@@ -29,36 +33,74 @@ struct Breakpoint
...
 
     Luau::DenseHashMap<std::string, std::shared_ptr<Ref>> loadedSources; // source path -> reference to chunk
 
     bool installBreakpoint(lua_State* L, Breakpoint& bp);
     void installPendingBreakpoints(lua_State* L);
 };
 
+struct Process
+{
+    explicit Process(lua_State* thread, Target& parentTarget, LaunchConfig config);
```
Vighnesh-V: Process is not strictly the right verbiage here - I think Thread might be more 'correct'. very nit

lute/debug/src/debugger.cpp
```diff
@@ -117,32 +214,139 @@ bool Target::installBreakpoint(lua_State* L, Breakpoint& bp)
...
+{
+    if (!resumeToken)
+        return false;
+    continueRequested = true;
+    resumeToken->complete(
+        [](lua_State* L)
+        {
+            return 0;
+        }
+    );
```
Vighnesh-V: I forgot to mention this before, but can you run clang-format over your changes. The invokcation is:
```
clang-format -i path/to/things
```
to just in place update the files.

Vighnesh-V (changes requested):

lute/debug/src/debugger.cpp
```diff
@@ -23,54 +24,137 @@ Breakpoint::Breakpoint(int id, std::string sourcePath, int line, BreakpointStatu
...
 {
+    std::optional<Breakpoint> preexistingBp = getBreakpointBySourceLine(sourcePath, line);
+    if (preexistingBp)
+        return *preexistingBp;
     int id = currentBreakpointId;
     currentBreakpointId++;
-    auto [it, _] = breakpoints.insert_or_assign(id, Breakpoint(id, sourcePath, line, BreakpointStatus::Pending));
-    installBreakpoint(runtime.GL, it->second);
+    std::unique_lock lock(breakpointsMutex);
+    auto [it, _] = breakpoints.insert_or_assign(id, Breakpoint(id, sourcePath, line, BreakpointStatus::PendingInstall));
```
Vighnesh-V:
```suggestion
    auto [it, _] = breakpoints.insert_or_assign(id, Breakpoint{id, sourcePath, line, BreakpointStatus::PendingInstall});
```
nit

lute/debug/src/debugger.cpp
```diff
@@ -23,54 +24,137 @@ Breakpoint::Breakpoint(int id, std::string sourcePath, int line, BreakpointStatu
...
+    // We schedule breakpoint installs to happen async
+    if (childRuntime)
+        childRuntime->schedule(
+            [this, id]()
+            {
+                std::unique_lock lock(breakpointsMutex);
+                auto it = breakpoints.find(id);
+                // check breakpoint has not been erased when this scheduled install has been fired
+                if (it != breakpoints.end())
```
Vighnesh-V:
```suggestion
                // check breakpoint has not been erased when this scheduled install has been fired
                if (auto it = breakpoints.find(id); it != breakpoints.end())
```

lute/debug/src/debugger.cpp
```diff
@@ -90,12 +175,24 @@ std::vector<Breakpoint> Target::getBreakpointsByStatus(BreakpointStatus status)
...
 
+std::optional<Breakpoint> Target::getBreakpointBySourceLine(std::string source, int line) const
+{
+    std::unique_lock lock(breakpointsMutex);
+    for (const auto& [_, bp] : breakpoints)
+    {
+        if (bp.sourcePath == source && bp.line == line)
+            return bp;
+    }
+    return std::nullopt;
```
Vighnesh-V: Could be written with std::find_if.

lute/debug/src/debugger.h
Vighnesh-V, line 30: nit: does this own the path or does something else. If this doesnt own it, can be a string_view.

lute/debug/src/debugger.h
wizhao96: it's a little bit confusing but im trying to emulate the structure used by LLDB [here](https://lldb.llvm.org/python_api.html). this results in some confusion (i.e. a Process will own a bunch of Threads eventually but each Thread object really represents a coroutine). DAP also makes https://microsoft.github.io/debug-adapter-protocol/specification#Requests_Threads requests (which will actually result in a list of coroutines), which partially also contributed to the naming scheme. 

lmk if this is similar to what you would expect? or if names should be changed to better reflect the underlying Runtime instead of the protocol

lute/debug/src/debugger.cpp
```diff
@@ -117,32 +214,139 @@ bool Target::installBreakpoint(lua_State* L, Breakpoint& bp)
...
+        {
+            bool installed = installBreakpoint(L, bp);
+            if (installed && launchConfig.onBreakpointInstall)
+            {
+                lock.unlock();
+                launchConfig.onBreakpointInstall(bp);
+                lock.lock();
+            }
+        }
     }
```
Vighnesh-V, line 248: This feels like when you release the lock, the iterator that [_,bp] points to might get invalidated? Can you make the critical section
{
   acquire:
   track breakpoint structures to call the callback on.
}

{
    call the callback for those breakpoints.
}
I think this is okay because breakpoints are just plain old objects. You can copy them over.

Vighnesh-V (changes requested):

lute/debug/src/debugger.cpp
```diff
@@ -117,32 +214,139 @@ bool Target::installBreakpoint(lua_State* L, Breakpoint& bp)
...
+    if (activeProcess != nullptr)
+        return nullptr;
+    childRuntime = std::make_shared<Runtime>(parentRuntime.reporter);
+    setupState(*childRuntime, nullptr);
+    launchConfig = config;
+
     std::ifstream file(sourcePath);
     if (!file.is_open())
-        return false;
+        return nullptr;
```
Vighnesh-V: if the file open fails, does the childRuntime ever get cleaned up?

Vighnesh-V (changes requested):

tests/src/debug.test.cpp
wizhao96: yep i guess i understand this but i don't really see any good way around it for testing async behavior

lute/debug/src/debugger.h
wizhao96: the Target also owns a copy to the Process; at least for things like checking if the process has already been spawned before launching

lute/debug/src/debugger.cpp
wizhao96: we can't really initialize resumeToken on construction time because we should fetch it every single time we actually run into a stop. otherwise, we add a pending token to the runtime. If this is never decremented (i.e. we like never hit a breakpoint), `runToCompletion()` on the runtime stalls because `hasWork()` always has an active token.

lute/debug/src/debugger.h
wizhao96: breakpoint should own the path

lute/debug/src/debugger.cpp
wizhao96: if this was in response to the cited line of code, i don't think any changes needed to be made

lute/debug/src/debugger.cpp
```diff
@@ -117,32 +219,153 @@ bool Target::installBreakpoint(lua_State* L, Breakpoint& bp)
...
+    };
+    runtime.addThreadCompletionHandler(thread, std::move(completion));
+}
+
+bool Process::continueProcess()
+{
+    if (!resumeToken)
+        return false;
+    // we are continuing on a breakpoint and so might need to flag continueRequestedBp.
+    if (bpHit)
```
wizhao96, line 375: effectively this set of new lines of code makes sure that if we hit a breakpoint, after the callback (which theoretically could remove a breakpoint), that breakpoint is still there. otherwise, `continueRequestedBp` is set to `true` and we skip a breakpoint that should actually be hit.

this behavior is tested by a newly added `bp3` in the `Debug_hitBreakpoint` test

lute/debug/src/debugger.h
```diff
@@ -97,10 +99,13 @@ struct Process
     lua_State* thread;
     Runtime& runtime;
     Target& parentTarget;
-    ResumeToken resumeToken;
     LaunchConfig config;
+
+    // continueMutex protects continueRequestedBp, resumeToken, and bpHit
+    mutable std::mutex continueMutex;
```
wizhao96: since we can have multiple continue requests it's also worthwhile to lock these with a mutex

lute/debug/src/debugger.h
Vighnesh-V: I think with that in mind, this is very reasonable naming.

lute/debug/src/debugger.cpp
Vighnesh-V: cool!

lute/debug/src/debugger.cpp
```diff
@@ -115,34 +186,206 @@ bool Target::installBreakpoint(lua_State* L, Breakpoint& bp)
...
+            Luau::format(
+                "breakpoint %d installed at line %d in %s that is queued for uninstall could not be removed", bp.id, bp.line, bp.sourcePath.c_str()
+            )
+        );
+        return false;
+    }
+    return true;
+}
+
+// This returns the list of installed and unisntalled vectors for use in callbacks
```
Vighnesh-V: nit: typo on uninstalled.

lute/debug/src/debugger.cpp
```diff
@@ -115,34 +186,206 @@ bool Target::installBreakpoint(lua_State* L, Breakpoint& bp)
...
+    std::vector<int> toErase;
+    for (auto& [id, bp] : breakpoints)
     {
-        if (bp.status == BreakpointStatus::Pending)
-            installBreakpoint(L, bp);
+        if (bp.status == BreakpointStatus::PendingInstall)
+        {
+            bool installed = installBreakpoint(L, bp);
+            if (installed && launchConfig.onBreakpointInstall)
+                installedBpsCallback.push_back(bp);
```
Vighnesh-V: use emplace_back when possible!

lute/debug/src/debugger.cpp
```diff
@@ -115,34 +186,206 @@ bool Target::installBreakpoint(lua_State* L, Breakpoint& bp)
...
+        return false;
+    }
+    return true;
+}
+
+// This returns the list of installed and unisntalled vectors for use in callbacks
+// later.
+std::pair<std::vector<Breakpoint>, std::vector<Breakpoint>> Target::modifyPendingBreakpoints(lua_State* L)
+{
+    std::vector<Breakpoint> installedBpsCallback, uninstalledBpsCallback;
```
Vighnesh-V: initialize on separate lines please.

lute/debug/src/debugger.cpp
```diff
@@ -115,34 +186,206 @@ bool Target::installBreakpoint(lua_State* L, Breakpoint& bp)
...
+        }
+        if (bp.status == BreakpointStatus::PendingUninstall)
+        {
+            bool uninstalled = uninstallBreakpoint(L, bp);
+            if (uninstalled)
+            {
+                if (launchConfig.onBreakpointUninstall)
+                    uninstalledBpsCallback.push_back(bp);
+                toErase.push_back(bp.id);
+            }
```
Vighnesh-V: if this is the only place you're using the boolean, just check the condition directly.

lute/debug/src/debugger.cpp
```diff
@@ -115,34 +186,206 @@ bool Target::installBreakpoint(lua_State* L, Breakpoint& bp)
...
     }
+    for (int id : toErase)
+        breakpoints.erase(id);
+    return {installedBpsCallback, uninstalledBpsCallback};
 }
 
-bool Target::launch(const std::vector<std::string>& args)
+bool Target::launch(const std::string& sourcePath, const std::vector<std::string>& args, LaunchConfig config)
 {
+    // launch() cannot be called twice from the same target.
```
Vighnesh-V: can you make this a bool on Target and  LUTE_ASSERT if you see this again?

lute/debug/src/debugger.cpp
```diff
@@ -115,34 +186,206 @@ bool Target::installBreakpoint(lua_State* L, Breakpoint& bp)
...
     luaL_sandboxthread(thread);
     luau_load(thread, sourcePath.c_str(), bytecode.c_str(), bytecode.size(), 0);
     loadedSources[sourcePath] = std::make_shared<Ref>(thread, -1);
-    installPendingBreakpoints(thread);
+    auto [installed, uninstalled] = modifyPendingBreakpoints(thread);
     for (const std::string& arg : args)
         lua_pushstring(thread, arg.c_str());
-    runtime.runningThreads.push_back({true, getRefForThread(thread), static_cast<int>(args.size())});
-    lua_pop(runtime.GL, 1);
+    childRuntime->runningThreads.push_back({true, getRefForThread(thread), static_cast<int>(args.size())});
```
Vighnesh-V: emplace_back here and just pass the arguments directly as a list, instead of `{}` constructing them.

lute/debug/src/debugger.cpp
```diff
@@ -115,34 +186,206 @@ bool Target::installBreakpoint(lua_State* L, Breakpoint& bp)
...
+    if (bpHit)
+    {
+        // we need to check if our breakpoint is still currently installed after
+        // onBreakpointHit() callback
+        std::optional<Breakpoint> currentBp = getBreakpointByIdHelper(bpHit->id);
+        if (currentBp && currentBp->status == BreakpointStatus::Installed)
+            continueRequestedBp = true;
+        bpHit = std::nullopt;
+    }
+    token = std::move(resumeToken);
```
Vighnesh-V: If you have a lock, why do you need to move the resumeToken off the Target?

lute/debug/src/debugger.cpp
```diff
@@ -115,34 +186,206 @@ bool Target::installBreakpoint(lua_State* L, Breakpoint& bp)
...
+    paused = false;
+    lock.unlock();
+    // complete() holdes a continuation mutex that we want to avoid holding while holding targetMutex()
```
Vighnesh-V: nit: typo.

lute/debug/src/debugger.h
```diff
@@ -29,36 +33,72 @@ struct Breakpoint
...
     // We do this because clients may 1) configure breakpoints before launching executables
     // 2) we load sources dynamically with @require that a client may want to debug.
-    // TODO: implement 2 and add some callback when breakpoints get installed
-    Breakpoint addBreakpoint(std::string sourcePath, int line);
+    // TODO: implement 2
+    //
+    // Guarantees for when breakpoints are installed:
+    // Any breakpoint that is placed when the target process is paused (including before launch) and that
+    // have a loaded source are guaranteed to be installed before the process is resumed. Breakpoints placed on a loaded source
+    // when the target script is running may not be installed until the next time that script is paused.
```
Vighnesh-V, line 59: Are these comments still up to date?

lute/debug/src/debugger.h
```diff
@@ -29,36 +33,72 @@ struct Breakpoint
...
+    // For processes:
+    bool launch(const std::string& sourcePath, const std::vector<std::string>& args, LaunchConfig config = {});
+    bool continueProcess();
 
 private:
-    Runtime& runtime;
-    std::string sourcePath;
+    // targetMutex protects the entire Target, which basically serializes calls into it
+    // This is ok because we are not looking for high performance but rather correctness.
+    mutable std::mutex targetMutex;
```
Vighnesh-V, line 76: are there const functions using the `mutex`?

lute/debug/src/debugger.h
```diff
@@ -29,36 +33,72 @@ struct Breakpoint
...
-    bool launch(const std::vector<std::string>& args);
+    // For processes:
+    bool launch(const std::string& sourcePath, const std::vector<std::string>& args, LaunchConfig config = {});
+    bool continueProcess();
 
 private:
-    Runtime& runtime;
-    std::string sourcePath;
+    // targetMutex protects the entire Target, which basically serializes calls into it
+    // This is ok because we are not looking for high performance but rather correctness.
```
Vighnesh-V: I think the more useful invariant to comment here is that Target is accessed between the main thread and the child runtime thread. Thus, any writes to this object must be synchronized.

tests/src/debug.test.cpp
```diff
@@ -1,119 +1,232 @@
...
 
         // valid breakpoint
-        debug::Breakpoint bp = target.addBreakpoint(fixturePath, 2);
+        debug::Breakpoint bp = target.setBreakpoint(fixturePath, 2);
         // invalid breakpoint
-        debug::Breakpoint bp2 = target.addBreakpoint(fixturePath, 100);
+        debug::Breakpoint bp2 = target.setBreakpoint(fixturePath, 100);
         // breakpoint that will be moved on installation due to whitespace
-        debug::Breakpoint bp3 = target.addBreakpoint(fixturePath, 3);
+        debug::Breakpoint bp3 = target.setBreakpoint(fixturePath, 3);
```
Vighnesh-V: you can just write `using namespace debug` here.

lute/debug/src/debugger.h
```diff
@@ -29,36 +33,72 @@ struct Breakpoint
...
+    Breakpoint setBreakpoint(std::string sourcePath, int line);
     bool removeBreakpoint(int bpId);
 
     std::vector<Breakpoint> getBreakpoints() const;
     std::vector<Breakpoint> getBreakpointsByStatus(BreakpointStatus status) const;
     std::optional<Breakpoint> getBreakpointById(int breakpointId) const;
+    std::optional<Breakpoint> getBreakpointBySourceLine(std::string source, int line) const;
 
-    bool launch(const std::vector<std::string>& args);
+    // For processes:
```
Vighnesh-V: nit: process doesn't exist anymore.

lute/debug/src/debugger.cpp
```diff
@@ -115,34 +186,206 @@ bool Target::installBreakpoint(lua_State* L, Breakpoint& bp)
...
+    return true;
+}
+
+void Target::installBpHitCallback()
+{
+    lua_Callbacks* cb = lua_callbacks(childRuntime->GL);
+    cb->userdata = this;
+    cb->debugbreak = [](lua_State* L, lua_Debug* ar)
+    {
+        Target* target = static_cast<Target*>(lua_callbacks(L)->userdata);
```
Vighnesh-V:
```suggestion
        auto target = static_cast<std::shared_ptr<Target*>>(lua_callbacks(L)->userdata);
```

For an example of this, check out the `retake` function, which casts a unique pointer.

lute/debug/src/debugger.cpp
```diff
@@ -115,34 +186,206 @@ bool Target::installBreakpoint(lua_State* L, Breakpoint& bp)
...
+    for (int id : toErase)
+        breakpoints.erase(id);
+    return {installedBpsCallback, uninstalledBpsCallback};
 }
 
-bool Target::launch(const std::vector<std::string>& args)
+bool Target::launch(const std::string& sourcePath, const std::vector<std::string>& args, LaunchConfig config)
 {
+    // launch() cannot be called twice from the same target.
+    std::unique_lock lock(targetMutex);
```
Vighnesh-V: one small nit: where possible, can you use std::scope_lock and specify an explicit scope:
e.g:
```
{
std::lock_guard(targetMutex)
}
```

Vighnesh-V: Sometimes this isn't possible, so don't worry about making those changes.

Vighnesh-V: this saves you from having to manually unlock

Vighnesh-V (changes requested):
left some comments - emplace_back is the big thing here, to avoid a construction.

lute/debug/src/debugger.cpp
wizhao96: `runningThreads` is actually a `Luau::VecDeque` not supporting `emplace_back`

lute/debug/src/debugger.h
wizhao96: yep. the breakpoints placed on a loaded source when target script is running won't be installed until next time that script is paused.

lute/debug/src/debugger.h
wizhao96: yep all `const` breakpoint getters generally take the mutex as well

lute/debug/src/debugger.cpp
wizhao96: ok yea this wasn't so good and has been changed

lute/debug/src/debugger.cpp
wizhao96: hmm i actually don't think this is good. `std::shared_ptr<Target>(static_cast<Target*>(raw))` would try to delete `Target` when the pointer goes out of scope. this is bad because our `Target` is externally owned and may not be tracked on the heap.

wizhao96: i know the above comment is not exactly the code above but that code above doesn't work because `lua_callbacks(L)->userdata` is not stored as a `std::shared_ptr` but is instead simply set to `this`. we shouldn't store as a `std::shared_ptr` because then we would have to manually manage the memory allocated by the smart pointer for its own management as it gets cast into a `void*`.

lute/debug/src/debugger.cpp
Vighnesh-V: ahh okay, good catch!

lute/debug/src/debugger.cpp
Vighnesh-V: to be clear, I am suggesting storing the shared pointer on the userdata itself. You can store shared_ptr as a void* since it's just memory. If you can keep a second std::shared_ptr<Target> around (maybe owned by the parent thread), this is safe to do.

Vighnesh-V (approved):
Approving this since I don't want to block progress. As we keep developing this, keep in mind opportunities to do refactors here to simplify (e.g the mutex design we talked about today).

lute/debug/src/debugger.cpp
wizhao96: i can kind of see it where we create a shared pointer by making sure that we use `std::enable_shared_from_this` and then having a second shared_ptr. but im not necessarily willing to enforce that `Target` has to be managed via `shared_ptr`. 

i was also reading from this thread: https://www.reddit.com/r/cpp_questions/comments/d2owlo/how_can_i_pass_shared_ownership_through_a_void/ but i'm not entirely sure how accurate/related to the situation it is. there seem to be solutions there involving creating a shared pointer on the heap but that seems like it's overcomplicating the situation.

i am frankly not an expert on smart pointer usage but at least i feel relatively more confident that if we pass in a raw `Target*` that is pointing to `this`, this should be safe and not lead to memory leaks since the callback's memory is tied to the `childRuntime`, which we explicitly destroy in the `Target` destructor (as in a childRuntime's lifetime is always less than a Target's runtime).

PR #1220, opened 2026-07-08
lute pkg: add overrides

afujiwara-roblox:
- Add overrides support to loom, allowing projects to force specific versions or sources for any transitive dependency (similar to npm/pnpm overrides)
- Two override modes: version pin (force an exact registry version) and source redirect (replace a registry package with a local path or github source)
- Lockfile version bumps to 4 only when overrides are present; projects without overrides stay on version 3

Overrides are declared at the project root in loom.config.luau:

```
return {
    package = { ... },
    overrides = {
        -- Pin a transitive dependency to an exact version
        citrus = { sourceKind = "registry", source = "citrus", version = "3.0.0" },
        -- Or redirect to a local fork
        citrus = { sourceKind = "path", source = "./citrus-fork" },
    },
}
```

The resolver removes conflicting constraints for overridden packages, builds a modified universe respecting the pins, and stitches source-redirected packages back into the dependency graph after solving. The lockfile records a fingerprint of
active overrides so changes trigger re-resolution.

### DIFF
--- a/lute/cli/commands/pkg/loom-core/src/resolution.luau
+++ b/lute/cli/commands/pkg/loom-core/src/resolution.luau
@@ -96,19 +96,6 @@ local function removeRequirementsForPackage(registryRequirements: { [string]: se
 	end
 end
 
-local function serializeOverrides(overrides: { [string]: manifest.DependencySpec }): { [string]: string }
-	local result: { [string]: string } = {}
-	for name, spec in overrides do
-		if spec.sourceKind == "registry" then
-			result[name] = spec.version
-		elseif spec.sourceKind == "path" then
-			result[name] = `path:{spec.source}`
-		elseif spec.sourceKind == "github" then
-			result[name] = `github:{spec.source}@{spec.rev}`
-		end
-	end
-	return result
-end
 
 local function buildEffectiveUniverse(
 	universe: semver.Universe,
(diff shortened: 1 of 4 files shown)


lute/cli/commands/pkg/loom-core/src/resolution.luau
```diff
@@ -88,6 +88,99 @@ export type ResolveOptions = {
...
 
+local function removeRequirementsForPackage(registryRequirements: { [string]: semver.Requirement }, packageName: string)
+	for reqKey, req in registryRequirements do
+		if req.name == packageName then
+			registryRequirements[reqKey] = nil
+		end
+	end
+end
+
+local function serializeOverrides(overrides: { [string]: manifest.DependencySpec }): { [string]: string }
```
jescherob: Nit: is this the same function as lockfile.luau? If so can they be refactored to one?

lute/cli/commands/pkg/loom-core/src/resolution.luau
```diff
@@ -88,6 +88,99 @@ export type ResolveOptions = {
...
+		if registryOverrides[pkgName] then
+			local filtered: { semver.PackageVersion } = {}
+			for _, v in versions do
+				if semver.satisfies(v.version, registryOverrides[pkgName]) then
+					table.insert(filtered, v)
+				end
+			end
+			assert(
+				#filtered > 0,
+				`Override constrains '{pkgName}' to {overrides[pkgName].version}, but no matching version exists in the registry`
```
jescherob: Lute check issue, even though version always exist since this is only accessed by registry override, need a cast.

jescherob (approved):

PR #1222, opened 2026-07-09
Add a minimal DAP adapter

wizhao96:
This adds a minimal DAP adapter, which is not currently callable by the `lute` CLI
* we handle I/O and JSON parsing correctly (and add some tests for this)
* add some minimal implementations for the debugger startup/terminate events

If we modify Mandolin, we can connect it to VSCode
Screenshot 2026-07-10 at 2 11 11 PM

### DIFF
--- a/lute/cli/commands/debug/dap.luau
+++ b/lute/cli/commands/debug/dap.luau
@@ -1,91 +1,100 @@
 local io = require("@std/io")
 local json = require("@std/json")
 
-local readBuf = ""
-local writeSeqNum = 1
+-- we want generic Reader/Writer functions to easily test + support
+-- different transport mechanisms
+type Reader = (string?) -> string
+type Writer = (string) -> ()
 
-local function readDap(): json.Object
-	while not readBuf:find("\r\n\r\n") do
-		readBuf ..= io.input()
+local function dapLoop(reader: Reader, writer: Writer)
+	local readBuf = ""
+	local seqNum = 1
+
+	local function readDap(): json.Object
+		while not readBuf:find("\r\n\r\n") do
+			readBuf ..= reader()
+		end
+
+		-- the header for a DAP request will always be of the form
+		-- "Content-Length: 119\r\n\r\n"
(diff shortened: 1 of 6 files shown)


lute/cli/commands/debug/dap.luau
```diff
@@ -0,0 +1,101 @@
+local json = require("@std/json")
+
+-- we want generic Reader/Writer functions to easily test + support
+-- different transport mechanisms
+type Reader = (string?) -> string
```
wizhao96: in actual usage, this will be `io.input()` and `io.write()`

wizhao96:
ok just a note that the code structure is getting refactored somewhat (parsing is getting moved into a separate file along with the creation of a new `init.luau`)

lute/cli/src/climain.cpp
```diff
@@ -60,6 +61,10 @@ Compile Options:
 	lute compile <entry.luau> [--output <executable>]
 		Compiles entry point and auto-discovered dependencies into a standalone executable.
 
+Debug Options:
+	lute debug serve
+		Serves a DAP server for Luau for use with a development environment.
+
```
wizhao96, line 67: i believe in the future there should be an option to start a really simple basic terminal-based debugger. that's why it's `lute debug serve` and not just `lute debug`.

lute/cli/commands/debug/dapStream.luau
```diff
@@ -0,0 +1,88 @@
...
+			request_seq = requestSeq,
+			success = success,
+			command = command,
+		}
+		if message then
+			response.message = message
+		end
+		if body then
+			response.body = body
+		end
```
Vighnesh-V: I think in this repo we usually use the style
```
response.body = if body then body else nil
```

or 
```
response.body = body or nil
```

Vighnesh-V (changes requested):
one change requested!

Vighnesh-V (approved):

PR #1223, opened 2026-07-10
Add `values` to `tableext`

sbi-n:
Returns an array of values from a dictionary `tbl` with the keys discarded. It would be useful for functions like `toSet`, `extend` of `TableExt`

### DIFF
--- a/tests/std/tableext.test.luau
+++ b/tests/std/tableext.test.luau
@@ -75,6 +75,14 @@ test.suite("TableExt", function(suite)
 		end
 	end)
 
+	suite:case("values", function(assert)
+		local a = { a = 1, b = 2, c = 3 }
+		local values = tableext.values(a)
+		for _, value in a do
+			assert.neq(table.find(values, value), nil)
+		end
+	end)
+
 	suite:case("reverse", function(assert)
 		-- in-place
 		local a = { 1, 2, 3 }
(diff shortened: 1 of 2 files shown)


tests/std/tableext.test.luau
```diff
@@ -75,6 +75,14 @@ test.suite("TableExt", function(suite)
 		end
 	end)
 
+	suite:case("values", function(assert)
+		local a = { a = 1, b = 2, c = 3 }
+		local values = tableext.values(a)
+		for _, value in a do
+			assert.neq(table.find(values, value), nil)
+		end
+	end)
```
skberkeley, line 85:
```suggestion
		for _, value in a do
			assert.neq(table.find(values, value), nil)
		end
		assert.eq(#values, 3)
	end)
```

skberkeley (commented):
Thank you for your contribution! Just the one comment on the test you added.

tests/std/tableext.test.luau
sbi-n: I've applied it, though I was a bit surprised since my test was mostly imitated `.keys` test

skberkeley: Fair enough, we should probably update that test too separately

skberkeley (approved):

PR #1226, opened 2026-07-12
`lute pkg`: Add user-defined credential hooks

jkelaty-rbx:
Adds a pluggable credential resolution system for authenticating content downloads (currently just github.com). Users can specify an array of paths in their `.config.luau` (under `lute.credentials.providers`) that each point to a single luau module that exports a provider interface. 

This is meant to provide an alternative to the plaintext token storage loom currently supports. This was inspired as an alternative to introducing a native keyring storage implementation: https://github.com/luau-lang/lute/pull/1184

Resolution order:
1. `GITHUB_TOKEN` environment variable
2. User-defined credential provider hook (`lute.credentials.providers` via `.config.luau`)
3. Plaintext auth store (`~/.loom/auth.luau` via `lute pkg auth`)

The resolver hook takes precedence over the plaintext store, partially becuase I intend to propose removing the plaintext store if this solution is tenable enough.

Example:
```luau
-- src/.config.luau
return {
    lute = {
        credentials = {
            providers = { "./providers/github.luau" }
        },
        lint = { ... },
    },
}

-- src/providers/github.luau
local stringext = require("@std/stringext")
return {
    host = "github.com",
    resolve = function(request)
        local result = process.run({ "gh", "auth", "token", "--hostname", request.host })
        return if result.ok then stringext.trim(result.stdout) else nil
    end,
}
```


lute/cli/commands/pkg/loom-core/src/credentials.luau
```diff
@@ -0,0 +1,160 @@
+local fs = require("@std/fs")
+local luau = require("@std/luau")
+local path = require("@std/path")
+local process = require("@std/process")
+local stringext = require("@std/stringext")
+
+export type CredentialRequest = {
+	read protocol: string,
+	read host: string,
+	read path: string?,
```
afujiwara-roblox, line 10: is path used for anything right now?

afujiwara-roblox (approved):

lute/cli/commands/pkg/loom-core/src/credentials.luau
jkelaty-rbx: Currently it's unused by the internals, but it would be used by any credential brokers that need to disambiguate credentials based on path, e.g. `git-credential` supports this as a parameter: https://git-scm.com/docs/git-credential#_typical_use_of_git_credential

PR #1227, opened 2026-07-14
Add a debug runtime

wizhao96:
Adds a debug runtime that is templated. Allows for installing breakpoints synchronously.

### DIFF
--- a/lute/cli/include/lute/requiresetup.h
+++ b/lute/cli/include/lute/requiresetup.h
@@ -1,5 +1,6 @@
 #pragma once
 
+#include "lute/runtime.h"
 #include "lute/userlandvfs.h"
 
 #include "Luau/DenseHash.h"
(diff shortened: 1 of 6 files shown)


lute/runtime/include/lute/runtime.inl
```diff
@@ -190,54 +212,59 @@ void Runtime::reportError(lua_State* L)
     reporter.reportError(error);
 }
 
-void Runtime::runContinuously()
+template<bool DebugEnabled>
+void RuntimeT<DebugEnabled>::runContinuously()
 {
     runLoopThreadStarted = uv_thread_create(
-        &runLoopThread,
```
wizhao96, line 196: this is a clang-format change

lute/cli/include/lute/requiresetup.h
```diff
@@ -1,5 +1,6 @@
 #pragma once
 
+#include "lute/runtime.h"
```
wizhao96, line 3: we can't forward declare with a `struct Runtime` since it conflicts with the `typedef Runtime = RuntimeT ... `

wizhao96:
ultimately, we decided not to use templates b/c it leads to a lot of complexity and just add something to the base `Runtime`

PR #1231, opened 2026-07-16
`lute self`: Support updating to latest nightly builds

jkelaty-rbx:
Pretty straightforward, implements a `--nightly` flag to update to the latest nightly build as opposed to the latest official release which is currently at `1.0.0`: https://api.github.com/repos/luau-lang/lute/releases/latest

### DIFF
--- a/lute/cli/commands/self/update.luau
+++ b/lute/cli/commands/self/update.luau
@@ -65,13 +65,10 @@ return function(...: string)
 		error("Automatic updates are not supported for nightly releases. Use --nightly or --version.")
 	end
 
-	local tag =
-	    if requested ~= nil then
-		    release.normalizeTag(requested)
-	    elseif nightly then
-		    release.queryLatestNightlyTag(REPO)
-	    else
-		    release.queryLatestTag(REPO)
+	local tag = if requested ~= nil
+		then release.normalizeTag(requested)
+		elseif nightly then release.queryLatestNightlyTag(REPO)
+		else release.queryLatestTag(REPO)
 
 	if current == tag then
 		print(`lute is up to date ({tag})`)
--- a/lute/cli/commands/self/update.luau
+++ b/lute/cli/commands/self/update.luau
@@ -14,13 +14,15 @@ local HELP_STRING = [[
 Update the host lute binary to the latest release.
 
 OPTIONS:
+  --nightly           Install the latest nightly build
   --version <version> Install a specific release (e.g. 1.0.1)
-  -h, --help         Show this help message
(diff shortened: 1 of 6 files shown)


lute/cli/commands/self/update.luau
```diff
@@ -45,6 +47,11 @@ return function(...: string)
 	end
 
 	local requested = args:get("version")
+	local nightly = args:has("nightly")
+
+	if requested ~= nil and nightly then
```
aatxe, line 52: The inconsistency seems weird here? Probably the condition should be:

```suggestion
	if requested ~= nil and nightly ~= nil then
```

lute/cli/commands/self/update.luau
```diff
@@ -54,11 +61,19 @@ return function(...: string)
 		print(`warning: could not read version from {destination}; proceeding with update`)
 	end
 
-	if requested == nil and current ~= nil and string.find(current, "-nightly.", 1, true) ~= nil then
-		error("Automatic updates are not supported for nightly releases. Use --version to select a nightly release.")
+	if requested == nil and not nightly and current ~= nil and release.isNightlyTag(current) then
```
aatxe, line 64: Likewise, the condition is inconsistent here:
```suggestion
	if requested == nil and nightly == nil and current ~= nil and release.isNightlyTag(current) then
```

lute/cli/commands/self/update.luau
```diff
@@ -54,11 +61,19 @@ return function(...: string)
...
+	end
+
+	local tag: string
+	if requested ~= nil then
+		tag = release.normalizeTag(requested)
+	elseif nightly then
+		tag = release.queryLatestNightlyTag(REPO)
+	else
+		tag = release.queryLatestTag(REPO)
 	end
```
aatxe: More opinionated, but 

```suggestion
	local tag =
	    if requested ~= nil then
		    release.normalizeTag(requested)
	    elseif nightly then
		    release.queryLatestNightlyTag(REPO)
	    else
		    release.queryLatestTag(REPO)
```

lute/cli/commands/self/update.luau
skberkeley: `args:has` returns a bool is probably why

aatxe: oh, blegh, i misread them.

lute/cli/commands/self/update.luau
aatxe: Misread has as get.

aatxe (approved):

PR #1232, opened 2026-07-16
`lute/debug`: initial bindings to Luau

wizhao96:
Create initial bindings for `lute/debug` to Luau, which primarily focuses on breakpoint behavior.

- Add some basic Luau tests
- A definitions file for the library for type checking
- Renames `debugger.cpp` to `debuginternals.cpp` to more clearly separate out the bindings from internal implementation


tests/lute/debugger.test.luau
```diff
@@ -0,0 +1,166 @@
...
+	-- this test cause mainly checks launching and hitting breakpoints
+	suite:case("launchAndHitBreakpoint", function(assert)
+		local target = debug.newTarget()
+		assert.eq(typeof(target), "Target")
+		local fixturePath = path.format(path.join(proc.cwd(), "tests/src/debug/loop.luau"))
+		local bp = target:setBreakpoint(fixturePath, 6)
+		assert.eq(bp.id, 0)
+		assert.eq(bp.line, 6)
+		assert.eq(bp.sourcePath, fixturePath)
+		assert.eq(bp.status, "pendingInstall")
```
wizhao96: if we do a type definition in `export type BreakpointStatus` is there type checking going on here relative to it being a raw string?

tests/lute/debugger.test.luau
```diff
@@ -0,0 +1,166 @@
...
+	end)
+	-- this test case focuses on adding, getting, and removing breakpoints
+	suite:case("setAndRemove", function(assert)
+		local target = debug.newTarget()
+		assert.eq(typeof(target), "Target")
+		local fixturePath = path.format(path.join(proc.cwd(), "tests/src/debug/loop.luau"))
+		local bp1 = target:setBreakpoint(fixturePath, 6)
+		local bp2 = target:setBreakpoint(fixturePath, 7)
+		local bp3 = target:setBreakpoint(fixturePath, 4)
+		local bpGet = target:getBreakpointById(bp1.id) :: debug.Breakpoint
```
wizhao96: is there a better way to do this type narrowing here? i think the normal `assert` function is being shadowed by the assert library we're using for testing?

lute/debug/src/debug.cpp
```diff
@@ -0,0 +1,331 @@
...
+
+    // read callbacks from index 4 (optional table)
+    debug::LaunchConfig config;
+    Runtime* runtime = getRuntime(L);
+    if (lua_istable(L, 4))
+    {
+        if (auto ref = getOptionalCallback(L, 4, "onBreakpointHit"))
+        {
+            // this schedules the handler to run on the parent runtime queue.
+            config.onBreakpointHit = [ref, runtime](const debug::Breakpoint& bp)
```
wizhao96: i guess i just wanted to highlight this as the mechanism for setting the actual client-defined callback on the parent runtime. The C++ library makes no such assumptions about where the client-defined callbacks are run.

These user-defined callbacks are different than something like `debugbreak`, which runs on the child runtime.

lute/debug/src/debug.cpp
```diff
@@ -0,0 +1,331 @@
...
+static void initializeTarget(lua_State* L)
+{
+    luaL_newmetatable(L, "Target");
+
+    lua_pushcfunction(
+        L,
+        [](lua_State* L)
+        {
+            const char* index = luaL_checkstring(L, -1);
+            auto it = kTargetMethods.find(index);
```
wizhao96: i see in previous code the function dispatch is primarily done with a bunch of if statements checking `index`. Since this object will have a bunch of methods, I decided to use an unordered hash map for code cleanliness but this might be less efficient. On efficiency too, lmk if I should use something like a `Luau::DenseHashMap` instead of the `unordered_map` default implementation.

lute/debug/include/lute/debug.h
```diff
@@ -0,0 +1,20 @@
+#pragma once
+
+#include "lute/library.h"
+
+#include "lua.h"
+#include "lualib.h"
+
+// open the library as a standard global luau library. note:
+// this has a special name bc otherwise, it conflicts with a preexisting Lua library
```
Vighnesh-V, line 9: you could just call this library debugger!

lute/require/src/lutevfs.cpp
```diff
@@ -36,6 +37,7 @@ const Luau::DenseHashMap<std::string, lua_CFunction> kLuteModules = []()
     map["@lute/task.luau"] = luteopen_task;
     map["@lute/time.luau"] = luteopen_time;
     map["@lute/vm.luau"] = luteopen_vm;
+    map["@lute/debug.luau"] = luteopen_debug;
```
Vighnesh-V: For now, let's not add this here. The debug library should only exist for the debugger cli command. In `setupCLICommandState`, you can register your debug library explicitly.

lute/debug/src/debug.cpp
afujiwara-roblox: Should we worried about a dangling `Runtime` pointer in all of these callbacks? This is probably safe today depending on the order things get deconstructed, but we might be able to be a little more defensive here.

lute/debug/include/lute/debug.h
```diff
@@ -0,0 +1,20 @@
...
+LUTE_API int luaopen_debug_lute(lua_State* L);
+// open the library as a table on top of the stack
+LUTE_API int luteopen_debug(lua_State* L);
+
+struct Debug : LuteLibrary<Debug>
+{
+    static constexpr const char kName[] = "debug";
+    static int pushLibrary(lua_State* L);
+    static const luaL_Reg lib[];
+    static const char* const properties[];
```
afujiwara-roblox, line 19: What is `properties[]` for? I don't see this used anywhere

lute/debug/src/debug.cpp
```diff
@@ -0,0 +1,331 @@
...
+                    [success](lua_State* L)
+                    {
+                        lua_pushboolean(L, success);
+                        return 1;
+                    }
+                );
+            };
+        }
+    }
+    bool removed = target->launch(source, args, config);
```
afujiwara-roblox: nit: removed is confusing here. Unless I'm missing something, nothing is being removed?

lute/debug/src/debug.cpp
afujiwara-roblox: nit: can probably add a helper function for the (const debug::Breakpoint& bp) callbacks instead of repeating the same lambda function

lute/debug/src/debug.cpp
```diff
@@ -0,0 +1,331 @@
...
+{
+    lua_getfield(L, tableIndex, field);
+    std::shared_ptr<Ref> ref;
+    if (lua_isfunction(L, -1))
+        ref = std::make_shared<Ref>(L, -1);
+    lua_pop(L, 1);
+    return ref;
+}
+
+static int target_launch(lua_State* L)
```
afujiwara-roblox, line 212: It would be good to guard against launching multiple times.

lute/debug/include/lute/debug.h
wizhao96: `properties[]` is needed by whatever code auto-generates the bindings even if it is empty. i'll just explicitly define to `{nullptr}` in `debug.cpp` to make that clearer.

lute/debug/src/debug.cpp
wizhao96: it's guarded against in `debuginternals.cpp`

lute/debug/src/debug.cpp
Vighnesh-V: I guess holding a resume token here makes sense then, because that itself stores a shared_ptr to the runtime, which prevents premature destruction.

lute/debug/src/debug.cpp
```diff
@@ -0,0 +1,331 @@
...
+#include <optional>
+#include <string>
+#include <unordered_map>
+#include <vector>
+
+static debug::Target* getTarget(lua_State* L, int index)
+{
+    auto* storage = static_cast<std::shared_ptr<debug::Target>*>(lua_touserdatatagged(L, index, kTargetTag));
+    if (!storage || !*storage)
+        luaL_errorL(L, "invalid target");
```
Vighnesh-V: Maybe a slightly more descriptive error message here.

lute/debug/src/debug.cpp
```diff
@@ -0,0 +1,331 @@
...
+        return debug::BreakpointStatus::Installed;
+    if (strcmp(status, "invalid") == 0)
+        return debug::BreakpointStatus::Invalid;
+    // on unknown status, simply return that bp is invalid
+    return debug::BreakpointStatus::Invalid;
+}
+
+static void pushBreakpoint(lua_State* L, const debug::Breakpoint& bp)
+{
+    lua_createtable(L, 0, 4);
```
Vighnesh-V, line 74: If you're pushing things onto the stack, make sure you have enough stack space to do this! You can use lua_checkstack to do this. See the type serialization code for examples.

Vighnesh-V: also, worth making push breakpoint return an int (which is always 1). Then you can just return pushBreakpoint(...)

tests/lute/debugger.test.luau
Vighnesh-V: The typechecker does have the ability to do type checking relative to raw strings. You should check this with lute check.

Additionally, you can rename the `assert` in the function argument something else to avoid clashing with `assert`.

Vighnesh-V (changes requested):

raymondnumbergenerator (commented):
would it be worth starting the `debug` page in the `@lute` library documentation and add to it as we implement more bindings?

lute/debug/src/debug.cpp
raymondnumbergenerator: `debuginternals` has this:
```
        // launch() cannot be called twice from the same target.
        LUTE_ASSERT(!launched);
```
should this just return a false instead of hitting a c++ assert?

raymondnumbergenerator: idk if production builds hit asserts so it might not actually be guarding this and the API is defined as returning false if it fails to launch anyway

lute/debug/src/debug.cpp
raymondnumbergenerator: what happens if target gets deconstructed while it's still holding a resume token?

wizhao96: this runtime is actually the parent runtime, not the child one. this makes sense because callbacks to the client should be queued on the parent runtime. destroying the parent runtime should stop all these user callbacks from running on itself if that makes sense, so nothing should dangle. then eventually the parent runtime's destructor will destroy both the current target and the child runtime 

future PR doesn't rely on the resume token anymore so we shouldn't rely on that behavior.

lute/debug/src/debug.cpp
wizhao96: ah ok sure let's change this if prod never hits `LUTE_ASSERT`

lute/debug/include/lute/debug.h
```diff
@@ -0,0 +1,20 @@
+#pragma once
```
wizhao96, line 1: ok so this file and `debug.cpp` could be named `debugger.h` and `debugger.cpp`. Unfortunately, when we do this this way, the Github PR system thinks `debuginternals.cpp` is a new file and that the renamed `debugger.cpp` is just a modification of the old `debugger.cpp` (which is now `debuginternals.cpp`). That makes it super messy to review.

lute/require/src/lutevfs.cpp
wizhao96: i guess for everybody else's information, we talked about this and there are some issues with tests if we do it this way, so we're allowing as a library unconditionally for now

wizhao96:
> would it be worth starting the `debug` page in the `@lute` library documentation and add to it as we implement more bindings?

from my perspective, i think we should do this later when we have more functionality in the library

wizhao96:
since `pause` is now merged, I added bindings for that as well in the rebase.

lute/debug/src/debug.cpp
```diff
@@ -0,0 +1,350 @@
...
+    case debug::BreakpointStatus::PendingInstall:
+        return "pendingInstall";
+    case debug::BreakpointStatus::PendingUninstall:
+        return "pendingUninstall";
+    case debug::BreakpointStatus::Installed:
+        return "installed";
+    case debug::BreakpointStatus::Invalid:
+        return "invalid";
+    }
+    return "unknown";
```
annieetang: maybe we could use `LUTE_ASSERT(false)` and/or `LUTE_UNREACHABLE()` here (as well as in `breakpointStringToStatus()`) instead of "unknown", since we shouldn't be reaching this point right, and we also don't have "unknown" as an option in `type BreakpointStatus`

lute/debug/src/debug.cpp
```diff
@@ -0,0 +1,350 @@
...
+        return debug::BreakpointStatus::PendingUninstall;
+    if (strcmp(status, "installed") == 0)
+        return debug::BreakpointStatus::Installed;
+    if (strcmp(status, "invalid") == 0)
+        return debug::BreakpointStatus::Invalid;
+    // on unknown status, simply return that bp is invalid
+    return debug::BreakpointStatus::Invalid;
+}
+
+static int pushBreakpoint(lua_State* L, const debug::Breakpoint& bp)
```
annieetang: (nit) if it's not too much work, could you add some comments for what we're expecting to push, it would help with future maintenance of the code! we have some in `type.cpp` and their equivalent in `definitions/luau.luau`. all good if not, would just be a nice to have :)

lute/debug/src/debug.cpp
annieetang: have we considered a table-based `__index` for this instead? where we create the method functions once and store them in the table, so each method lookups could be a single table access and without a new allocation each time.

I was referencing `lute/time/src/time.cpp`, if you check out `init_duration_lib()` for example, but your `kTargetMethods` hashmap would make this way cleaner than that file, something like 

```
create the __index table
loop over `kTargetMethods`
   push the function 
   set the field
set the __index field
```

lute/debug/src/debug.cpp
```diff
@@ -0,0 +1,350 @@
...
+    auto target = getTarget(L, 1);
+    bool continued = target->continueProcess();
+    checkStack(L, 1);
+    lua_pushboolean(L, continued);
+    return 1;
+}
+
+static int target_pauseProcess(lua_State* L)
+{
+    auto target = getTarget(L, 1);
```
annieetang: (nit) could we specify the type instead of using `auto` for clarity? I saw you did for some of the `target_xxx` functions already, there are just a couple other ones that also use `auto`

lute/debug/src/debug.cpp
```diff
@@ -0,0 +1,350 @@
...
+    bool paused = target->pauseProcess();
+    checkStack(L, 1);
+    lua_pushboolean(L, paused);
+    return 1;
+}
+
+static int debug_newTarget(lua_State* L)
```
annieetang, line 302: ah also a nit, but could we have some of these functions be in a `namespace debug` or something? I've been referencing the semantics/styling we use in `fs` and other libraries in `lute` so keeping it similar could be great, but this is also just styling, feel free to ignore/address later

annieetang (commented):
Looking pretty solid to me! thanks for the work on this 😄 left a couple comments about some things that are mostly related to matching the semantics we have in other lute libraries. 

Also, looks like you have some good tests, but would love to see some more coverage! I'm noticing we aren't testing/covering some of the APIs yet? could be in follow up PRs, but things like testing `getBreakpointsByStatus()`, `getBreakpointBySourceLine()`, and for false/failure cases if it's possible would be nice (i.e.,  if we try to call `launch` twice on the same target, if we set a breakpoint on an invalid id, etc.)

wizhao96:
> Also, looks like you have some good tests, but would love to see some more coverage! I'm noticing we aren't testing/covering some of the APIs yet? could be in follow up PRs, but things like testing `getBreakpointsByStatus()`, `getBreakpointBySourceLine()`, and for false/failure cases if it's possible would be nice (i.e., if we try to call `launch` twice on the same target, if we set a breakpoint on an invalid id, etc.)

ok so on testing, I think I'll probably try to cover at least every binding in one test in `debug.test.luau` (this wasn't the case exactly before for `getBreakpointsByStatus` and `getBreakpointBySourceLine`. i've also done some buffing up of edge cases on the C++ tests here specifically on `getBreakpointBySourceLine`: https://github.com/luau-lang/lute/pull/1249

lute/debug/src/debug.cpp
```diff
@@ -0,0 +1,383 @@
...
+
+    // read args from index 3
+    std::vector<std::string> args;
+    if (lua_istable(L, 3))
+    {
+        int n = lua_objlen(L, 3);
+        for (int i = 1; i <= n; i++)
+        {
+            lua_rawgeti(L, 3, i);
+            args.push_back(luaL_checkstring(L, -1));
```
Vighnesh-V: where possible, use emplace_back. It might not really be necessary here, but if you use it by default, youll avoid random constructions.

Vighnesh-V (approved):

annieetang (approved):
look good to me, ty!!

wizhao96:
ok im removing the `pauseProcess` luau test for now because it's flaky. i think with some additional debugger features implemented we can make it reliably pass but rn it is hard to do so.

PR #1233, opened 2026-07-17
`lute/debug`: connect pausing to Runtime

wizhao96:
Previously, we used `resumeTokens` to do pausing in the debugger. This is unsustainable for multi-coroutine applications. Instead, we introduce a`debugMode` to `Runtime` that enables pausing all coroutines at once. We also add a `pauseProcess` method in the debugger itself that can be called outside of breakpoints.

The pausing pathway is now:
1. Trigger the presently queued running coroutine to break out of execution. The easiest way to do this is to set a callback in the VM and then call `lua_break`
2. Use the runtime's `stopDebug()` to actually stop queuing of further coroutines.

### DIFF
--- a/lute/debug/src/debugger.cpp
+++ b/lute/debug/src/debugger.cpp
@@ -285,6 +285,8 @@ bool Target::launch(const std::string& sourcePath, const std::vector<std::string
         lua_pop(childRuntime->GL, 1);
 
         scriptThread = thread;
+        lua_Callbacks* cb = lua_callbacks(childRuntime->GL);
+        cb->userdata = this;
         installBpHitCallback();
         installExitCallback();
         // All VM setup happens synchronously before runContinuously starts the background thread.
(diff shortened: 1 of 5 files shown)


lute/runtime/src/runtime.cpp
```diff
@@ -193,29 +216,29 @@ void Runtime::reportError(lua_State* L)
...
-                        }
-                    );
-                }
-
-                self->runToCompletion();
-            }
-        },
-        this
-    ) == 0;
+                               &runLoopThread,
```
wizhao96: this is just a formatting change from `clang-format`

lute/debug/src/debugger.cpp
```diff
@@ -285,6 +285,8 @@ bool Target::launch(const std::string& sourcePath, const std::vector<std::string
         lua_pop(childRuntime->GL, 1);
 
         scriptThread = thread;
+        lua_Callbacks* cb = lua_callbacks(childRuntime->GL);
+        cb->userdata = this;
```
wizhao96, line 289: realistically this should be here not in installBpHitCallbacks because many other callbacks need `cb->userdata`

Vighnesh-V (approved):

PR #1234, opened 2026-07-17
`lute check`: Support loading definition files

jkelaty-rbx:
## Summary
Adds support for user-specified type definition files in `lute check`. Definition files (`.d.luau`) declare globals and types that are injected into the type checker's global scope, allowing user code to reference them without explicit imports.

## Configuration
Definition files are specified in `.config.luau` under lute.definitions as a map:
```luau

return {
    lute = {
        check = {
            definitions = {
                roblox = "./types/roblox.d.luau",
                custom = "./types/custom-globals.d.luau",
            },
        },
    },
}
```


tests/src/typecheck/definitions_invalid/.config.luau
```diff
@@ -0,0 +1,7 @@
+return {
+    lute = {
+        definitions = {
```
jkelaty-rbx: I could also make this `lute.check.definitions` if we'd prefer

Vighnesh-V: I think that might be better, since I'm adding support for selecting the solver mode too.

```
lute = { check = { mode = true, definitions = {...}}}
```

lute/cli/src/tc.cpp
```diff
@@ -117,6 +249,39 @@ int typecheck(const std::vector<std::string>& sourceFilesInput, LuteReporter& re
         return 1;
     }
 
+    std::vector<DefinitionEntry> definitionEntries;
+    std::string configDir;
+
+    std::optional<std::string> configPath = configPathOverride ? configPathOverride : findProjectConfig();
```
Vighnesh-V: Quick question - for config.luau parsing in require, is there a function for doing this? The reason I ask, is that the process to find a .config.luau is defined in the rfc already, and we should be respecting that behaviour.

Vighnesh-V: If so, Varun might have written a function to do this already, and we can just lean on that.

Vighnesh-V (changes requested):
one small request - if not, I'll re-review the PR as is.

tests/src/typecheck/definitions_invalid/.config.luau
jkelaty-rbx: Updated!

lute/cli/src/tc.cpp
jkelaty-rbx: It seems the config discovery is generally implemented ad-hoc in places: https://github.com/luau-lang/lute/blob/primary/lute/luau/src/configresolver.cpp#L24-L31 and https://github.com/luau-lang/lute/blob/primary/lute/require/src/filevfs.cpp I couldn't find any explicit general purpose helper for config discovery from my initial search

jkelaty-rbx: @vrn-sn do you have any thoughts on this?

hgoldstein: Varun is OOO. If we're already implementing it ad-hoc then it doesn't seem horrible to keep doing so, but maybe @Vighnesh-V writes up a common implementation.

Vighnesh-V: yeah it's not ideal to not have a single implementation, but I don't see any reason we can't add one at a later time.

lute/cli/src/tc.cpp
```diff
@@ -107,7 +108,169 @@ std::vector<std::string> processSourceFiles(const std::vector<std::string>& sour
...
+    {
+        const std::string* key = k.get_if<std::string>();
+        if (!key)
+            return {{}, "configuration keys in \"lute.check.definitions\" must be strings"};
+
+        const std::string* value = v.get_if<std::string>();
+        if (!value)
+            return {{}, "configuration values in \"lute.check.definitions\" must be strings (file paths)"};
+
+        entries.push_back({*key, normalizePath(joinPaths(configDir, *value))});
```
Vighnesh-V:
```suggestion
        entries.emplace_back(*key, normalizePath(joinPaths(configDir, *value)));
```

lute/cli/src/tc.cpp
```diff
@@ -129,6 +299,17 @@ int typecheck(const std::vector<std::string>& sourceFilesInput, LuteReporter& re
...
     Luau::registerBuiltinGlobals(frontend, frontend.globals);
+
+    if (!definitions.entries.empty())
+    {
+        int defFailures = loadDefinitions(frontend, definitions.entries, reporter);
+        if (defFailures > 0)
+        {
+            Luau::freeze(frontend.globals.globalTypes);
+            return 1;
+        }
```
Vighnesh-V: if you want richer error messages / richer types, I think making this an optional<Parse/TcFailure> will be nicer.

lute/cli/src/tc.cpp
```diff
@@ -129,6 +299,17 @@ int typecheck(const std::vector<std::string>& sourceFilesInput, LuteReporter& re
...
+    }
+
     Luau::freeze(frontend.globals.globalTypes);
```
Vighnesh-V, line 317: It looks like this freeze happens unconditionally. Can you hoist this?

lute/cli/src/tc.cpp
```diff
@@ -107,7 +108,169 @@ std::vector<std::string> processSourceFiles(const std::vector<std::string>& sour
...
+    std::vector<DefinitionEntry> entries;
+    std::string error;
+
+    bool ok() const
+    {
+        return error.empty();
+    }
+};
+
+static DefinitionExtractResult extractDefinitionEntries(const Luau::ConfigTable& configTable, const std::string& configDir)
```
Vighnesh-V: maybe put this as a static function on the DefinitionExtractResult for code organization?

lute/cli/src/tc.cpp
```diff
@@ -107,7 +108,169 @@ std::vector<std::string> processSourceFiles(const std::vector<std::string>& sour
...
+{
+    if (!configTable.contains("lute"))
+        return {};
+
+    const Luau::ConfigTable* luteTable = configTable.find("lute")->get_if<Luau::ConfigTable>();
```
Vighnesh-V: omega nit: auto is quite nice here

lute/cli/src/tc.cpp
```diff
@@ -107,7 +108,169 @@ std::vector<std::string> processSourceFiles(const std::vector<std::string>& sour
...
+    if (!luteTable)
+        return {{}, "configuration value for key \"lute\" must be a table"};
```
Vighnesh-V: I think you could define two constructors for this, one that takes a vector, str error, and one that just takes an error and use the latter one here for brevity.

lute/cli/src/tc.cpp
Vighnesh-V: I'm also not sure why this succeeds? If the key doesn't exist, find("lute") will return nullpointer. Don't you need to check the result of `configTable.find("lute")` first before invoking get_if on it?

Vighnesh-V: oh I see. You do configTable.contains before.

Can you simplify this code a bit by doing:
```
auto prop = configTable.find("...")

if (!prop)
    return { "error" }

auto nestedProp = prop->get_if....
if (!nestedProp)
    return { "error" }

```

That way you don't pay for two lookups.

lute/cli/src/tc.cpp
```diff
@@ -107,7 +108,169 @@ std::vector<std::string> processSourceFiles(const std::vector<std::string>& sour
...
+
+    if (!luteTable->contains("check"))
```
Vighnesh-V: ditto!

lute/cli/src/tc.cpp
```diff
@@ -107,7 +108,169 @@ std::vector<std::string> processSourceFiles(const std::vector<std::string>& sour
...
+        std::string parseError;
+        std::optional<Luau::ConfigTable> configTable = Luau::extractConfig(*contents, {}, &parseError);
+        if (!configTable)
+            return {{}, "Error parsing " + *configPathOverride + ": " + parseError};
+
+        std::optional<std::string> configDir = getParentPath(*configPathOverride);
+        return extractDefinitionEntries(*configTable, configDir.value_or(""));
+    }
+
+    for (auto dir = getCurrentWorkingDirectory(); dir; dir = getParentPath(*dir))
```
Vighnesh-V, line 189: Could this be auto& to avoid a copy or no?

lute/cli/src/tc.cpp
```diff
@@ -107,7 +108,169 @@ std::vector<std::string> processSourceFiles(const std::vector<std::string>& sour
...
+            continue;
+
+        std::string parseError;
+        std::optional<Luau::ConfigTable> configTable = Luau::extractConfig(*contents, {}, &parseError);
+        if (!configTable)
+            return {{}, "Error parsing " + candidate + ": " + parseError};
+
+        DefinitionExtractResult result = extractDefinitionEntries(*configTable, *dir);
+        if (!result.ok())
+            return {{}, "Error in " + candidate + ": " + result.error};
```
Vighnesh-V: nit: These error messages should ideally indicate a) which definition file failed to load, b) what the parse error / tc error was.

lute/cli/src/tc.cpp
```diff
@@ -107,7 +108,169 @@ std::vector<std::string> processSourceFiles(const std::vector<std::string>& sour
...
+
+        if (!result.entries.empty())
+            return result;
```
Vighnesh-V: would it be ok to just return the result here? Could be that they wrote a lute.definitions but the table itself was empty?

lute/cli/src/tc.cpp
```diff
@@ -117,6 +280,13 @@ int typecheck(const std::vector<std::string>& sourceFilesInput, LuteReporter& re
         return 1;
     }
 
+    DefinitionExtractResult definitions = findDefinitions(configPathOverride);
+    if (!definitions.ok())
```
Vighnesh-V: usually the C++ style we use in lute is to return an optional instead of checking a `.ok()` method. You can add a std::string& errorMessage parameter and then use that to report if findDefinitions returns a optional.

tests/src/typecheck/definitions_invalid/.config.luau
```diff
@@ -0,0 +1,9 @@
+return {
+    lute = {
+        check = {
+            definitions = {
```
Vighnesh-V, line 4: Two more test requests:

1) User writes `check = { } `
2) User writes `check = { definitions = { } }`

Vighnesh-V (changes requested):

tests/src/typecheck/definitions_invalid/.config.luau
jkelaty-rbx: Added!

lute/cli/src/tc.cpp
jkelaty-rbx: My understanding is that loadDefinitions may mutate it, so we can't call it earlier than here

lute/cli/src/tc.cpp
```diff
@@ -129,6 +318,17 @@ int typecheck(const std::vector<std::string>& sourceFilesInput, LuteReporter& re
     Luau::Frontend frontend(Luau::SolverMode::New, &fileResolver, &configResolver, frontendOptions);
 
     Luau::registerBuiltinGlobals(frontend, frontend.globals);
+
+    if (!definitions->empty())
+    {
+        int defFailures = loadDefinitions(frontend, *definitions, reporter);
+        if (defFailures > 0)
+        {
+            Luau::freeze(frontend.globals.globalTypes);
```
jkelaty-rbx: This freeze might not even be necessary tbh

lute/cli/src/tc.cpp
jkelaty-rbx: Done! Although I chose not to return error when the field was nil/unspecified

lute/cli/src/tc.cpp
jkelaty-rbx: Any errors that we encounter during definition loading are reported into the `reporter` so messages should at least propagate as expected, this is just returning the status code for control-flow purposes

lute/cli/src/tc.cpp
jkelaty-rbx: done!

lute/cli/src/tc.cpp
jkelaty-rbx: auto& didn't seem to work here, though I can't really articulate the details of why very clearly lol I'm going to say this is probably fine as-is though since I see other call sites also perform a copy

lute/cli/src/tc.cpp
jkelaty-rbx: a-hoisted

lute/cli/src/tc.cpp
jkelaty-rbx: I updated the error messages, lmk if there's any specific coverage I missed

lute/cli/src/tc.cpp
jkelaty-rbx: I did a pass over the control-flow handling here, it should be better now, but lmk if you see anything I could improve

lute/cli/src/tc.cpp
jkelaty-rbx: I did as you suggested, but i felt like the code was harder to understand control-flow invariants in terms of optional and an out parameter, so I switched it in my latest commit to a result<> shaped solution. Lmk what you think about it, I don't mind switching it back if you prefer we stick with the out param

lute/cli/src/tc.cpp
```diff
@@ -107,7 +108,212 @@ std::vector<std::string> processSourceFiles(const std::vector<std::string>& sour
     return files;
 }
 
-int typecheck(const std::vector<std::string>& sourceFilesInput, LuteReporter& reporter)
+template<typename T>
+struct Result
```
Vighnesh-V: Should the result type also be parameterized on the error? Either way, can you move the implementation here to a place other parts of the code could use?

lute/cli/src/tc.cpp
```diff
@@ -107,7 +108,212 @@ std::vector<std::string> processSourceFiles(const std::vector<std::string>& sour
...
+{
+    std::optional<T> value;
+    std::string error;
+
+    Result() = default;
```
Vighnesh-V: I don't think Result should be default initializable. A default constructed `Result` would by default both be err (empty value) and ok(), empty string.

A few more notes:
- Can you make static factory methods like:
```
static Result success(T value);
static Result failure(error message);
```

- Can you make the state private?
- Can you replace the two members with just `Luau::Variant<T, E>`. That way you only have to allocate one thing instead of two.

lute/cli/src/tc.cpp
```diff
@@ -107,7 +108,212 @@ std::vector<std::string> processSourceFiles(const std::vector<std::string>& sour
...
+    if (!definitions.ok())
+    {
+        reporter.formatError("%s\n", definitions.error.c_str());
+        return 1;
+    }
+
+    if (!definitions.value->empty())
+    {
+        int defFailures = loadDefinitions(*definitions.value, frontend, reporter);
+        if (defFailures > 0)
```
Vighnesh-V:
```suggestion
        if (int defFailures = loadDefinitions(*definitions.value, frontend, reporter); defFailures > 0)
```

for terseness?

Vighnesh-V (approved):
Approved, but left one set of comments on the Result data structure that should be addressed.

lute/cli/src/tc.cpp
Vighnesh-V: fair enough! I trust your judgement here.

lute/cli/src/tc.cpp
jkelaty-rbx: done!

lute/cli/src/tc.cpp
jkelaty-rbx: done!

lute/cli/src/tc.cpp
jkelaty-rbx: done!

lute/cli/src/tc.cpp
jkelaty-rbx: I held off on this for now

Vighnesh-V (approved):

PR #1238, opened 2026-07-20
`lute toolchain`: Dispatch to config-pinned lute version

jkelaty-rbx:
## Summary

Adds project-local version pinning for lute. When you run lute in a project that pins a version in `.config.luau`, the running binary transparently downloads (if needed) and re-execs into the pinned version. `lute self` and `lute toolchain` always run on the invoked binary, or `LUTE_SKIP_DISPATCH=1` can be set to opt out of the dispatch.

## Example

```luau
-- .config.luau
return {
    lute = {
        toolchain = {
            lute = "1.0.1",
        },
    },
}
```
Now any lute <command> run in that directory tree uses `v1.0.1`, downloading it to `~/.lute/versions/lute/v1.0.1/lute` on first use.


jkelaty-rbx:
Closing this for now as I don't see this being used for myself anymore, at least until further consensus is reached on https://github.com/luau-lang/lute/issues/1266

PR #1242, opened 2026-07-21
add timeout opt to process.run / process.system

wmccrthy:
Adds a timeout option to process.run and process.system. When specified (in seconds), the child process is sent SIGTERM after the deadline. If it doesn't exit within a 2-second grace period, it's sent SIGKILL. The result includes `timedOut: boolean?` (true only when killed by timeout) and ok is set to false.

  - Spawns with UV_PROCESS_DETACHED when timeout is set, so the child becomes its own process group leader
  - Kills the process group via uv_kill(-pid, sig) with a fallback to uv_process_kill for platforms that don't support negative PIDs (Windows)
  - Grace period is handled by a second uv_timer_t that only initializes if the SIGTERM timer actually fires

Consumption will look like:
```luau
local r = process.run({ "some-command" }, { timeout = 5 })
if r.timedOut then
  -- process was killed after 5 seconds
end
```


lute/process/src/processhandle.cpp
```diff
@@ -270,14 +343,14 @@ void ProcessHandle::completeProcessExecution()
     {
         int64_t finalExitCode = state.getExitCode();
         int finalTermSignal = state.getTermSignal();
-        // TODO: should we put any leftover stdin data into the output here?
         std::string finalStdout = state.stdoutData;
```
wmccrthy, line 348:
```suggestion
		// TODO: should we put any leftover stdin data into the output here?
        std::string finalStdout = state.stdoutData;
```

lute/process/src/processhandle.cpp
```diff
@@ -270,14 +343,15 @@ void ProcessHandle::completeProcessExecution()
...
+		// TODO: should we put any leftover stdin data into the output here?
         std::string finalStdout = state.stdoutData;
         std::string finalStderr = state.stderrData;
         std::string finalSignalStr = finalTermSignal ? std::to_string(finalTermSignal) : "";
+        bool finalTimedOut = timedOut;
         resumeToken->complete(
             [=](lua_State* L)
             {
-                lua_createtable(L, 0, 5); // ok, exitCode, stdout, stderr, signal
+                lua_createtable(L, 0, 6);
```
wmccrthy:
```suggestion
                lua_createtable(L, 0, 6); // ok, exitCode, stdout, stderr, signal, timeout
```

lute/process/src/processhandle.cpp
```diff
@@ -270,14 +343,15 @@ void ProcessHandle::completeProcessExecution()
     {
         int64_t finalExitCode = state.getExitCode();
         int finalTermSignal = state.getTermSignal();
-        // TODO: should we put any leftover stdin data into the output here?
+		// TODO: should we put any leftover stdin data into the output here?
         std::string finalStdout = state.stdoutData;
```
wmccrthy:
```suggestion
        // TODO: should we put any leftover stdin data into the output here?
        std::string finalStdout = state.stdoutData;
```

lute/process/include/lute/processhandle.h
```diff
@@ -27,6 +27,7 @@ struct ProcessOptions
     std::string stdioKind = kStdioKindDefault;
     std::map<std::string, std::string> env;
     std::string customShell; // only used by system()
+    double timeout = 0; // seconds; 0 means no timeout
```
Vighnesh-V: can you make this in ms instead?

lute/process/include/lute/processhandle.h
```diff
@@ -124,13 +125,22 @@ struct ProcessHandle
     std::vector<std::string> environmentStrings;
     std::vector<char*> environmentVarString;
 
+    uv_timer_t timeoutTimer;
+    uv_timer_t killTimer;
```
Vighnesh-V, line 129: Ariel added a signal handle wrapper - would you be able to use that here instead?

Vighnesh-V (commented):
one comment about a refactor...

Vighnesh-V (approved):

PR #1243, opened 2026-07-21
Reenable a suppressed flag

skberkeley:
This broke typechecking somewhere, but should now be fixed after the latest Luau update.

### DIFF
--- a/lute/cli/src/luauflags.cpp
+++ b/lute/cli/src/luauflags.cpp
@@ -33,7 +33,6 @@ void setLuauFlags()
 {
     enableAllLuauFlags();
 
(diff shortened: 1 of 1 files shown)


Vighnesh-V (approved):

hgoldstein:
Apparently not :( y'all might need a newer Luau version?

aatxe:
We just updated to 731, so worth a shot now.

skberkeley:
> We just updated to 731, so worth a shot now.

I don't think a fix exists yet, parking in draft for now

hgoldstein:
I just landed a (very hacky) fix that would go out in ... 733 I think.

PR #1244, opened 2026-07-21
`lute/debug`: fix destructor when in infinite loop

wizhao96:
Fix `Target` destructor in infinite loop case as part of splitting https://github.com/luau-lang/lute/pull/1233 up.
- If the debugger is running an infinite loop, that coroutine would never break and so the destructor will stall infinitely.
- This PR fixes this by making sure that always coroutines break out of execution in our destructor by setting the `interrupt` callback. This interrupt callback is called on back edges of loops or function calls/returns (which realistically will always occur when the thread is running an infinite loop)

### DIFF
--- a/lute/debug/src/debuginternals.cpp
+++ b/lute/debug/src/debuginternals.cpp
@@ -47,6 +47,7 @@ Target::~Target()
     // any infinite/long-running coroutines.
     if (launched)
     {
+        childRuntime->continueDebug();
         lua_Callbacks* cb = lua_callbacks(childRuntime->GL);
         cb->interrupt = [](lua_State* L, int gc)
         {
--- a/lute/debug/src/debuginternals.cpp
+++ b/lute/debug/src/debuginternals.cpp
@@ -43,6 +43,18 @@ Target::~Target()
     // We want to stop the runtime so nothing runs while we are destroying the target but first
     // we need to clear all sources (which are stored as refs in the runtime).
     loadedSources.clear();
+    // this interrupts execution of runToCompletion() in order to stop
+    // any infinite/long-running coroutines.
+    if (launched)
+    {
+        lua_Callbacks* cb = lua_callbacks(childRuntime->GL);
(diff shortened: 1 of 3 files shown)


lute/runtime/src/runtime.cpp
```diff
@@ -40,8 +40,22 @@ Runtime::~Runtime()
...
+        // this interrupts execution of runToCompletion() in order to stop
+        // any infinite/long-running coroutines.
+        if (runLoopThreadStarted)
+        {
+            lua_Callbacks* cb = lua_callbacks(GL);
+            cb->interrupt = [](lua_State* L, int gc)
+            {
+                if (gc != -1)
+                    return;
+                lua_break(L);
```
wizhao96: this does `lua_break` rather than `lua_yield` since we don't want to do a `lua_resume` in the future

lute/debug/src/debuginternals.cpp
```diff
@@ -43,6 +43,18 @@ Target::~Target()
     // We want to stop the runtime so nothing runs while we are destroying the target but first
     // we need to clear all sources (which are stored as refs in the runtime).
     loadedSources.clear();
+    // this interrupts execution of runToCompletion() in order to stop
+    // any infinite/long-running coroutines.
+    if (launched)
+    {
+        lua_Callbacks* cb = lua_callbacks(childRuntime->GL);
+        cb->interrupt = [](lua_State* L, int gc)
```
raymondnumbergenerator, line 52: is installing an interrupt callback sufficient? e.g. a paused runtime waiting on `debugStoppedCv` never gets notified and also never runs this callback

i suppose it is a different case than this hang though

wizhao96: yea good point. originally this interrupt change was going to go into `Runtime` but then got moved into `debuginterrals` but we should make sure to keep the runtime continued now

Vighnesh-V (approved):

PR #1245, opened 2026-07-21
`lute/runtime`: clang-format

wizhao96:
Run `clang-format` (v22.1.8) on `runtime` as a part of splitting https://github.com/luau-lang/lute/pull/1233 into three.


wizhao96:
just a quick q: why does `check-format` not check `clang-format` on our cpp files?

Vighnesh-V:
@wizhao96 good question. We should add it.

Vighnesh-V (approved):

PR #1246, opened 2026-07-22
Fix nightly release missing actions:read permission

jkelaty-rbx:


jkelaty-rbx:
Seems to be no longer needed

PR #1248, opened 2026-07-23
`lute/debug`: adds multisource support  

wizhao96:
Add support for multiple sources in the debugger
* Intercept calls to loading new modules by modifying the `RequireCtx`
* also add `getLoadedSources` to debugger Target + bindings

### DIFF
--- a/lute/debug/include/lute/debuginternals.h
+++ b/lute/debug/include/lute/debuginternals.h
@@ -70,7 +70,7 @@ struct Target
     std::optional<Breakpoint> getBreakpointBySourceLine(std::string source, int line) const;
 
     // For actively running scripts:
-    bool launch(const std::string& sourcePath, const std::vector<std::string>& args, LaunchConfig config = {});
+    bool launch(std::string sourcePath, const std::vector<std::string>& args, LaunchConfig config = {});
     bool continueProcess();
     bool pauseProcess();
 
--- a/lute/debug/include/lute/debuginternals.h
+++ b/lute/debug/include/lute/debuginternals.h
@@ -28,6 +28,7 @@ enum class BreakpointStatus
 struct Breakpoint
 {
     int id;
+    // This will use forward slashes instead of backwards.
     std::string sourcePath;
     int line;
     BreakpointStatus status;
@@ -48,7 +49,9 @@ struct Target
     explicit Target(Runtime& parentRuntime);
     ~Target();
 
-    // Get list of sources
+    // Get list of sources, with sources using forward slahes consistently.
+    // Our principle in path format is that we accept any path format as input but will
+    // internally use and then output with paths that use exclusively forward slashes.
     std::vector<std::string> getLoadedSources();
 
     // Setting breakpoints is a two step process. We add them to our Target. If they
--- a/lute/debug/include/lute/debuginternals.h
+++ b/lute/debug/include/lute/debuginternals.h
@@ -1,5 +1,6 @@
 #pragma once
 
+#include "lute/require.h"
 #include "lute/runtime.h"
 
 #include "Luau/DenseHash.h"
@@ -46,6 +47,9 @@ struct Target
 {
     explicit Target(Runtime& parentRuntime);
     ~Target();
+
+    // Get list of sources
+    std::vector<std::string> getLoadedSources();
 
     // Setting breakpoints is a two step process. We add them to our Target. If they
     // involve a source that has already been loaded by the VM, we attempt to install that
@@ -95,6 +99,9 @@ private:
 
     // our stopped thread that we need to requeue when we continue
     lua_State* stoppedThread = nullptr;
+
+    // for require contexts
+    std::unique_ptr<RequireCtx> requireCtx;
 
     // private methods are meant for internal calls, so these don't lock targetMutex
     std::optional<Breakpoint> getBreakpointBySourceLineHelper(std::string source, int line) const;
--- a/lute/debug/include/lute/debuginternals.h
+++ b/lute/debug/include/lute/debuginternals.h
@@ -56,7 +56,6 @@ struct Target
     // breakpoint. Otherwise, it exists as a pending breakpoint until new sources are loaded.
     // We do this because clients may 1) configure breakpoints before launching executables
     // 2) we load sources dynamically with @require that a client may want to debug.
-    // TODO: implement 2
     //
     // Guarantees for when breakpoints are installed:
     // Any breakpoint that is placed when the target process is paused (including before launch) and that
--- a/lute/debug/src/debug.cpp
+++ b/lute/debug/src/debug.cpp
@@ -181,11 +181,10 @@ static int target_getLoadedSources(lua_State* L)
 {
     Target* target = getTarget(L, 1);
     std::vector<std::string> sources = target->getLoadedSources();
-    checkStack(L, 1);
+    checkStack(L, 2);
     lua_createtable(L, sources.size(), 0);
     for (int i = 0; i < (int)sources.size(); i++)
     {
-        checkStack(L, 1);
         lua_pushstring(L, sources[i].c_str());
         lua_rawseti(L, -2, i + 1);
     }
--- a/lute/debug/src/debug.cpp
+++ b/lute/debug/src/debug.cpp
@@ -175,6 +175,23 @@ static int target_getBreakpointBySourceLine(lua_State* L)
     return pushBreakpoint(L, *bp);
 }
 
+// target.getLoadedSources()
+// returns a table of strings
+static int target_getLoadedSources(lua_State* L)
+{
+    Target* target = getTarget(L, 1);
+    std::vector<std::string> sources = target->getLoadedSources();
+    checkStack(L, 1);
+    lua_createtable(L, sources.size(), 0);
+    for (int i = 0; i < (int)sources.size(); i++)
+    {
+        checkStack(L, 1);
+        lua_pushstring(L, sources[i].c_str());
+        lua_rawseti(L, -2, i + 1);
+    }
+    return 1;
+}
+
 static std::shared_ptr<Ref> getOptionalCallback(lua_State* L, int tableIndex, const char* field)
 {
     lua_getfield(L, tableIndex, field);
@@ -297,7 +314,7 @@ static int target_pauseProcess(lua_State* L)
     return 1;
 }
 
-// debugger.mewTarget()
+// debugger.newTarget()
 // returns Target
 static int debug_newTarget(lua_State* L)
 {
@@ -319,6 +336,7 @@ static const std::unordered_map<std::string, lua_CFunction> kTargetMethods = {
     {"launch", debug::target_launch},
     {"continueProcess", debug::target_continueProcess},
     {"pauseProcess", debug::target_pauseProcess},
+    {"getLoadedSources", debug::target_getLoadedSources},
 };
 
 static void initializeTarget(lua_State* L)
--- a/lute/debug/src/debuginternals.cpp
+++ b/lute/debug/src/debuginternals.cpp
@@ -45,12 +45,20 @@ Target::~Target()
     childRuntime.reset();
 }
 
-static std::string getChunkFromSource(std::string sourcePath)
+static std::string normalizeSeparators(std::string sourcePath)
+{
+    for (char& c : sourcePath)
+        if (c == '\\')
+            c = '/';
+    return sourcePath;
+}
+
+static std::string getChunkFromSource(const std::string& sourcePath)
 {
     return '@' + sourcePath;
 }
 
-static std::string getSourceFromChunk(std::string chunkname)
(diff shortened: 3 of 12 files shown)


lute/debug/src/debuginternals.cpp
```diff
@@ -260,7 +289,29 @@ bool Target::launch(const std::string& sourcePath, const std::vector<std::string
...
+        // Set up require system before launch.
+        requireCtx = std::make_unique<RequireCtx>(std::make_unique<RequireVfs>());
+        requireCtx->compileOptions.optimizationLevel = 1;
+        requireCtx->compileOptions.debugLevel = 2;
+        requireCtx->onLoad = [this](const std::string& chunkName, lua_State* ML)
+        {
+            std::unique_lock lock(targetMutex);
+            // this strips the potential leading @ from the chunkName for consistency when returning to DAP
+            loadedSources[getSourceFromChunk(chunkName)] = std::make_shared<Ref>(ML, -1);
+            auto [installed, uninstalled] = modifyPendingBreakpoints(ML);
```
wizhao96: i feel like this should be refactored into some like `onStopped` helper function since it appears a lot but also im not sure of the wisdom of having a function that requires a lock and then unlocks it

lute/debug/src/debuginternals.cpp
```diff
@@ -42,6 +45,18 @@ Target::~Target()
     childRuntime.reset();
 }
```
wizhao96, line 48: the require system uses chunks but our `Target` should use plain source paths, since debuggers generally want that

lute/debug/include/lute/debuginternals.h
```diff
@@ -47,12 +49,16 @@ struct Target
     explicit Target(Runtime& parentRuntime);
     ~Target();
 
+    // Get list of sources, with sources using forward slahes consistently.
+    // Our principle in path format is that we accept any path format as input but will
+    // internally use and then output with paths that use exclusively forward slashes.
```
wizhao96, line 54: lmk if this is not the right principle here

lute/debug/include/lute/debuginternals.h
```diff
@@ -67,7 +73,7 @@ struct Target
     std::optional<Breakpoint> getBreakpointBySourceLine(std::string source, int line) const;
 
     // For actively running scripts:
-    bool launch(const std::string& sourcePath, const std::vector<std::string>& args, LaunchConfig config = {});
+    bool launch(std::string sourcePath, const std::vector<std::string>& args, LaunchConfig config = {});
```
wizhao96, line 76: it turns out either way we need to make a copy of this so it's no longer a constant reference. and this matches what we're doing with `getBreakpointBySourceLine` + other functions anyways

lute/debug/src/debuginternals.cpp
```diff
@@ -42,9 +45,30 @@ Target::~Target()
...
 
+static std::string normalizeSeparators(std::string sourcePath)
+{
+    for (char& c : sourcePath)
+        if (c == '\\')
+            c = '/';
+    return sourcePath;
+}
```
Vighnesh-V: Luau/FileUtils.h implements a path normalization function that you should lean on here. Works cross platform too.

lute/debug/src/debuginternals.cpp
```diff
@@ -251,19 +276,56 @@ std::pair<std::vector<Breakpoint>, std::vector<Breakpoint>> Target::modifyPendin
...
+std::vector<std::string> Target::getLoadedSources()
+{
+    std::lock_guard lock(targetMutex);
+    std::vector<std::string> sources;
+    sources.reserve(loadedSources.size());
+    for (auto& [path, _] : loadedSources)
+    {
+        // we discard the empty sentinel value for loadedSources
+        if (path != "")
+            sources.emplace_back(path);
```
Vighnesh-V: can we avoid having a sentinel here?

lute/debug/src/debug.cpp
```diff
@@ -175,6 +175,23 @@ static int target_getBreakpointBySourceLine(lua_State* L)
...
+    checkStack(L, 1);
+    lua_createtable(L, sources.size(), 0);
+    for (int i = 0; i < (int)sources.size(); i++)
+    {
+        checkStack(L, 1);
+        lua_pushstring(L, sources[i].c_str());
+        lua_rawseti(L, -2, i + 1);
+    }
+    return 1;
+}
```
Vighnesh-V: You can just checkStack here just once (for size two). I think because you're always pushing one auxiliary thing and popping it immediately

Vighnesh-V (changes requested):

lute/debug/src/debuginternals.cpp
wizhao96: yep we can

lute/debug/src/debuginternals.cpp
```diff
@@ -42,9 +46,22 @@ Target::~Target()
...
 }
 
+static std::string getChunkFromSource(const std::string& sourcePath)
+{
+    return '@' + sourcePath;
+}
+
+static std::string getSourceFromChunk(const std::string& chunkname)
+{
+    if (chunkname.size() > 0 && chunkname[0] == '@')
```
Vighnesh-V: can be replaced with chunkname.rfind(pos = 0, '@') ( I don't remember the exact function invokcation)

lute/debug/src/debuginternals.cpp
```diff
@@ -251,19 +269,52 @@ std::pair<std::vector<Breakpoint>, std::vector<Breakpoint>> Target::modifyPendin
...
 
-bool Target::launch(const std::string& sourcePath, const std::vector<std::string>& args, LaunchConfig config)
+std::vector<std::string> Target::getLoadedSources()
+{
+    std::lock_guard lock(targetMutex);
+    std::vector<std::string> sources;
+    sources.reserve(loadedSources.size());
+    for (auto& [path, _] : loadedSources)
+        sources.emplace_back(path);
+    return sources;
```
Vighnesh-V, line 279: if loadedSources owns the path strings, can this function just give out a string view instead, to avoid copying data around.

lute/debug/src/debuginternals.cpp
```diff
@@ -279,8 +330,11 @@ bool Target::launch(const std::string& sourcePath, const std::vector<std::string
         std::string bytecode = Luau::compile(source, debugOptions);
         lua_State* thread = lua_newthread(childRuntime->GL);
         luaL_sandboxthread(thread);
-        luau_load(thread, sourcePath.c_str(), bytecode.c_str(), bytecode.size(), 0);
+
+        std::string chunkname = getChunkFromSource(sourcePath);
+        luau_load(thread, chunkname.c_str(), bytecode.c_str(), bytecode.size(), 0);
         loadedSources[sourcePath] = std::make_shared<Ref>(thread, -1);
+
         std::tie(installedBps, uninstalledBps) = modifyPendingBreakpoints(thread);
```
Vighnesh-V, line 345: nit, could this just be:
`auto& [installedBps, uninstalledBps] = ..`

lute/require/include/lute/require.h
```diff
@@ -42,4 +44,7 @@ struct RequireCtx
     RequireCtx(std::unique_ptr<IRequireVfs> vfs);
 
     std::unique_ptr<IRequireVfs> vfs;
+
+    Luau::CompileOptions compileOptions = copts();
```
Vighnesh-V: Can you initialize this in the constructor as well.

lute/debug/src/debuginternals.cpp
```diff
@@ -279,8 +330,11 @@ bool Target::launch(const std::string& sourcePath, const std::vector<std::string
         std::string bytecode = Luau::compile(source, debugOptions);
         lua_State* thread = lua_newthread(childRuntime->GL);
         luaL_sandboxthread(thread);
-        luau_load(thread, sourcePath.c_str(), bytecode.c_str(), bytecode.size(), 0);
+
+        std::string chunkname = getChunkFromSource(sourcePath);
+        luau_load(thread, chunkname.c_str(), bytecode.c_str(), bytecode.size(), 0);
```
Vighnesh-V: what is this luau_load line doing here again? Is there a reason you aren't checking the result of luau_load here?

lute/debug/src/debuginternals.cpp
Vighnesh-V: I could see that, but I don't think it's the end of the world

lute/debug/src/debuginternals.cpp
```diff
@@ -251,19 +269,52 @@ std::pair<std::vector<Breakpoint>, std::vector<Breakpoint>> Target::modifyPendin
...
-        setupState(*childRuntime, nullptr);
+        // Set up require system before launch.
+        requireCtx = std::make_unique<RequireCtx>(std::make_unique<RequireVfs>());
+        requireCtx->compileOptions.optimizationLevel = 1;
+        requireCtx->compileOptions.debugLevel = 2;
+        requireCtx->onLoad = [this](const std::string& chunkName, lua_State* ML)
+        {
+            std::unique_lock lock(targetMutex);
+            // this strips the potential leading @ from the chunkName for consistency when returning to DAP
+            loadedSources[getSourceFromChunk(chunkName)] = std::make_shared<Ref>(ML, -1);
```
Vighnesh-V: I guess the access to `loadedSources` must be protected, but can you compute the source from chunk before this?

lute/debug/src/debuginternals.cpp
```diff
@@ -251,19 +269,52 @@ std::pair<std::vector<Breakpoint>, std::vector<Breakpoint>> Target::modifyPendin
...
+            auto [installed, uninstalled] = modifyPendingBreakpoints(ML);
+            lock.unlock();
```
Vighnesh-V: I think writing this using scoped_lock is much more clear. The explicit lock.unlock() is a bit noisy.

lute/require/include/lute/require.h
```diff
@@ -42,4 +44,7 @@ struct RequireCtx
...
+    std::function<void(const std::string& chunkName, lua_State* L)> onLoad;
```
Vighnesh-V: should swap argument order here. Maybe rename to `onChunkLoad`?

lute/debug/src/debuginternals.cpp
```diff
@@ -251,19 +269,52 @@ std::pair<std::vector<Breakpoint>, std::vector<Breakpoint>> Target::modifyPendin
...
         LUTE_ASSERT(!launched);
         if (launched)
             return false;
         childRuntime = std::make_unique<Runtime>(parentRuntime.reporter, true);
-        setupState(*childRuntime, nullptr);
+        // Set up require system before launch.
+        requireCtx = std::make_unique<RequireCtx>(std::make_unique<RequireVfs>());
+        requireCtx->compileOptions.optimizationLevel = 1;
+        requireCtx->compileOptions.debugLevel = 2;
+        requireCtx->onLoad = [this](const std::string& chunkName, lua_State* ML)
```
Vighnesh-V: Can you add a constructor to requirectx for this?

Vighnesh-V (changes requested):

lute/debug/src/debuginternals.cpp
wizhao96: i don't think so b/c loadedSources can rehash and then invalidate your `string_view` if you add more sources

lute/debug/src/debuginternals.cpp
wizhao96: this change and the `scoped_lock` change go against each other because by creating a new scope we need to declare `installedBps` and `uninstalledBps` outside of the lock. Then, if we have already preexisting vaariables, we have to use `std::tie` instead of `auto [...]`.

I've generally gone with using `scoped_lock` since I generally believe that's the more pressing issue.

lute/debug/src/debuginternals.cpp
wizhao96: so i was planning to check this in https://github.com/luau-lang/lute/pull/1251#discussion_r3648462903 but let's just move that to this PR then.

`luau_load` pushes the compiled code onto the top of `thread`'s stack so that it can be run. i've added a comment saying that we should surface compilation errors since `luau_load` pushes the error to the top of the stack if it is an error.

lute/require/include/lute/require.h
```diff
@@ -39,7 +41,13 @@ class IRequireVfs
 
 struct RequireCtx
 {
-    RequireCtx(std::unique_ptr<IRequireVfs> vfs);
+    RequireCtx(
+        std::unique_ptr<IRequireVfs> vfs,
+        Luau::CompileOptions compileOptions = copts(),
```
Vighnesh-V: no default arguments for this please.Probaly add a second constructor that lets you initialize this stuff for the debugger specifically

Vighnesh-V (changes requested):
one last change sorry!

Vighnesh-V (approved):

PR #1250, opened 2026-07-24
`lute/runtime`: fix scheduling issue with blocking on uv event loop

wizhao96:
Fixes scheduled continuations and callbacks blocking on `uv_run(getEventLoop(), UV_RUN_ONCE)` in `runOnce`.  
* When testing the DAP adapter, I was seeing behavior where callbacks such as `onExit` would not fire until after receiving I/O from the VSCode debugging client. 
* This uses `uv_async` to attempt to fix that.
* Empirically it fixes the DAP problem above.

### DIFF
--- a/lute/runtime/include/lute/runtime.h
+++ b/lute/runtime/include/lute/runtime.h
@@ -154,6 +154,7 @@ private:
     std::condition_variable runLoopCv;
     uv_thread_t runLoopThread = {};
     bool runLoopThreadStarted = false;
+    uv_async_t wakeup;
 
     std::atomic<int> activeTokens;
     uv_loop_t eventLoop;
--- a/lute/runtime/include/lute/runtime.h
+++ b/lute/runtime/include/lute/runtime.h
@@ -139,6 +139,7 @@ struct Runtime
 
     // for debug mode only:
     const bool debugMode;
+    std::atomic<int> numLaunchedDebuggees = 0;
     void stopDebug();
     void continueDebug();
     // Same as scheduleLuauCallback but we since we don't call this within a libuv completion callback
--- a/lute/runtime/include/lute/runtime.h
+++ b/lute/runtime/include/lute/runtime.h
@@ -141,6 +141,9 @@ struct Runtime
     const bool debugMode;
     void stopDebug();
     void continueDebug();
+    // Same as scheduleLuauCallback but we since we don't call this within a libuv completion callback
+    // we need to wake up the libuv event loop.
+    void scheduleDebugLuauCallback(std::shared_ptr<Ref> callbackRef, std::function<int(lua_State*)> argPusher);
 
 private:
     bool runThreadCompletionHandler(lua_State* L, int status);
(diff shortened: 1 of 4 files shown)


lute/runtime/src/runtime.cpp
```diff
@@ -32,8 +32,18 @@ Runtime::Runtime(LuteReporter& reporter, bool debugMode)
 
     if (uv_loop_init(&eventLoop) < 0)
     {
-        LUTE_ASSERT("Couldn't initialize runtime event loop");
```
wizhao96, line 35: this did not do anything before since "Couldn't initialize runtime event loop" is a truthy value. i'm not sure if that was intentional or not?

Vighnesh-V: o. You should leave the assert in and prepend the message with a !

lute/runtime/src/runtime.cpp
```diff
@@ -273,6 +290,9 @@ void Runtime::schedule(std::function<void()> f)
     continuations.push_back(std::move(f));
 
     runLoopCv.notify_one();
+    // We may be blocked on uv_run(getEventLoop(), UV_RUN_ONCE) in runOnce().
+    // This signals the uv_loop and unblocks it to run the continuation.
+    uv_async_send(&wakeupEventLoop);
```
Vighnesh-V: This feels strange to me. if a continuation doesn't schedule any new work to run on the event loop, then it shouldn't be yielding?? I think what you want is a way to signal from your containing process to the child process to wake up the event loop. Does that make sense?

Vighnesh-V (changes requested):

wizhao96:
realistically i'm wishing for a way to keep some of this `uv_async` logic locked behind `debugMode` but we can't since we shouldn't force the parent runtime to be in debug mode (consider if we had someone write a script using the package, then the runtime running that package would also have to be in `debugMode` somehow)

lute/runtime/src/runtime.cpp
```diff
@@ -338,6 +351,14 @@ void Runtime::scheduleLuauCallback(std::shared_ptr<Ref> callbackRef, std::functi
     runLoopCv.notify_one();
 }
 
+void Runtime::scheduleDebugLuauCallback(std::shared_ptr<Ref> callbackRef, std::function<int(lua_State*)> argPusher)
```
Vighnesh-V, line 354: make sure to assert here that we are in debug mode!

Vighnesh-V (approved):
one comment, but should be fine!

lute/runtime/src/runtime.cpp
```diff
@@ -338,6 +351,14 @@ void Runtime::scheduleLuauCallback(std::shared_ptr<Ref> callbackRef, std::functi
...
+{
+    scheduleLuauCallback(callbackRef, argPusher);
```
Vighnesh-V: std::move argPusher?

Vighnesh-V (commented):
one more comment ...

lute/runtime/include/lute/runtime.h
```diff
@@ -139,6 +139,7 @@ struct Runtime
 
     // for debug mode only:
     const bool debugMode;
+    std::atomic<int> numLaunchedDebuggees = 0;
```
wizhao96, line 142: this is an int b/c theoretically we could have multiple `Target` objects in one script. it's a little bit unlikely but if we can check for it ig we should.

PR #1251, opened 2026-07-24
`lute/debug`: add prototype DAP adapter for breakpoints

wizhao96:
A basic implementation of a DAP adapter that can handle breakpoints, continuing, and pausing
* Also changes `onBreakpointInstall` to signal when installation fails as well

Screenshot 2026-07-24 at 3 22 39 PM
(Note: line highlighting is provided by stack frames, which is why it's stuck at line 1 above. really the important thing to look at is the breakpoints bar in the bottom left, where we can see that we've hit a bp)

### DIFF
--- a/definitions/debugger.luau
+++ b/definitions/debugger.luau
@@ -18,7 +18,8 @@ export type Breakpoint = {
 export type LaunchConfig = {
 	--- called when a breakpoint `bp` is hit; execution is paused until continueProcess() is called
 	onBreakpointHit: ((bp: Breakpoint) -> ())?,
-	--- called when breakpoint `bp` is successfully installed in the VM
+	--- called when there is has been an attempt to install the breakpoint `bp` on the VM, regardless
+	--- of success and failure. check the status of `bp` to see if the attempt has succeeded or failed.
 	onBreakpointInstall: ((bp: Breakpoint) -> ())?,
 	--- called when a breakpoint `bp` is uninstalled from the VM
 	onBreakpointUninstall: ((bp: Breakpoint) -> ())?,
--- a/lute/debug/src/debuginternals.cpp
+++ b/lute/debug/src/debuginternals.cpp
@@ -72,10 +72,10 @@ Breakpoint Target::setBreakpoint(std::string sourcePath, int line)
     // they are scheduled for pending installs.
     if (childRuntime && paused)
     {
-        bool installed = installBreakpoint(childRuntime->GL, it->second);
+        installBreakpoint(childRuntime->GL, it->second);
         Breakpoint bpCopy = it->second;
         lock.unlock();
-        if (installed && launchConfig.onBreakpointInstall)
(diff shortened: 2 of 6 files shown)


tests/cli/debug.test.luau
```diff
@@ -107,6 +119,60 @@ test.suite("LuteDebug", function(suite)
 		dapStream.dapLoop(reader, write)
 		checkEquivalence(getWritten())
 	end)
+	-- tests specifically for DAP set breakpoints format
```
wizhao96: it's actually quite hard to write end-to-end tests in this format because the DAP adapter is async. our `chunkReader` and `chunkWriter` are not yielding and generally can't support the asynchronous input/output. thus, we can't for example test `onBreakpointHit` and the subsequent `continueProcess` easily

definitions/debugger.luau
```diff
@@ -17,7 +17,8 @@ export type Breakpoint = {
 export type LaunchConfig = {
 	--- called when a breakpoint `bp` is hit; execution is paused until continueProcess() is called
 	onBreakpointHit: ((bp: Breakpoint) -> ())?,
-	--- called when breakpoint `bp` is successfully installed in the VM
+	--- called when there is has been an attempt to install the breakpoint `bp` on the VM, regardless
+	--- of success and failure. check the status of `bp` to see if the attempt has succeeded or failed.
```
wizhao96, line 22: we change this definition so that a IDE can actually see when a bp fails to install for whatever reason

lute/debug/src/debuginternals.cpp
```diff
@@ -279,7 +280,11 @@ bool Target::launch(const std::string& sourcePath, const std::vector<std::string
         std::string bytecode = Luau::compile(source, debugOptions);
         lua_State* thread = lua_newthread(childRuntime->GL);
         luaL_sandboxthread(thread);
-        luau_load(thread, sourcePath.c_str(), bytecode.c_str(), bytecode.size(), 0);
+        if (luau_load(thread, sourcePath.c_str(), bytecode.c_str(), bytecode.size(), 0) != 0)
```
wizhao96: this checks for the file actually compiling or not

afujiwara-roblox (commented):
Looks pretty good. Could we add some tests for all of the commands. I think we're missing `continue`, `pause`, `loadedSources`, `disconnect`, and `unknown`.

wizhao96:
yea I think generally it's generally hard to test `pause`, `loadedSources`, and `continue` due to https://github.com/luau-lang/lute/pull/1251#discussion_r3648284310 (basically we can't tell in our fixture when callbacks happen).

but i've updated tests to do what we can do easily such as `disconnect` and `unknown`.

afujiwara-roblox (approved):

wizhao96:
actually we can test `loadedSources` so i've added that in just now. but still `pause` and `continue` are hard to test

PR #1252, opened 2026-07-24
`lute/fs`: implement `fs.readIntoBuffer` API

annieetang:
to support partial reads into buffers and more, this PR introduces a new fs API 

```lua
--- Reads up to `numBytes` bytes from `handle` at `fileOffset` into `buf` starting at `bufferOffset`.
--- Returns the number of bytes actually read (may be less than `numBytes` if EOF is reached).
--- - `bufferOffset` defaults to 0.
fs.readIntoBuffer(handle: FileHandle, buf: buffer, fileOffset: number, numBytes: number, bufferOffset: number?): number
```


definitions/fs.luau
```diff
@@ -77,6 +77,19 @@ function fs.close(handle: FileHandle): ()
...
 end
 
+--- Reads up to `count` bytes from `handle` at `fileOffset` into `buf` starting at `bufferOffset`.
+--- Returns the number of bytes actually read (may be less than `count` if EOF is reached).
+--- - `bufferOffset` defaults to 0.
+function fs.readIntoBuffer(
+	handle: FileHandle,
+	buf: buffer,
+	fileOffset: number,
+	count: number,
```
Vighnesh-V:
```suggestion
numBytes: number,
```

lute/fs/src/fs_impl.cpp
```diff
@@ -607,6 +607,19 @@ struct FSCopyDirectory
     }
 };
 
+struct FSReadIntoBuffer : FSRequest
```
Vighnesh-V, line 610: should this extend FSReadRequest? It looks like it's doing many of the same things?

Vighnesh-V (changes requested):

lute/fs/src/fs_impl.cpp
```diff
@@ -766,6 +779,48 @@ int close_impl(lua_State* L, UVFile* handle)
...
+    r->succeed(
+        [bytesRead](lua_State* L)
+        {
+            lua_pushnumber(L, static_cast<double>(bytesRead));
+            return 1;
+        }
+    );
+}
+
+int readIntoBuffer_impl(lua_State* L, UVFile* handle, void* buf, size_t bufLen, int64_t fileOffset, size_t count, size_t bufferOffset)
```
skberkeley: why is `fileOffset` signed?

lute/fs/src/fs_impl.h
```diff
@@ -54,4 +56,5 @@ int mkdir_impl(lua_State* L, const char* path, int mode);
 int rmdir_impl(lua_State* L, const char* path);
 int listdir_impl(lua_State* L, const char* path);
 
+
```
skberkeley: ?

lute/fs/src/fs_impl.cpp
annieetang: was following the `uv_fs_read` API https://docs.libuv.org/en/v1.x/fs.html#c.uv_fs_read

lute/fs/src/fs_impl.cpp
annieetang: talked offline and `FSRead` and `FSReadIntoBuffer` are pretty different from each other so I wanted to keep them separate. tldr is bc FSRead loops until EOF without knowing the total size upfront, while FSReadInfoBuffer is a fixed size single read call into an external buffer

tests/std/fs.test.luau
```diff
@@ -424,4 +424,88 @@ test.suite("FsSuite", function(suite)
...
+
+		local success = pcall(function()
+			fs.readIntoBuffer(h, readBuf, 0, 2, 3)
+		end)
+		assert.eq(success, false)
+
+		success = pcall(function()
+			fs.readIntoBuffer(h, readBuf, -1, 1)
+		end)
+		assert.eq(success, false)
```
Vighnesh-V: isn't there an assert.that?

Vighnesh-V (changes requested):
nit on tests, but ow looks fine

tests/std/fs.test.luau
annieetang: o true ty

Vighnesh-V (approved):

PR #1255, opened 2026-07-26
`lute format`: CLI scaffolding

jkelaty-rbx:
This PR implements a minimal CLI surface for `lute format` which is currently implemented as a no-op.

The "format" implementation is stubbed as a no-op for now.


tests/cli/format.test.luau
```diff
@@ -0,0 +1,122 @@
...
+		assert.eq(result.exitcode, 0)
+		assert.strContains(result.stdout, "of 1 file(s)")
+
+		fs.removeDirectory(dir, { recursive = true })
+	end)
+
+	suite:case("leaves a file byte-identical when the formatter reports no change", function(assert)
+		local dir = makeDir("format_noop")
+		local source = "local   x=1\nprint(  x  )\n"
+		local target = writeFile(childPath(dir, "messy.luau"), source)
```
skberkeley: i'm not a huge fan of this, it's not obvious looking at this line that `writeFile` is returning the result of `childPath`, and someone new working on this file might not know to use this pattern

tests/cli/format.test.luau
```diff
@@ -0,0 +1,122 @@
...
+
+	suite:case("errors when a path holds no Luau files", function(assert)
+		local dir = makeDir("format_empty")
+		writeFile(childPath(dir, "notes.txt"), "not luau\n")
+
+		local result = process.run({ lutePath, "format", dir })
+		assert.eq(result.exitcode, 1)
+		assert.strContains(result.stdout, "no Luau files found")
+
+		fs.removeDirectory(dir, { recursive = true })
```
skberkeley: can we use `case:beforeEach` and `case:afterEach` to handle creating and tearing down a test directory instead? `afterEach` has the guarantee that it'll run even if the test case itself errors, whereas here `fs.removeDirectory` won't run if any of the asserts fail

tests/cli/format.test.luau
```diff
@@ -0,0 +1,122 @@
...
+	end)
+
+	suite:case("accepts -c as an alias for --check", function(assert)
+		local dir = makeDir("format_alias")
+		local target = writeFile(childPath(dir, "clean.luau"), "local x = 1\n")
+
+		local result = process.run({ lutePath, "format", "-c", target })
+		assert.eq(result.exitcode, 0)
```
skberkeley: can you leave a todo noting that this test should be updated once `--check` isn't a no-op?

lute/cli/src/climain.cpp
skberkeley, line 1: Can you note that `lute format` is experimental or in beta or something until it's actually implemented?

lute/cli/commands/format/init.luau
```diff
@@ -0,0 +1,105 @@
+local cli = require("@batteries/cli")
+local fs = require("@std/fs")
+local path = require("@std/path")
+local process = require("@std/process")
+
+local files = require("./lib/files")
+
+local HELP_MESSAGE = [[Usage: lute format [options] [files...]
```
skberkeley: did you try using `cli.help` to generate the help message?

tests/cli/format.test.luau
skberkeley, line 1: can you also add a test checking that the contents of a `.gitignore` file are respected?

lute/cli/commands/format/init.luau
```diff
@@ -0,0 +1,105 @@
...
+		print(HELP_MESSAGE)
+		return
+	end
+
+	local check = args:has("check")
+	local verbose = args:has("verbose")
+
+	local inputPaths = args:variadics()
+	if #inputPaths == 0 then
+		inputPaths = { path.format(process.cwd()) }
```
skberkeley: is there a reason why we format here if we're formatting later anyway?

lute/cli/commands/format/init.luau
```diff
@@ -0,0 +1,105 @@
...
+			continue
+		end
+
+		local ok, formatted = pcall(formatSource, source :: string)
+		if not ok then
+			table.insert(failed, `{target}: {formatted}`)
+			continue
+		end
+
+		if formatted ~= source then
```
skberkeley, line 76: this comparison seems potentially expensive, maybe `formatSource` should return a bool as well indicating whether the source was modified?

lute/cli/commands/format/init.luau
```diff
@@ -0,0 +1,105 @@
...
+	end
+
+	local sourceFiles = files.getSourceFiles(inputPaths, verbose)
+	if #sourceFiles == 0 then
+		print("Error: no Luau files found")
+		process.exit(1)
+	end
+
+	local failed: { string } = {}
+	local unformatted: { string } = {}
```
skberkeley: could you give this a better name? if i'm reading this correctly, this contains files that are formatted if we're not in check mode

lute/cli/commands/format/init.luau
```diff
@@ -0,0 +1,105 @@
...
+			if #unformatted == 0 then
+				print(`All {#sourceFiles} file(s) are formatted.`)
+			else
+				print(`Checked {#sourceFiles} file(s)`)
+				print(`{#unformatted} file(s) were not formatted`)
+			end
+		else
+			print(`Formatted {#unformatted} of {#sourceFiles} file(s).`)
+		end
+	end
```
skberkeley, line 98:
```suggestion
	if verbose then
		if check then
			print(`Checked {#sourceFiles} file(s).`)
			if #unformatted == 0 then
				print(`All {#sourceFiles} file(s) were formatted properly.`)
			else
				print(`{#unformatted} file(s) were not formatted properly.`)
			end
		else
			print(`Formatted {#unformatted} of {#sourceFiles} file(s).`)
		end
	end
```

lute/cli/commands/format/init.luau
```diff
@@ -0,0 +1,105 @@
...
+
+	if check and #unformatted > 0 then
+		print(`\n{#unformatted} file(s) not formatted:\n{table.concat(unformatted, "\n")}`)
```
skberkeley:
```suggestion
		print(`\n{#unformatted} file(s) not formatted:\n\t- {table.concat(unformatted, "\n\t- ")}`)
```
i think this should present as a bulleted list?

tests/cli/format.test.luau
```diff
@@ -0,0 +1,122 @@
...
+		local dir = makeDir("format_walk")
+		fs.createDirectory(childPath(dir, "nested"))
+		writeFile(childPath(dir, "top.luau"), "local x = 1\n")
+		writeFile(childPath(childPath(dir, "nested"), "deep.luau"), "local y = 2\n")
+		writeFile(childPath(dir, "ignored.txt"), "not luau\n")
+
+		local result = process.run({ lutePath, "format", "--verbose", dir })
+		assert.eq(result.exitcode, 0)
+		-- The count proves the walk descended into `nested` and skipped the .txt.
+		assert.strContains(result.stdout, "of 2 file(s)")
```
skberkeley: This comment is misleading, can you print `Trying to format <filename>` under verbose mode and check that instead?

skberkeley (changes requested):
thanks for taking a run at this, i'm excited to see `lute format` take shape!

Vighnesh-V:
@jkelaty-rbx Would you be able to convert these tests to snapshot tests instead? Testing cli commands (and updating their output in tests) in a fast changing repo can be a bit of the pain, and if you take the time to set this up now, it'll probably make testing later much easier. It also makes changes a lot easier to review because we can diff actual .luau artifacts. You might have to write yourself a little test harness, but once you've done that, it'll make the rest of testing much faster. There are some examples of these tests for the lints, and you can run `lute test snapshots` to run all of them.

lute/cli/commands/format/init.luau
jkelaty-rbx: I think this should be O(1) for any string size due to luau's string interning

lute/cli/commands/format/init.luau
jkelaty-rbx: Not particularly, I removed the extra format

lute/cli/commands/format/init.luau
jkelaty-rbx: Updated to use cli.help/usage

batteries/cli.luau
skberkeley, line 1: Can I ask that you split these changes out into a separate PR?

lute/cli/commands/format/format.luau
```diff
@@ -0,0 +1,10 @@
+local stringext = require("@std/stringext")
+
+local function formatSource(source: string): string
+	local trimmed = stringext.trim(source)
+	return if trimmed == "" then trimmed else `{trimmed}\n`
```
skberkeley: Wasn't this a no-op before? I'd prefer keeping it that way if this PR is just the scaffolding for the command

lute/cli/commands/format/init.luau
```diff
@@ -0,0 +1,157 @@
...
+local configs = require("./lib/configs")
+local files = require("./lib/files")
+local parseIgnores = require("./lib/parseIgnores")
+
+local DESCRIPTION = [[
+
+Format Luau source files. Experimental: formatting is still under development, and its output will change.
+]]
+
+local function findConfiguredIgnores(): parseIgnores.GitignoreData?
```
skberkeley: this and `filterIgnored` seem like they could go in `commands/lib` somewhere

lute/cli/commands/format/init.luau
```diff
@@ -0,0 +1,157 @@
...
+	for _, sourcePath in sourceFiles do
+		local target = path.format(sourcePath)
+
+		if verbose then
+			print(`Trying to format {target}`)
+		end
+
+		local read, source = pcall(fs.readFileToString, sourcePath)
+		if not read then
+			table.insert(failed, `{target}: {source}`)
```
skberkeley, line 63:
```suggestion
            if verbose then
                print(`Read of {target} failed`)
            end
			table.insert(failed, `{target}: {source}`)
```
just to make potential debugging easier

lute/cli/commands/lib/configs.luau
```diff
@@ -0,0 +1,38 @@
...
+			end
+		end
+
+		return nil :: any, nil :: any
+	end
+end
+
+return {
+	walkConfigPaths = walkConfigPaths,
+}
```
skberkeley: could you do:
```
local configLib = {}
function configLib.walkConfigPaths ...
return table.freeze(configLib)
```
It makes it easier to tell what the export surface of the file is, especially as it grows

lute/cli/commands/format/format.luau
jkelaty-rbx: If this is a no-op then I cannot meaningfully test some parts of the scaffolding, such as --check mode

batteries/cli.luau
jkelaty-rbx: I can but for what purpose?

skberkeley: They could exist independently of your other changes and limiting the scope of the PR makes it easier to review

jkelaty-rbx: https://github.com/luau-lang/lute/pull/1281

.github/workflows/ci.yml
```diff
@@ -206,6 +206,9 @@ jobs:
       - name: Run Lute Lint
         run: ${{ steps.build_lute.outputs.exe_path }} lint -v .
 
+      - name: Run Lute Format
+        run: ${{ steps.build_lute.outputs.exe_path }} format -c -v .
+
```
skberkeley: nit: this might make more sense as its own job

lute/cli/commands/format/init.luau
```diff
@@ -0,0 +1,113 @@
...
+		print(args:usage("lute format"))
+		print(DESCRIPTION)
+		print(args:help())
+		return
+	end
+
+	const check = args:has("check")
+	const verbose = args:has("verbose")
+
+	local inputPaths = args:variadics()
```
skberkeley, line 34: do we need to freeze `args:variadics()`?

lute/cli/commands/lib/configs.luau
```diff
@@ -0,0 +1,77 @@
...
+function configLib.walkConfigPaths(startDir: string?): () -> (string, { [any]: any })
+	local current = path.format(path.normalize(startDir or process.cwd()))
+
+	return function(): (string, { [any]: any })
+		while current do
+			const candidate = path.format(path.join(current, ".config.luau"))
+
+			const parent = path.format(path.dirname(current))
+			if parent == current then
+				current = nil
```
skberkeley, line 19: does this only happen if we've hit the root, or are there other cases?

lute/cli/commands/lib/configs.luau
```diff
@@ -0,0 +1,77 @@
...
+		elseif typeof(luteConfig) ~= "table" then
+			error(`{configPath}: lute must be a table`, 0)
+		end
+
+		const commandConfig = luteConfig[command]
+		if commandConfig == nil then
+			continue
+		elseif typeof(commandConfig) ~= "table" then
+			error(`{configPath}: lute.{command} must be a table`, 0)
+		end
```
skberkeley, line 54: @wmccrthy fyi

tests/cli/format.test.luau
skberkeley, line 1: I might be misreading, but is there duplication between some of the tests here and the snapshot tests?

lute/cli/commands/format/init.luau
jkelaty-rbx: I feel like freezing probably should happen at the producer (inside cli.luau) not at the consumer

jkelaty-rbx: I don't think we need to scope any more cli changes in this PR

lute/cli/commands/lib/configs.luau
jkelaty-rbx: This can happen if there are no `.config.luau` files in the cwd's hierarchy

jkelaty-rbx:
@skberkeley can this be approved? I didn't see any comments that would justify not approving this PR

skberkeley (approved):

lute/cli/commands/lib/configs.luau
wmccrthy: thx for the notice, I think this should be perfectly compatible with the existing lint config structure. I also see we're not using this module yet in the lute lint code, which seems like the right call and easier to address + review in a follow up

wizhao96:
sorry this is my flaky test :( just feel free to re-run

tests/cli/format.test.luau
jkelaty-rbx: I don't think there's much duplication. I see at least the `--check` exit code is tested in both the suite and the unformatted snapshot, but the snapshot is there for the (verbose) stdout, and the suite case is there to assert the file isn't rewritten, so they technically test for different properties

jkelaty-rbx: The `lute format --help` snapshot is also different from the the top-level `lute --help` that's in this file

PR #1258, opened 2026-07-26
`lute format`: Doc IR builders and renderer

jkelaty-rbx:


lute/cli/commands/format/lower.luau
jkelaty-rbx, line 1: This file should be the focus for the majority of the rest of the implementation

jkelaty-rbx: This list should get burned down as the lowering stage is implemented: https://github.com/luau-lang/lute/pull/1311/changes#diff-af7eb1e476713bed4a4d440fccbde40c44cd27e8f68fb6b0603f873abf951c48

PR #1263, opened 2026-07-28
`lute/debug`: inspect threads

wizhao96:
Adds thread/coroutine inspection for multiple co-routines (part 1/3 of PRs dealing with inspecting into state)
* Creates data structures to keep track of threads getting created and destroyed while working with bps
* Make sure `onPause` and `onBreakpointHit` take in threadId, keeping in align with DAP protocol
* Add `getStoppedThread`, `getThreads` and `getMainThread` for inspection
* (maybe?) replace `task.spawn()` behavior with `task.defer()` in debug mode to allow for correct pausing
* switch everything to use 1-indexing to match Luau

btw: i use threads and coroutines interchangeably even though they are semantically different since we have `Thread` objects tracking coroutines

Screenshot 2026-07-31 at 11 36 06 AM

### DIFF
--- a/lute/debug/include/lute/debuginternals.h
+++ b/lute/debug/include/lute/debuginternals.h
@@ -86,8 +86,9 @@ struct Target
     std::optional<Breakpoint> getBreakpointBySourceLine(std::string source, int line) const;
 
     // For inspection while paused:
-    // About multiple coroutines: we don't currently handle task.spawn() but do handle task.defer().
-    // The difference is that task.spawn() tries to run the coroutine inline with our current execution, resulting
+    // About multiple coroutines: we don't currently handle the original implementation of task.spawn(). Calling
+    // task.spawn() in the debugger calls task.defer() instead.
+    // The difference is that the original task.spawn() tries to run the coroutine inline with our current execution, resulting
     // in issues when trying to pause. Similar methods that also run coroutines inline with our current execution will not work.
     // In contrast, in task.defer(), the coroutine is added to set of running coroutines but execution
     // is deferred. Our pause mechanism will then work correctly.
@@ -110,7 +111,7 @@ private:
     int currentBreakpointId = 0;
     bool paused = true;
     bool launched = false;
-    std::unordered_map<int, Breakpoint> breakpoints; // breakpoint id -> breakpoint object (this is unordered_map to support erase)
+    std::unordered_map<int, Breakpoint> breakpoints;    // breakpoint id -> breakpoint object (this is unordered_map to support erase)
     std::unordered_set<lua_State*> continueRequestedBp; // if the thread's lua_State* is in this set, we skip the next bp it hits
     std::optional<Breakpoint> bpHit;
     LaunchConfig launchConfig;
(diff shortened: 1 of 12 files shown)


lute/task/src/task.cpp
```diff
@@ -185,6 +185,11 @@ struct LuaThread
 int lua_spawn(lua_State* L)
 {
     LuaThread newThread{L, "task.spawn"};
+    // In debug mode, if we run new thread inline, our pausing mechanism
```
wizhao96, line 188: i'm not sure we want to do this. im also fine with reverting this change and basically saying somewhere "the debugger has only limited multi-coroutine ability and doesn't support task.spawn. please use task.defer if you want debugger to work"

lute/debug/include/lute/debuginternals.h
```diff
@@ -107,6 +127,11 @@ struct Target
     // for require contexts
     std::unique_ptr<RequireCtx> requireCtx;
 
+    // thread information
+    int threadId = 0;
```
wizhao96: lmk if you would prefer this one indexed

wizhao96: actually i literally just think one-indexing is better so i'll change that

wizhao96: i've also changed breakpoint ids to start at 1. there's a lot of reasons to 1-index for a Luau debugger so let's just keep that consistent

lute/debug/include/lute/debuginternals.h
```diff
@@ -88,11 +108,11 @@ struct Target
     Runtime& parentRuntime;
     std::unique_ptr<Runtime> childRuntime;
 
-    int currentBreakpointId = 0;
+    int currentBreakpointId = 1;
```
wizhao96, line 112: see https://github.com/luau-lang/lute/pull/1263#discussion_r3678633939

lute/debug/include/lute/debuginternals.h
```diff
@@ -107,6 +128,11 @@ struct Target
     // for require contexts
     std::unique_ptr<RequireCtx> requireCtx;
 
+    // thread information
+    int threadId = 1;
+    std::unordered_map<lua_State*, Thread> stateToThread; // lua_State* -> thread information about that state
```
wizhao96, line 133: this set of maps might seem a little strange but basically you can find a `lua_State*` from thread id and then use that to find a `Thread` object itself. this way if you have one of a `lua_State*`, thread id, or `Thread` object you can always find any other necessary piece of information.

`Thread` objects don't store `lua_State*` because that should not be exposed to users.

afujiwara-roblox (approved):

lute/task/src/task.cpp
Vighnesh-V: task.spawn has this behaviour in engine (it runs the thread until it yields and then queues it). We wanted to preserve that for compatibility. An equivalent thing we could do, is enqueue it onto the threads vector, but in front of all the other threads. That way, on the next iteration of the runtime main loop, this thread will get picked up first and execute.

Vighnesh-V (commented):
@wizhao96 You could address this in a follow up I think. Instead of `defer` you could add a `deferFront` :)

PR #1264, opened 2026-07-29
Handle tables with __iter metamethod in `deepClone`

wmccrthy:
`tableext.deepClone` failed to correctly clone tables with `__iter` metamethods that didn't perform conventional key, value iteration. I discovered this while cloning the CST returned by lute's `parser` APIs, and found that `CstPunctuated` nodes were missing the `separators` field due to their `__iter` method.

This PR updates `deepClone`'s iteration over a table to use `pairs`, such that all key, value pairs of cloned tables are handled.

### DIFF
--- a/lute/std/libs/tableext.luau
+++ b/lute/std/libs/tableext.luau
@@ -84,7 +84,7 @@ function deepClone<T>(tbl: T, seen: { [T]: T }, opts: DeepCloneOptions): T
 
 	seen[tbl] = clone
 
-	for k, v in pairs(tbl :: any) do
+	for k, v in pairs(tbl) do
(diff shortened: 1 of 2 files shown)


lute/std/libs/tableext.luau
```diff
@@ -84,7 +84,7 @@ function deepClone<T>(tbl: T, seen: { [T]: T }, opts: DeepCloneOptions): T
 
 	seen[tbl] = clone
 
-	for k, v in tbl :: any do
+	for k, v in next, tbl :: any do
```
wmccrthy:
```suggestion
	for k, v in pairs(tbl :: any) do
```

skberkeley (approved):

PR #1267, opened 2026-07-30
`lute/debug`: add stepping

wizhao96:
Incorporate stepping to `lute/debug`
* Add capabilities for `stepIn`, `stepOver`, and `stepOut`
* Also add `getLine()` to streamline testing

https://github.com/user-attachments/assets/2916d967-50b7-43ea-b7de-d9703d780e6b


wizhao96:
so in reality this does have some dependency on the thread code since you can step on an individual thread, so let's get that threading code in first.

definitions/debugger.luau
```diff
@@ -33,6 +43,8 @@ export type LaunchConfig = {
 	onExit: ((success: boolean) -> ())?,
 	--- called when a pauseProcess() interrupts `thread` and then actually stops all threads from execution.
 	onPause: ((thread: Thread) -> ())?,
+	--- called when our script finishes stepping on `thread`. `info` reflects the information about our stepping command.
+	onStepStop: ((thread: Thread, info: StepInfo) -> ())?,
```
wizhao96, line 49: ok so `StepInfo` in the C++ code already contains `thread` for implementation convenience sake but I separate them out on the Luau code to make the definitions of callbacks in the Luau code consistent

lmk if we should just merge everything together in Luau to make it `onStepStop: ((info: StepInfo) -> ())`

lute/cli/commands/debug/dap.luau
```diff
@@ -181,7 +203,7 @@ local function dapLoop(reader: dapStream.Reader, writer: dapStream.Writer)
 						id = 0,
 						name = "main",
 						source = createSource(launchPath.program),
-						line = 1,
+						line = target:getLine(),
```
raymondnumbergenerator, line 206: wait isn't this a bug? we're always returning `launchPath` as the source in the stackframe so this is gonna point to the correct line but the wrong file if you break outside of the entry file

wizhao96: this is kind of a temporary measure as https://github.com/luau-lang/lute/pull/1280 fully addresses this. in reality the normal flow of information for DAP would be to do something like `getStackFrame` for both source and line information.

i added `getLine()` since it's useful for testing and might be something that users in `lute/debug` actually want.  i am using it in this PR for line information but this isn't intended to be permanent

wizhao96: realistically maybe we want something like `getStoppedInfo()` instead of `getLine()` that returns both source and line number though

raymondnumbergenerator: nevermind i just read the other PR, looks like that's dealt with in the actual stackframes pr

raymondnumbergenerator: > realistically maybe we want something like `getStoppedInfo()` instead of `getLine()` that returns both source and line number though

but ya that sounds sensible to me, maybe makes more sense to do it in the stacktrace PR?

raymondnumbergenerator (approved):

PR #1275, opened 2026-08-03
`lute/debug`: surface compile errors on launch

wizhao96:
instead of returning `true/false`, surface compile errors on launch to the user

this is what it looks like:
Screenshot 2026-08-03 at 9 46 46 AM


tests/src/debug/compile_error.luau
```diff
@@ -0,0 +1 @@
+local _x = )
```
wizhao96, line 1: this is a `.txt` to avoid CI being angry (otherwise it fails `stylua`)

wizhao96:
we should prob also find a way to surface runtime errors but that might rely on a different pathway with like `output` events. or maybe that gets done with exception breakpoints later

tests/src/debug/compile_error.luau
afujiwara-roblox: You can add this file to the toplevel .styluaignore or use inline comment directives to have stylua ignore this file

afujiwara-roblox (approved):

PR #1276, opened 2026-08-03
infra: enable CI for merge groups

aatxe:
We've got some interest in trying out a merge queue for lute, which means that we need to modify our CI to actually _run_ in the event that we have a merge queue going. Otherwise, everything will just fail immediately.


Vighnesh-V (approved):
hopefully this helps with merge times :D

PR #1280, opened 2026-08-04
`lute/debug`: add stack traces

wizhao96:
Adds stack frames/stack traces (part 2/3 of PRs dealing with inspecting into state)
* Adds `getStackDepth`, `getStackFrame`, and `getStackTrace` to debugger
* Modify `getLine` to `getStoppedLocation` to include source path
* Allows for delayed stack loading in DAP
* In DAP by supporting stack frames we allow for line highlighting

https://github.com/user-attachments/assets/ea426af3-2952-4141-803f-0797ce2f2f37

### DIFF
--- a/definitions/debugger.luau
+++ b/definitions/debugger.luau
@@ -99,7 +99,7 @@ export type Target = {
 	--- Returns `true` if target was paused before stepping and `false` otherwise.
 	stepOut: (self: Target, threadId: number) -> boolean,
 	--- Gets our current stopped location, consisting of (sourcePath: string, line: number)
-	--- Returns nil if we are not stopped.
+	--- Returns (nil, nil) if we are not stopped.
 	getStoppedLocation: (self: Target) -> (string?, number?),
 	--- Gets a list of all Luau coroutines that are being used by the script.
 	getThreads: (self: Target) -> { Thread },
--- a/definitions/debugger.luau
+++ b/definitions/debugger.luau
@@ -20,6 +20,16 @@ export type Thread = {
 	name: string,
 }
 
+--- A stack frame of the execution. New stack frames are recreated every time execution is paused, and
+--- ids are reused.
+export type StackFrame = {
+	id: number,
+	name: string,
+	sourcePath: string,
+	line: number,
+	column: number,
+}
+
 --- A set of callbacks that we configure before launching our debugger
 export type LaunchConfig = {
 	--- called when a breakpoint `bp` is hit on `thread`; execution is paused until continueProcess() is called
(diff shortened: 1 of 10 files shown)


lute/debug/src/debuginternals.cpp
```diff
@@ -422,8 +422,9 @@ void Target::installBpHitCallback()
             return;
         }
         lua_Debug info = {};
-        lua_getinfo(L, 0, "sl", &info);
-        int line = info.currentline;
+        lua_getinfo(L, 0, "s", &info);
```
wizhao96, line 445: this is redundant with some code in the stepping PR. i'll deal with merge conflicts later.

tests/src/debug/recursion.luau
```diff
@@ -0,0 +1,14 @@
+local b: (n: number) -> number
```
wizhao96, line 1: there are two separate tests, where `pow.luau` is tied to the Luau test and has a simpler test while `recursion.luau` is tied to C++ and is more comprehensive.

definitions/debugger.luau
```diff
@@ -88,15 +98,25 @@ export type Target = {
 	--- Steps out of the current function call in thread id `threadId`.
 	--- Returns `true` if target was paused before stepping and `false` otherwise.
 	stepOut: (self: Target, threadId: number) -> boolean,
-	--- Gets our current line of execution when we are stopped.
-	--- Returns -1 if we are not stopped.
-	getLine: (self: Target) -> number,
+	--- Gets our current stopped location, consisting of (sourcePath: string, line: number)
```
wizhao96, line 101: this typing is quite awkward but I don't think we can do (string, number)? because then the unpacking logic doesn't work well if the function can return one or two values.

lute/cli/commands/debug/dap.luau
```diff
@@ -3,7 +3,10 @@ local dapTypes = require("./dapTypes")
 local debugger = require("@lute/debugger")
 local json = require("@std/json")
 
-local function createSource(sourcePath: string): dapTypes.Source
+local function createSource(sourcePath: string): dapTypes.Source?
+	if sourcePath == "" or sourcePath == "=[C]" then
```
wizhao96, line 7: the dap display breaks if our sourcePath is =[C] if we're in a C function. this resolves that by just forcing the VSCode adapter to display an unknown source.

lute/debug/include/lute/debuginternals.h
```diff
@@ -160,19 +173,28 @@ struct Target
     std::unordered_map<lua_State*, Thread> stateToThread; // lua_State* -> thread information about that state
     std::unordered_map<int, lua_State*> threadIdToState;  // thread id -> lua_State*
 
+    // stack frame information
+    // note: stack frames are copies between these two data structures, not pointers. That's ok because the debugger
+    // should never modify the stack frames themselves.
+    // stack frame ID information is reset upon every continue(). The base id resets to 1 as well.
+    int stackframeId = 1;
+    std::unordered_map<int, std::unordered_map<int, StackFrame>> stateToStackFrame; // thread id -> level -> stackFrame
```
wizhao96, line 181: realistically this is kinda of janky but `pair` has no built in hash function and I want to stick to `unordered_map` over `map`

afujiwara-roblox (approved):

Issue #1282, opened 2026-08-05
Bug: Parser mis-attaches closing `)` as trailing trivia on type aliases with parenthesized return packs


Gravemind1142:
### Summary

When parsing a type alias whose return type is a parenthesized type pack, `@std/syntax` leaves the final closing `)` outside the typed AST. It is attached as trailing trivia on the `typealias` statement node and incorrectly tagged as `whitespace`.

This breaks consumers that remove or replace the statement node and then re-print/re-parse the source, because the stray `)` remains in the output.

### Minimal reproduction

```luau
local luau = require("@std/syntax")
local printer = require("@std/syntax/printer")
local trivia = require("@std/syntax/utils/trivia")

local source = "export type resource<A..., R...> = (A...) -> (dispose, R...)\n"
local parseResult = luau.parse(source)

for _, st in parseResult.root.statements do
	if st.tag == "typealias" then
		print(printer.printNode(st))
		for _, t in trivia.rightmostTrivia(st) do
			print(t.tag, string.format("%q", t.text))
		end
	end
end
```

### Actual behavior

```
export type resource<A..., R...> = (A...) -> (dispose, R...)
whitespace ")\n"
```

- `printer.printNode(st)` includes the full source text, including the final `)`.
- `trivia.rightmostTrivia(st)` contains a trivia item with:
  - `tag = "whitespace"`
  - `text = ")\n"`

So the closing `)` of `(dispose, R...)` is not represented as a normal token inside the type AST. It is orphaned as trailing trivia on the statement node.

### Expected behavior

The closing `)` should be part of the parsed type structure for the parenthesized return pack `(dispose, R...)`, not emitted separately as trailing trivia on the `typealias` node.

At minimum:

- `trivia.rightmostTrivia(typealias)` should contain only actual whitespace/newline trivia (e.g. `"\n"`), not syntax tokens.
- Replacing/removing the `typealias` node and re-printing should not leave a dangling `)`.

### Impact

Any tool that:

1. Parses Luau with `@std/syntax`
2. Replaces whole AST nodes (e.g. to strip types)
3. Re-prints via the visitor/printer and re-parses

can produce invalid Luau from otherwise valid source.

PR #1287, opened 2026-08-10
Fast tracks bugfixes from Luau 0.734+735 to unblock internal users of Lute

### LINKED ISSUE #1237: Closing ) is left behind when modifying a function type annotation
Replacing the function type annotation in the following Luau source with an empty string leaves the final closing parenthesis:
`local test: () -> (number, ...number)`
The output becomes:
`local test )`

### Reproduction
Input:
`local test: () -> (number, ...number)`
Replace the relevant node with ""
Actual output:
`local test )`
Expected output:
`local test `
**Used require("@std/syntax/printer").printFile(parserResult, replacements) to transform the code.**

When replacing it with:
`{ content = "", preserveTrivia = false }`
the final ) is correctly removed

### Expected behavior
The closing ) should be included in the replaced node’s source range.

### Tested with
Lute version: 1.0.1-nightly.20260717
Operating system: macOS
Minimal input: local test: () -> (number, ...number)

Vighnesh-V:
This PR fast tracks a fix made in Luau 0.734 to unblock internal users of Lute. The bug in question is caused by type aliases for functions with pack returns having extraneous trivia that should actually be part of the pack, not the overall statement. This caused issues because replacing type packs would not replace the parenthesis that closed the pack.

The bulk of this PR is implementing a patching system into tune files, that allows for the application of one or more patches to each of the dependencies of lute. I've added this change as one patch that's been concatenated from two smaller subpatches because the changes are related and mostly in the same place. 

Resolves #1237 and #1282.


tools/luthier.luau
```diff
@@ -21,6 +21,8 @@ type Tune = {
 	},
 }
 
+local patches = { "luau-patch-1.patch" }
```
aatxe: A static list of patches in this file and a folder of patches doesn't seem like the right way to go about implementing this to me. I'd expect to see the patches defined in the toml (tune) files for each dependency instead, and for the patches themselves to just live in `extern/` or at least `extern/patches`.

aatxe (changes requested):

tools/luthier.luau
Vighnesh-V: Good call. I did it this way originally because of the bootstrap script - I wanted to keep it more uniform between the two since I don't foresee us wanting to patch boringssl. I've updated the tune files to implement this, but the bootstrap script will just unconditionally apply the patches to extern/luau directly.

extern/luau.tune
```diff
@@ -3,3 +3,4 @@ name = "luau"
 remote = "https://github.com/luau-lang/luau.git"
 branch = "0.733"
 revision = "ca128af4c531310d6f5c1b354df4b79fdd782ede"
+patches = ["luau-001.patch"]
```
aatxe:
```suggestion
patches = ["luau-001.patch"]

```

trailing new line

tools/luthier.luau
```diff
@@ -703,6 +718,23 @@ local function areTuneFilesUpToDate(): boolean
...
+	for _, file in files do
+		if string.find(file.name, "%.tune$") then
+			local dep: Tune = readTuneFile(projectRelative("extern", file.name))
+			local patches = dep.dependency.patches
+			if not patches then
+				continue
+			end
+			assert(patches)
+			for _, patch in patches do
+				applyLuauPatch(`extern/{patch}`)
```
```suggestion
				check(applyLuauPatch(`extern/{patch}`))
```

I think we want to call check here to bubble up patch failures, right?

aatxe: Alternatively, we could make this whole function return a status code like the others and then `check(applyPatches())` later, which would be more consistent.

Vighnesh-V: I ended up doing the latter!

aatxe (approved):

PR #1288, opened 2026-08-11
`lute/debug`: inspect variables

wizhao96:
Enable variable + scope inspection for `lute/debug`. Part 3/3 of a series of PRs on inspection.

* Generally, you take a stack frame and first drill into a `VariableScope` (a scope represents something like a group of local variables or upvalue variables)
* Each scope has its own `variableReference` and you can get a set of `Variable` objects per scope.
* For tables, the `Variable` objects themselves may have a `variableReference`, allowing for additional lazy inspection of deeper levels of the table

https://github.com/user-attachments/assets/f0346606-3961-429d-92c5-111a4f4fc8aa


wizhao96:
whoops i accidentally closed and reopened this

tests/src/debug/variables.luau
```diff
@@ -0,0 +1,14 @@
+local a = 12
```
afujiwara-roblox, line 1: Can we check really big and small numbers too. Maybe really precise float points to see that the conversions work well

tests/src/debug/variables.luau
```diff
@@ -0,0 +1,14 @@
...
+local _b = 1.2
+local _c = "test"
```
afujiwara-roblox, line 3: Do we handle escape characters properly? Something like "\nsdas" should be printed as is, not as a new line.

afujiwara-roblox (approved):
Looks good to me. In general I think this PR tests the happy paths well, but it could cover edges more comprehensively

tests/src/debug/variables.luau
wizhao96: both the comments about escape chars and precision had problems and have been fixed

PR #1289, opened 2026-08-12
fix: requiring a bare directory without init file should not cause ambiguity

### LINKED ISSUE #1279: Bug: `resolveModule` treats directories without `init` as module candidates, causing false ambiguity
## Summary

On case-insensitive filesystems (e.g. default macOS APFS), `luau.resolveModule` fails with `(ambiguous)` when a **file module** and a **sibling directory** share the same case-insensitive basename — even if the directory has no `init.luau` / `init.lua` and therefore is not a valid module target.

Because exact-case disambiguation is not reliably available on case-insensitive filesystems, the resolver should not treat bare directories (those without an init file) as competing module candidates.

## Environment

- **OS:** macOS (darwin), case-insensitive APFS (default)
- **API:** `luau.resolveModule(path, fromchunkname)` via `@std/luau` / `@lute/luau`

## Minimal reproduction

Project layout:

```
project/
├── root.luau
├── foo.luau
└── Foo/
    └── bar.luau
```

`root.luau`:

```luau
local a = require("./foo")
local b = require("./Foo")
```

Repro script:

```luau

local root = path.format(path.join("project", "root.luau"))

for _, req in { "./foo", "./Foo", "./Foo/bar" } do
    local ok, result = pcall(luau.resolveModule, req, root)
    print(req, ok, result)
end
```

**Actual output:**

```
./foo       false   ... could not resolve child component "foo" (ambiguous)
./Foo       false   ... could not resolve child component "Foo" (ambiguous)
./Foo/bar   false   ... could not resolve child component "Foo" (ambiguous)
```

## Expected behavior

A directory should only be considered a **module candidate** if it contains `init.luau` or `init.lua`. A bare directory — one with child modules but no init — is not a valid target for `require("./Dir")`.

| Require path | Expected result |
|---|---|
| `./foo` | Resolve to `foo.luau` (sole valid module candidate) |
| `./Foo` | **Fail** — `Foo/` has no init file, so it is not a module |
| `./Foo/bar` | Resolve to `Foo/bar.luau` |

This layout should **not** be ambiguous. `foo.luau` is the only valid module at that basename; `Foo/` is a directory container, not a module.

## Why this matters

On case-insensitive filesystems, a file and directory with differently-cased names can coexist. A common layout is a lowercase file module alongside a PascalCase package directory containing submodules. Today, **all** requires at that basename fail — including `./foo` → `foo.luau`, which has exactly one valid target.

Exact-case disambiguation is not a viable fix on these filesystems, but requiring an init file to treat a directory as a module candidate is — and it matches runtime behavior, where `require("./Foo")` only works when `Foo/init.luau` exists.

## Suggested fix

When enumerating module candidates for a path component in a directory:

1. Always consider file modules: `{name}.luau`, `{name}.lua`.
2. Only consider a **directory** as a module candidate if it contains `init.luau` or `init.lua`.
3. Do not treat a bare directory (no init) as competing with a sibling file module.
4. Reserve `(ambiguous)` for cases where multiple **valid module targets** actually exist (e.g. both `{name}.luau` and `{name}/init.luau`).

With those rules, the reproduction above would yield:

- `./foo` → `foo.luau` ✅
- `./Foo` → clear failure (no init), not ambiguous ✅
- `./Foo/bar` → `Foo/bar.luau` ✅

annieetang:
fixes https://github.com/luau-lang/lute/issues/1279

When a file like `foo.luau` has a sibling directory named `foo/`, we previously reported an ambiguity error unconditionally. But if the directory had no `init.luau` or `init.lua` file, we shouldn't be reporting an ambiguity error because we know that we aren't requiring the `foo/` directory directly. 

This fix defers the ambiguity check until after we scan the directory for init files, so now if we `require(foo)` with:
* `foo.luau` + `foo/` with no init = require resolves to `foo.luau` (and we can still require any files inside the directory)
* `foo.luau` + `foo/` with `init.luau` results in an ambiguous error

### DIFF
--- a/lute/require/src/modulepath.cpp
+++ b/lute/require/src/modulepath.cpp
@@ -132,17 +132,17 @@ ResolvedRealPath ModulePath::getRealPath() const
             }
         }
 
-        if (hasInit)
-        {
-            if (resolvedType)
-                return {NavigationStatus::Ambiguous};
+        // Path is a directory with an init file but there is also a sibling file with the same name (e.g. foo.luau and foo/init.luau), so this is ambiguous.
+        if (hasInit && resolvedType)
+            return {NavigationStatus::Ambiguous};
 
-            resolvedType = ResolvedRealPath::PathType::File;
-        }
-        else if (!resolvedType)
-        {
+        // Path is a directory with no init file and no sibling file, so we know this is a directory.
+        if (!resolvedType)
             resolvedType = ResolvedRealPath::PathType::Directory;
-        }
+
+        // Path is a directory with an init file and no sibling file, so we know this is a file (the init file).
+        if (hasInit)
+            resolvedType = ResolvedRealPath::PathType::File;
     }
 
     if (!resolvedType)
--- a/lute/require/src/modulepath.cpp
+++ b/lute/require/src/modulepath.cpp
@@ -119,7 +119,6 @@ ResolvedRealPath ModulePath::getRealPath() const
     if (isADirectory(partialRealPath))
     {
         bool hasInit = false;
-        std::string_view initSuffix;
 
         for (std::string_view potentialSuffix : kInitSuffixes)
         {
@@ -129,7 +128,7 @@ ResolvedRealPath ModulePath::getRealPath() const
                     return {NavigationStatus::Ambiguous};
 
                 hasInit = true;
-                initSuffix = potentialSuffix;
+                suffix = potentialSuffix;
             }
         }
 
@@ -139,7 +138,6 @@ ResolvedRealPath ModulePath::getRealPath() const
                 return {NavigationStatus::Ambiguous};
 
             resolvedType = ResolvedRealPath::PathType::File;
-            suffix = initSuffix;
         }
         else if (!resolvedType)
         {
--- a/lute/require/src/modulepath.cpp
+++ b/lute/require/src/modulepath.cpp
@@ -118,23 +118,33 @@ ResolvedRealPath ModulePath::getRealPath() const
 
     if (isADirectory(partialRealPath))
     {
-        if (resolvedType)
-            return {NavigationStatus::Ambiguous};
+        bool hasInit = false;
+        std::string_view initSuffix;
 
         for (std::string_view potentialSuffix : kInitSuffixes)
         {
             if (isAFile(partialRealPath + std::string(potentialSuffix)))
             {
-                if (resolvedType)
+                if (hasInit)
                     return {NavigationStatus::Ambiguous};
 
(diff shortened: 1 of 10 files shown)


lute/require/src/modulepath.cpp
```diff
@@ -118,23 +118,33 @@ ResolvedRealPath ModulePath::getRealPath() const
...
             if (isAFile(partialRealPath + std::string(potentialSuffix)))
             {
-                if (resolvedType)
+                if (hasInit)
                     return {NavigationStatus::Ambiguous};
 
-                resolvedType = ResolvedRealPath::PathType::File;
-                suffix = potentialSuffix;
+                hasInit = true;
+                initSuffix = potentialSuffix;
```
Vighnesh-V: nit: you don't need to set or store initSuffix at all? you could just say suffix = initSuffix, since hasInit is true here. If you end up having multiple init.luau / init.lua, you would return Ambiguous?

Vighnesh-V (changes requested):
I think the testing is a bit sparse in this PR. Can you also add:
1) The exact directory structure as mentioned in #1279, paying attention to the casing.
2) a std/luau.test.luau test for resolveModule

lute/require/src/modulepath.cpp
annieetang: yea fair point, fixing!

lute/require/src/modulepath.cpp
```diff
@@ -118,23 +118,31 @@ ResolvedRealPath ModulePath::getRealPath() const
...
+        {
+            if (resolvedType)
+                return {NavigationStatus::Ambiguous};
+
+            resolvedType = ResolvedRealPath::PathType::File;
+        }
+        else if (!resolvedType)
+        {
             resolvedType = ResolvedRealPath::PathType::Directory;
+        }
```
Vighnesh-V: I think this entire block might be written in a slightly confusing way. Suppose partial real path is: `/path/to/dir`.

Then right before this if you know the following:
1) if resolvedType is not nil, it must be File
2) the directory /path/to/dir exists.

I think a slightly better way to write this is the following:
```
if (hasInit && resolvedType)
    return Ambiguous

if (!resolvedType)
    // At this point you know dir/init.luau doesn't exists, dir/ exists and dir.luau is empty
   resolvedType = directory

if (hasInit)
    // At this point you know dir/init.luau exists, dir/ exists and resolvedType is not true
    resolvedType = file

```
Might be worth commenting as well to explain this.

Vighnesh-V (changes requested):
One more logic change that might make this more readable?

Vighnesh-V (approved):

PR #1290, opened 2026-08-12
`lute/debug`: evaluate expressions

wizhao96:
Adds ability for the debugger to evaluate expressions within a certain stack frame

* Spawns a new thread, injects the relevant globals, local variables and upvalues, and then evaluates the result
* Any errors are reported back as a string
* Enables watch pane, debug console, and hovers 

https://github.com/user-attachments/assets/c2d3603d-6249-4718-9537-fb57eb529c21


definitions/debugger.luau
```diff
@@ -166,6 +166,10 @@ export type Target = {
 	--- `scopeType` should be `locals` or `upvalues`, not `table`.
 	--- Return `nil` if script is running or if such variables can not be found.
 	getVariablesByScopeType: (self: Target, frameId: number, scopeType: VariableScopeType) -> { Variable }?,
+	--- Evaluates an expression within the scope of a given stack frame id.
+	--- If `frameId` is -1 or absent, evaluate in the global scope.
```
afujiwara-roblox, line 174: when might be frameId be -1 or absent?

lute/debug/src/debuginternals.cpp
```diff
@@ -1009,6 +1005,143 @@ std::optional<std::vector<Variable>> Target::getVariablesByScopeType(int frameId
...
+        {
+            cb->userthread = savedUserthread;
+            lua_gc(global, LUA_GCRESTART, 0);
+        }
+    };
+    Luau::CompileOptions debugOptions;
+    debugOptions.optimizationLevel = 1;
+    debugOptions.debugLevel = 2;
+    std::string bytecode = Luau::compile("return " + expression, debugOptions);
+    CallbackGuard callbackGuard(childRuntime->GL);
```
Vighnesh-V: curly braces here pls `{}`.

definitions/debugger.luau
wizhao96: i think generally it should be -1 or absent if you want to evaluate in the global scope without any care for local variables, but generally i added it for compatibility with https://microsoft.github.io/debug-adapter-protocol/specification#Requests_Evaluate

afujiwara-roblox: Gotcha, generally I'm not a huge fan of null values having implicit behavior that's different, but we should follow the spec here

afujiwara-roblox (approved):

PR #1291, opened 2026-08-13
`lute/debug`: exception breakpoints

wizhao96:
Enables exception breakpoints for the debugger
* "Uncaught exceptions" will catch runtime errors and pause the runtime, allowing for inspection.
* "Caught exceptions" will catch caught exceptions 

https://github.com/user-attachments/assets/e61e701b-b454-4caf-a385-e81223014298

### DIFF
--- a/lute/debug/include/lute/debuginternals.h
+++ b/lute/debug/include/lute/debuginternals.h
@@ -174,7 +174,7 @@ struct Target
     std::optional<Breakpoint> getBreakpointBySourceLine(std::string source, int line) const;
 
     // Exception breakpoints are mostly separate from normal breakpoints. They have negative
-    // breakpoint IDs for DAP purposes. 
+    // breakpoint IDs for DAP purposes.
     ExceptionBreakpointInfo setExceptionBreakpoint(bool caught, bool uncaught);
 
     // For inspection:
@@ -288,7 +288,14 @@ private:
 
     void computeStoppedLocation(lua_State* L);
     void unsetStoppedLocation();    
-    void stoppedSetState(lua_State* L);
+    // Controls how stoppedSetState decides between the yielding and non-yielding stop pathways.
+    enum class StopYieldMode
+    {
+        Auto,
+        ForceNoYield,
+        Neither,
+    };
+    void stoppedSetState(lua_State* L, StopYieldMode yieldMode = StopYieldMode::Auto);
     void stoppedDispatchCallback(std::function<void()> debugStopCallback);
 
     Variable makeVariable(lua_State* L, const std::string& name);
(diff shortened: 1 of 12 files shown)


lute/debug/include/lute/debuginternals.h
```diff
@@ -102,7 +110,7 @@ struct Variable
     std::string value;
     std::string type;
     int variableReference = 0;
-    bool isTrue();
+    bool isTruthy();
```
wizhao96, line 116: i forgot to do this in the previous PR :( so here it is now

lute/debug/include/lute/debuginternals.h
```diff
@@ -48,6 +48,14 @@ struct Breakpoint
     Breakpoint(int id, std::string sourcePath, int line, BreakpointConfig config, BreakpointStatus status);
 };
 
+struct ExceptionBreakpointInfo
+{
+    bool uncaughtExceptions = false;
+    const int uncaughtId = -1;
```
wizhao96, line 56: negative numbers are viable dap adapter ids and are an easy way not to conflict with actual breakpoint ids. unfortunately, in dap, breakpoints and exceptions pretty much share the same id space

lute/debug/src/debuginternals.cpp
```diff
@@ -779,7 +847,7 @@ std::vector<Thread> Target::getThreads() const
     std::vector<Thread> result;
     for (auto& [L, thread] : stateToThread)
     {
-        if (lua_costatus(childRuntime->GL, L) != LUA_COFIN && lua_costatus(childRuntime->GL, L) != LUA_COERR)
+        if (L == stoppedThread || (lua_costatus(childRuntime->GL, L) != LUA_COFIN && lua_costatus(childRuntime->GL, L) != LUA_COERR))
```
wizhao96, line 864: this change is guided by once you have an uncaught exception the `lua_costatus` of the thread is LUA_COERR but we still want to show it in the display.

lute/debug/include/lute/debuginternals.h
Vighnesh-V: why are some of these fields randomy marked const? If possible, please also define a constructor here.

Vighnesh-V (commented):
just one comment, otherwise C++ changes look reasonable.

lute/debug/include/lute/debuginternals.h
wizhao96: at least for why the `caughtId` and `uncaughtId` are marked `const` i always want the ids to tied to these filters to stay the same. at least in dap, exception catching works where you can have certain exception filters. and users can't add filters but can toggle ones surfaced by the adapter on and off. so at least from my pov exception filters should have built-in immutable ids that shouldn't be changed by the users.

wizhao96: but i have added a constructor for `ExceptionBReakpointInfo`

Vighnesh-V (approved):

raymondnumbergenerator (approved):

PR #1298, opened 2026-08-17
`lute/debug`: add tracking for user-defined global variables

wizhao96:
This updates `lute/debug` to track user-defined global variables. We still don't track in-built globals because this clogs up the interface and generally these in-built globals are unlikely to be changed by the user.

https://github.com/user-attachments/assets/1863efd3-e42b-498b-b8e6-80fd599b9588

### DIFF
--- a/definitions/debugger.luau
+++ b/definitions/debugger.luau
@@ -46,7 +46,7 @@ export type VariableScopeType = "locals" | "upvalues" | "globals" | "table"
 
 --- A VariableScope contains the contextual information for a collection of Variables
 --- such as what variable reference ID number to refer to this collection, what type of collection
---- it is, etc. Variable reference IDs are reset upon continuing execution. 
+--- it is, etc. Variable reference IDs are reset upon continuing execution.
 --- Generally, VariableScope will have a type `locals`, `upvalues`, or `globals`. `table` is used in the C++ implementation
 --- but generally is not used in Luau code currently.
 export type VariableScope = {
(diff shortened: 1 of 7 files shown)


lute/debug/src/debuginternals.cpp
```diff
@@ -718,11 +723,23 @@ std::optional<std::vector<VariableScope>> Target::getScopesHelper(int threadId,
     VariableScope locals = VariableScope::makeLocals(variableRefId, threadId, level);
     variableContexts.insert_or_assign(variableRefId, locals);
     variableRefId++;
-    contexts.push_back(locals);
```
wizhao96, line 721: generally we should prefer `emplace_back` to `push_back` so this changes that across the codebase

definitions/debugger.luau
```diff
@@ -40,18 +40,22 @@ export type StackFrame = {
 	column: number,
 }
 
---- The type of a VariableScope (i.e. whether it contains local variables, upvalues, or
+--- The type of a VariableScope (i.e. whether it contains local variables, upvalues, globals, or
 --- values from a table)
-export type VariableScopeType = "locals" | "upvalues" | "table"
+export type VariableScopeType = "locals" | "upvalues" | "globals" | "table"
```
aatxe: Probably all of these should be made singular tbf, `local` scope, `upvalue` scope, `global` scope, `table` scope. Could be done in its own PR though.

aatxe (approved):
Looks good to me, just a comment about the scope names.

definitions/debugger.luau
wizhao96: ok i've changed the names for VariableScopeType in this PR but the actual display names are still plural (since it looks better on VSCode that way)

aatxe: Because VS Code is just rendering the names as entire section headings, right? That seems fine to me!

wizhao96: yea exactly

PR #1299, opened 2026-08-17
`lute/debug`: fix thread ordering

wizhao96:
when we continue execution, always restart execution from the currently stopped coroutine
* while the previous behavior wasn't objectively wrong, it was weird in the sense that pausing your code yielded your stopped coroutine, leading sometimes to counterintuitive results
* also fixes a bug in debug tests leading to flaky Windows tests

### DIFF
--- a/lute/debug/src/debuginternals.cpp
+++ b/lute/debug/src/debuginternals.cpp
@@ -1068,7 +1068,7 @@ void Target::continueProcessHelper()
                 continueRequestedBp.insert(stoppedThread);
             bpHit = std::nullopt;
         }
-        childRuntime->runningThreads.push_back({true, stoppedThreadRef, 0});
+        childRuntime->runningThreads.push_front({true, stoppedThreadRef, 0});
         // This schedule() wakes up the runtime in runContinuously() to re-run runToCompletion() in case that has exited. This is a no-op if
         // runToCompletion() has not exited.
         childRuntime->schedule([]() {});
--- a/lute/debug/src/debuginternals.cpp
+++ b/lute/debug/src/debuginternals.cpp
@@ -439,7 +439,7 @@ std::optional<std::string> Target::launch(std::string sourcePath, const std::vec
 
         scriptThread = thread;
         scriptThreadRef = getRefForThread(scriptThread);
-        childRuntime->runningThreads.push_back({true, scriptThreadRef, static_cast<int>(args.size())});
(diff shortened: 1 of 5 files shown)


lute/debug/src/debuginternals.cpp
```diff
@@ -1041,7 +1041,7 @@ void Target::continueProcessHelper()
                 continueRequestedBp.insert(stoppedThread);
             bpHit = std::nullopt;
         }
-        childRuntime->runningThreads.push_back({true, stoppedThreadRef, 0});
+        childRuntime->runningThreads.push_front({true, stoppedThreadRef, 0});
```
Vighnesh-V: we should have an emplace back/front now!

Vighnesh-V (changes requested):

lute/runtime/include/lute/runtime.h
```diff
@@ -30,6 +30,8 @@ struct ThreadToContinue
     std::shared_ptr<Ref> ref;
     int argumentCount = 0;
     std::function<void()> cont;
+    ThreadToContinue() = default;
```
wizhao96, line 33: i need constructors for ThreadToContinue to use emplace instead of push

tests/src/debug.test.cpp
```diff
@@ -510,7 +510,7 @@ TEST_SUITE("Debug")
         Breakpoint bp3 = target.setBreakpoint(fixturePath, 22);
 
         int maxThreadsSeen = 0;
-        int hitsBp1 = 0, hitsBp2 = 0;
+        int hitsBp1 = 0, hitsBp2 = 0, total_sum_val = 0;
```
Vighnesh-V: one per line pls!

tests/lute/debugger.test.luau
```diff
@@ -305,7 +305,7 @@ test.suite("LuteDebugLibrarySuite", function(suite)
 						if otherThread.id == 3 then
 							local otherTrace = assert(target:getStackTrace(otherThread.id))
 							local line = otherTrace[#otherTrace].line
-							check.that(line >= 12 and line <= 15)
+							check.that(line >= 12 and line <= 16)
```
wizhao96, line 308: im hopeful that these changes fixed the flaky windows ci since it has now passed 5 times in a row

Vighnesh-V (changes requested):

Vighnesh-V (approved):

PR #1300, opened 2026-08-19
`lute/debug`: enable stopping in non-yieldable contexts

wizhao96:
Enable stopping in non-yieldable contexts (such as metatable `__index` or inside error handlers)
* Our previous stopping mechanism required yielding our current coroutine in Lute, but this fails inside certain non-yieldable contexts, such as above
* Add an alternative stopping mechanism in these cases that relies on stopping the child runtime by waiting on a condition variable inside the runtime

let me know whether we want to proceed with this change. it's somewhat messy architecturally but solves real bugs we encountered in the bug bash, since previously these cases would result in the debugger crashing/behaving unexpectedly.

https://github.com/user-attachments/assets/8f4abe4c-37c2-4367-bc98-188cad9ffa9b

### DIFF
--- a/lute/debug/src/debuginternals.cpp
+++ b/lute/debug/src/debuginternals.cpp
@@ -497,7 +497,10 @@ void Target::installBpHitCallback()
             target->stoppedThreadRef = getRefForThread(L);
             // Clear out stepping when this happens.
             lua_callbacks(L)->debugstep = nullptr;
-            lua_break(L);
+            if (lua_isyieldable(L))
+                lua_break(L);
+            else
+                target->stoppedNoYield = true;
             auto [installed, uninstalled] = target->modifyPendingBreakpoints(target->scriptThread);
             debug::Thread thread = target->stateToThread.at(L);
             lock.unlock();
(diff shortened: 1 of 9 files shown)


lute/debug/src/debuginternals.cpp
```diff
@@ -497,7 +497,10 @@ void Target::installBpHitCallback()
             target->stoppedThreadRef = getRefForThread(L);
             // Clear out stepping when this happens.
             lua_callbacks(L)->debugstep = nullptr;
-            lua_break(L);
+            if (lua_isyieldable(L))
+                lua_break(L);
+            else
+                target->stoppedNoYield = true;
```
Vighnesh-V, line 503: I see this a few times - should this be a helper?

Vighnesh-V (approved):
one nit.

lute/debug/src/debuginternals.cpp
wizhao96: there's a `stoppedHelper()` function that is being implemented in https://github.com/luau-lang/lute/pull/1291 that deals with this and a bunch more duplicated stopping code. i'll fold it into that!

PR #1301, opened 2026-08-20
`lute/debug`: source breakpoints

wizhao96:
This allows additional capability to breakpoints, allowing it to conform to the DAP `sourceBreakpoint` specification
* It adds conditional breakpoints, hit condition breakpoints, and logpoints
* Conditional breakpoints only stop once a condition evaluates to non-false
* Hit conditional breakpoints stop according to an expression based on the number of times that bp has been hit
* Logpoints don't stop execution and instead log a statement to the debug console (this statement can do evaluation interpolation)

conditional bps:
https://github.com/user-attachments/assets/6702c791-7b7d-433d-95e1-453108e43730

hit conditional bps:
https://github.com/user-attachments/assets/f37a11f0-97e8-4777-afc2-4d1ca8daf033

logpoints:
https://github.com/user-attachments/assets/6fe8909f-15c0-464f-a7a5-36f342f3587f

### DIFF
--- a/definitions/debugger.luau
+++ b/definitions/debugger.luau
@@ -6,11 +6,18 @@ local debugger = {}
 export type BreakpointStatus = "pendingInstall" | "pendingUninstall" | "installed" | "invalid"
 
 --- A breakpoint that stops execution when encountered during runtime
+--- if condition is not "", the condition must evaluate to true before stopping.
+--- if hitCondition is not "", a condition involving the hitCount must be true before stopping.
+--- if logMessage is not "", then instead of stopping we simply output an interpolated log message to the relevant callback.
 export type Breakpoint = {
 	id: number,
 	line: number,
 	--- this source path will use forward slashes
 	sourcePath: string,
+	condition: string,
+	hitCondition: string,
+	hitCount: number,
+	logMessage: string,
 	status: BreakpointStatus,
 }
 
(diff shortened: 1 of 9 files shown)


definitions/debugger.luau
```diff
@@ -5,12 +5,32 @@ local debugger = {}
 --- operation is queued to happen when the script is next paused.
 export type BreakpointStatus = "pendingInstall" | "pendingUninstall" | "installed" | "invalid"
 
+--- A config for advanced breakpoint features
+export type BreakpointConfig = {
```
afujiwara-roblox, line 9: Do these types need to be nullable? The comments only suggest behavior for when the string is not "".

lute/debug/src/debuginternals.cpp
```diff
@@ -1207,9 +1329,7 @@ void Target::continueProcessHelper()
         childRuntime->schedule([]() {});
         stoppedThread = nullptr;
         stoppedThreadRef = nullptr;
-        stoppedLine = -1;
-        stoppedLocation = "";
-        stoppedPc = nullptr;
+        unsetStoppedLocation();
```
afujiwara-roblox, line 1387: curious why this is it's own function, but some of the other cleanup isn't

wizhao96: we need to run `computeStoppedLocation` so that we can run `evaluateExpression` accurately in conditional breakpoints. in the case we don't actually break, we clear that specific set of changes to state here https://github.com/luau-lang/lute/pull/1301/changes#diff-d230239784ea520392bd9e20de5c01ec4bd3baacf21ae78f9ac4add81f56e582R598. if we do break in that bp, we make additionally changes to state that we need to clear separately (and in addition to `unsetStoppedLocation`.

wizhao96: i mean i guess we can also just have the helper function do all the cleanup. then, we just do redundant work here https://github.com/luau-lang/lute/pull/1301/changes#diff-d230239784ea520392bd9e20de5c01ec4bd3baacf21ae78f9ac4add81f56e582R598 even though some of the values haven't actually been set

definitions/debugger.luau
wizhao96: yea i intend for these types to be nullable because I don't want to force the user to specify all config options when they really only want to set one option. i'll change the comment

lute/debug/src/debuginternals.cpp
```diff
@@ -76,6 +79,11 @@ VariableScope VariableScope::makeTable(int variableReference, int luaref)
     return VariableScope(variableReference, VariableScopeType::Table, "Table", -1, -1, luaref);
 }
 
+bool Variable::isTrue()
```
Vighnesh-V, line 82: nit: isTruthy - that's what it's called in Luau.

Vighnesh-V (approved):
cpp changes looks good, one nit. Can you get @raymondnumbergenerator to sign off on the .luau changes?

raymondnumbergenerator (approved):

PR #1303, opened 2026-08-20
`lute/debug`: set variable and set expression

wizhao96:
Implements `setVariable` and `setExpression` so the debugger can change values of variables during runtime
* refactors `evaluateExpressionHelper` to 1) be able to transfer variables to another thread and 2) return multiple (or no) return values
* refactors `getLocal, setLocal, injectLocal` and `getUpvalue, setUpvalue, injectUpvalue` to use a shared `forAllLocals` and `forAllUpvalues`

https://github.com/user-attachments/assets/fb24a07d-d03a-4b2c-be15-250cafbb3592


lute/debug/src/debuginternals.cpp
```diff
@@ -376,7 +377,7 @@ std::optional<std::string> Target::launch(std::string sourcePath, const std::vec
         childRuntime = std::make_unique<Runtime>(parentRuntime.reporter, true);
         // Set up require system before launch.
         Luau::CompileOptions debugOptions;
-        debugOptions.optimizationLevel = 1;
+        debugOptions.optimizationLevel = 0;
```
wizhao96, line 453: otherwise sometimes setVariable/setExpression doesn't work well with optimizationLevel = 1 with like different slots for upvalues and variables

lute/debug/src/debuginternals.cpp
```diff
@@ -66,14 +67,14 @@ VariableScope VariableScope::makeUpvalues(int variableReference, int threadId, i
     return VariableScope(variableReference, VariableScopeType::Upvalue, "Upvalues", threadId, level, -1);
 }
 
-VariableScope VariableScope::makeGlobals(int variableReference)
+VariableScope VariableScope::makeGlobals(int variableReference, int threadId, int level)
```
wizhao96, line 73: ok so there unfortunately has to be a different global `VariableScope` per stack frame so that we can set a global variable from an expression evaluated from the variables available to that stack frame.

lute/debug/src/debuginternals.cpp
```diff
@@ -66,14 +67,14 @@ VariableScope VariableScope::makeUpvalues(int variableReference, int threadId, i
...
 {
-    return VariableScope(variableReference, VariableScopeType::Global, "Globals", -1, -1, -1);
+    return VariableScope(variableReference, VariableScopeType::Global, "Globals", threadId, level, -1);
 }
 
-VariableScope VariableScope::makeTable(int variableReference, int luaref)
+VariableScope VariableScope::makeTable(int variableReference, int threadId, int level, int luaref)
```
wizhao96, line 78: Tables need to track which threadId and level they are in for a similar reason as above. We want to be able to set a value within a table from the variables available to that stack frame.

lute/debug/src/debuginternals.cpp
```diff
@@ -918,46 +921,76 @@ std::vector<Variable> Target::getLocalsHelper(lua_State* L, int level)
...
 }
 
 std::vector<Variable> Target::getGlobalsHelper()
 {
-    return getTableHelper(scriptThread, LUA_GLOBALSINDEX);
+    return getTableHelper(scriptThread, LUA_GLOBALSINDEX, -1);
 }
 
-std::vector<Variable> Target::getTableHelper(lua_State* L, int idx)
+std::vector<Variable> Target::getTableHelper(lua_State* L, int idx, int parentRef)
```
wizhao96, line 1236: we use `parentRef` to keep track of which threadId and level each table belongs to

lute/debug/src/debuginternals.cpp
```diff
@@ -1165,7 +1202,198 @@ EvaluateResult Target::evaluateExpression(std::string expression, int frameId)
...
+            if (std::holds_alternative<Variable>(var))
+                lua_setupvalue(L, -2, n);
+            variableCache.clear();
+            result = var;
+            return false;
+        }
+    );
+    return result;
+}
+
```
wizhao96, line 1515: ok so when we look up in the table, we sometimes get names like "[true]" or "[2]" and we need to 1) strip the brackets and 2) determine what actual type this key is to push onto the Luau thread

lute/cli/commands/debug/dap.luau
raymondnumbergenerator: btw i think you missed a rename here to `local`

lute/debug/src/debuginternals.cpp
```diff
@@ -867,43 +860,53 @@ static std::string printTable(lua_State* L, int idx, int levelsToPrint)
...
 }
 
-Variable Target::makeVariable(lua_State* L, const std::string& name)
+Variable Target::makeVariable(lua_State* L, int stackSlot, const std::string& name, int parentRef)
 {
     Variable var;
     var.name = name;
-    var.type = lua_typename(L, lua_type(L, -1));
+    var.type = lua_typename(L, lua_type(L, stackSlot));
     switch (lua_type(L, -1))
```
raymondnumbergenerator:
```suggestion
    switch (lua_type(L, stackSlot))
```
is this also supposed to be stackSlot

definitions/debugger.luau
```diff
@@ -211,6 +211,15 @@ export type Target = {
 	--- If `frameId` is -1 or absent, evaluate in the global scope.
 	--- Returns a `Variable` if expression could be evaluated or an error string if not.
 	evaluateExpression: (self: Target, expression: string, frameId: number?) -> Variable | string,
+	--- Sets a variable named `name` within the scope with reference `variableRef` equals to the value
+	--- that `setExpression` evaluates to. If this operation is not successful, it returns an error string.
+	--- Otherwise, it returns a `Variable` containing the new value.
+	setVariable: (self: Target, varRef: number, varName: string, setExpression: string) -> Variable | string,
+	--- Sets the l-expression that `lExpression` is equal to to the value that `setExpression` evaluates to.
```
Vighnesh-V: l-expression?

lute/debug/include/lute/debuginternals.h
```diff
@@ -116,6 +116,8 @@ struct Variable
     bool isTruthy();
 };
 
+// evaluate multi result exists to evaluate expressions that can return any amount of variables (including zero)
+using EvaluateMultiResult = std::variant<std::vector<Variable>, std::string>;
```
Vighnesh-V: use Luau::Variant instead of std::variant here. We've spent more effort making that one good :)

lute/debug/src/debuginternals.cpp
Vighnesh-V: makes sense!

lute/debug/src/debuginternals.cpp
```diff
@@ -1418,7 +1466,198 @@ EvaluateResult Target::evaluateExpression(std::string expression, int frameId)
...
+        lua_getfenv(thread, -1);
+        EvaluateResult result = setTableEntryHelper(thread, lua_gettop(thread), contextLevel, varName, setExpression);
+        lua_pop(thread, 2);
+        return result;
+    }
+    else if (context.type == VariableScopeType::Table)
+    {
+        lua_rawgeti(thread, LUA_REGISTRYINDEX, context.luaref);
+        EvaluateResult result = setTableEntryHelper(thread, lua_gettop(thread), contextLevel, varName, setExpression);
+        lua_pop(thread, 1);
```
Vighnesh-V, line 1594: does this popThread happen as a consequence of setTableEntryHelper? Or is it from lua_rawgeti?

Vighnesh-V: if it's the former, the lua_pop should belong in that function

Vighnesh-V (commented):
some comments :)

definitions/debugger.luau
wizhao96: oops i misremember the term for this (it should be lvalues in C++) but will change to something better that is also not C++ specific

lute/debug/include/lute/debuginternals.h
wizhao96: i've switched generally to using Luau::variant throughout the codebase. Just a note: it would be nice to have some place where you could easily find out where there are Luau:: versions of C++ structures (like DenseHashMap for unordered_map)

lute/debug/src/debuginternals.cpp
wizhao96: it's the latter, we're popping from the lua_rawgeti

Vighnesh-V (approved):

Issue #815, opened 2026-02-12
Subcommands with CLI battery


vocksel:
I'm interested in writing a more complex CLI for Flipbook's build script which would include subcommands for building specific chunks of the codebase.

A couple examples of what I'm looking for are:

```
lute scripts/build.luau plugin --channel prod
lute scripts/build.luau package --channel prod --member package-name
```

I could maybe hack something together with positional arguments but I would much rather have first-class support that could also integrate with the `help` output.

If the CLI battery could be updated to support this use case it would make my life easier :)

annieetang:
closing as completed as https://github.com/luau-lang/lute/pull/1304 is merged!

PR #1304, opened 2026-08-21
Adds subcommand support to the CLI battery

astrorehan:
#815 asks for first-class subcommands in `@batteries/cli`, so a build script can expose `build.luau plugin --channel prod` and `build.luau package --channel prod --member foo` without hand-rolling positional parsing.

`cli.subcommand` registers one and hands back a parser for it:

```luau
local args = cli.parser()
args:add("verbose", "flag", { help = "Enable verbose output", aliases = { "v" } })

local plugin = args:subcommand("plugin", { help = "Build the plugin" })
plugin:add("channel", "option", { help = "Release channel", default = "dev" })

local package = args:subcommand("package", { aliases = { "pkg" }, help = "Build a single package" })
package:add("member", "option", { help = "Package name" })

args:parse({ ... })

local name, selected = args:selected()
```

The returned value is a normal `Parser`, so a subcommand gets flags, options, positionals, a variadic argument, and subcommands of its own for free. `options` is optional on `subcommand`, and on `add` as well, so `args:subcommand("plugin")` and `args:add("verbose", "flag")` both work.

### Parsing

When `parse` reaches a bare argument that names a registered subcommand or one of its aliases, the rest of the arguments go to that subcommand's parser and the parent stops reading. Arguments before it stay with the parent. `cli.selected` returns the canonical name and the parser, or `nil` when no subcommand ran.

That makes the command name a boundary: an argument declared on the parent has to be given before it, the way `git` and `docker` treat their global options. A missing one now says so instead of only naming the argument:

```
Missing required argument: root (it has to be given before the 'build' command)
```

An argument that matches no subcommand keeps its existing meaning: a positional, then a variadic element, then a forwarded argument. But a parser that declares subcommands and neither a positional nor a variadic has nothing else the word could have been, so a typo throws `Unknown command: buld` rather than being quietly forwarded. Parsers that declare either one keep forwarding, since there an unrecognized word is not necessarily a command at all. Nothing changes for a parser that registers no subcommands.

### Duplicate names

Both lookups return the first match, so a repeated name or alias used to register fine and then sit unreachable, and this gets easier to hit once a parser has commands as well as arguments. `add` and `subcommand` now throw on a name or an alias that collides with one already registered, on an alias equal to its own name, and on the same alias listed twice in one call.

Variadics turned out to have the same shape of bug: a second one overwrote `variadicArg` and left the first registered but unreachable. That throws now too.

An argument and a command may still share a name, since `--build` and `build` never collide in `parse`, and aliases may repeat across nested parsers.

### Usage and help

`cli.usage` gains a `<command>` placeholder, ahead of the positionals so it matches the order `parse` tries them in, and `cli.help` grows a `Commands:` section above `Options:`:

```
Usage: build <command> [options]

Commands:
  plugin - Build the plugin
  package (pkg) - Build a single package

Options:
  --verbose (-v) - Enable verbose output
```

`hidden` works on subcommands the way it already does on arguments, and both sections are omitted when nothing is registered, so the existing help snapshots are untouched.

### Testing

`tests/batteries/cli.test.luau` gains a `CLI subcommands` suite covering dispatch, aliases, nesting, precedence against positionals, argument ownership on both sides of the command, defaults and required arguments inside a subcommand, variadic and forwarded arguments, unknown commands against each of the parsers that can absorb them, the ordering rule and its error message, and the usage placeholder with and without positionals. A `CLI duplicate names` suite covers the validation above. A new snapshot covers the `Commands:` rendering including aliases and hidden entries.

`lute test tests/batteries` reports 182 passed, 0 failed, and the four CLI snapshots pass. `lute check` and `lute lint` are clean on the changed files.

### Notes

I have not migrated any of lute's own commands onto this. `lute self`, `lute pkg` and the `snapshots` branch inside `lute test` each dispatch by hand today and would all shrink, but that is a bigger and more opinionated diff than the battery change, and it seemed better to land the primitive first. Happy to follow up if you want it.

Addresses #815. I cannot set labels on this repository, so this one wants the `batteries` label.


astrorehan:
The Windows `run-lutecli-luau-tests` failure is `LuteDebugLibrarySuite.multithread`, which this branch does not touch. Everything else passed, including the same suite on macOS and both Ubuntu configurations, and the base commit bfe25db9 is green on `primary`.

The only assertion that failed is the five second `waitUntil` for the debuggee to exit at `tests/lute/debugger.test.luau:361`. The counters checked right after it, `bpHit1`, `bpHit2` and `maxThreads`, all passed, so every breakpoint was hit and the thread bookkeeping was correct; only the exit notification missed the deadline.

The change itself cannot reach that test. It is confined to `@batteries/cli`, and a parser that registers no subcommands walks an identical path through `parse`: the new lookup runs against an empty list and falls straight through to the existing positional, variadic and forwarded handling.

Running `lute test tests/lute/debugger.test.luau` on Windows here fails `multithread`, `inspectVariables` and `evaluateExpressions`, all with the same `eq: false ~= true` from a `waitUntil` timeout. That is against a stock `1.0.1-nightly.20260814` binary which contains none of this branch, though it also predates some of the recent debugger work, so take it as supporting rather than conclusive.

Could someone re-run the job? I do not have permission to. Happy to open a separate issue about the `waitUntil` deadlines if that would be useful.

wizhao96:
hi! thanks for contributing! i reran the tests for you. can you make an issue about the tests? is it timing out persistently or is it flaky? are they always the same set of tests that fail? or are you running the tests on `1.0.1-nightly.20260814`, in which case I think failure of these test cases is expected.

im so sorry about this but as much info as possible is appreciated because im not working in windows

astrorehan:
Thanks for the rerun, it went green.

You were dead right about the nightly. rokit.toml pins 1.0.1-nightly.20260814, which
is older than #1290, #1298 and #1299, so that binary cannot run the debugger suite
here at all. It falls over with attempt to call missing method 'evaluateExpression' of
Target, and all three failures I quoted are downstream of that. Ignore them, they told
us nothing.

The CI one is real though, and it is not a timeout either, which is the other thing I
got wrong. Digging into the attempt 1 log, this shows up just before the failure:

    D:/a/lute/lute/tests/src/debug/task.luau:7: attempt to perform arithmetic (add) on function and number

task.luau:7 is total_sum += 1, and total_sum never holds anything but a number. It is
also the exact line bp1 sits on. bpHit1, bpHit2 and maxThreads all passed, so the
breakpoints themselves were fine and this lands at or near the last hit. The debuggee
dies there, onExit never gets a truthy status, and the waitUntil at :361 is just the
five second echo.

Which I think puts it right in the middle of your #1308, since a function turning up
where a number should be is what you would expect from a thread resuming on the wrong
instruction. One thing worth a second look: the guard there keys on stoppedNoYield,
and that came in with #1300 at 21:38 on the 21st, while this run failed at 04:40 that
morning. So my build had the old level == 0 && L == stoppedThread condition and
stoppedNoYield was not in the picture yet. The savedpc swap in getLocalsHelper still
looks reachable on the ordinary path.

Wrote it up in #1310 with the full log excerpts. 

Also, you are chasing a Windows-only flake without a Windows
machine, which is the hard way round by a mile. I'm on Windows with the repo set up,
so if you want a build run, a patch tried, or to hammer that suite a few
hundred times, tag me and I'll turn it around.

batteries/cli.luau
```diff
@@ -59,6 +75,30 @@ function cli.add(self: Parser, name: string, kind: ArgKind, options: ArgOptions)
 	end
 end
 
+function cli.subcommand(self: Parser, name: string, options: CommandOptions): Parser
+	local parser = cli.parser()
+
+	table.insert(self.commands, {
+		name = name,
+		parser = parser,
+		options = options or { aliases = {} },
```
annieetang: I see `options: CommandOptions` is a non-optional parameter, but we have this fallback for `options or { aliases = {} }`, which would only reach if the caller passes `nil` for `options`, but it doesn't seem that's allowed, so would we ever reach this fallback?

We could either make options `CommandOptions?` to allow something like `p:subcommand("build")`, or we could drop the `or` if we want the arg to be required, thoughts? 

Also, I understand `cli.add` also has this fallback style, but I wonder if it's necessary or not

batteries/cli.luau
```diff
@@ -132,6 +172,14 @@ function cli.parse(self: Parser, args: { string }): ()
 			continue
 		end
 
+		-- if the argument names a subcommand, that subcommand takes over the rest of the arguments
+		local command = findCommand(self, arg)
+		if command then
+			self.selectedCommand = command
+			command.parser:parse(table.move(args, i + 1, #args, 1, {}))
+			break
```
annieetang, line 219: If we find a subcommand, I see we `break`, but this only exits the parse loop, and we continue to check parent's `self.arguments`. So it looks like we still enforce a parent's required args, even when a subcommand is selected, and I wonder if confusing behaviors could happen when we mix the two together:

Given something like,

```lua
p:add("root", "option", { required = true })
p:subcommand("build", {})
p:parse({ "build" })   -- currently throws "Missing required argument: root"
```

then trying, 

```lua
p:parse({ "build", "--root", "./src" })
```

We would get `Unknown argument: root`, right? So we're implicitly requiring that parent args have to come before the subcommand. Is that intended? If so, should we make that explicit somewhere, or maybe improve the expected error, because `Unknown argument: root` in this case might be misleading.

batteries/cli.luau
```diff
@@ -132,6 +172,14 @@ function cli.parse(self: Parser, args: { string }): ()
 			continue
 		end
 
+		-- if the argument names a subcommand, that subcommand takes over the rest of the arguments
+		local command = findCommand(self, arg)
```
annieetang, line 215: Mostly a thought/open-ended question: you mentioned leaving it to the caller to decide what to do with an unrecognized command, which makes sense given a parser can have both subcommands and positionals, but it seems a little hard for the caller to clearly distinguish an unknown command from no command at all via  `cli.selected()`. 

For example, when a typo happens:

```lua
p:subcommand("build", {})
p:parse({ "buld" }) -- typo, but no positional to absorb it
p:selected() -- ends up as nil, same as p:parse({})
```

we silently forward it, and it can't be easily rejected currently. In the earlier parts of `cli.parse`, unknown flags throw (see `assert(found, "Unknown argument: " .. name)`). Would it make sense to either surface something the caller could check, or throw an `Unknown command: ...` in the case where a subcommand exists but there's no positional/variadic to process it?

batteries/cli.luau
```diff
@@ -194,6 +251,13 @@ function cli.usage(self: Parser, name: string): string
 		end
 	end
 
+	for _, command in self.commands do
+		if not command.options.hidden then
+			table.insert(parts, "<command>")
```
annieetang: Similar topic about ordering with positional and subcommands, this would give us `Usage: tool <file> <command> [options]` right, because it's being inserted after positionals and before [options]? Which could seem a bit misleading 

Could we also add a couple of test cases for these situations? for the usage message but also parsing

annieetang (commented):
Looking great so far, thanks for the work on this! 

General question (in addition the commented ones), but what happens when we register two subcommands with the same name? Or a command whose name overlaps with an alias of another? Do we want to handle those cases (or add test cases for this)
^I notice we don't really handle duplicate arg names in `cli.add`, but could become more of a concern once we add subcommand support...

batteries/cli.luau
astrorehan: I made it optional on both subcommand and add, so p:subcommand("build") works and the fallback isn't dead anymore. seemed nicer than dropping it.

batteries/cli.luau
astrorehan: it is intended. parent args have to come before the subcommand name, since everything after it gets handed to the subcommand's own parser. the error said nothing about that though, which is the actual problem. now:

Missing required argument: root (it has to be given before the 'build' command)
and parse has a comment stating the rule.

the Unknown argument: root one is real but sort of inherent. build owns the rest of the line and has never heard of --root, which is what argumentsAfterSubcommandBelongToIt already pins down. added a test for it so it's written somewhere. having the parent pick up its own args after the command name is a bigger change, can do that separately if you want it.

batteries/cli.luau
astrorehan: it throws "Unknown command: buld" now, but only when the parser declares no positional and no variadic at all, so there's genuinely nothing else the word could have been. declare either one and it forwards like before, since an unrecognized word there isn't necessarily a command.

i had the condition too broad at first and it broke forwarding for mixed parsers once the positionals were full, hence the second commit.

batteries/cli.luau
astrorehan: moved <command> ahead of the positionals so it matches the order parse tries them in. gives Usage: tool <command> [file] [options]. tests added for the usage string and the parsing cases.

left one thing alone: when a parser has both, a positional and a command are alternatives rather than a sequence, so any flat ordering still reads like you pass both. tool <command> | [file] [options] would be honest but noisier. lmk if you'd rather have that.

astrorehan:
both were silently ignored before. the lookups return the first match, so the second registration never got reached. add and subcommand reject a colliding name or alias now, and catch an alias that's the same as its own name or listed twice in one call.

variadics turned out to have the same problem, which i only noticed writing tests for this. a second one replaced the first and left it registered but unreachable. throws now.

didn't block an argument and a command sharing a name though, since --build and build never collide in parse.

pushed as dffd1f50, rest of the answers are inline.

batteries/cli.luau
annieetang: sounds good! thanks for updating `add` as well

batteries/cli.luau
annieetang: throwing the unknown command message sounds good to me now, thx for updating that, it'll definitely help give a clearer error message in these cases

batteries/cli.luau
annieetang: The error update seems great to me and more informative. We don't need the parent to pick up its own args after the command because requiring globals before the subcommand matches how git/docker/etc work, so this is reasonable to match, and we can always revisit if it becomes an issue later on.

batteries/cli.luau
annieetang: `Usage: tool <command> [file] [options]` looks good and lining up with parse order makes sense. No need for the other version, I don't think it would come up that often to justify needing the noise. Thanks!

batteries/cli.luau
```diff
@@ -38,17 +52,57 @@ function cli.parser(): Parser
...
+	for _, command in self.commands do
+		local aliases = command.options.aliases or {}
+
+		if command.name == name or table.find(aliases, name) then
+			return command
+		end
+	end
+
+	return nil
+end
```
annieetang, line 88: nit: the these two helpers as well as the two inline loops in `parse` all seem to repeat the `name == x or table.find(aliases, x)` check, so it would be great if we can share this lookup and cut the repetition somehow, but not a blocking comment

annieetang (approved):
Looks great, thanks for the updates! Could you update the PR description with the new updates, including the changes you made to `add` (i.e., duplicate commands now throw) 

and there's also one more nit below

batteries/cli.luau
astrorehan: done in 335e11a6. the check lives in isNamed now and both find helpers call it.

the two loops in parse turned out to be worse than repetition, they were findArgument written out by hand, so they're gone entirely and call it instead. -41 +28 on the file. left the ones in add and subcommand alone though. those run before the argument is registered, so there's nothing for isNamed to look at yet. I also updated the description like you asked.

annieetang: great! thanks so much for contributing, will merge in!

PR #1308, opened 2026-08-21
`lute/debug`: solve race condition bug involving stopping the debugger

### LINKED ISSUE #1310: Windows: LuteDebugLibrarySuite.multithread flakes when the debuggee errors at the breakpoint line
LuteDebugLibrarySuite.multithread fails intermittently on Windows. It gets reported as
a waitUntil timeout, but the debuggee is actually erroring at the breakpoint line, and
the timeout is just the knock-on effect.

Opening this because #1308 goes after the same flake and says the cause is not
certain. There is more in the log than the failure line suggests.

## Reproduction

1. Build from source on Windows with cl.exe, matching the
   run-lutecli-luau-tests (windows-latest, cl.exe, all) job.
2. Run `lute test tests/lute/debugger.test.luau`.
3. Repeat. It passes most of the time.

Concrete occurrence: https://github.com/luau-lang/lute/actions/runs/32447447407,
attempt 1, job 96669559876. Same commit, same job, attempt 1 failed and attempt 2
passed. That was on #1304, whose diff only touches batteries/cli.luau.

## What the log shows

The reported failure:

    FAIL LuteDebugLibrarySuite.multithread
        tests/lute/debugger.test.luau:361
        eq: false ~= true

Line 361 is the waitUntil(function() return onFinished end) check, so it reads as a
slow debuggee hitting the 5 second poll limit. It is not that. Earlier in the same log:

    D:/a/lute/lute/tests/src/debug/task.luau:7: attempt to perform arithmetic (add) on function and number
    stacktrace:
    D:/a/lute/lute/tests/src/debug/task.luau:7

tests/src/debug/task.luau:7 is total_sum += 1. total_sum is local total_sum = 0 on
line 2, captured by both deferred closures, and nothing in that file ever assigns a
non-number to it. Line 7 is also where the test puts bp1:

    local bp1 = target:setBreakpoint(mainPath, 7)

bpHit1 == 5, bpHit2 == 5 and maxThreads == 3 all passed, so every breakpoint fired the
expected number of times and this lands at or near the final hit. The debuggee then
errors, onExit never receives a truthy status, onFinished stays false, and the
waitUntil is what ends up in the report five seconds later.

## Where it probably comes from

A function sitting where a number should be is what you would get if the thread
resumed on the wrong instruction. The thing on that path is Target::getLocalsHelper,
which the test reaches through getVariablesByScopeType(threadStack.id, "local") on
that same line:

    const Instruction* original = L->ci->savedpc;
    if (level == 0 && L == stoppedThread)
    {
        L->ci->savedpc = stoppedPc;
        fixedSavedpc = true;
    }

#1308 adds !stoppedNoYield to that condition. Worth flagging that stoppedNoYield
arrives with #1300, which merged at 21:38 on 2026-08-21, while this run failed at
04:40 the same morning. So this failure happened under the older condition and the new
guard would not have changed it. Whatever makes savedpc wrong here looks reachable
without the non-yieldable path.

## What I could not check

No repro outside CI here. rokit.toml pins 1.0.1-nightly.20260814, which predates
#1290, #1298 and #1299, and that binary cannot run this suite locally at all (attempt
to call missing method 'evaluateExpression' of Target). Building from source needs
MSVC, which I do not have set up, so I could not bisect or confirm the mechanism
directly.

I cannot set labels on this repository. This one wants bug and runtime.

wizhao96:
Fixes https://github.com/luau-lang/lute/issues/1310. In order to do so, in yieldable contexts, we don't fire the callback immediately when receiving notification of hitting a breakpoint. instead, we wait for the current coroutine to yield and then fire this callback (after which the runtime will now be paused).

**Bug**
The root cause of the issue is that in the child runtime we schedule a debug callback on the parent runtime that inspects the child runtime in this time. This can race when the child runtime unwinds from `luau_callhook` leading to a corruption of the instruction pointer of the VM.

In particular, the issue is that in `getLocalVars()` in `debuginternals.cpp`
```
const Instruction* original = L->ci->savedpc;   // (A) read live pc
... L->ci->savedpc = stoppedPc;                 // (B)
... lua_getlocal loop ...
L->ci->savedpc = original;                      // (C) restore what A read
```

can race with in `luau_callhook` in `lvmexecute.cpp` in the Luau VM

```
oldsavedpc = L->ci->savedpc;      // = break_pc (line 185)
L->ci->savedpc++;                 // now break_pc+1   (line 188)
hook(L, &ar);                     // <-- our debugbreak runs here, schedules onBreakpointHit, returns (line 199)
L->ci->savedpc = oldsavedpc;      // restore break_pc (line 201)
```

Line 201 and (C) can race, resulting in L->ci->savedpc being set to break_pc + 1 instead of break_pc.

**Fix**
We fire the callback after the child runtime unwinds from `luau_callhook` and has already yielded the current coroutine. We do this right after yielding and right before the pause freezes further coroutines from firing on the runtime. 

Thus, since `luau_callhook` has already finishing calling and the thread it is being called on is yielded, there is no more opportunity for a race condtion.

### DIFF
--- a/lute/debug/src/debuginternals.cpp
+++ b/lute/debug/src/debuginternals.cpp
@@ -913,7 +913,7 @@ std::vector<Variable> Target::getLocalsHelper(lua_State* L, int level)
     // when hitting a bp we try to re-enter
     bool fixedSavedpc = false;
     const Instruction* original = L->ci->savedpc;
-    if (level == 0 && L == stoppedThread)
+    if (level == 0 && !stoppedNoYield && L == stoppedThread)
     {
         L->ci->savedpc = stoppedPc;
         fixedSavedpc = true;
@@ -1038,7 +1038,7 @@ void Target::injectLocals(lua_State* L, int level, lua_State* eval, int evalTabl
 {
     bool fixedSavedpc = false;
     const Instruction* original = L->ci->savedpc;
-    if (level == 0 && L == stoppedThread && L->ci)
+    if (level == 0 && !stoppedNoYield && L == stoppedThread && L->ci)
     {
         L->ci->savedpc = stoppedPc;
         fixedSavedpc = true;
--- a/lute/debug/src/debuginternals.cpp
+++ b/lute/debug/src/debuginternals.cpp
@@ -361,6 +361,47 @@ void Target::computeStoppedLocation(lua_State* L)
         stoppedLocation = "";
 }
 
+void Target::stoppedSetState(lua_State* L)
+{
+    paused = true;
+    childRuntime->stopDebug();
+    computeStoppedLocation(L);
+    stoppedThread = L;
+    stoppedThreadRef = getRefForThread(L);
+    stepInfo = std::nullopt;
+    // Clear out stepping and pausing when this happens.
+    lua_callbacks(L)->debugstep = nullptr;
+    lua_callbacks(L)->interrupt = nullptr;
+    // When in a yieldable state, we call lua_break to notify the runtime
+    // to yield our current thread of execution and then stop running further coroutines in the runtime.
+    // Otherwise, in a nonyieldable state, we simply set stoppedNoYield to be true. This
+    // will later force us to wait on a condition variable, preventing execution from continuing in our
+    // current coroutine.
+    if (lua_isyieldable(L))
+        lua_break(L);
+    else
+        stoppedNoYield = true;
+}
+
+void Target::stoppedDispatchCallback(std::function<void()> debugStopCallback)
+{
+    if (!stoppedNoYield)
+    {
+        // In a yieldable state, we add a pending callback that the runtime runs
+        // only after runOnce() returns. Thus, it is guaranteed that the entire runtime is paused.
+        // If we call immediately, the debuggee coroutine still needs to unwind and is still running
+        // leading to possible race conditons.
+        childRuntime->pendingDebugStopNotification = std::move(debugStopCallback);
+    }
+    else
+    {
+        // We are already inside the debug hook so the runtime cannot run any other code during this time.
+        // We can thus run the callbacks inline and then block here until continued.
+        debugStopCallback();
+        childRuntime->waitForDebugContinue();
+    }
+}
+
 std::optional<std::string> Target::launch(std::string sourcePath, const std::vector<std::string>& args, LaunchConfig config)
 {
     std::vector<Breakpoint> installedBps;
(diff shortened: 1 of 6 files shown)


tests/lute/debugger.test.luau
```diff
@@ -304,13 +304,17 @@ test.suite("LuteDebugLibrarySuite", function(suite)
 					for _, otherThread in threads do
 						if otherThread.id == 3 then
 							local otherTrace = assert(target:getStackTrace(otherThread.id))
-							local line = otherTrace[#otherTrace].line
-							check.that(line >= 12 and line <= 16)
+							if #otherTrace > 0 then
```
wizhao96, line 307: this offer additional protection on this test but i don't think it's the root cause of the issue

lute/runtime/src/runtime.cpp
```diff
@@ -407,10 +414,9 @@ void Runtime::continueDebug()
 }
 
 // stopDebug() actually allows for some C bookkeeping when we are unwinding the stack
-// when it is called within a callback such as debugbreak or debugInterrupt, which you could imagine might
```
wizhao96, line 410: this faulty assumption documented here is the root cause of the bug

lute/debug/src/debuginternals.cpp
```diff
@@ -361,6 +361,47 @@ void Target::computeStoppedLocation(lua_State* L)
         stoppedLocation = "";
 }
 
+void Target::stoppedSetState(lua_State* L)
```
wizhao96, line 364: theoretically we could also refactor getting the modified breakpoints and the current thread into this but I want to keep this exclusive into modifying the current state of the runtime and the debugger

lute/debug/src/debuginternals.cpp
```diff
@@ -913,7 +948,7 @@ std::vector<Variable> Target::getLocalsHelper(lua_State* L, int level)
     // when hitting a bp we try to re-enter
     bool fixedSavedpc = false;
     const Instruction* original = L->ci->savedpc;
-    if (level == 0 && L == stoppedThread)
+    if (level == 0 && !stoppedNoYield && L == stoppedThread)
```
wizhao96, line 951: We don't need to complete this action when in non-yielding stopping, and while I don't think this is actually dangerous, I would rather not modify `L->ci->savedpc` if it's not needed

Vighnesh-V (approved):

PR #1315, opened 2026-08-26
`lute/debug`: fix global scoping across files

wizhao96:
Fixes global scopes across files, since each file has its own set of globals. Previously, global scopes always defaulted to the main script file's globals.

https://github.com/user-attachments/assets/632e665f-4c24-406e-af8e-8ab39d8358cd

### DIFF
--- a/lute/debug/src/debuginternals.cpp
+++ b/lute/debug/src/debuginternals.cpp
@@ -58,22 +58,22 @@ VariableScope::VariableScope(int variableReference, VariableScopeType type, std:
 
 VariableScope VariableScope::makeLocals(int variableReference, int threadId, int level)
 {
-    return VariableScope(variableReference, VariableScopeType::Local, "Locals", threadId, level, -1);
+    return VariableScope{variableReference, VariableScopeType::Local, "Locals", threadId, level, -1};
 }
 
 VariableScope VariableScope::makeUpvalues(int variableReference, int threadId, int level)
 {
-    return VariableScope(variableReference, VariableScopeType::Upvalue, "Upvalues", threadId, level, -1);
+    return VariableScope{variableReference, VariableScopeType::Upvalue, "Upvalues", threadId, level, -1};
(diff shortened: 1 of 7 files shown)


lute/debug/src/debuginternals.cpp
```diff
@@ -764,18 +764,10 @@ std::optional<std::vector<VariableScope>> Target::getScopesHelper(int threadId,
...
-    }
-    else
-    {
-        VariableScope globals = VariableScope::makeGlobals(variableRefId);
-        variableContexts.insert_or_assign(variableRefId, globals);
-        globalVariableRef = variableRefId;
-        variableRefId++;
-        contexts.emplace_back(globals);
-    }
+    VariableScope globals = VariableScope::makeGlobals(variableRefId, threadId, level);
```
wizhao96, line 767: in order for this fix to work, globals need to have unique scopes. this was generally going to happen anyway in the set variable/expression change

lute/debug/src/debuginternals.cpp
```diff
@@ -66,9 +66,9 @@ VariableScope VariableScope::makeUpvalues(int variableReference, int threadId, i
     return VariableScope(variableReference, VariableScopeType::Upvalue, "Upvalues", threadId, level, -1);
 }
 
-VariableScope VariableScope::makeGlobals(int variableReference)
+VariableScope VariableScope::makeGlobals(int variableReference, int threadId, int level)
 {
-    return VariableScope(variableReference, VariableScopeType::Global, "Globals", -1, -1, -1);
+    return VariableScope(variableReference, VariableScopeType::Global, "Globals", threadId, level, -1);
```
Vighnesh-V: nit, please use {} initialization instead of (). Makes grepping consistently easier.

Vighnesh-V (approved):
left some comments

raymondnumbergenerator (approved):

PR #1319, opened 2026-08-28
`lute/debug`: add onSourceLoad

wizhao96:
adds `onSourceLoad` to inform when sources are loaded. this is used in dap to send the `loadedSource` event.
* this is apparently what the emacs `dape` adapter uses to track sources instead of the `loadedSources`request and is easily implementable


lute/debug/src/debuginternals.cpp
```diff
@@ -483,7 +485,13 @@ std::optional<std::string> Target::launch(std::string sourcePath, const std::vec
             childRuntime.reset();
             return error;
         }
+        paused = false;
+        launched = true;
+        parentRuntime.numLaunchedDebuggees++;
```
wizhao96, line 507: this statements needs to moved earlier so we can fire the `onSourceLoad`'s `debugScheduleCallback` with numLaunchedDebuggees > 0

Vighnesh-V (approved):

